PR #21, opened 2013-03-31
Proposed spectrum arithmetic tools

keflavich:
(I'll need to fix some of the commit messages)

This PR includes a proposed spectrum arithmetic toolset.  It will allow you to do something like:

```
# assume Spectrum1D() magically creates a valid spectrum...
sp1 = Spectrum1D()
sp2 = Spectrum1D()
# this works iff all(sp1.disp == sp2.disp)
sp3 = sp1+sp2
# this works if sp1.disp ~ sp2.disp
sp1._arithmetic_threshold(1,'angstrom')
sp3 = sp1-sp2
```

There probably needs to be a `force` option as well, to override the dispaxis agreement entirely.

This PR includes at least one assumption that is not necessarily satisfied: that `disp` has a `units` attribute (it should, but perhaps not by that name)

In the comments, let's discuss first whether this is a good idea at all (I think so), then implementation if we give it a go-ahead.

Also, tell me if `nddata` will do all this implicitly... I am still unfamiliar with its capabilities.

### DIFF
--- a/specutils/spectrum1d.py
+++ b/specutils/spectrum1d.py
@@ -305,5 +305,73 @@ class Spectrum1D(NDData):
         # Which are all common NDData objects, therefore I am (perhaps
         # reasonably) assuming that __slice__ will be a NDData base function
         # which we will inherit.
-        raise NotImplemented('Will presumeably implemented in core NDDATA')
+        # raise NotImplemented('Will presumeably implemented in core NDDATA')
+        return self.__slice__(start_index, stop_index)
+        
+    # PROPOSAL:
+    def _operation_wrapper(operation):
+        """
+        Perform an operation (addition, subtraction, mutiplication, division,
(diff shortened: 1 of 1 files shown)


jason-neal:
Hi,  Are these arithmetic operations going to be included any time soon?

keflavich:
@jason-neal Unlikely.  Specutils is in a somewhat unmaintained state right now, so we don't know what direction it is going to take and these sorts of features can't be merged until some bigger pictures are resolved.

jason-neal:
@keflavich Thanks for the quick reply. Ok I will use my own creation.

nmearl:
Closing due to backwards-incompatible break with the new `specutils` implementation. Please reopen this PR against the [`legacy-specutils`](https://github.com/astropy/specutils/tree/legacy-specutils) branch if you wish to keep this discussion open.

PR #110, opened 2014-12-03
Fix extinction units

JALusk:
This is my first pull-request, so I hope I'm doing this right! I faced some problems when trying to use the extinction laws included in specutils. Specifically, the code as written introduced extraneous units to the extinction laws, returning astropy quantities with units of Angstroms instead of the expected dimensionless quantities.

These changes aim to fix that problem. I have removed code which introduces the wavelength units to the extinction, and changed certain aspects of the docstrings which indicated functionality not yet implemented (specifically, the freedom to specify either A(V) or E(B-V): the extinction functions as written expect to be passed A(V), and the functinality to check the inputs and convert E(B-V) to A(V) doesn't seem like it's been integrated into the function calls, though I might be wrong about that.)

I have not figured out how to run the test suite yet, so I don't know if I've broken something subtle, but I have been able to import and use the reddening() function in my research code and got expected results - something I could not do before my changes (since the exponent was not unitless before these changes.)

### DIFF
--- a/specutils/extinction.py
+++ b/specutils/extinction.py
@@ -215,7 +215,7 @@ class ExtinctionF99(object):
         # Analytic function in the UV.
         uvmask = wave < (2700. * u.angstrom)
         if np.any(uvmask):
-            res[uvmask] = cextinction.f99uv(wave[uvmask].value, self.a_v, self.r_v) * wave.unit
+            res[uvmask] = cextinction.f99uv(wave[uvmask].value, self.a_v, self.r_v)
 
         # Spline in the Optical/IR
         oirmask = ~uvmask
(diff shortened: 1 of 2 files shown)


JALusk:
Just want to poke the maintainers about this PR. It looks like it's still passing the CI builds, but I can double-check by testing against the latest dev build of astropy, if you like.

Cheers,
Jeremy

wkerzendorf:
@JALusk I made a PR on your PR (let me know if that is unclear). to add a test for the now dimensionless Quantity

JALusk:
I merged your PR with the regression test included.

WIll that change automatically be reflected in my PR, or do I have to re-submit?

wkerzendorf:
no need to resubmit. `astropy` just got a new requirement and I need to fix it before TRAVIS will pass again. Should be done tomorrow.

wkerzendorf:
@JALusk, so there was a recent change in the astropy-core `nddata` package that `specutils` relies on. I have changed `specutils` to be compliant to this new structure (see PR #113). Can you rebase onto the new master. If this doesn't mean anything to you, we can skype and I can give you a quick step-by-step tour. Sorry, that this has been a bit of an involved process.

JALusk:
I think I have the basic idea, but that's what I always think right before I make a mess with git.

I'd appreciate the help at your convenience. I'll email you my skype id.

Cheers,
Jeremy

JALusk:
Rebased onto the new master, awaiting Travis build.

Cheers,
Jeremy

wkerzendorf:
@JALusk Thanks for doing this!!

PR #112, opened 2015-01-09
Heliocentric velocity corrections for astronomical sources

andycasey:
Hola,

This pull request adds a utility function that will calculate heliocentric velocity corrections for astronomical sources. The `helcorr` and `baryvel` functions have been updated from the [astrolibpy](https://code.google.com/p/astrolibpy/) package which was written by Sergey Koposov (@segasai). Those astrolibpy functions were originally Fortran functions from the triassic period before they were updated to modern IDL. In `helcorr.py` I have credited authors who have contributed to these functions over the years, and @segasai has granted permission for these updated functions to be included in AstroPy.

If people have strong opinions about the input to the `helcorr` function then I'd like to hear them. At the moment it takes observatory longitude, latitude, altitude, source RA, source Dec and MJD, but in reality these could be replaced with just a `astropy.coordinates.SkyCoord` and a `astropy.time.Time` object with a specified `location`. What do people think?

We also need to drum up some test cases for this code. Perhaps someone would be kind enough to generate a few test cases using IRAF?

### DIFF
--- a/specutils/helcorr.py
+++ b/specutils/helcorr.py
@@ -249,7 +249,7 @@ def baryvel(dje):
 # Similarly lon/lat/alt/jd could be replaced with a single astropy.units.Time
 # class.
 
-def helcorr(lon, lat, alt, ra, dec, jd):
+def helcorr(lon, lat, alt, ra, dec, mjd):
     """
     Calculate the heliocentric radial velocity corrections for an astronomical 
     source.
@@ -269,8 +269,8 @@ def helcorr(lon, lat, alt, ra, dec, jd):
         Right ascension of the object for epoch J2000 (if float, in degrees).
     dec : `~astropy.coordinates.Angle` or float
         Declination of the object for epoch J2000 (if float, in degrees).
-    jd : float
-        Julian date for the middle of exposure.
+    mjd : float
+        The modified Julian date for the middle of exposure.
 
     Returns
     -------
@@ -295,7 +295,7 @@ def helcorr(lon, lat, alt, ra, dec, jd):
 
     # Here we specify the location so that we can easily calculate the mean
     # local siderial time later on
-    time = astropy.time.Time(2.4e6 + jd, format="jd", location=(lon, lat, alt))
+    time = astropy.time.Time(2.4e6 + mjd, format="jd", location=(lon, lat, alt))
     epoch = time.datetime.year + time.datetime.month/12. \
         + time.datetime.day/365.
 
--- a/specutils/helcorr.py
+++ b/specutils/helcorr.py
@@ -1,0 +1,343 @@
+# coding: utf-8
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+""" Heliocentric velocity corrections """
+
+from __future__ import division, print_function
+
+# This code has been modified from legacy codes. Credits are chronologically
+# attributable to: Sergey Koposov, Kochukho, Kudryavtsev, W. Landsman, 
+# Chris McCarthy, and Jeff Valenti.
+
+# Third-party
+import numpy as np
+
+import astropy.constants as constants
+import astropy.coordinates as coord
+import astropy.time
+import astropy.units as u
+
+__all__ = ["helcorr", "baryvel"]
+
+def baryvel(dje):
+    """
+    Calculates the heliocentric and barycentric velocity components of Earth.
+
+    Parameters
+    ----------
+    dje : `~astropy.time.Time` or float
+        The Julian ephemeris date.
+
+    Returns
+    -------
+    dvelh : `~astropy.units.Quantity`
+        The heliocentric velocity components in (X, Y, Z) coordinates.
+
+    dvelb : `~astropy.units.Quantity`
+        The barycentric velocity components in (X, Y, Z) coordinates.
+    """
+
+    if isinstance(dje, astropy.time.Time):
+        dje = dje.jd
+
+    # Prepare for the pain.
+    dcto = 2415020.
+    dcjul = 36525. # Days in Julian year
+    dc1 = 1.
+
+    # Constants dcfel(i,k) of fast changing elements.
+    dcfel = np.array([
+        1.7400353e00, 6.2833195099091e02,  5.2796e-6, 
+        6.2565836e00, 6.2830194572674e02, -2.6180e-6,
+        4.7199666e00, 8.3997091449254e03, -1.9780e-5,
+        1.9636505e-1, 8.4334662911720e03, -5.6044e-5, 
+        4.1547339e00, 5.2993466764997e01,  5.8845e-6, 
+        4.6524223e00, 2.1354275911213e01,  5.6797e-6, 
+        4.2620486e00, 7.5025342197656e00,  5.5317e-6, 
+        1.4740694e00, 3.8377331909193e00,  5.6093e-6]).reshape(8, 3)
+
+    # Constants dceps and ccsel(i,k) of slowly changing elements.
+    dceps = np.array([4.093198e-1, -2.271110e-4, -2.860401e-8])
+    ccsel = np.array([
+        1.675104e-2, -4.179579e-5, -1.260516e-7,
+        2.220221e-1,  2.809917e-2,  1.852532e-5,
+        1.589963e00,  3.418075e-2,  1.430200e-5,
+        2.994089e00,  2.590824e-2,  4.155840e-6,  
+        8.155457e-1,  2.486352e-2,  6.836840e-6,
+        1.735614e00,  1.763719e-2,  6.370440e-6, 
+        1.968564e00,  1.524020e-2, -2.517152e-6, 
+        1.282417e00,  8.703393e-3,  2.289292e-5,
+        2.280820e00,  1.918010e-2,  4.484520e-6, 
+        4.833473e-2,  1.641773e-4, -4.654200e-7,
+        5.589232e-2, -3.455092e-4, -7.388560e-7, 
+        4.634443e-2, -2.658234e-5,  7.757000e-8,  
+        8.997041e-3,  6.329728e-6, -1.939256e-9,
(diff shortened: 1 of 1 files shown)


andycasey:
Travis is failing on 5/7 jobs in the build matrix, none of which seem to be related to the code in this PR.

wkerzendorf:
@andycasey it is likely to be a change in modeling which is currently undergoing restructuring. I'll have a look.

keflavich:
I'm entirely in favor of adding this functionality, but I thought ERFA was supposed to include some of these conversions as low-level methods?  I might be entirely wrong about that, but I wanted to raise the point in case someone knows of another implementation.  pyslalib (https://github.com/scottransom/pyslalib) might also be useful for generating test cases.

wkerzendorf:
@andycasey it's the new jinja2 requirement from astropy. Will add to TRAVIS.

wkerzendorf:
@andycasey the structure of `nddata` has changed and I needed to update specutils to that ( #113 ). I'm merging this now - can you rebase your branch?

wkerzendorf:
Ah: tests. Can you add some?

specutils/helcorr.py
```diff
@@ -0,0 +1,343 @@
...
+    Returns
+    -------
+    dvelh : `~astropy.units.Quantity`
+        The heliocentric velocity components in (X, Y, Z) coordinates.
+
+    dvelb : `~astropy.units.Quantity`
+        The barycentric velocity components in (X, Y, Z) coordinates.
+    """
+
+    if isinstance(dje, astropy.time.Time):
```
wkerzendorf: I would only allow an `astropy.time.Time` object. If float is really important, please use ducktyping to differentiate between time and float.

andycasey: I was following the convention set by the `astropy.time` and `astropy.coordinates` module where they would allow astropy objects or a float. I can ducktype for floatst if necessary, but checking against the `Time` baseclass would be sufficient to see if it's an astropy object or not

specutils/helcorr.py
```diff
@@ -0,0 +1,343 @@
...
+    mjd : float
+        The modified Julian date for the middle of exposure.
+
+    Returns
+    -------
+    helcorr : `~astropy.units.Quantity`
+        The heliocentric velocity correction.
+    """
+
+    if not isinstance(lon, coord.Longitude):
```
wkerzendorf: I would suggest to require that these things are quantities or directly coordinates, to make sure that users definitely know what they are inputting. Maybe using an example section in the input:

```python

Example
-------

helcor(20*u.deg, 155*u.deg, ....)```

File: specutils/helcorr.py [hunk 3]
```diff
@@ -0,0 +1,343 @@
...
+        lon = coord.Longitude(lon * u.deg)
+
+    if not isinstance(lat, coord.Latitude):
+        lat = coord.Latitude(lat * u.deg)
+
+    if not isinstance(alt, u.Quantity):
```
wkerzendorf: If we don't require coordinates, the best way to write this is `alt = Quantity(alt, u.m)`

specutils/helcorr.py
```diff
@@ -0,0 +1,343 @@
...
+
+    if not isinstance(ra, u.Quantity):
+        ra *= u.deg
+
+    if not isinstance(dec, u.Quantity):
+        dec *= u.deg
+
+    # Here we specify the location so that we can easily calculate the mean
+    # local siderial time later on
+    time = astropy.time.Time(2.4e6 + mjd, format="jd", location=(lon, lat, alt))
```
wkerzendorf: I think `mjd` is a format that is offered (@mhvk @taldcroft ?)

specutils/helcorr.py
```diff
@@ -0,0 +1,343 @@
...
+        coord.FK5(equinox="J{}".format(epoch)))
+
+    # Convert geodetic latitude into geocentric latitude to correct for rotation
+    # of the Earth
+    dlat = ((-11. * 60. + 32.743) * np.sin(2 * lat) + 1.1633 * np.sin(4 * lat) \
+        - 0.0026 * np.sin(6 * lat)) * u.degree
+    geocentric_lat = lat + dlat / 3600.
+
+    # Calculate distance of observer from Earth center
+    r = alt + 6378160.0 * u.m * (0.998327073 \
```
wkerzendorf: I think, here one should use a constant (I don't know if that's defined). But, for a correct assessment one really should use a geoid (@taldcroft @mhvk @eteq does ERFA provide any of this?)

specutils/helcorr.py
mhvk: Yes!  And as written this would be wrong, since `mjd = jd -2400000.5`

specutils/helcorr.py
mhvk: Yes, the geoid is defined. See http://docs.astropy.org/en/latest/api/astropy.coordinates.EarthLocation.html#astropy.coordinates.EarthLocation
It can also passed on directly to `Time` if that is more helpful (sorry, have not looked at all at the code here, just answering @wkerzendorf's query).

mhvk: More general: I'm pretty sure the whole correction can be calculated using `erfa`. There was some draft code to do this from @Cadair -- see
https://github.com/astropy/astropy/issues/2244#issuecomment-64900470

Cadair: I think you meant to ping @StuartLittlefair

specutils/helcorr.py
wkerzendorf: For ducktyping I would use this piece of code (not doing type checks, such as `type` or `isinstance`:

```python
if hasattr(dje, 'jd'):
    jd = dje.jd
else:
   jd = dje
```

wkerzendorf:
@andycasey For now there are no things left to do from my side, right? Let me know if there's other questions.

andycasey:
Yep, that looks fine; thanks for the feedback @wkerzendorf et al. I will get to this "soon".

wkerzendorf:
you mean "soon"(TM). :wink: . On a more serious note - there's no rush.

rickyegeland:
This functionality looks useful.  What's left to do?

wkerzendorf:
@rickyegeland see the inline comments

crawfordsm:
Any chance this can updated,  @andycasey ?  Or plans to?   This is something I think a number of people would be interested in and a great thing to have a version of even if it eventually gets replaced with something else.

mhvk:
@crawfordsm - the machinery is now in place in `astropy.coordinates.solar_system` to get velocities -- all that is missing is a handy convenience function or method; see https://github.com/astropy/astropy/pull/5231#issuecomment-238963891 (though note that the API was changed slightly; one gets position and velocity using `get_body_barycentric_posvel`). Indeed, I think the main question is where the convenience function should reside: `Time`? `SkyCoord`? Both?

StuartLittlefair:
@crawfordsm in fact, @mhvk is being a bit modest. Thanks to work he's put in for astropy 1.3 this is actually very, very easy. We might get around the difficulty about where to locate a convenience function by simply documenting how to do it. 

With the code @mhvk has added, it is as simple as:

```python
from astropy.time import Time
from astropy.coordinates import SkyCoord, solar_system, EarthLocation

def velcorr(time, coord, observatory):
    # ICRS position and velocity of Earth's geocenter
    ep, ev = solar_system.get_body_barycentric_posvel('earth', time)
    # GCRS position and velocity of observatory
    op, ov = observatory.get_gcrs_posvel(time)
    # ICRS and GCRS are axes-aligned. Can add the velocities
    velocity = ev + ov # relies on PR5434 being merged
    return sc.cartesian.dot(velocity).to(u.km/u.s) # similarly requires PR5434

t = Time('2016-03-20T12:30:00')  # near Spring equinox
obs = EarthLocation.of_site('greenwich')
sc = SkyCoord(['12h00m00s', '18h00m00s'], ['+0d00m00s', '-23d00m00s'])
print(velcorr(t, sc, obs))
[  0.45060386  29.61127524] km / s
```

crawfordsm:
@mhvk  Oh good, I saw that was still on-going but had missed that was merged.   My guess is with the coordinates is probably fine, however the need for the convenience function is critical.

StuartLittlefair:
Following the additions @mhvk has made to astropy 1.3, the convenience function is very simple. 

This [gist](https://gist.github.com/StuartLittlefair/5aaf476c5d7b52d20aa9544cfaa936a1) should cover it. Although you might want to change the function definition, for example to allow the user to set the ephemeris used to calculate the velocities, a la ```light_travel_time```.

In terms of where we put this, just for consistency maybe it ought to be a method of the ```Time``` class, like ```light_travel_time```?

crawfordsm:
Okay, perhaps we should move the discussion of where it should go back to https://github.com/astropy/astropy/issues/3544, where it might be more useful to have it then here (Also yes truly awesome work by @mhvk and others to see this finally get implemented and how relatively few lines of code it takes). 

I'll leave it up to @andycasey if he would like to see this still merged or closed.  I think there is some value in having this in terms of another implementation at the very least for testing/comparison purposes (which will need to be extensive for this functionality) even if it gets replaced quickly by other implementations.

migueldvb:
@andycasey the `helcorr` function in this PR gives the radial velocity correction between topocentric and heliocentric frames, correct? I am comparing the results with the solar_system implementation.

andycasey:
@migueldvb yes, that's right.

Now that there is a better, Correct(tm) way to do barycentric/similar correction in astropy, then I think this should be closed. Part of the motivation for opening this was because I knew this would be something that people would immediately want when switching to astropy, and didn't already exist. At the time there was talk that this could be incorporated within the astropy ecosystem in a better way, but there wasn't a clear path forward. Hopefully this code was useful to some in the meantime.

I'd am going to close this, and recommend that a new PR is opened that expands on @StuartLittlefair's gist. Re-open if there is disagreement! Thanks all

PR #113, opened 2015-01-14
added jinja2 requirement

wkerzendorf:
This should fix the `jinja2` problem with TRAVIS. @keflavich - I will merge if it passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ install:
     - source activate test
 
     # CORE DEPENDENCIES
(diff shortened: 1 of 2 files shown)


wkerzendorf:
@keflavich This time it was not the `modeling` but the new `nddata` framework

wkerzendorf:
@mwcraig Can you have a look if this is how a subclass of `nddata` should behave? It now passes all the tests (at least on my machine).

mwcraig:
@wkerzendorf -- yes, though it looks like there are a couple of mentions of `FlagCollection` you could drop (in an import and in docs).

PR #118, opened 2015-06-01
testing the current build

wkerzendorf:
Testing to see if the current build would succeed with a minimal change (@profjxp let's see what it does)

### DIFF
--- a/specutils/tests/test_spectrum1d.py
+++ b/specutils/tests/test_spectrum1d.py
@@ -7,9 +7,8 @@ from numpy import testing as nptesting
 def test_spectrum1d_fromarray_quantity():
     test_spec = Spectrum1D.from_array(np.arange(3000, 9000) * u.angstrom,
                           np.random.random(6000))
-
+    wavelength = test_spec.wavelength
     assert hasattr(test_spec, 'wavelength')
-
     assert test_spec.dispersion_unit == u.angstrom
 
 def test_spectrum1d_fromarray_quantity2():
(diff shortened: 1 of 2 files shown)


wkerzendorf:
@embray I think something in `astropy.modeling` changed and `specutils` is unhappy.

nden:
@wkerzendorf This is caused by a change in astropy#3702. The evaluate method is not a class method any more. The scaling of input for polynomials was done in the `__call__` method and wasn't happening if they are part of a compound model. Is there a reason you are using `evaluate` and not `__call__` in specutils?

embray:
IIRC specutils current implements its own `__call__` since it also wants to pass a unit argument into the evaluate.  I've suggested that it might just make the unit available as an instance attribute (though I also want to update the `evaluate()` protocol to allow additional arguments (that are not necessarily inputs and/or parameters) to be passed in, similar to the `args` argument accepted by many of the scipy optimizers.

In the meantime though that's fine.  But @nden is right here that assuming `Spectrum1DPolynomialWCS` has `Polynomial1D` as its base class, the evaluate in that is no longer a classmethod due to astropy/astropy@3702.  I'll see about that extra args functionality.

wkerzendorf:
@nden @embray thanks for the explanation! @profjx well that seems to be a problem. I'll try to fix this ASAP. Currently travelling.

wkerzendorf:
@nden @embray that should fix it. Can you guys have a look? In the short/medium term future it would be great to get these spectral WCSes implemented into GWCS.

Issue #11, opened 2013-03-14
How to handle echelle spectra


hamogu:
This issue is here to raise the issue of echelle spectra. They are also 1D spectra in the sense that they only have one axis (wavelength), however, they have multiple orders. They should be kept in one object, as all the echelle order typically have the same metadata (GAIN, OBS-TIME etc.).

We could subclass `Spectrum1D` for that purpose or just allow dispersion and flux to be 2 D arrays and add a method like `flatten(some options here)` to merge all echelle orders.

Clearly `flatten` is a major undertaking that requires many different options and methods (cross-correltate features to improve wavelength solution? How to set the merge points? Rebin to common grid?), but the first step is to decide if echelle spectra need a separate object.

RogueAstro:
Hey folks, shall I unearth this 3-year old issue? I will start working on an echelle spectra extraction code for a new instrument, and I thought this might be a good opportunity to contribute to specutils. Any suggestions on where to start? What about this Fortran code?

wkerzendorf:
@RogueAstro what specifically do you need for your echelle spectra.

RogueAstro:
@wkerzendorf Sorry for not being specific, I later realized that my project probably does not fit well to the objectives of specutils, I think. I will mostly deal with the initial data reduction (calibration and so on), and then later move on to writing a echelle spectral extraction routine.

crawfordsm:
@RogueAstro You may want to check out ccdproc and pyhrs for deailing with reduction of echelle spectra.   I'd be happy to discuss your needs with your data reductions needs with you.

https://github.com/astropy/ccdproc
https://github.com/saltastro/pyhrs

jgussman:
Heyyo! Is there a way to read echelle spectra now that are in fits format now? Or is still the "best" way making a 1Dspectra list? I don't need to reduce the data. 
Thanks

matiruizdiaz:
Hello! I didn't understood at all how to read a echelle spectra. Apparently it is trivial according the comments but there is no way to get a image of my spectra. There is some documentation or anything? Thank you

hamogu:
@matiruizdiaz Thanks for asking. As you may see from the dates and the text above, this is an issue where we discussed how to handle echelle spectra in the code about a decade ago. That's been implemented, and thus the discussion is closed.

General documentation on specutils is here: https://specutils.readthedocs.io/en/stable/
Unfortunately, I can't help you in more detail, because your post does not provide any details of what you want.

[...]

PR #121, opened 2015-08-19
Feature to load IRAF's equispec spectrum

indiajoe:
This is to load fits equispec spectrum.
IRAF's apall outputs spectrum in this format when it stores extracted sky, variance etc. (the extras=yes flag in apall). 
They are 3 dimensional. The first dimension contains flux, sky value, variance etc. Second dimension contains aperture (typically size 1 if it is not multispec). And third dimension is the dispersion axis.

Ideally, I think we would like to have 2 dimensional flux array to represent this kind of spectrum to avoid repetition of meta data (as I understand from the discussion in #11).

As a temporary and quick solution, this PR creates multiple Spectrum1D for each rows of spectrum (similar to current implementation for multispec format).

For a normal equispec spectrum, it returns a list of Spectrum1D.
For a multispec equispec spectrum, it returns a list of list Spectrum1D.


wkerzendorf:
hi @indiajoe looks good. Do you want to add your files as a test?

indiajoe:
@wkerzendorf . Sure, shall I put the fits file in specutils/data/ and add the tests in tests/test_spectrum1d.py ?

wkerzendorf:
sounds great!

wkerzendorf:
@indiajoe hi this looks great - sorry it took such a long time to merge.

PR #129, opened 2016-06-28
Introduce new generic spectrum object

nmearl:
In an effort to begin unifying spectrum utilities, the space telescope development forum has converted their base spectrum objects to a new `Spectrum1D`-based object. This new base object incorporates many of the astropy mixins. We attempted to be as minimally invasive to the current specutils structure as possible.


specutils/core/generic.py
```diff
@@ -0,0 +1,122 @@
...
+        if self._dispersion is None:
+            self._dispersion = np.arange(self.data.size)
+
+            try:
+                crval = self.wcs.wcs.crval[0]
+
+                # RuntimeWarning: cdelt will be ignored since cd is present
+                try:
+                    cdelt = self.wcs.wcs.cd[0][0]
+                except:
```
keflavich: best not to have a generic exception catch here.  I expect this could be an `AttributeError` if `cd` is missing

specutils/core/generic.py
```diff
@@ -0,0 +1,122 @@
...
+                    cdelt = self.wcs.wcs.cdelt[0]
+
+                end = self.data.shape[0] * cdelt + crval
+                num = (end - crval) / cdelt
+
+                # TODO: the values for the keywords are not guaranteed to be
+                #  at the first index
```
keflavich: can use `wcs.sub([WCS_SPECTRAL])` to avoid this, I think (though I may be misinterpreting this comment)

specutils/core/generic.py
```diff
@@ -0,0 +1,122 @@
...
+                if hasattr(self.wcs.wcs, 'ctype') and "log" \
+                        in self.wcs.wcs.ctype[-1].lower():
+                    self._dispersion = np.logspace(crval, end, num)
+                else:
+                    self._dispersion = np.arange(crval, end, cdelt)
+            except:
```
keflavich: ditto above: no generic exceptions

specutils/core/generic.py
```diff
@@ -0,0 +1,122 @@
...
+        return self.multiply(other, handle_meta='first_found')
+
+    def __truediv__(self, other):
+        return self.divide(other, handle_meta='first_found')
+
+    def __len__(self):
+        return len(self.data)
+
+
+if __name__ == '__main__':
```
keflavich: is this needed?

specutils/core/generic.py
nmearl: I think you're interpreting it correctly, but wouldn't the `WCSSUB_SPECTRAL` object reference the spectral axis instead of the dispersion axis?

specutils/core/generic.py
nmearl: Nope -- will remove!

specutils/core/generic.py
keflavich: as far as I understand, the spectral axis is always the dispersion axis - is there a case where that is not true?

nmearl: Mmm... I'm not familiar with the `sub` method, and any attempts to extract the coordinate description given `WCSSUB_SPECTRAL` turns up empty (on various data sets). Maybe this is something to look into in the future.

nmearl:
Is this PR acceptable as is (would you prefer I file an issue to explore more thoroughly the use of `sub`); or, is the idea to wait until that has been done to accept the PR? (I only press the issue because we're sprinting currently with specutils as a dependency.)

keflavich:
@nmearl I'm fine with merging it now.  @wkerzendorf, any comments?

(sorry about the delay; at a conference)

nmearl:
I don't have write access to merge, so please go ahead and merge whenever you feel comfortable.

astrofrog:
@nmearl discussed this on telecon and there are some fixes that are needed first, will comment in detail soon.

specutils/core/generic.py
```diff
@@ -0,0 +1,119 @@
...
+        super(GenericSpectrum1D, self).__init__(flux=data,
+                                                wcs=wcs,
+                                                uncertainty=uncertainty,
+                                                indexer="UNUSED",
+                                                *args, **kwargs)
+        self._dispersion = dispersion
+        self._dispersion_unit = dispersion_unit
+        self.name = name or "New Data Object"
+
+    def from_self(self, copy=True, **kwargs):
```
astrofrog: This should be a class method. I think that in other parts of Astropy we would call this method ``clone``, or ``copy`` - @eteq, what do you think? I wonder if this could instead be ``def copy(self, deepcopy=True, **kwargs)`` and if ``deepcopy`` is ``False`` the arrays themselves are not copied?

specutils/core/generic.py
```diff
@@ -0,0 +1,119 @@
...
+        self_kwargs.update(kwargs)
+
+        return self.__class__(**self_kwargs, copy=copy)
+
+    @classmethod
+    def from_array(cls, data, *args, **kwargs):
+        return cls(name="Array Data", data=data, **args, **kwargs)
+
+    @property
+    def dispersion(self):
```
astrofrog: Could you include a comment as to why you are redefining this property?

specutils/core/generic.py
```diff
@@ -0,0 +1,119 @@
...
+import numpy as np
+import logging
+
+from astropy.nddata import NDArithmeticMixin, NDSlicingMixin, NDIOMixin
+import astropy.units as u
+from astropy.wcs import WCS
+
+
+class GenericSpectrum1D(NDIOMixin, NDSlicingMixin, NDArithmeticMixin,
+                        Spectrum1D):
```
astrofrog: @mwcraig @crawfordsm @MSeifert04 - do you think this is the correct name for this class, or would you suggest something else? (in relation to NDData)

specutils/core/generic.py
wkerzendorf: I think this also sounds very optical focused. I think we had this in one of our earliest drafts.

specutils/core/generic.py
```diff
@@ -0,0 +1,119 @@
...
+                       "uncertainty": self._uncertainty, "mask": self.mask,
+                       "meta": self.meta}
+
+        self_kwargs.update(kwargs)
+
+        return self.__class__(**self_kwargs, copy=copy)
+
+    @classmethod
+    def from_array(cls, data, *args, **kwargs):
+        return cls(name="Array Data", data=data, **args, **kwargs)
```
astrofrog: This appears to bypass all the code in the original ``from_array`` - is this desirable?

specutils/core/generic.py
```diff
@@ -0,0 +1,119 @@
...
+        return self.add(other, handle_meta='first_found')
+
+    def __sub__(self, other):
+        return self.subtract(other, handle_meta='first_found')
+
+    def __mul__(self, other):
+        return self.multiply(other, handle_meta='first_found')
+
+    def __truediv__(self, other):
+        return self.divide(other, handle_meta='first_found')
```
astrofrog: @mwcraig @MSeifert04  - is there a reason these weren't included in ``NDArithmeticMixin``? If so, do the same reasons apply here, or do you think it's fine to have these? I just want to make sure that different packages are consistent in their use of NDData.

specutils/core/generic.py
astrofrog: Actually dispersion does exist in the base class, I'm just curious why it's redefined here.

specutils/core/generic.py
nmearl: Yes. We're not using the custom WCS that the `specutils` package implements, but the Astropy WCS. I also need to include the `name` attribute.

astrofrog:
I think it would be nice for @mwcraig @crawfordsm @eteq to take a look at this before we merge just to make sure it's going in the right direction.

@nmearl - for now, until this is merged, can you include the subclass directly in MOSViz?

specutils/core/generic.py
nmearl: Because dispersion array is being constructed from the WCS information.

astrofrog: I guess what I mean is, is there a reason this should be here and not in the base class? (not saying it should be, just trying to understand if it has limitations)

specutils/core/generic.py
nmearl: `NDArithmeticMixin` does not implement these methods, I would need to override them even if they did exist in order to set the `handle_meta` attribute on the function call that `NDArithmeticMixin` does actually implement.

specutils/core/generic.py
nmearl: Having to implement in the base class would mean overhauling the custom WCS structure, which is way out of scope for what I'm hoping to do right now. When Steven Crawford (@crawfordsm) and I spoke, he proposed a sort of "non-interventionist" approach to making changes to specutils by creating separate, quasi-exclusive subclasses.

specutils/core/generic.py
nmearl: Okay, I'll implement this a static method and just pass instances to it.

specutils/core/generic.py
mwcraig: This name seems fine to me. I don't think there is any prescription for how things that subclass from `NDData` should be named, and **not** incorporating `Data` in the name honestly seems better at this point.

specutils/core/generic.py
mwcraig: [IIRC](http://www.acronymfinder.com/Slang/IIRC.html) this was not implement in the `NDArithmeticMixin` to avoid forcing objects to interact like numpy arrays. When I tried adding these in ccdproc a couple years ago I ran into problems with trying to for the returned object to have the proper class if I, say, added a `Quantity` to a `CCDData` object. 

In `NDArithmeticMixin` the situation would be messier because you don't know a priori what else the objects might be combined with in operations or which type should get priority.

In some cases I could imagine these methods being very handy, but think it makes sense to implement them in derived classes like`GenericSpectrum1D` rather than at a higher level.

mwcraig:
Keeping in mind I've spent a lifetime total of about 10 minutes looking at the codebase for specutils, this PR looks fine. 😄

astrofrog:
@nmearl - can you rebase this to include the fixes in master? (which should make the tests pass)

nmearl:
> @nmearl - for now, until this is merged, can you include the subclass directly in MOSViz?

Just saw this -- sorry! Both MOSViz and the next release of SpecViz are waiting on this PR, so I'd rather not triplicate the effort by wedging this separately in both of those. Waiting isn't a big deal for now.

specutils/core/generic.py
```diff
@@ -0,0 +1,124 @@
...
+from astropy.nddata import NDArithmeticMixin, NDSlicingMixin, NDIOMixin
+import astropy.units as u
+from astropy.wcs import WCS
+
+
+class GenericSpectrum1D(NDIOMixin, NDSlicingMixin, NDArithmeticMixin,
+                        Spectrum1D):
+    """
+    A generalized spectrum data object that implements several `NDData`
+    features such as IO, arithmetic, and slicing.
```
eteq: I think it would be useful to add just a bit more here about the differences.  In particular it should be noted here that the wcs is *different* from the `Spectrum1D` wcs.

specutils/core/generic.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        Override parent method in order to construct dispersion array from
+        Astropy's WCS object instead of the custom specutils WCS object.
+        """
+        if self._dispersion is None:
+            self._dispersion = np.arange(self.data.size)
+
+            if isinstance(self.wcs, WCS):
+                try:
+                    crval = self.wcs.wcs.crval[0]
```
eteq: I'm confused why this is necessary.  Shouldn't it just work to be able to do something like ``self.wcs.all_pix2world(np.arange(len(data)))``?  Isn't that the point of spectral WCS, after all?

(I admit I haven't tested this... but it seems weird to not use the WCS machinery...)

eteq: Or put another way, why not use the machinery described here: http://specutils.readthedocs.io/en/latest/specutils/read_fits.html#reading-simple-linear-1d-wcs ?

eteq: Or alternatively, why not use the machinery described here: http://specutils.readthedocs.io/en/latest/specutils/read_fits.html#reading-simple-linear-1d-wcs ?  That's *not* the same as using the spectral WCS like I'm saying above, but seems more like what this machinery is doing?

eteq:
Ok, @nmearl - I looked over this and most of it looks fine... But see my comment above about the wcs - I don't understand why you're hand-coding the dispersion to work differently from `Spectrum1D` instead of just using the wcs as-is, or using the existing specutils machinery for linear WCS.

There's also the question of the name.  I discussed this a bit with @nmearl out-of-band, but I think it's more that this is ``NDData``-compliant.  That suggests a name like ``NDSpectrum1D``, but of course that's just silly 😉 ... Maybe ``Spectrum1DData``? 


I also think it might make sense down the line to just merge this into ``Spectrum1D``.  But for the purposes of facilitating `specviz`/`MOSViz` releases I think it's find to merge this for now as a separate class and deal with that later before the next specutils release.

specutils/core/generic.py
nmearl: > Or alternatively, why not use the machinery described here:

I don't see the point in this, really. Either we get the relevant information from the wcs, or directly from the fits header. In both cases (theirs and ours), we're doing essentially the same thing, but just that we're not incorporating an object dependency that would otherwise never be used in our case (e.g. `specwcs.Spectrum1DPolynomialWCS`).

But, using `all_pix2world` is definitely an improvement, please let me know if the changes are satisfactory.

nmearl:
@eteq I don't have strong feelings about `Spectrum1DData`, but @mwcraig raised the point:

> This name seems fine to me. I don't think there is any prescription for how things that subclass from NDData should be named, and not incorporating Data in the name honestly seems better at this point. 

Perhaps a bit more brainstorming?

eteq:
`Spectrum1DND`? Seems silly, but it does get across the idea that it's "spectrum1d w/ added NDDAta" capabilities

nmearl:
That might be misleading because `Spectrum1D` is already an `NData` object, `GenericSpectrum1D` is just incorporating a few more bells and whistles.

nmearl:
How about this: Astropy has an `NData` class, that is the basic implementation, but then has another class that incorporates `NDData` and all these mixins called [`NDDataRef`](http://docs.astropy.org/en/stable/api/astropy.nddata.NDDataRef.html#astropy.nddata.NDDataRef). Why don't we, for now, call this spectrum object `Spectrum1DRef` in the same vein?

crawfordsm:
Thanks @nmearl -- this looks really useful!   Now that I have had a chance to sit down, read it, and understand it, it looks okay to me.  However, a few thoughts:
1) Please add some tests
2) Can you leave a comment here for why the change in dispersion and how the WCS is handled (just for future reference)?
3) Can you add some documentation for when to use this vs. Spectrum1D, and how to use this with writing in and out. 

If 3) would hold up the release of other software, I think that could be added later although it would be much better to add that now. 

The `Spectrum1DRef` seems pretty reasonable name.

nmearl:
> Can you leave a comment here for why the change in dispersion and how the WCS is handled

@crawfordsm, do you mean *why the change* in reference to the changes I made today, or *why the change* in comparison with the way `Spectrum1D` does it?

And what do you mean by how the WCS is handled?

crawfordsm:
Why the change as compared to Spectrum1D -- ie why are you overriding the specutils wcs method?

nmearl:
Mainly, because we want to switch to using Astropy's `WCS` object, since, according to APE 7, the `NDData` object will eventually require the `WCS` attribute be an Astropy `WCS` object (or so it was).

Another reason has to do with the fact that in many of our cases, this spectrum object is going to be a product of ifu data cubes, meaning that all the relevant `WCS` information should be able to be parsed and applied on the fly without every interacting directly with a spectrum object. Specutil's WCS object seems to be a very custom way of implementing a transformation based on a model (going so far as to inherit directly from an Astropy `Model` object, which seems a bit extreme). It seems to do this by asking the user to *specify* a particular subclass of `WCS` that, along with being the `WCS` data structure, is also the thing that acts to *interpret* the `WCS` data structure. Perhaps down the road, these two things can be separated, and the Astropy `WCS` object can be adopted as the data structure, and a separate transformation library can be implemented as way to interpret it.

In the same vein, there then should be some kind of parity of functionality between the 2d and 1d data objects that will be associated with each other in the high-level applications we're creating. That is, if WCS transformations for the spatial dimensions of a data cube are implemented as astropy `WCS` functions, then so too should the transformations for the spectral axis. Same notion with having to interact with the data objects' `WCS` information.

eteq:
One bit to add about the Astropy WCS: you shouldn't interpet anything about WCS as being the *current* i.e., "FITS-WCS") WCS object as the end-all be-all thing. There's a desparate need for a new more general WCS scheme (or at least a new standard Python-level *interface*), perhaps along the lines of `gwcs`.  The specutils WCS is actually *closer* to what that might look like (i.e. `gwcs` is based on composing models, which is a much easier-to-maintain approach than the FITS WCS).

Of course that hasn't yet materialized, so I think this makes sense as an interim solution... except that I still don't understand why you're not using http://specutils.readthedocs.io/en/latest/specutils/read_fits.html#reading-simple-linear-1d-wcs  instead?  That seems to accomplish the same aim, and uses the same Astropy WCS object...?

eteq:
Oh, and on the name: ``Spectrum1DRef`` sounds good, @nmearl !  Just be sure to update the class docstring to note what that means (basically just what you said in your comment above).

nmearl:
> except that I still don't understand why you're not using http://specutils.readthedocs.io/en/latest/specutils/read_fits.html#reading-simple-linear-1d-wcs instead?

I think I might be missing something... I don't see how this uses the Astropy `WCS` object? The `Spectrum1DPolynomialWCS` is still a `Model` object, and (as the example shows) requires parsing the fits header directly to get the linear dispersion values where it then constructs the custom `WCS` object (which, internally, is used quite differently from how the `Spectrum1DRef` interacts with its `WCS` object). It isn't that different from what we were doing originally (parsing the values out of the `WCS` object, and then just constructing the polynomial ad hoc), but this method requires bringing in an entirely new object *just* to construct a linear or logarithmic dispersion array, which in this case is stored as a custom `WCS` object which none of the internal of the `Spectrum1DRef` object would consider interacting like this in order to *set* the dispersion attribute. I guess I don't understand how or why we would use it?

mwcraig:
> Why don't we, for now, call this spectrum object Spectrum1DRef in the same vein?

Sounds good!

eteq:
Had some out-of-band discussion with @nmearl about the WCS situation, and will make an issue with a bit more about that and some future directions.  But in the meantime, will merge this now.  Thanks @nmearl

PR #135, opened 2016-07-14
Add new tag for astroconda incorporation

keflavich:
The astroconda maintainers usually use tags when incorporating packages into the anaconda channel. Obviously, the v0.2 tag is outdated with our current changes (not to mention, the fix for `setup.py` is not in the tagged version, meaning it suffers from the python 3.5.2 `ConfigParser()` regression).

A new tag should be created for the current version of specutils (v0.2.1?) encompassing the new Spectrum1DRef object and the `setup.py` fixes.


nmearl:
@keflavich Can you make a tag for this so the astroconda channel can build against it?

keflavich:
@nmearl done.  I had forgotten that the merge didn't include the tag.

PR #151, opened 2017-05-25
WCS wrapper implementation

nmearl:
My approach was to create a meta class that keeps a registry of adapter classes that are defined to be associated with a particular WCS class (see https://github.com/nmearl/specutils/blob/master/specutils/wcs/wcs_adapter.py#L9). 

WCS adapters inherit from an abstract [`WCSAdapter` class](https://github.com/nmearl/specutils/blob/master/specutils/wcs/wcs_adapter.py#L28), and are automatically added to the [meta registry](https://github.com/nmearl/specutils/blob/master/specutils/wcs/adapters/fitswcs_adapter.py). 

Then, I define a [`WCSWrapper` class](https://github.com/nmearl/specutils/blob/master/specutils/wcs/wcs_wrapper.py#L8), that takes any WCS object and finds the associated adapter from the registry.

You can imagine then that in the `Spectrum1D` class, we have a line that simply has `wcs = WCSWrapper(wcs)`, and the returned object would be some adapter that ensures the WCS can interact with the `Spectrum1D` object.

```python
 >>> from specutils.spectra import Spectrum1D
 >>> from astropy.wcs import WCS
 >>> fwcs = WCS()
 >>> spec = Spectrum1D(np.random.sample(100), wcs=fwcs)
 >>> print(type(spec.wcs), type(spec.wcs.axes.spectral))
<class 'specutils.wcs.adapters.fitswcs_adapter.FITSWCSAdapter'> <class 'astropy.wcs.wcs.WCS'>
```

There are parts I'm not quite sure of the best way to approach (e.g. see the [`WCSAxes` namedtuple object](https://github.com/nmearl/specutils/blob/master/specutils/wcs/wcs_adapter.py#L5)).

**tl;dr**: Aside from defining the abstract base class for WCS adapters, it just provides a simple interface to a registry. Any adapter class defined by inheriting from `WCSAdapter` will be added to the registry. When any arbitrary WCS is passed into `Spectrum1D`, the `WCSWrapper` class will look through the registry and return the appropriate adapter WCS object, like the `FITSWCSAdapter` object shown in the example. It’s really just an effort to remove the responsibility for defining an adapter instance before instantiating a `Spectrum1D` object.


specutils/wcs/adapters/fitswcs_adapter.py
```diff
@@ -0,0 +1,31 @@
...
+    WCSSUB_CUBEFACE, WCSSUB_STOKES, WCSSUB_CELESTIAL
+
+
+__all__ = ['FITSWCSAdapter']
+
+
+class FITSWCSAdapter(WCSAdapter):
+    wrapped_class = WCS
+    axes = None
+
```
keflavich: related to the `.system` discussion:

```
def __repr__(self):
      return "< WCS with axes {self.axes.names} and frames {self.ctype} >"
```
where 'axes.names' probably needs some work to infer, but `self.ctype` is almost exactly `self.wcs.ctype`

keflavich: (I have no idea how to do this for gwcs but it should look similar)

specutils/wcs/adapters/fitswcs_adapter.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, wcs):
+        super(FITSWCSAdapter, self).__init__(wcs)
+
+        self.axes = WCSAxes(
+            longitude=self.wcs.sub([WCSSUB_LONGITUDE]),
+            latitude=self.wcs.sub([WCSSUB_LATITUDE]),
+            cubeface=self.wcs.sub([WCSSUB_CUBEFACE]),
+            spectral=self.wcs.sub([WCSSUB_SPECTRAL]),
+            stokes=self.wcs.sub([WCSSUB_STOKES]),
+            celestial=self.wcs.sub([WCSSUB_CELESTIAL])
```
keflavich: In this case, each `axes` object is a complete WCS object of its own.  Is that what we want?  Should these perhaps also be "adapters", e.g. `longitude=FITSWCSAdapter(self.wcs.sub([WCSSUB_LONGITUDE]))`?

keflavich:
I like this framework.  We should keep discussing details as you develop it further.

(the `NamedTuple` link is dead)

nmearl:
@keflavich This is the first-pass implementation of most of the stuff from the APE, and I think it's ready to be merged. The [tests](https://github.com/astropy/specutils/pull/151/files#diff-538a2d2a3e44de04822e1ed05990c13c) cover the basic use cases discussed in the APE. After this PR, I'll open another with the work on the arithmetic, and a final one for ensuring compatibility with collections of spectrum 1d objects.

specutils/wcs/wcs_wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+import astropy.units as u
+from gwcs import coordinate_frames as cf
+
+from .adapters import *
+from .wcs_adapter import WCSAdapter
+
+__all__ = ['WCSWrapper']
+
+
+class WCSWrapper:
```
keflavich: do we need `class WCSWrapper(object):`

nmearl: In python 3, `class WCSWrapper` and `class WCSWrapper` are equivalent. But you're correct for the python 2 case: `object` should be explicitly inherited from.

keflavich: I suppose we're fine with not supporting py2 from the start?

specutils/spectra/spectrum1d.py
```diff
@@ -18,53 +15,52 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
...
+                              "been given. Please provide one.")
+
+        if not isinstance(flux, u.Quantity):
+            flux = u.Quantity(flux, unit=unit or "Jy")
+
+        self._velocity_convention = None
+
+        # Currently, only a fits wcs object stores the rest wavelength or
+        # frequency information in the wcs object. In any other case, the user
+        # will be given an error to provide these.
```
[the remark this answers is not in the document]
nmearl: Oops, technically the comment is out of date. The GWCS object should simply log a warning telling the user to set the `rest_value` explicitly on the `Spectrum1D` object, since the GWCS wrapper provides properties (that simply return 0).

specutils/wcs/wcs_adapter.py
```diff
@@ -0,0 +1,107 @@
...
+            wcs_class = dct.get('wrapped_class', None)
+
+            if wcs_class is not None:
+                logging.info("Added {} to adapter registry.".format(wcs_class))
+                cls.registry[wcs_class] = cls
+
+        super(WCSAdapterMeta, cls).__init__(name, bases, dct)
+
+
+@six.add_metaclass(type('WCSAdapterMetaProxy',
```
Cadair: Given comments below on py2 support, you probably don't need this.

Also, you could go Python 3.6 only and drop the whole metaclass (astropy/astropy#7144)

nmearl: Yes, definitely. I suppose the question is whether to be python 3+ only, or python 3.6+ only?

Cadair: Go 3.6+ people should upgrade 😆

PR #178, opened 2018-05-02
XraySpectrum1D class

eblur:
First pass at XraySpectrum1D class. 

Currently working on example cases to test that the spectrum and response files play nicely.


specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,542 @@
...
+
+    # Convenience function for Xray people
+    @property
+    def counts(self):
+        return self.flux
+
+    def assign_arf(self, arf_inp):
+        """
+        Assign an ARF object to the XraySpectrum1D object
+
```
stscicrawford: Can you spell out the acronym?

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,542 @@
...
+        if isinstance(arf_inp, str):
+            self.arf = ARF(arf_inp)
+        else:
+            self.arf = arf_inp
+        return
+
+    def assign_rmf(self, rmf_inp):
+        """
+        Assign an RMF object to the XraySpectrum1D object
+
```
stscicrawford: Can you spell out the acronym?

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,542 @@
...
+        Modifies the XraySpectrum1D.rmf attribute
+        """
+        if isinstance(rmf_inp, str):
+            self.rmf = RMF(rmf_inp)
+        else:
+            self.rmf = rmf_inp
+        return
+
+    def apply_resp(self, mflux, exposure=None):
+        """
```
stscicrawford: apply_response is likely more clear than 'resp'

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,542 @@
...
+            result = self.rmf.apply_rmf(mrate)
+        else:
+            #print("Caution: no RMF file specified")
+            result = mrate
+
+        return result
+
+## ----  Supporting response file objects
+
+class RMF(object):
```
stscicrawford: Please use the full name for the class

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,542 @@
...
+        # anxillary information
+        hdulist = fits.open(filename)
+        self.filename = filename
+
+        # get all the extension names
+        extnames = np.array([h.name for h in hdulist])
+
+        # figure out the right extension to use
+        if "MATRIX" in extnames:
+            h = hdulist["MATRIX"]
```
stscicrawford: As suggestion, make "MATRIX" a variable here so it isn't hard coded then the user can specify what the keyword for this is.  It could default to MATRIX

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,542 @@
...
+                                                                np.float(source_bin_i)
+                    # iterate the response index for next round
+                    resp_idx += current_num_chans
+
+
+        return counts[:self.detchans]
+
+
+class ARF(object):
+
```
stscicrawford: Please spell out the full name here for the class

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,542 @@
...
+            Average exposure time for the dataset
+            (takes telescope dithering into account)
+        """
+        # open the FITS file and extract the MATRIX extension
+        # which contains the redistribution matrix and
+        # anxillary information
+        hdulist = fits.open(filename)
+        self.filename = filename
+
+        h = hdulist["SPECRESP"]
```
stscicrawford: Have 'SPECRESP' has a keyword option

stscicrawford:
A first pass of some comments, but will have to look at some more.   

Depending on the size of the files, we should use the astropy remote data for the testing.

Having scipy as an optionally dependency is okay.   If it is for tests, that can be added so a test is skipped when scipy is missing.  I can look an example if you need one

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,542 @@
...
+
+    arf
+
+    rmf
+    """
+    def __init__(self, bin_lo, bin_hi, bin_unit, counts, exposure,
+                 arf=None, rmf=None, rest_value=0.0 * u.angstrom, **kwargs):
+        try:
+            axis_unit = u.Unit(bin_unit)
+        except:
```
nmearl: Could you catch the _specific_ error you expect here (`ValueError`, I believe, in the case where the unit cannot be parsed)?

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,542 @@
...
+        bin_mid = 0.5 * (bin_lo + bin_hi) * axis_unit
+        Spectrum1D.__init__(self, spectral_axis=bin_mid, flux=counts,
+                            rest_value=rest_value, **kwargs)
+
+        self.bin_lo = bin_lo
+        self.bin_hi = bin_hi
+        self.exposure = exposure
+        self.assign_rmf(rmf)
+        self.assign_arf(arf)
+        return
```
nmearl: Just a minor thing, the `return` could be removed.

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,542 @@
...
+
+        Returns
+        -------
+        Modifies the XraySpectrum1D.arf attribute
+        """
+        if isinstance(arf_inp, str):
+            self.arf = ARF(arf_inp)
+        else:
+            self.arf = arf_inp
+        return
```
nmearl: (Minor) The `return` is not necessary.

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,542 @@
...
+            result = mrate
+
+        return result
+
+## ----  Supporting response file objects
+
+class RMF(object):
+    def __init__(self, filename):
+        self._load_rmf(filename)
+        pass
```
nmearl: (Minor) `pass` isn't necessary.

specutils/spectra/xrayspectrum1d.py
eblur: I'm spelling out the acronym in the comment string. I suggest we keep the "assign_arf" function name as it is because this is the common command string used across many X-ray spectral analysis tools. It is intuitive to X-ray astronomers.

specutils/spectra/xrayspectrum1d.py
eblur: Same as ARF. Spelling out the acronym in the comment string but not the function call.

specutils/spectra/xrayspectrum1d.py
eblur: The user should not ever have to think about this, and the extension depends on the instrument or data reduction tool used to produce the RMF.

For now, I am adding an "extension" keyword that defaults to None. The `_load_rmf` will check if the user has provided a value, and use it. Otherwise, it searches for "MATRIX" or "SPECRESP MATRIX". If extension=None, and the two common extensions are not found, a message is printed and the function quits.

specutils/spectra/xrayspectrum1d.py
eblur: Pushing back on this gently because it is so standard. I can't think of a reason a user would ever want to change this value. It is so standard, we named the attribute after it.

Pinging @dhuppenkothen for comments, since this portion is widely her doing.

specutils/spectra/xrayspectrum1d.py
eblur: I spelled out the class names to `AreaResponse` and `ResponseMatrix`.

However, these file types are so standard that I urge we keep most of the attribute and function names in the shortened version. X-ray astronomers even refer to them as "arfs" and "rmfs" verbally. I spelled out the acronyms in many of the docstrings to aid in clarity for a new user.

specutils/spectra/xrayspectrum1d.py
eblur: Changed to AreaResponse

eblur:
Review comments that I gave a thumbs up to were implemented without any further comment from me.

@nmearl the response classes are really for loading files that are complex enough that a user would never specify these by hand. How would you recommend initializing the class for a user who is stubborn on faking their own arfs and rmfs? If it's just a few lines of code, it (maybe) can't hurt. (Famous last words?)

@stscicrawford Please send me instructions for where to upload test files. @dhuppenkothen can you send me a link or remind me where we put the PCA and Chandra ACIS test data?  I have only tested this so far on Chandra HETG data.

specutils/spectra/xrayspectrum1d.py
dhuppenkothen: It's been a while since I looked at this, but I have the vague memory that there might be instruments that don't call this `SPECRESP`, in which case having it as a keyword might be helpful?

The code I wrote was never meant to be fully inclusive of every instruments; it was a prototype I wanted to get work on a single type of data, and the possibly extend it to more types of data. As such, don't assume that I've necessarily picked the most general way of implementing things everywhere. :)

dhuppenkothen:
@eblur I thought we'd uploaded the test data, but perhaps not. I can try to dig it out for you?

specutils/spectra/xrayspectrum1d.py
eblur: Ahh, okay. I thought we had tried it on enough telescopes. I will add the keyword and change the attribute name to something informative, like `effective_area`

eblur:
@dhuppenkothen I will do some digging as well

crawfordsm:
Okay just opened #223 to remind myself to set up some instructions.  One question would be how big are the data?    If they are small, they could be added directly here, otherwise they will have to be stored elsewhere.

eblur:
Not particularly small, but not always big. ~10 - 100 MB, I think ...

Proposition: I'll collect the various test files, write the tests, then ping you @crawfordsm (@nmearl ?) Probably by the end of the month.

We could set aside a time to Skype and get everything running together. Should take ~60 minutes (if the `specutils-datasets` repo is not set up yet).

nmearl:
I'm personally not a big of having a separate repository for data. I'd rather use something like [git lfs](https://git-lfs.github.com/), which just requires some simple setup on our side, and contributors don't have to think about submitting data to a separate repo.

nmearl:
> @nmearl the response classes are really for loading files that are complex enough that a user would never specify these by hand. How would you recommend initializing the class for a user who is stubborn on faking their own arfs and rmfs? If it's just a few lines of code, it (maybe) can't hurt. (Famous last words?)

The issue I raised is purely a syntactical python issue. That is, initializing instance variables outside of an `__init__` is generally proscribed. Rather, something like:

```python
class ResponseMatrix(object):
    def __init__(self, filename):
        self._load_rmf(filename)
        self.filename = "" # Initialize filename so that this variables reference exists when the instance is created.

    def _load_rmf(self, filename, extension=None):
        """
        ...
        """
        # open the FITS file and extract the MATRIX extension
        # which contains the redistribution matrix and
        # anxillary information
        hdulist = fits.open(filename)
        self.filename = filename
```

This is not a strict absolute, but if ever a user comes back to this in the future, they won't be aware of what the class variables are without combing through the entire class definition and finding all the `self.` declarations. Also, it's possible to get caught in the situation where the user is asking for a variable that hasn't been initialized yet, but is expected to be a part of the class. E.g. the `filename` attribute -- instead of raising an `AttributeError`, should return `None`, or an empty string if it hasn't been initialized.

crawfordsm:
@eblur If you have data that would be on Zenodo, that would also be okay to include.   We can always decide to copy it to a different location for better access.

nmearl:
@eblur Is there a chance that you can take another look at this soon?

eblur:
@nmearl How soon is soon? I tabled this task for code sprinting at an Astropy coordinating meeting in December. Is there a deadline you're hoping to meet? If so I'll see what I can do.

I'll be traveling Sept 17 - Oct 11, so if urgent, I'll see what I can get done this week.

eteq:
@eblur - we are trying to do a specutils release this week. But the hope is for the next release to be in December, so that's actually well timed for coordination meeting sprinting.  With that in mind I'll re-tag this for the next milestone.

eblur:
Perfect.

crawfordsm:
Alternatively, what is the outstanding work for this?   I think it was mainly testing?   Could this be rebased and part of this release with testing added as an issue to be addressed in the future?  Or is there substantial more work for that is needed?

eblur:
Nope, just tests. And perhaps maybe making it compatible with any of the recent changes from the latest release.

specutils/spectra/xrayspectrum1d.py
hamogu: Going with @eblur : calling is resp is common across the standard tools in X-ray astronomy. That does not mean that we need to keep it, but we should either consistently use what X-ray astronomy tools do, or consistently spell out everything.

eblur:
Hey! This branch has been updated to the latest **specutils/master** and most of the tests are passing (aside from some docstring indentation issues).

@nmearl @hamogu @stscicrawford Can one of you do the final review?

hamogu:
@eblur I appreciate the trust you put in me, put I don't have the authority to do final reviews in specutils.

specutils/tests/test_xrayspectrum1d.py
```diff
@@ -0,0 +1,34 @@
...
+    return
+
+def test_apply_model():
+    # Test that one can evaluate XraySpectrum1D with a model
+    test_spec = test_create_from_arrays()
+    new_model = PowerLaw1D(amplitude=5.e-3, alpha=1.8, x_0=1.e3)
+    model_flux = new_model(test_spec.spectral_axis.value)
+    ymodel = test_spec.apply_response(model_flux)
+    assert len(ymodel) == len(test_spec.spectral_axis)
+    return
```
hamogu: no empty return needed. Unless that's the spectutils style - some specutils maintainer will have to tell us how they want it.

specutils/tests/test_xrayspectrum1d.py
```diff
@@ -0,0 +1,34 @@
...
+    assert len(test_spec.bin_hi) == len(test_spec.spectral_axis)
+    return
+
+def test_apply_model():
+    # Test that one can evaluate XraySpectrum1D with a model
+    test_spec = test_create_from_arrays()
+    new_model = PowerLaw1D(amplitude=5.e-3, alpha=1.8, x_0=1.e3)
+    model_flux = new_model(test_spec.spectral_axis.value)
+    ymodel = test_spec.apply_response(model_flux)
+    assert len(ymodel) == len(test_spec.spectral_axis)
```
hamogu: I would feel better if this not only tested that the calculation works, but also that the numbers are right. Can you add a 
`ymodel[234] = 234` just for one or two bins? Either get the number from a known good program (ISIS is clearly absolutely 100% bug free people in my corridor tell me) or at the very least from a case you looked at by hand that looks good so that we notice if some other change affects these numbers.

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+        return u.eV
+    if unit_string in KEV:
+        return u.keV
+    if unit_string in ANGS:
+        return u.angstrom
+
+class XraySpectrum1D(Spectrum1D):
+    """
+    Spectrum properties specific to X-ray data
+
```
hamogu: Really: Binned counts spectra. Could be a Gamma ray spectrum... but that's just a side-note.

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+    Parameters
+    ----------
+    bin_lo : numpy.ndarray
+        The left edges for bin values
+
+    bin_hi : numpy.ndarray
+        The right edges for bin values
+
+    bin_unit : string OR astropy.units.Unit
```
hamogu: Should this be 
```
`astropy.units.Unit`
```
so that sphinx build the links to the astropy docs?

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+        Unit for the bin values
+
+    counts : astropy.Quantity array
+        Counts histogram for the X-ray spectrum
+
+    exposure : float
+        Exposure time for the dataset
```
hamogu: Since exposure is a number and not a quantity, what's the unit?

specutils/spectra/xrayspectrum1d.py
hamogu: Same is true for all later docs, in particular for those that relate to other classes in the same file, e.g arf or rmf.

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+    arf : specutils.AreaResponse or string, default None
+        Strings will be passed to AreaResponse.__init__
+        All other input types are stored as arf attribute
+
+    rmf : specutils.ResponseMatrix or string, default None
+        Strings will be passed to ResponseMatrix.__init__
+        All other input types are stored as rmf attribute
+
+    rest_value : astropy.units.Quantity, default 0 Angstrom
+        See Spectrum1D rest_value input
```
hamogu: Again, perfect for 
```
See `~specutils.spectra.Spectrum1D`
```
(or whereever Spectrum1D is defined. I don't know on top of my head).

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+
+    bin_hi
+
+    counts
+
+    exposure
+
+    arf
+
+    rmf
```
hamogu: I suggest list formatting with a an explanation:
```
       - bin_lo : Lower bound of spectral bin (np.array)
       - bin_hi : Upper bounds of spectral bins (np array)
```
etc.

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+    """
+    def __init__(self, bin_lo, bin_hi, bin_unit, counts, exposure,
+                 arf=None, rmf=None, rest_value=0.0 * u.angstrom, **kwargs):
+        try:
+            axis_unit = u.Unit(bin_unit)
+        except ValueError:
+            axis_unit = _unit_parser(bin_unit)
+
+        bin_mid = 0.5 * (bin_lo + bin_hi) * axis_unit
+        Spectrum1D.__init__(self, spectral_axis=bin_mid, flux=counts,
```
hamogu: perfect use case for `super().__init__(...)`

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+        """
+        Assign an area response file (ARF) object to the XraySpectrum1D object
+
+        Input
+        -----
+        arf_inp : string
+            File name for the area response file (FITS file)
+
+        Returns
+        -------
```
hamogu: No "Returns" block. This function does not return anything.

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+        Assign a response matrix file (RMF) object to the XraySpectrum1D object
+
+        Input
+        -----
+        rmf_inp : string
+            File name for the response matrix file (FITS file)
+
+        Returns
+        -------
+        Modifies the XraySpectrum1D.rmf attribute
```
hamogu: No "returns" block.

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+
+        If no ARF file exists, it will return the model flux after applying the RMF
+        If no RMF file exists, it will return the model flux after applying the ARF (with a warning)
+        If no ARF and no RMF, it will return the model flux spectrum (with a warning)
+        """
+
+        if self.arf is not None:
+            mrate  = self.arf.apply_arf(mflux, exposure=exposure)
+        else:
+            #print("Caution: no ARF file specified")
```
hamogu: Remove print statements.

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+        return
+
+    def apply_response(self, mflux, exposure=None):
+        """
+        Given a model flux spectrum, apply the response. In cases where the
+        spectrum has both an area response file (ARF) and a response matrix
+        file (RMF), apply both. Otherwise, apply whatever response is in RMF.
+
+        The model flux spectrum *must* be created using the same units and
+        bins as in the ARF (where the ARF exists)!
```
hamogu: I wonder if there is a way to enforce that somehow. Seems like a classic point of failure.

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+        # find the place where the tlmin keyword is defined
+        t = np.array(["TLMIN" in k for k in keys])
+
+        # get the index of the TLMIN keyword
+        tlmin_idx = np.hstack(np.where(t))[0]
+
+        # get the corresponding value
+        tlmin = np.int(list(hdr.items())[tlmin_idx][1])
+
+        return tlmin
```
hamogu: I'm confused. Can't you get the same output as this function simply with `return int(h.header['TLMIN'])`?

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+
+        # find all non-zero groups
+        nz_idx = (n_grp > 0)
+
+        # stack all non-zero rows in the matrix
+        matrix_flat = np.hstack(matrix[nz_idx])
+
+        # stack all nonzero rows in n_chan and f_chan
+        #n_chan_flat = np.hstack(n_chan[nz_idx])
+        #f_chan_flat = np.hstack(f_chan[nz_idx])
```
hamogu: I think this should either be used (if it should be used) or deleted.

specutils/spectra/xrayspectrum1d.py
```diff
@@ -0,0 +1,573 @@
...
+        then need to loop over the number of channel sets. For each
+        channel set, we look up the first channel into which flux
+        will be distributed as well as the number of channels in the
+        group. We then need to also loop over the these channels and
+        actually use the corresponding elements in the redistribution
+        matrix to redistribute the photon flux into channels.
+
+        All of this is basically a big bookkeeping exercise in making
+        sure to get the indices right.
+
```
hamogu: In fact, it's so painful as a bookkeeping exercise that I won't do it by hand now. That's why it's so important to test with at least one real-world rmf in the tests below.

eblur:
Replaced this PR with #427

Issue #179, opened 2018-05-03
Unable to load FITS files with Linear Solution


simontorres:
# how I'm reading
I'm trying to load sample fits files.

```python
import os
from specutils import Spectrum1D

path = '/home/simon/development/astropy/specutils/specutils/tests/data'

test_file = 'L5g_0355+11_Cruz09.fits'

spec1d = Spectrum1D.read(os.path.join(path, test_file), format='wcs1d-fits')


# This works well
print(spec1d.flux)

# this fails
print(spec1d.wavelength)
```

# The error message

```ipython
In [38]: ccd.wavelength
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-38-f28127121b9c> in <module>()
----> 1 ccd.wavelength

~/anaconda3/envs/astroconda/lib/python3.5/site-packages/specutils/spectra/spectrum1d.py in wavelength(self)
    117     @property
    118     def wavelength(self):
--> 119         return self.spectral_axis.to(u.AA, u.spectral())
    120 
    121     @property

~/anaconda3/envs/astroconda/lib/python3.5/site-packages/astropy/utils/decorators.py in __get__(self, obj, owner)
    709                 return val
    710             else:
--> 711                 val = self.fget(obj)
    712                 obj.__dict__[self._key] = val
    713                 return val

~/anaconda3/envs/astroconda/lib/python3.5/site-packages/specutils/spectra/spectrum_mixin.py in spectral_axis(self)
     59         # TODO: Should applying the spectral axis unit occur in the adapter?
     60         spectral_axis = self.wcs.pixel_to_world(
---> 61             np.arange(self.flux.shape[0])) * self.wcs.spectral_axis_unit
     62 
     63         return spectral_axis

~/anaconda3/envs/astroconda/lib/python3.5/site-packages/specutils/wcs/adapters/fitswcs_adapter.py in pixel_to_world(self, pixel_array)
     40         Method for performing the pixel to world transformations.
     41         """
---> 42         return self.axes.spectral.all_pix2world(pixel_array, 0)[0]
     43 
     44     @property

~/anaconda3/envs/astroconda/lib/python3.5/site-packages/astropy/wcs/wcs.py in all_pix2world(self, *args, **kwargs)
   1278 
   1279     def all_pix2world(self, *args, **kwargs):
-> 1280         return self._array_converter(
   1281             self._all_pix2world, 'output', *args, **kwargs)
   1282     all_pix2world.__doc__ = """

[...]

PR #184, opened 2018-05-03
added unittest for Spectrum1D.read

simontorres:
I wrote a test that checks that `Spectrum1D.read ` is working properly it asserts.

related to issue #179

### DIFF
--- a/specutils/tests/test_spectrum1d.py
+++ b/specutils/tests/test_spectrum1d.py
@@ -121,8 +121,8 @@ def test_read_linear_solution():
     assert isinstance(spec, Spectrum1D)
 
     assert isinstance(spec.flux, u.Quantity)
-    assert isinstance(spec.wavelength, u.Quantity)
+    assert isinstance(spec.spectral_axis, u.Quantity)
 
     assert spec.flux.size == spec.data.size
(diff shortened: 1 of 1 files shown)


specutils/tests/test_spectrum1d.py
```diff
@@ -110,3 +111,18 @@ def test_create_with_uncertainty():
...
+def test_read_linear_solution():
+    file_path = os.path.join(os.path.dirname(__file__),
+                             'data/L5g_0355+11_Cruz09.fits')
+
+    spec = Spectrum1D.read(file_path, format='wcs1d-fits')
+
+    assert isinstance(spec, Spectrum1D)
+
+    assert isinstance(spec.flux, u.Quantity)
+    assert isinstance(spec.wavelength, u.Quantity)
```
nmearl: I think we should check this with `spec.spectral_axis`, because using `.wavelength`, `.frequency`, `.energy` will attempt to put the dispersion solution through astropy quantities, which itself should be a different test.

(I say this because there's no `spectral_axis_unit` in the test data, so `spec.wavelength` fails because it cannot convert to a length physical type.)

PR #189, opened 2018-05-03
Wcs wcs cunit

simontorres:
Apparently there was a problem with spec_axis being 1-based instead of 0, with this it works now.


nmearl:
Good catch. This has to do with the way the WCS slicing works. I think a better fix is #190, because it fixes it at the source. You may want to remove your index fix, but keep the test.

simontorres:
Excelent! It works now. I think the test is from #184  and I propagated it to here because I didn't branch from master. Should I close this pull request?

PR #212, opened 2018-05-29
Unit Conversion Patch and Tests

robelgeda:
Simple tests to check if unit conversion is functioning properly. 
Patches:
- `isinstance(unit, u.Unit)` -> `isinstance(unit, u.UnitBase)`
- Move the location of tests. Remove a folder I created for testing and merge the new tests into the already existing `/tests/test_spectrum1d.py`.

Flux Unit Tests:
- Check what unit is assigned when flux units are not specified.
- Simple Conversion from `Jy` to `uJy`.
- Raises an error when trying to convert to incompatible units.
- Pass custom equivalencies.
- Check if suppressing the unit conversion works

Spectral Axis Unit Tests:
- Check what unit is assigned when spectral units are not specified.

CC: @nmearl @brechmos-stsci @eteq


specutils/wcs/adapters/gwcs_adapter.py
```diff
@@ -79,7 +79,7 @@ def rest_wavelength(self):
     def with_spectral_unit(self, unit, rest_value=None, velocity_convention=None):
         """
         """
-        if isinstance(unit, u.Unit) and unit.is_equivalent(
```
nmearl: What failure are you seeing using `isinstance(unit, u.Unit)`?

robelgeda: for example, `isinstance(u.m, u.Unit)` will return false.

PR #228, opened 2018-06-08
SpectrumCollection object implementation

nmearl:
This is the result of discussions on an implementation of a spectrum collection object in which users add several spectra with differing dispersion axes to a container object that then allows analysis over the entire set.

The use case involved specifically the use of STIS echelle data. It became clear that data added to the spectrum had to be implicitly deposited onto a grid using several different possible methods (e.g. using the largest delta dispersion, or smallest, or a user-defined grid).

The object behaves as a would a `Spectrum1D` object, allowing for its use in analysis functions that expect a `Spectrum1D` object.

### DIFF
--- a/specutils/tests/test_analysis.py
+++ b/specutils/tests/test_analysis.py
@@ -73,7 +73,7 @@ def test_snr_multiple_flux(simulated_spectra):
     """
 
     np.random.seed(42)
-    
+
     #
     #  Set up the data and add the uncertainty and calculate the expected SNR
     #
@@ -140,7 +140,7 @@ def test_snr_two_regions(simulated_spectra):
     #
 
     regions = [SpectralRegion(0.52*u.um, 0.59*u.um), SpectralRegion(0.8*u.um, 0.9*u.um)]
-    
+
     #
     #  Set up the data
     #
@@ -155,8 +155,8 @@ def test_snr_two_regions(simulated_spectra):
     spec_snr_expected = []
     for region in regions:
 
-        l = np.nonzero(wavelengths>region.lower)[0][0]
-        r = np.nonzero(wavelengths<region.upper)[0][-1]
+        l = np.nonzero(wavelengths >= region.lower)[0][0]
+        r = np.nonzero(wavelengths <= region.upper)[0][-1]
 
         spec_snr_expected.append(np.mean(flux[l:r] / (uncertainty.array[l:r]*uncertainty.unit)))
 
--- a/specutils/tests/test_slicing.py
+++ b/specutils/tests/test_slicing.py
@@ -1,8 +1,5 @@
 import astropy.units as u
-import astropy.wcs as fitswcs
-import gwcs
 import numpy as np
-import pytest
 from numpy.testing import assert_allclose
 
 from ..spectra.spectrum1d import Spectrum1D
@@ -15,7 +12,6 @@ def test_spectral_axes():
     sliced_spec1 = spec1[0]
 
     assert isinstance(sliced_spec1, Spectrum1D)
-    assert sliced_spec1.wcs == spec1.wcs
     assert_allclose(sliced_spec1.wcs.pixel_to_world(np.arange(10)), spec1.wcs.pixel_to_world(np.arange(10)))
 
     flux2 = np.random.sample((10, 49)) * 100
@@ -26,6 +22,52 @@ def test_spectral_axes():
     sliced_spec2 = spec2[0]
 
     assert isinstance(sliced_spec2, Spectrum1D)
-    assert sliced_spec2.wcs == spec2.wcs
     assert_allclose(sliced_spec2.wcs.pixel_to_world(np.arange(10)), spec2.wcs.pixel_to_world(np.arange(10)))
     assert sliced_spec2.flux.shape[0] == 49
+
+
+def test_slicing():
+
+    # Create the initial spectrum
+    spec = Spectrum1D(spectral_axis=np.arange(10) * u.um, flux=2*np.arange(10)*u.Jy)
+
+    # Slice it.
+    sub_spec = spec[4:8]
+
+    # Check basic spectral_axis property
+    assert sub_spec.spectral_axis.unit == u.um
+    assert np.allclose(sub_spec.spectral_axis.value, np.array([4, 5, 6, 7]))
+    assert np.allclose(sub_spec.flux.value, np.array([8, 10, 12, 14]))
+
+    assert sub_spec.wavelength.unit == u.AA
+    assert np.allclose(sub_spec.wavelength.value, np.array([40000., 50000., 60000., 70000.]))
+
+    assert sub_spec.frequency.unit == u.GHz
+    assert np.allclose(sub_spec.frequency.value, np.array([74948.1145, 59958.4916, 49965.40966667, 42827.494]))
+
+    # Do it a second time to confirm the original was not modified.
+    sub_spec2 = spec[1:5]
+
+    # Check basic spectral_axis property
+    assert sub_spec2.spectral_axis.unit == u.um
+    assert np.allclose(sub_spec2.spectral_axis.value, np.array([1, 2, 3, 4]))
+    assert np.allclose(sub_spec2.flux.value, np.array([2, 4, 6, 8]))
+
+    assert sub_spec2.wavelength.unit == u.AA
+    assert np.allclose(sub_spec2.wavelength.value, np.array([10000., 20000., 30000., 40000.]))
+
+    assert sub_spec2.frequency.unit == u.GHz
+    assert np.allclose(sub_spec2.frequency.value, np.array([299792.458, 149896.229,  99930.81933333,  74948.1145]))
+
+    # Going to repeat these to make sure the original spectrum was
+    # not modified in some way
+    assert spec.spectral_axis.unit == u.um
+    assert np.allclose(spec.spectral_axis.value, np.array(np.arange(10)))
+    assert np.allclose(spec.flux.value, np.array(2*np.arange(10)))
+
+    assert spec.wavelength.unit == u.AA
+    assert np.allclose(spec.wavelength.value, np.array(10000*np.arange(10)))
+
+    assert sub_spec.frequency.unit == u.GHz
+    assert np.allclose(sub_spec.frequency.value, np.array([74948.1145, 59958.4916, 49965.40966667, 42827.494]))
+
--- a/specutils/wcs/adapters/gwcs_adapter.py
+++ b/specutils/wcs/adapters/gwcs_adapter.py
@@ -1,6 +1,9 @@
 import logging
+import copy
+
 from gwcs.wcs import WCS
 import astropy.units as u
+from astropy.modeling import models
 
 from ..wcs_adapter import WCSAdapter
 
@@ -30,6 +33,44 @@ class GWCSAdapter(WCSAdapter):
         if unit is not None and unit.is_equivalent(self._unit,
                                                    equivalencies=u.spectral()):
             self._unit = unit
+
+    def __getitem__(self, item):
+        """
+        This is a bit of a hack in order to fix the slicing of the WCS
+        in the spectral dispersion direction.  The NDData slices properly
+        but the spectral dispersion result was not.
+
+        There is code slightly downstream that sets the *number* of entries
+        in the dispersion axis, this is just need to shift to the correct
+        starting element.
+
+        When WCS gets the abillity to do slicing then we might be able to
+        remove this code.
+        """
+
+        # Create shift of x-axis
+        if isinstance(item, int):
+            shift = item
+        elif isinstance(item, slice):
+            shift = item.start
+        else:
+            raise TypeError('Unknown index type {}, must be int or slice.'.format(item))
+
+        # Create copy as we need to modify this and return it.
+        newwcs = copy.deepcopy(self)
+
+        if shift == 0:
+            return newwcs
+
+        shifter = models.Shift(shift)
+
+        # Get the current forward transform
+        forward = newwcs._wcs.forward_transform
+
+        # Set the new transform
+        newwcs.wcs.set_transform(newwcs._wcs.input_frame, newwcs._wcs.output_frame, shifter|forward)
+
+        return newwcs
 
     def world_to_pixel(self, world_array):
         """


bsipocz:
@nmearl - #227 has dealt with the docs failures, so they should go away once you rebase this.

profxj:
Will checkout the SpectrumCollections class later.  

Let me only add right now that the resampling can and
should avoid the ugly `for loop` that it is using.  

See:

https://github.com/linetools/linetools/blob/master/linetools/spectra/utils.py

lines 197-355

specutils/spectra/spectrum_collection.py
```diff
@@ -9,6 +9,9 @@
 
 __all__ = ['SpectrumArray', 'SpectrumCollection']
 
+__doctest_skip__ = ['specutils.spectra.spectrum_collection.ResampleMixin', 
+                    'astropy.nddata.mixins.ndio.NDIOMixin']
```
bsipocz: why do you need to list the astropy class here?

nmearl: I'm not sure. The failing doc test was raising it as a warning (`/home/travis/build/astropy/specutils/docs/api/specutils.SpectrumCollection.rst:6: WARNING: py:class reference target not found: astropy.nddata.mixins.ndio.NDIOMixin`). It's still doing it, so this didn't really do anything (I'll be reverting that commit).

bsipocz: ohh, right. So you can skip these by creating a `docs/nitpick-exceptions` file, and list this there as 
`py:class astropy.nddata.mixins.ndio.NDIOMixin`. Reason is some non straightforward inheritances, apparently the issue didn't solve itself over the years.

Example is in e.g ccdproc, https://github.com/astropy/ccdproc/pull/221

bsipocz: (Also, doctests are run during the normal pytest test runs but not during the docs building job. The latter though is picky about links, and bases classes are usually picked up by their full path rather than by the simplest namespace where their docs page is generated. Ideally this is something the astropy intersphinx mapping should deal with, but apparently it doesn't)

docs/spectrum_collection.rst
```diff
@@ -0,0 +1,102 @@
...
+
+.. code:: python
+
+    >>> print(spec1.flux.shape)
+    (25,)
+    >>> print(spec_coll.flux.shape)
+    (2, 24)
+
+The items stored in the :class:`~specutils.SpectrumCollection` object are the
+*original* input spectra. Iterating over a :class:`~specutils.SpectrumCollection`
```
keflavich: capslock

docs/spectrum_collection.rst
```diff
@@ -0,0 +1,102 @@
...
+
+    >>> print(spec1.flux)  # Input spectrum values for flux
+    >>> print(spec_coll.flux[0])  # Output spectrum values for flux
+
+
+Defining output grids
+---------------------
+
+Output grids can be defined one of three ways: as a string indicating whether
+to use the finest or coarsest sampling from the input spectra; an explicitly
```
keflavich: explicit (drop "ly")

docs/spectrum_collection.rst
```diff
@@ -0,0 +1,102 @@
...
+array of values to use as the output; or as a three-tuple indiciating the
+start bin, end bin, and bin size.
```
keflavich: Is there room for a special case `'same'`, which would raise an exception *unless* all input spectra are on identical grids?  That would be more efficient for some cases, and I expect those cases to be reasonably common

specutils/analysis/resample.py
```diff
@@ -0,0 +1,219 @@
...
+
+    resamp_mat = resamp_mat / delta_fin
+
+    return resamp_mat
+
+
+def _nonuniform_matrix(orig_lamb, fin_lamb, extrapolate=False):
+    """
+    Compute re-sampling matrix R_o2r, useful to convert a spectrum sampled at
+    wavelengths lorig to a new grid lresamp. Here, there is no necessity to
```
keflavich: there are variables referred to here that are not in the code.  Presumably `lorig`->`orig_lamb`?

specutils/analysis/resample.py
```diff
@@ -0,0 +1,219 @@
...
+    have constant gris as on :func:`_uniform_matrix`. This is adapted from code
```
keflavich: gris->grids?

specutils/analysis/resample.py
```diff
@@ -0,0 +1,219 @@
...
+    lr_low = np.zeros(len(fin_lamb))
+    lr_low[1:] = (fin_lamb[1:] + fin_lamb[:-1]) / 2
+    lr_low[0] = fin_lamb[0] - (fin_lamb[1] - fin_lamb[0]) / 2
+
+    lr_upp = np.zeros(len(fin_lamb))
+    lr_upp[:-1] = lr_low[1:]
+    lr_upp[-1] = fin_lamb[-1] + (fin_lamb[-1] - fin_lamb[-2]) / 2
+
+    # Iterate over resampled fin_lamb vector
+    for i in range(len(fin_lamb)):
```
keflavich: could you use a >=2 character iteration variable here?

specutils/analysis/resample.py
```diff
@@ -0,0 +1,219 @@
...
+            aux = 0
+
+            d_lr = lr_upp[i] - lr_low[i]
+            d_lo = lo_upp[j] - lo_low[j]
+            d_ir = lo_upp[j] - lr_low[i]  # common section on the right
+            d_il = lr_upp[i] - lo_low[j]  # common section on the left
+
+            # Case 1: resampling window is smaller than or equal to the
+            # original window. This is where the bug was: if an original bin
+            # is all inside the resampled bin, then all flux should go into it,
```
keflavich: Is this a comment ported from the old code?

specutils/spectra/spectrum_collection.py
```diff
@@ -0,0 +1,180 @@
...
+        if isinstance(grid, np.ndarray):
+            resample_grid = grid
+        elif isinstance(grid, list):
+            resample_grid = np.array(grid)
+        elif isinstance(grid, tuple):
+            resample_grid = np.arange(*grid)
+        else:
+            if isinstance(grid, str) and len(self._items) > 0:
+                start_bin = min([x.spectral_axis[0] for x in self._items]).value
+                end_bin = max([x.spectral_axis[-1] for x in self._items]).value
```
keflavich: Note that these are the 'centers' of the start/end bins.  You probably instead want something like:
```
start_bin = min([x.spectral_axis[0]-np.abs((x.spectral_axis[1]-x.spectral_axis[0])/2) for x in self._items]).value
```
to get the edges.  This is important since you're optionally changing the bin size below.

specutils/spectra/spectrum_collection.py
```diff
@@ -0,0 +1,180 @@
...
+                if grid == 'coarse':
+                    largest_bin = max([np.max(np.diff(x.spectral_axis))
+                                    for x in self._items])
+                    resample_grid = np.arange(
+                        start_bin, end_bin, largest_bin.value) * largest_bin.unit
+                elif grid == 'fine':
+                    smallest_bin = min([np.min(np.diff(x.spectral_axis))
+                                        for x in self._items])
+                    resample_grid = np.arange(
+                        start_bin, end_bin, smallest_bin.value) * smallest_bin.unit
```
keflavich: Following previous commend, this would then be
```
np.arange(start_bin+smallest_bin.value/2, end_bin+smallest_bin.value, smallest_bin.value)
```
or similar.  (note the `+` in `end_bin` is to make the last bin inclusive)

specutils/spectra/spectrum_collection.py
```diff
@@ -0,0 +1,180 @@
...
+    @output_grid.setter
+    def output_grid(self, value):
+        """
+        If the output grid value is changed, force the `SpectrumCollection` to
+        re-evaluate the resampled grid.
+        """
+        self._output_grid = value
+        self._evaluate_grids()
+
+    def _evaluate_grids(self):
```
keflavich: This seems potentially expensive.  I haven't thought through it, but can this be cached...?

stis_specutils_loader.ipynb
```diff
@@ -0,0 +1,514 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "WARNING:root:GWCS does not store rest frequency information. Please define the rest value explicitly in the `Spectrum1D` object.\n",
```
keflavich: If this notebook is going in the repo, it would be helpful to catch & silence these warnings.

nmearl: Oops, it's not supposed to be there. It was test notebook from the STIS team.

docs/spectrum_collection.rst
nmearl: I'm not sure I understand this comment? 😅

keflavich: originaL -> original

nmearl: I think I might be blind... ? It all looks lowercase to me.

docs/spectrum_collection.rst
nmearl: Certainly. I supposed in this case, then, it's just a matter of defining the resample grid as the unique values of the concatenated spectral axes.

specutils/analysis/resample.py
nmearl: Good catch.

specutils/analysis/resample.py
nmearl: Yes, removing.

docs/spectrum_collection.rst
keflavich: this... might be something weird about github then?

image


you see that L at the end?  But I think you're right, it's all correct in the text...

specutils/spectra/spectrum_collection.py
nmearl: `_evaluate_grids` should only be called in a few cases:

1. Then the `SpectrumCollection` is created
2. When an item has been added, removed, or replaced in the `SpectrumCollection`.

Those seem like reasonable times to re-evaluate the grid?

keflavich: Yes.  It shows up in many lines of code, so I wasn't sure if there were more cases.

specutils/analysis/resample.py
nmearl: How do you mean? Only iterate if the final grid has more than a single point?

docs/spectrum_collection.rst
nmearl: Super weird. Might be a type hinting artifact?

eteq:
@nmearl - can you clarify why this includes the resampling?  I think that's a separate effort that we need to consider, because there's very much *not* one way to do resampling.

nmearl:
Well, we need resampling in some form for the STIS team to find anything useful about `specutils` (at least for the use cases they've been asking me about), as well as for the splice implementation. I think, as it stands the formalism is fine and should not defunct future efforts.

To be honest, I don't think a `SpectrumCollection` object is even useful without resampling. That is, what's the point and/or how would it even work?

eteq:
@nmearl well the discussion around splice has made it clear that this is *not* sufficient for some of those use cases. And the implementation here is much harder to generalize than the one we've been discussing there, which uses reampking *functions* with specific algorithm/unit choices. I can go into more on that if you want but I'm trying to keep the discussion sane by not mixing that with the `SpectralCollection` *data structure* (which is what the title implies this PR is about).

So while I might be convinced this is the way to go, that will take time, abd I don't want that to derail the other stuff here.

nmearl:
@eteq I'd prefer you did. I'm not exactly following, so if you could expound a bit on these resampling functions that would not be able to follow the formalism of the `Resample` class shown here (i.e. instantiation with a wanted output spectral axis; call method with input spectrum axis). Not to mention that the title of this PR (and its description) certainly detail the implicit necessity of rebinning the data in the collection.

If we don't want resample in this, I'd then rather just wait until we've decided on a resampling formalism before doing the `SpectrumCollection` class. Otherwise, we may as well close this and start over since resampling is taken as a core aspect of the collection.

eteq:
@nmearl - I've tried to summarize more specific thoughts on this in more specific issues so we can keep the discussion sane (and some of it is more general).  In particular, https://github.com/astropy/specutils/issues/255#issuecomment-409686415 (and I think @SaOgaz is  on this so that we don't mix things too much #255), and #266.  In the language from #266 I *think* you've implemented something that's sort of case 4 as a wrapper around case 2? (can discuss in more detail there).


One additional point I'm confused about now that I've looked more closely at the code, @nmearl : why does `SpectrumCollection` *require* resampling?  I understand the point that if you have multiple spectra of different lengths they need to be resampled, but the raise d'etre of `SpectralCollection` is for cases where they already *are* square: chief among those cases are fiber spectrographs where each spectrum is the same number of *pixels* but with different wavelength solutions (similar deal for multi-slit spectrographs or echelle spectra with multiple orders).  For all of those resampling is neither required nor desirable...?

Relatedly: it looks to me like the core data in `SpectrumArray` (and therefore `SpectrumCollection`)  is a *list* of 1D arrays instead of a 2D array.  That seems to defeat the point of having the `SpectrumCollection` as a "performance shortcut"? Or am I missing something?

profxj:
I encourage you to look at David Kirkby's `speclite` and also
my implementation in `XSpectrum1D`  (about to become XSpectra1D
as I start refactoring to use the new `Spectrum1D`).

nmearl:
The list is not used in operations concerning the `SpectrumCollection`, for the obvious performance reasons. The list architecture is to ensure we get useful dunder methods that allow slicing/inserting/deleting from the collection. Further, lists allow inhomogeneous generic objects, since an `NDData` is not inherently a numpy array. The result of these methods, however, are ndarrays which themselves would be used for performant analytic operations.

Second, it seems maybe I was trying to incorporate use cases that clearly aren't immediately desired. For instance, getting statistical information by using a spectral region on a `SpectrumCollection` would require that there be a single dispersion solution that the provided bounds would use to get the appropriate region from each spectrum in the collection. I assumed that there _needed_ to be some type of rebinning for operations that treat the entire `SpectrumCollection` as a single object.

If we're just building toward the more singular case of spectra with square shapes for use in operations that treat values in the collection as unrelated along the dispersion axis (until they are explicity resampled), then certainly, the resampling portion can be removed from this PR.

PR #248, opened 2018-07-18
Added in SNR of spectrum and excise of region

### LINKED ISSUE #239: Signal to noise measurement
Calculate the signal to noise of the input spectrum.

brechmos-stsci:
This is the first pass at the spectral SNR calculation.  The SNR can be calculated based on the uncertainties in the spectrum (default) or by specifying a 'noise region' in the spectrum (from which the standard deviation of the flux will be calculated). 

This also enables SNR calculate over the whole spectrum or over a region (or regions) specified by region 'tuples'.

Included are tests for the new functionality.

Fixes #239


brechmos-stsci:
@eteq Could you please look through this PR. It has the spectral region and will be the basis for a few of the other stats implementations.

specutils/utils/__init__.py
```diff
@@ -1 +1 @@
-
+from .spectral_region import SpectralRegion
```
eteq: I would say `SpectralRegion` belongs in `spectrum`, *not* in utils - this is a key concept for interfacing with spectra, after all.  And it should be importable directly as `from specutils import SpectralRegion` since it's a basic user interaction.

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+
+    def __init__(self, lower, upper):
+        """
+        Lower and upper values for the interval.
+
+        Parameter
+        --------
+
+        lower: number with ``astropy.units`` unit
```
eteq: Here (and in upper) I think you want to say:
```
lower : Scalar `~astropy.units.Quantity` with pixel or ``spectral_axis`` units
```
That is, a "number with ``astropy.units`` unit" *is* a `Quantity` 

Alternatively, you could do:
```
lower : Scalar `~astropy.units.Quantity` 
     The lower bound of the region. Should have units of `~astropy.units.pixel` or units that are convertable to match the `spectral_axis`.
```

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+           The lower bound of the region.
+
+        upper: number with ``astropy.units`` unit
+           The upper bound of the region.
+
+        Notes
+        -----
+
+        The ``astropy.units`` for each do not have to be the same, but to retain
+        sanity it might be easiest to have them the same.  The will need to be
```
eteq: I don't think you want to say the "retain sanity" part - there's no reason *not* to do this, and for some science cases it may be useful. (e.g., a cosmic ray where you know the location in pixels, but the other side you want your standard wavelength window for the line you're measuring).

Also I'd say the "dispersion as opposed to flux" should go in the description of the parameter as I noted above.

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+        dispersion units as opposed to flux units.
+        """
+
+        if not isinstance(lower, u.Quantity):
+            raise TypeError('Lower bound of the region must have an astropy.units unit')
+
+        if not isinstance(upper, u.Quantity):
```
eteq: Instead of doing the type checking here *and* in the setter, I think you can get away with just doing:
```
self.lower = lower
self.upper = upper
```
in the ``__init__`` and relying on the setter to do the checking?

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+    @property
+    def lower(self):
+        return self._lower
+
+    @lower.setter
+    def lower(self, value):
+        """
+        Lower bound for the interval.
+
+        Parameter
```
eteq: The docstring here isn't visible, I believe, because it's only the first docstring that is considered the "property" docstring.

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+
+    In the future, there might be more functionality added in here and there
+    is some discussion that this might/could move to astropy.region.
+    """
+
+    def __init__(self, lower, upper):
+        """
+        Lower and upper values for the interval.
+
+        Parameter
```
eteq: ``Parameter`` -> ``Parameters`` - I think the docs turn out wrong if you don't do that.

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    SpectralRegion is currently a container class enables some simplicty
+    in defining and passing a region (interval) for a spectrum.
+
+    In the future, there might be more functionality added in here and there
+    is some discussion that this might/could move to astropy.region.
+    """
+
+    def __init__(self, lower, upper):
+        """
```
eteq: The usual convention we follow in astropy is to put the initializer arguments (i.e., the paramters section) in the *class* docstring.  That makes it more visible to the user because the average user doesn't know about (not care) that ``__init__`` is where the parameters are defined.  Also the documentation machinery doesn't recognize it if it's in the ``__init__`` docstring. (I think?)

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+
+        self._lower = value
+
+    @property
+    def upper(self):
+        return self._upper
+
+    @upper.setter
+    def upper(self, value):
+        """
```
eteq: (same as for ``lower`` re the docstring) and unit-checking

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+
+        Parameter
+        --------
+
+        value: number with ``astropy.units`` unit
+           The lower bound of the region.
+        """
+
+        if not isinstance(value, u.Quantity):
+            raise TypeError('Lower bound of the region must have an astropy.units unit')
```
eteq: You also want to check that the units are valid units.  E.g., they must be either `pixel`, or *if* the `spectral_axis` has physical units, it has to be something that can be converted to that using the `spectral` equivalency.  I think the easiest way to do this is:
```
if not (value.unit.is_equivalent(u.pixel)  or
           value.unit.is_equivalent(u.angstrom, equivalencies=u.spectral)):
    raise u.UnitError('Not a spectral region unit')
```
You could potentially also drop the type error with this, because if a user passes in a non-unit object, it will still raise an error that says ``does not have unit``, basically.  Arguably that's better because it's duck-typing, so a quantity-like thing that's not literally a quantity will still work... but your call which you think is safer.

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+        value: number with ``astropy.units`` unit
+           The upper bound of the region.
+        """
+
+        if not isinstance(value, u.Quantity):
+            raise TypeError('Upper bound of the region must have an astropy.units unit')
+
+        self._upper = value
+
+    def excise(self, spectrum):
```
eteq: Hmm... I don't like this name, because if you look at my py-in-the-sky document you'll see I used `excise` to mean something subtly different: getting a new spectrum like the old one but where the regions have been excised.  That is, returning what's *left* instead of what's excised.

Maybe `extract`? Or `extract_from_spectrum`?

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        Excise a spectral region from the spectrum.
```
eteq: As per my comment above, I think this needs to be a smidge more verbose since "excise" is an ambiguous term.

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+
+        Parameters
+        ----------
+        spectrum: ``specutils.spectra.spectrum1d.Spectrum1D``
```
eteq: Use `~specutils.spectra.spectrum1d.Spectrum1D` here.  double-backticks make it a "code snippet", while single-backticks make it a link to that object's API doc page.  The ``~`` makes it appear as "Spectrum1D" instead of "specutils.spectra.spectrum1d.Spectrum1D" (which is desirable because we're hiding the exact location of the class from the user)

eteq: Note: same goes for other places where this idiom is used - I saw it a few other places but don't want to fill everywhere.

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+            The spectrum object from which the region will be excised.
+
+        Return
+        ------
+        spectrum: ``specutils.spectra.spectrum1d.Spectrum1D``
+            Excised spectrum.
```
eteq: Again, a bit more detail needed: something like "a new spectrum" covering only the range from ``lower`` to ``upper`` (there double-backticks are correct because you're referencing a name as-used-in-code rather than something in the documented API)

specutils/analysis/snr.py
```diff
@@ -0,0 +1,106 @@
...
+    # List of regions
+    elif isinstance(region, list):
+
+        snrs = []
+        for reg in region:
+            snrs.append(_snr(spectrum, region=reg, noise_region=noise_region))
+        return snrs
+
+
+def _snr(spectrum, region=None, noise_region=None):
```
eteq: Maybe call this `_snr_single_region` or something?  At first I was confused why it's separted from `snr`

specutils/analysis/snr.py
```diff
@@ -0,0 +1,106 @@
...
+    Notes
+    -----
+    The spectrum will need to have the uncertainty defined in order
+    for the SNR to be calculated.
+
+    Two methods implemented:
+      1) SNR based on uncertainty
+          This will compute the mean of the flux / uncertainty.
+
+      2) SNR based on noise_region
```
eteq: I don't think this function should implement the `noise_region` option.

I agree the functionality should exist, though.  So I think this function should only do option 1 but at the same time there should be a function `uncertainty_from_noise_region`.  That would take a region and yield a *new* spectrum with the uncertainty set by that region, which the user would then pass to this function to acheive what you've got here.  That would also be more performant for the list-of-regions case because the noise calculation need only be done once.  If you agree with this,  @brechmos-stsci, feel free to add that function here, or in a follow-on PR.

The argument being that we need to resist having a few functions that do a bunch of things if the attitude for the analysis tools is a "toolbox" - in a tool box you have lots of tools to do a specific job well, not a single swiss army knife (obviously this is not as clear-cut as that, but still).

specutils/analysis/snr.py
```diff
@@ -0,0 +1,106 @@
...
+    # Single region
+    elif isinstance(region, SpectralRegion):
+        return _snr(spectrum, region=region, noise_region=noise_region)
+
+    # List of regions
+    elif isinstance(region, list):
+
+        snrs = []
+        for reg in region:
+            snrs.append(_snr(spectrum, region=reg, noise_region=noise_region))
```
eteq: minor:  this would be *slightly* faster to do as a list comprehension.  I.e.:
``return [_snr(spectrum, region=reg, noise_region=noise_region) for reg in region]``

specutils/tests/test_analysis.py
```diff
@@ -4,8 +4,12 @@
 import numpy as np
 import pytest
 
+from ..tests.spectral_examples import simulated_spectra
```
eteq: This is *only* a "scalar" spectrum, right?  I.e. one spectral axis and one flux?  We do not this function to work for the one spectral axis, many-flux case... But perhaps it's best to get this in first and then worry about that?  (Or maybe @nmearl can advise on whether it still requires decisions to be made in `Spectrum1D`?)

specutils/analysis/snr.py
brechmos-stsci: I would suggest that both "versions" fit well in the `snr()` method. The method `snr` does one thing, it calculates the S/N.  The methods to define which uncertainty to use is very well separated by `noise_region` being None or not.  

Is the `uncertainty_from_noise_region()` method that  creates a new spectrum going to be used a lot outside of `snr()`?  If not, then I think it makes more sense to have the `noise_region` parameter for `snr()`.

So, I think it is a simpler, cleaner implementation to have `noise_region` as a parameter. But, I will do what you guys wish as you might have other thoughts of how this all fits together.  Maybe ping someone else for their opinion on this topic specifically?

specutils/utils/spectral_region.py
brechmos-stsci: Fixed this will be in next merge

specutils/utils/spectral_region.py
brechmos-stsci: changed and will be in next merge

specutils/utils/spectral_region.py
brechmos-stsci: should be corrected now, will be in next merge

specutils/utils/spectral_region.py
brechmos-stsci: I removed the note altogether as it really isn't needed.

specutils/analysis/snr.py
brechmos-stsci: renamed it, in next merge

specutils/analysis/snr.py
brechmos-stsci: modified. In next merge.

specutils/utils/spectral_region.py
brechmos-stsci: more info was added and will be in the next merge

specutils/utils/spectral_region.py
```diff
@@ -0,0 +1,116 @@
...
+import astropy.units as u
+
+
+class SpectralRegion:
+    """
+    SpectralRegion is currently a container class enables some simplicty
+    in defining and passing a region (interval) for a spectrum.
+
+    In the future, there might be more functionality added in here and there
+    is some discussion that this might/could move to astropy.region.
```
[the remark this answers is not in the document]
brechmos-stsci: added link to the package, will be in the next merge

specutils/utils/spectral_region.py
brechmos-stsci: added more detail, in next merge

specutils/tests/test_analysis.py
brechmos-stsci: Yes, this is for a scalar spectrum only. I understood the many-flux case is not implemented yet.  If hte many-flux case is fully implemented then I should change this, if not, then I'll get this in first and more changes could be done in another PR.

specutils/utils/spectral_region.py
brechmos-stsci: I'll ck

brechmos-stsci:
To be more clear. Everywhere I said merge above, I meant push.  My head is in vacation mode.

specutils/analysis/snr.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # If a noise_region is defined then calculate the standard deviation
+    # in the noise_region and use that as the uncertainty.
+    if noise_region is not None:
+        spectrum_noise_region = noise_region.extract(spectrum)
+        uncertainty = np.std(spectrum_noise_region.flux)
+
+    # If noise_region is not defined then use the Uncertainty in the spectrum.
+    else:
+        uncertainty = calc_spectrum.uncertainty.array * spectrum.uncertainty.unit
```
eteq: I think this should be a bit more loose in its definition of uncertainty.  I think a lot of users will provide a "quantity-like" uncertainty in many cases, at least at this stage in the game (when the current uncertainty machinery is a bit limited).  So something like the following, following general duck-typing rules:
```
if hasattr(calc_spectrum.uncertainty, 'array'):
    uncertainty = calc_spectrum.uncertainty.array * spectrum.uncertainty.unit
else:
    uncertainty = calc_spectrum.uncertainty
```

specutils/spectra/spectral_region.py
```diff
@@ -0,0 +1,152 @@
...
+    A `SpectralRegion` is a container class enables some simplicty
+    in defining and passing a region (interval) for a spectrum.
+
+    In the future, there might be more functionality added in here and there
+    is some discussion that this might/could move to
+    `Astropy Regions <http://astropy-regions.readthedocs.io/en/latest/>`_.
+    """
+
+    @staticmethod
+    def from_center(center=None, width=None):
```
eteq: I would think this should be a *class* method, rather than a static method, is this is basically what classmethod was invented for.  That is:
```
@classmethod
def from_center(cls, center=None, width=None):
    ...
    return cls(center - width, center + width)
```
that way sub-classes still get the correct `from_center` behavior.

specutils/spectra/spectral_region.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class SpectralRegion:
+    """
+    A `SpectralRegion` is a container class enables some simplicty
+    in defining and passing a region (interval) for a spectrum.
+
+    In the future, there might be more functionality added in here and there
+    is some discussion that this might/could move to
+    `Astropy Regions <http://astropy-regions.readthedocs.io/en/latest/>`_.
```
eteq: docstring for the ``__init__`` should go here rather than in ``__init__`` (and a corresponding convention is to put ``__init__`` at the top, before ``from_center`` so that "it's" docstring is as close as possible

specutils/spectra/spectral_region.py
```diff
@@ -0,0 +1,152 @@
...
+        SpectralRegion class method that enables the definition of a `SpectralRegion`
+        from the center and width rather than lower and upper bounds.
+
+        Parameters
+        ----------
+
+        center: Scalar `~astropy.units.Quantity` with pixel or ``spectral_axis`` units
+           The center of the spectral region.
+
+        Width: Scalar `~astropy.units.Quantity` with pixel or ``spectral_axis`` units
```
eteq: (typo): ``Width``-> ``width``

specutils/spectra/spectral_region.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @lower.setter
+    def lower(self, value):
+        """
+        Lower bound for the interval.
+
+        Parameters
+        ----------
+
+        value: Scalar `~astropy.units.Quantity` with pixel or ``spectral_axis`` units
```
eteq: I may have used words like this before, but reading this now I think it might be confusing to users to say `` ` ` spectral_axis` ` units``. That is, they might not understand what that means exactly.  Maybe something like "any unit that is a valid unit for a ``spectral_axis``"?

specutils/tests/test_analysis.py
eteq: I think it's fine to leave addressing the many-fluxes one-dispersion case for a later PR.  Might be worth putting in the docstrings a note that this only works for the single-flux case, though?

specutils/analysis/snr.py
eteq: The point isn't so much the implemenation - I agree it's straightforward and *could* make sense here.  It's that I think the typical user thinks of the uncertainty in a spectrum as being a part of the spectrum, *not* a thing you calculate as part of the operation of computing the S/N.  In fact, in a lot of UV/optical/IR spectra, you get incorrect results from choosing a noise region like this because the uncertainty varies *a lot* across the spectrum (sky lines are the most relevant thing here, but so is the contribution to the uncertainty from, say a line itself). So in my mind having the ``noise_region`` favors a *particular* way of computing the noise, which is *known* to be wrong for one of the most astronomical use cases.

> Is the uncertainty_from_noise_region() method that creates a new spectrum going to be used a lot outside of snr()?

Yes, I would say so - if you aren't just stopping at computing `snr` (which most users aren't), the uncertainty will also be used in lots of other ways - nearly every analysis task.  E.g., weighting for resampling or line-fitting, judging if the line fluxes measured from a flux measurement are significant, etc.  So if the user *does* need this particular algorithm, they'll *need* `uncertainty_from_noise_region` anyway for many of their other analysis tasks.

specutils/analysis/snr.py
```diff
@@ -0,0 +1,103 @@
+import numpy as np
+from ..spectra import SpectralRegion
+
+__all__ = ['snr']
+
+
+def snr(spectrum, region=None, noise_region=None):
+    """
+    Calculate the mean S/N of the spectrum based on the flux and uncertainty in the spectrum. This will
```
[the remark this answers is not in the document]
brechmos-stsci: Corrected. will be in next push.

specutils/spectra/spectral_region.py
brechmos-stsci: how about '... or spectral axis unit from `~astropy.units`' ?  (the suggested alternatively would make the line very long)

specutils/tests/test_analysis.py
brechmos-stsci: @nmearl Could you point me to the tests for many fluxes, one dispersion in specutils?  I would like to see how it is defined and used.

specutils/analysis/snr.py
brechmos-stsci: This now assumes uncertainty is set so no assignment is necessary here.

specutils/spectra/spectral_region.py
eteq: Well, there's nothing in `astropy.units` that's specific to spectral axes... Maybe just ".. or any valid ``spectral_axis`` unit" ? That's only slightly longer,  (FWIW, this is a situation where I generally let the 80-char limit go by the wayside, since it's better to be specific enough that the user understands)

specutils/analysis/snr.py
```diff
@@ -0,0 +1,81 @@
...
+    Calculate the mean S/N of the spectrum based on the flux and uncertainty
+    in the spectrum. This will be calculated over the regions, if they
+    are specified.
+
+    Parameters
+    ----------
+    spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
+        The spectrum object overwhich the equivalent width will be calculated.
+
+    region: `~specutils.utils.SpectralRegion` or list of `~specutils.utils.SpectralRegion`
```
eteq: Something occurs to me here: as a convenience for users, perhaps we should also allow "anything that can be converted to a `SpectralRegion`?  That is, if you just pass it through the `SpectralRegion` initializer as ``SpectralRegion(*region), the user can do ``snr(spec, (12*u.pixel, 24*u.pixel)`` or similar, and have it work.  If you like I can just make a PR against your branch to look at, as I was testing with it locally anyway...

eteq: In some out-of-band discussion, @brechmos-stsci and I discussed and agreed this could be a follow-on PR

eteq:
I implemented the small thing I mentioned above myself but otherwise, good to go.  Merging!

PR #272, opened 2018-08-10
Spectrum1D slicing bug fix

### LINKED ISSUE #271: Spectral_axis/wavelength property confusing after slicing
I believe there appears to be a bug when slicing a spectrum object.  For example, if I create a spectrum as:

```
import numpy as np
import astropy.units as u
from specutils.spectra import Spectrum1D

np.random.seed(42)

spec = Spectrum1D(spectral_axis=np.arange(10) * u.AA, flux=np.random.sample(10)*u.Jy)
print(spec.spectral_axis)
print(spec.flux)
```
then the output is (expected):
```
[0. 1. 2. 3. 4. 5. 6. 7. 8. 9.] Angstrom
[0.37454012 0.95071431 0.73199394 0.59865848 0.15601864 0.15599452
 0.05808361 0.86617615 0.60111501 0.70807258] Jy
```
Now, when I slice the spectrum using the `[]` and look at the output:
```
>>> sub_spec = spec[4:8]
>>> sub_spec.spectral_axis
 <Quantity [0., 1., 2., 3.] Angstrom>
>>> sub_spec.wavelength
 <Quantity [0., 1., 2., 3.] Angstrom>
>>> sub_spec.flux
<Quantity [0.15601864, 0.15599452, 0.05808361, 0.86617615] Jy>
```

The flux sub spectrum output (bottom line) is correct, or at least expected (ie it **is** the 4-8 elements). BUT, the spectral_axis and wavelength lines (5th and 3rd last line) is not correct, or at least what I would have expected. It shows [0, 1, 2, 3] and I was expecting `[4, 5, 6, 7] Angstrom`.

The wcs appears to be copied over as I can do:
```
>>> sub_spec.wcs.world_to_pixel(np.array(1*u.AA))
1.0
```

but I don't think that helps me.

So, I think there is a bug in the slicing or at the very least it isn't clear (to me).

@eteq @nmearl @crawfordsm

brechmos-stsci:
There appeared to be a bug in the slicing of Spectrum1D objects. The functionality relied on NDData, which all worked, but the `spectrum.spectral_axis` was not updated correctly. 

To fix this, I copied out the spectral_axis, applied the super() methods and then set the spectral_axis back again.  There might be a more efficient way of doing it, feel free to suggest (this works).

Tests were added for slicing to check for errors. The tests check `spectrum.spectral_axis`, `spectrum.wavelength` and `spectrum.frequency`.

Fixes #271


specutils/spectra/spectrum1d.py
```diff
@@ -194,6 +194,27 @@ def __truediv__(self, other):
...
+        __setitem__ and __delitem__ are not duplicated in this way
+        as Spectrum1D does not support item assignment.
+        """
+
+        # Get a copy of the current spectral axis
+        spectral_axis = self.spectral_axis
+
+        # Call the sub-class (in particular, here call the NDData which
+        # does most fo the work.
+        self = super().__getitem__(key)
```
eteq: Generally it's not a good idea to overwrite the `self` variable name - you might just use `other`.  And if you do that there's no need to "save" `spectral_axis` above - you can just do ``other.spectral_axis = self.spectral_axis[key]`` below.

eteq:
See brechmos-stsci/specutils#1 for the PR to this branch I mentioned in https://github.com/astropy/specutils/pull/272#issuecomment-412181275

specutils/spectra/spectrum1d.py
brechmos-stsci: No problem on all, there are lots of subtleties here (at least to me).

nmearl:
@brechmos-stsci the mixin _does_ have a `__getitem__`, more specifically, the `NDSlicingMixin` class which handles slicing the wcs, uncertainties, and the mask.

There are two issues at play here: 1) FITS WCS should support slicing. The reason this still fails on a FITS WCS is because the `wcs` attribute on the `Spectrum1D` class is a reference to an adapter class instance, not the actual FITS WCS instance, a `spec.wcs.wcs[2:4]`, for instance, will return the appropriately sliced FITS WCS slice.

2) The GWCS class does not support slicing in the normal sense. I believe in previous discussions I've had, it's going to take an implementation on the adapter class to sort out the slicing, with the GWCS using something like the [region selector](https://gwcs.readthedocs.io/en/latest/#module-gwcs.selector).

nmearl:
This should not be implemented on the `Spectrum1D` class, since the slicing is handled by the `NDData` mixin. This is particular to the wcs, and should be handled there.

brechmos-stsci:
This PR is almost ready. The current bugs in the checks are related to https://github.com/astropy/specutils/issues/287.  I am going to try and fix that first, then will re-push here for a clean test run.

brechmos-stsci:
@eteq @nmearl This should now be ready for review. Basically it just adds a slicer into the gwcs object. I talked with Nadia about the way the slicing was done on the wcs model and she seemed good with how I am using Shift.

specutils/tests/test_slicing.py
```diff
@@ -15,7 +12,7 @@ def test_spectral_axes():
     sliced_spec1 = spec1[0]
 
     assert isinstance(sliced_spec1, Spectrum1D)
-    assert sliced_spec1.wcs == spec1.wcs
+    #assert sliced_spec1.wcs == spec1.wcs
```
eteq: Probably should just delete these instead of commenting?

brechmos-stsci: 👍

brechmos-stsci:
If this is good with everyone, it would be great to have it merged as other test failures will be fixed by this....

nmearl:
@brechmos-stsci Can you make a test for the case of using a fits wcs object. Just to note that slicing should _not_ work with a fits wcs right now since `spec.wcs` returns a `FITSAdapter`, not a `FITSWCS` object (as would be returned via `spec.wcs.wcs`).

Most likely, you'll want to implement the `__getitem__` method for the `FITSAdapter` as well, though this should be just a relay to the stores `spec.wcs.wcs` objects slicing.

eteq:
@brechmos-stsci - @nmearl and I were just taking about this out-of-band - I'm going to go ahead and merge this since it works as-is for anyone with a `gwcs`-based WCS, and @nmearl will make a separate issue about the FITS case.

brechmos-stsci:
Oh, originally, I understood from @nmearl that only the GWCS adapter was broken. I can take a look at the FITS one too.

PR #276, opened 2018-08-17
Return quantity objects in pix2world and world2pix transformations

nmearl:
From a conversation with @eteq, `pixel_to_world` should return a `Quantity` object instead of the `Quantity` being created when the `spectral_axis` is evaluated in the `Spectrum1D` object.

### DIFF
--- a/.eggs/README.txt
+++ b/.eggs/README.txt
@@ -1,6 +1,0 @@
-This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
-
-This directory caches those eggs to prevent repeated downloads.
-
-However, it is safe to delete this directory.
-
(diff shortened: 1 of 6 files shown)


eteq:
(note the test failure looks real, @nmearl , so I'm waiting on that to review this)

nmearl:
@eteq Do you know if `world_to_pixel` return a `Quantity` with no unit (according to APE14)?

eteq:
@nmearl - as I mentioned yesterday out-of-band, the canonical place to look is the implementation for FITS-WCS.  That's astropy/astropy#7325 + astropy/astropy#7326 . There it turns out the pixel coordinates are *not* `Quantity` objects, rather raw arrays.

.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
```
eteq: This file should *not* be included.  In 9448fe2fc7f282afed471b4afe67e42c86b8f859 I added `.eggs` to gitignore, but in the meantime this should not be checked in in this PR.

eteq:
LGTM now - thanks @nmearl !

Issue #264, opened 2018-07-30
Drop `unit` and `spectral_axis_unit` keywords to `Spectrum1D` initializer?


eteq:
The title basically says it all.  I can implement this in a PR myself but wanted to first make sure I'm not missing something critical.

To justify it: I'd say the ``unit`` and ``spectral_axis_unit`` keywords are superfluous - if you want a unit on either you need to provide a `Quantity`, which can actually be created in *fewer* characters than ``unit=`` or ``spectral_axis_unit=`` (all that's needed is `` * `` or ``.to``).  Is there some important use case I'm missing by saying that?

(Note also that I think there might be some inconsistencies in what the docstring says and how it actually behaves if you give both a `Quantity` and a ``unit`` keyword.  But I'll only bother to sort that out if the conclusion is that these keywords should *stay* since it's a moot point otherwise.)

cc @nmearl @keflavich @crawfordsm

keflavich:
So you're saying you want:
```
spectral_axis = np.linspace(...)
flux = np.random.randn(...)
Spectrum1D(spectral_axis=spectral_axis, flux=flux, spectral_axis_unit='km/s', unit='Jy')
```
to not be supported, instead requiring
```
spectral_axis = np.linspace(...)*u.km/u.s
flux = np.random.randn(...)*u.Jy
Spectrum1D(spectral_axis=spectral_axis, flux=flux)
```

correct?

I'm mildly opposed to this as I prefer to have more flexibility in the specification.

eteq:
@keflavich - yes that is correct.  My argument is that there is *no* added flexibility in the current situation, but there is  added *complexity*.  That is, compare the following cases (the first is your  first example - current behavior - and the latter is what I'm proposing be standard):
```
Spectrum1D(spectral_axis=spectral_axis, flux=flux, spectral_axis_unit='km/s', unit='Jy')

Spectrum1D(spectral_axis=spectral_axis*u.km/u.s, flux=flux*u.Jy)
Spectrum1D(spectral_axis=u.Quantity(spectral_axis, 'km/s'), flux=u.Quantity(flux, 'Jy'))
```
the latter two convey the same information as the former but with fewer characters *and* without us having to document for the user what to do when the units and various other inputs conflict. As I said before there's already a subtle/non-trivial bug (I think) in that, demonstrating how difficult it might be both to keep it consistent and communicate with users.

[...]

PR #277, opened 2018-08-17
Remove unit and spectral_axis_unit from initializer

nmearl:
This PR removes unnecessary keywords from the `Spectrum1D` initializer as discussed in #264.


eteq:
Tests failed due to the doctests failing. Turns out there were some corrections needed there due to this PR anyway...

nmearl:
I cannot figure out why [this is failing](https://travis-ci.org/astropy/specutils/jobs/427365884#L1283). In previous iterations, it was failing because travis could not import matplotlib for the plotting directives in the docs. I added matplotlib as a dependency in the travis yaml file, but still erroring...

eteq:
Aha, I know at least why the error is only happening in those two cases: #310 added those tests partly because they address a bug fixed in astropy master that prevents the `.rst` files from being run as doctests.  So those two builds are the ones where the doctests get run, and this error comes from running the doctests (i.e. `getting_started.rst`).

The error *itself* is still mysterious to me though - I can't reproduce it locally even if I do `python setup.py test -t docs/getting_started.rst` (which does indeed run the part that's erroring on Travis)

eteq:
Oh, I think I figured it out: matplotlib is trying to pop up a window on travis, which fails because travis runs "headless" (unlike my local testing machine).  A work-around is to just force matplotlib into "agg" mode.  The commit I just pushed up *should* do just that (although we shall see if the tests actually pass...)

nmearl:
@eteq Looks like your changes were the solution. 👍

eteq:
Awesome, LGTM so merging!

PR #279, opened 2018-08-20
Weirdness with multidimensional Spectrum1D

drdavella:
The documentation says that I should be able to have multiple spectra that all use the same spectral axis in a `Spectrum1D` object. Naively, I would expect to be able to do this:

```python
frequencies = np.linspace(1, 10, 20) * u.um
flux = np.ndarray((3, 20)) * u.Jy
s1 = Spectrum1D(spectral_axis=frequencies, flux=flux)
```

However, the following assertions will fail:

```python
assert frequencies == s1.spectral_axis
assert len(frequencies) == s1.spectral_axis
```

This is highly unintuitive.

Doing the following *will* preserve the dimensions of the spectral axis:

```python
frequencies = np.linspace(1, 10, 20) * u.um
flux = np.ndarray((3, 20)) * u.Jy
s1 = Spectrum1D(spectral_axis=frequencies, flux=flux.T)
```

But this doesn't make any sense to me. It also has the consequence of making internal computations have to perform some unnecessary contortions.

Is this a bug or am I just misunderstanding how this is supposed to work?

### DIFF
--- a/specutils/analysis/__init__.py
+++ b/specutils/analysis/__init__.py
@@ -2,4 +2,4 @@ from __future__ import absolute_import
 
 from .equivalent_width import *  # noqa
 from .snr import snr  # noqa
-from .width import sigma # noqa
+from .width import sigma_full_width # noqa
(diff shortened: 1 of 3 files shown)


drdavella:
Sorry, this should be an issue, not a PR. I had a momentary lapse of reason.

Issue #268, opened 2018-08-03
Revamp equivalent_width to use spectral region


eteq:
Now that #248 is in we have a SpectrumRegion class which should be used with the `equivalent_width` function.

Although it might also be worth waiting on this issue until #266 is resolved since then both can be done at once.

Issue #280, opened 2018-08-20
Weirdness with multidimensional Spectrum1D


drdavella:
The documentation says that I should be able to have multiple spectra that all use the same spectral axis in a `Spectrum1D` object. Naively, I would expect to be able to do this:

```python
frequencies = np.linspace(1, 10, 20) * u.um
flux = np.ndarray((3, 20)) * u.Jy
s1 = Spectrum1D(spectral_axis=frequencies, flux=flux)
```

However, the following assertions will fail:

```python
assert frequencies == s1.spectral_axis
assert len(frequencies) == s1.spectral_axis
```

This is highly unintuitive.

Doing the following *will* preserve the dimensions of the spectral axis:

```python
frequencies = np.linspace(1, 10, 20) * u.um
flux = np.ndarray((3, 20)) * u.Jy
s1 = Spectrum1D(spectral_axis=frequencies, flux=flux.T)
```

But this doesn't make any sense to me. It also has the consequence of making internal computations have to perform some unnecessary contortions.

Is this a bug or am I just misunderstanding how this is supposed to work?

brechmos-stsci:
confirmed, at least here, and I agree it is a kind of odd way to define it for things to work out, but I might be missing something too.

@eteq @stscicrawford  @nmearl

Issue #266, opened 2018-08-01
Consistent style of API for analysis and manipulation tools


eteq:
The proximate cause of this is discussion is design considerations for several features (#224, #228, #255), particularly an in-person discussion related to #255 (involving @hcferguson @SaOgaz @crawfordsm and myself).  But it leads to a broader item: we should provide a consistent way of implementing/exposing the API the analysis and manipulation tools so that users and devs don't get interface whiplash (a la the matplotlib functional/OOP divide).  While there have been a few separate in-person discussions about this with various parties, the goal of this issue is to summarize the logic/decisions at the high-level so that it can be seen by all.

While at the basic level this is a functional vs object-oriented question, there's some `specutils`-specific considerations. These include:
* The need to support multiple algorithms for what seems to be the same thing but where they have different physical interpretations and therefore different science results (e.g. https://github.com/astropy/specutils/issues/255#issuecomment-409686415)
* Several operations that require "weights" - both in the sense of per-spectrum weights and per-pixel weights. Those weights *often* are, but are not *always* encoded in the uncertainty.  But they only make sense for a *specific* operation (not the spectrum as a whole)
* Many operations have several options/parameters that affect how they operate

I think the options are better when put out as actual pseudo-code examples.  So here are wait I see as the key 5 competing options (although of course others could add more):

1. (purely functional, each algorithm its own function): ``result_or_new_spectrum = operation_with_algorithm1(single_or_list_of_specs, weights=weights, options=...)``
2. (purely functional, each operation its own function): ``result_or_new_spectrum = operation(single_or_list_of_specs, weights=weights, algorithm='name_of_algorithm', options=...)``
3. (object-oriented, each algorithm its own class): 
```
op = OperationWithAlgorithm1(options=...)
result_or_new_spectrum = op(single_or_list_of_specs, weights)
```
4. (object-oriented, each operation its own class, data stored in operation) - similar to the `ccdproc` combiner classes: 
```
op = Operation(single_or_list_of_specs, weights)
result_or_new_spectrum = op.do_algorithm1(options=...)
```
5. (operations/algorithms live as *methods* of the collection objects): 
```
specc = SpectrumCollection(...)
result_or_new_spectrum.operation_with_algorithm1(weights, options=...)
```

[...]

PR #281, opened 2018-08-20
Add width statistics

### LINKED ISSUE #238: FWHM measurement
Calculate the full width at half maximum of the input spectrum.

### LINKED ISSUE #241: Gaussian width (sigma)
1-sigma full width of an equivalent Gaussian (Can be determined from moments or fitting). splot has some algorithms. 

This might be overtaken or use some work in the spectral fitting as it will likely use spectral fitting (which uses the astropy.modeling work).

drdavella:
This implements full width half max and full width based on gaussian sigma. It resolves #238 and resolves #241.

~Currently it does not properly handle `Spectrum1D` objects with multiple spectra (see #280).~ This is no longer true thanks to #283.

**Update**
~This now resolves issue #268 as well.~ Issue #268 should be addressed in a separate PR since as @eteq pointed out below, `equivalent_width` doesn't really belong with the other "width" calculations.

While this PR does not directly address #266, I believe it will provide a useful starting point for refactoring analysis operations.

### DIFF
--- a/docs/basic_analysis.rst
+++ b/docs/basic_analysis.rst
@@ -1,3 +1,5 @@
+.. currentmodule:: specutils.analysis
+
 ==============
 Basic Analysis
 ==============
@@ -75,6 +77,61 @@ And if the spectrum contains a continuum, then it should be subtracted first:
     >>> continuum_flux = continuum_baseline(spec.spectral_axis.value) #doctest:+SKIP
     >>> continuum = Spectrum1D(spectral_axis=spec.spectral_axis, flux=continuum_flux) #doctest:+SKIP
     >>> c = centroid(spec-continuum, region=None) #doctest:+SKIP
+
+Width
+-----
+
+There are several width statistics that are provided by the
+`~specutils.analysis` submodule.
+
+The `~gaussian_sigma_width` function estimates the width of the spectrum by
+computing an approximation of the standard deviation.
+
+The `~gaussian_fwhm` function estimates the width of the spectrum at half max,
+again by computing an approximation of the standard deviation.
+
+Both of these functions assume that spectrum is approximately gaussian and that
+it is centered on the spectral axis.
+
+The function `~fwhm` provides an estimate of the full width of the spectrum at
+half max that does not assume the spectrum is gaussian or centered on the
+spectral axis. It locates the maximum, and then locates the value closest to
+half of the maximum on either side, and measures the distance between them.
+
+Consider the following noisy gaussian spectrum as an example:
+
+.. plot::
+   :include-source: true
+
+   >>> import numpy as np
+   >>> from astropy import units as u
+   >>> from astropy.modeling import models
+   >>> from specutils import Spectrum1D
+   >>> from specutils.analysis import gaussian_sigma_width
+   >>> np.random.seed(0)
+
+   >>> spectral_axis = np.linspace(0., 10., 200) * u.GHz
+   >>> # Note that the gaussian is centered on the spectral axis
+   >>> spectral_model = models.Gaussian1D(amplitude=3*u.Jy, mean=5*u.GHz, stddev=0.8*u.GHz)
+   >>> flux = spectral_model(spectral_axis)
+   >>> # Add noise
+   >>> flux += np.random.normal(0., 0.2, spectral_axis.shape) * u.Jy
+   >>> noisy_gaussian = Spectrum1D(spectral_axis=spectral_axis, flux=flux)
+
+   >>> import matplotlib.pyplot as plt
+   >>> plt.plot(noisy_gaussian.spectral_axis, noisy_gaussian.flux)
+
+Each of the width analysis functions are applied to this spectrum below:
+
+.. code-block:: python
+
+   >>> from specutils.analysis import gaussian_sigma_width, gaussian_fwhm, fwhm
+   >>> gaussian_sigma_width(noisy_gaussian)
+   <Quantity 1.59661941 GHz>
+   >>> gaussian_fwhm(noisy_gaussian)
+   <Quantity 1.87987569 GHz>
+   >>> fwhm(noisy_gaussian)
+   <Quantity 1.85929648 GHz>
 
 Reference/API
 -------------
--- a/docs/basic_analysis.rst
+++ b/docs/basic_analysis.rst
@@ -85,7 +85,7 @@ There are several width statistics that are provided by the
 `~specutils.analysis` submodule.
 
 The `~gaussian_sigma_width` function estimates the width of the spectrum by
-computing an approximation of the standard deviation.
+computing a second-moment-based approximation of the standard deviation.
 
 The `~gaussian_fwhm` function estimates the width of the spectrum at half max,
 again by computing an approximation of the standard deviation.
@@ -125,12 +125,12 @@ Each of the width analysis functions are applied to this spectrum below:
 .. code-block:: python
 
    >>> from specutils.analysis import gaussian_sigma_width, gaussian_fwhm, fwhm
-   >>> gaussian_sigma_width(noisy_gaussian)
-   <Quantity 1.59661941 GHz>
-   >>> gaussian_fwhm(noisy_gaussian)
-   <Quantity 1.87987569 GHz>
-   >>> fwhm(noisy_gaussian)
-   <Quantity 1.85929648 GHz>
+   >>> gaussian_sigma_width(noisy_gaussian) # doctest: +FLOAT_CMP
+   <Quantity 0.93853592 GHz>
+   >>> gaussian_fwhm(noisy_gaussian) # doctest: +FLOAT_CMP
+   <Quantity 2.21008321 GHz>
+   >>> fwhm(noisy_gaussian) # doctest: +FLOAT_CMP
+   <Quantity 1.65829146 GHz>
 
 
 Reference/API
--- a/docs/basic_analysis.rst
+++ b/docs/basic_analysis.rst
@@ -90,13 +90,12 @@ computing an approximation of the standard deviation.
 The `~gaussian_fwhm` function estimates the width of the spectrum at half max,
 again by computing an approximation of the standard deviation.
 
-Both of these functions assume that spectrum is approximately gaussian and that
-it is centered on the spectral axis.
+Both of these functions assume that spectrum is approximately gaussian.
 
 The function `~fwhm` provides an estimate of the full width of the spectrum at
-half max that does not assume the spectrum is gaussian or centered on the
-spectral axis. It locates the maximum, and then locates the value closest to
-half of the maximum on either side, and measures the distance between them.
+half max that does not assume the spectrum is gaussian. It locates the maximum,
+and then locates the value closest to half of the maximum on either side, and
+measures the distance between them.
 
 Consider the following noisy gaussian spectrum as an example:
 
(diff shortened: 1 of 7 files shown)


brechmos-stsci:
@drdavella and I talked through lots of the code here.  @eteq and @stscicrawford definitely check through the *science* of the equations to confirm you guys are good with it from an astronomy point of view.

specutils/analysis/width.py
```diff
@@ -0,0 +1,190 @@
...
+
+    Returns
+    -------
+    full_width_half_max : `~astropy.units.Quantity` or list (based on region input)
+        Approximate full width of the signal at half max
+    """
+    return _computation_wrapper(_compute_full_width_half_max, spectrum, region)
+
+
+def equivalent_width(spectrum, region=None):
```
eteq: Actually I think `equivalent_width` should *not* be in this PR.  While the name would seem to suggest otherwise, EQW is actually a measure of the *flux* of a line more than a *width*.  It's basically the flux normalized by the continuum level.  So I'd say it should instead be in its own module for flux measurements (which we will later populate with `flux` and related things).

specutils/analysis/width.py
```diff
@@ -0,0 +1,190 @@
...
+    if region is not None:
+        calc_spectrum = region.extract(spectrum)
+    else:
+        calc_spectrum = spectrum
+
+    flux = calc_spectrum.flux
+    wavelength = calc_spectrum.wavelength
+
+    # Continuum is always assumed to be 1.0
+    avg_cont = np.median(flux)
```
eteq: I would say instead of assuming the `median`, have it be a user-supplied value, which should default to ``1.0``.  Might leave in the `median` option by being a special case if the user gives the string `median`, though.

specutils/analysis/width.py
```diff
@@ -0,0 +1,190 @@
...
+    # Single region
+    elif isinstance(region, SpectralRegion):
+        return func(spectrum, region=region)
+
+    # List of regions
+    elif isinstance(region, list):
+        return [func(spectrum, region=reg) for reg in region]
+
+
+def sigma_full_width(spectrum, region=None):
```
eteq: While it may seem a bit odd, I think *this* function is the  one that should use what you currently have for the FWHM.  But it should be called something like `gaussian_sigma_width`.  Using that convention, the current `full_width_half_max` could become `gaussian_fwhm` or similar.

(More on this in my review summary)

eteq:
To record some in-person discussion: we agreed to move the eqw part of this *out* of this PR into a separate one (see #314 for more), but the rest will be updated as discussed above.

docs/basic_analysis.rst
```diff
@@ -75,6 +77,95 @@ And if the spectrum contains a continuum, then it should be subtracted first:
...
+`~specutils.analysis` submodule.
+
+The `~gaussian_sigma_width` function estimates the width of the spectrum by
+computing an approximation of the standard deviation.
+
+The `~gaussian_fwhm` function estimates the width of the spectrum at half max,
+again by computing an approximation of the standard deviation.
+
+Both of these functions assume that spectrum is approximately gaussian and that
+it is centered on the spectral axis.
```
eteq: This might be a good spot to put an example of using one of these with `SpectralRegion.from_center` to get the centering right?

specutils/analysis/width.py
```diff
@@ -0,0 +1,212 @@
...
+    Estimate the full width of the spectrum based on an approximate value of
+    sigma.  This will be calculated over the regions, if they are specified.
+
+    Parameters
+    ----------
+    spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
+        The spectrum object over which the width will be calculated.
+
+    region: `~specutils.utils.SpectralRegion` or list of `~specutils.utils.SpectralRegion`
+        Region within the spectrum to calculate the gaussian sigma width.
```
eteq: Here and below, maybe add something like "If region is None, this will be calculated over the entire spectrum"?

specutils/analysis/width.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # List of regions
+    elif isinstance(region, list):
+        return [func(spectrum, region=reg) for reg in region]
+
+
+def gaussian_sigma_width(spectrum, region=None):
+    """
+    Estimate the full width of the spectrum based on an approximate value of
+    sigma.  This will be calculated over the regions, if they are specified.
```
eteq: Might reword this a bit to something like "Estimate the width of the spectrum using a second-moment analysis, scaled to match the sigma/standard deviation parameter of a standard Gaussian profile."? I'm not sure the grammar is sound in what I wrote there, but I think it at least hits on all the key mathematical points... 😉

specutils/analysis/width.py
```diff
@@ -0,0 +1,212 @@
...
+    -------
+    full_width : `~astropy.units.Quantity` or list (based on region input)
+        Approximate full width of the signal
+    """
+    return _computation_wrapper(_compute_gaussian_sigma_width, spectrum, region)
+
+
+def gaussian_fwhm(spectrum, region=None):
+    """
+    Estimate the full width half max of the spectrum assuming it is gaussian.
```
eteq: As above, perhaps "Estimate the width of the spectrum using a second-moment analysis, scaled to match the  full width at half max for a standard Gaussian profile."

docs/basic_analysis.rst
eteq: That said... I don't think this has to be that sensitive to the centering (see my comment in the implementation function further below about using `centroid`).  What it *is* sensitive to might be the continuum subtraction, though: I think they assume the continuum has been subtracted.

specutils/analysis/width.py
```diff
@@ -0,0 +1,212 @@
...
+
+    if region is not None:
+        calc_spectrum = region.extract(spectrum)
+    else:
+        calc_spectrum = spectrum
+
+    flux = calc_spectrum.flux
+    spectrum = calc_spectrum.spectral_axis
+
+    dx = spectrum - np.mean(spectrum)
```
eteq: Piggy-backing on the comment above in the docs: I think this is the magic part that could make it relatively insensitive to the centering.  If instead of doing ``np.mean`` here, you instead do ``np.mean(spectrum*flux)/np.sum(flux)``, you get get a centroid estimate of the center of the line.

Alternatively, you can just use the centroid function *directly* from `specutils.analysis.centroid` - that might not be quite as efficient (it has to do the region-extraction twice), but is a better re-use of code.

One step even further is to change this line to ``dx = spectrum - center_func(calc_spectrum)`` and have ``center_func`` be a user-choosable keyword, which you would have default to `specutils.analysis.centroid`.

(You've basically been getting my stream-of-conciousness here... But TL;DR I think you should do what I said in the last paragraph unless it's going to take too long)

specutils/analysis/width.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+def _arghalf(array, halfval):
+    """
+    Find index of the item closest to the half max
+    """
+    return np.abs(array - halfval).argmin()
+
+
+def _compute_single_fwhm(flux, spectrum):
```
eteq: (readability nitpick, don't need to both with it if you don't want to): 
`spectrum` is a bit confusing of a variable name because it seems like you're talking about the `Spectrum1D`, when actually this is the `spectral_axis`.  Maybe `specax`?

specutils/analysis/width.py
```diff
@@ -0,0 +1,212 @@
...
+
+    argmax = np.argmax(flux)
+    halfval = flux[argmax] / 2
+
+    max_idx = len(flux) - 1
+    l_idx = _arghalf(flux[:argmax], halfval) if argmax > 0 else 0
```
eteq: Doesn't this end up finding the *left-most* cross over point instead of the one closest to the center? (But maybe just changing it to ``.argmax()`` instead of ``.argmin()`` is the fix?)

specutils/analysis/width.py
drdavella: This makes sense. The duplication of region extraction occurred to me too, but this is probably something we should handle when addressing #266.

specutils/analysis/width.py
drdavella: Yes it does, this is what I mentioned to you in person yesterday. It's quick and dirty and intended to be efficient. Using `argmin` means we find the index of the minimum difference, whereas `argmax` would return the index of the maximum difference, which is not what we want. 

The better thing to do here is probably to explicitly traverse this array out from the center and look for the minimum difference, which is not quite as efficient, but might provide better results.

specutils/analysis/width.py
drdavella: Sounds good. Btw, I wonder what you think about removing the docstrings for the private methods below. They mostly just repeat what the public functions say already.

docs/basic_analysis.rst
drdavella: I added an `xfail` unit test for the centered case. Using `centroid` here causes the existing unit tests to fail, so I'm not sure what's going on.

drdavella:
I just pushed a fix for one test failure, but the non-centered sigma width calculation is still not working (marked as `xfail`).

docs/basic_analysis.rst
```diff
@@ -76,6 +78,61 @@ And if the spectrum contains a continuum, then it should be subtracted first:
...
+   >>> import matplotlib.pyplot as plt #doctest:+SKIP
+   >>> plt.plot(noisy_gaussian.spectral_axis, noisy_gaussian.flux) #doctest:+SKIP
+
+Each of the width analysis functions are applied to this spectrum below:
+
+.. code-block:: python
+
+   >>> from specutils.analysis import gaussian_sigma_width, gaussian_fwhm, fwhm
+   >>> gaussian_sigma_width(noisy_gaussian)
+   <Quantity 1.59661941 GHz>
```
eteq: Travis is currently failing here.  You probably want a ``  # doctest: +FLOAT_CMP`` regardless (for other architectures, etc), but I think the actual *value* is also different? (should be more like 0.8, right?)

docs/basic_analysis.rst
```diff
@@ -76,6 +78,61 @@ And if the spectrum contains a continuum, then it should be subtracted first:
...
     >>> c = centroid(spec-continuum, region=None) #doctest:+SKIP
 
+Width
+-----
+
+There are several width statistics that are provided by the
+`~specutils.analysis` submodule.
+
+The `~gaussian_sigma_width` function estimates the width of the spectrum by
+computing an approximation of the standard deviation.
```
eteq: "an approximation" -> "a second-moment based approximation"

specutils/analysis/width.py
```diff
@@ -0,0 +1,179 @@
...
+    spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
+        The spectrum object over which the width will be calculated.
+
+    region: `~specutils.utils.SpectralRegion` or list of `~specutils.utils.SpectralRegion`
+        Region within the spectrum to calculate the gaussian sigma width. If
+        region is `None`, computation is performed over entire spectrum.
+
+    Returns
+    -------
+    full_width : `~astropy.units.Quantity` or list (based on region input)
```
eteq: I think this is should be called `sigma` or maybe `sigma_estimate`, right?  That is, it's not a full width (sigma is sort of a "half width", although it's a specific scaling)

specutils/analysis/width.py
```diff
@@ -0,0 +1,179 @@
...
+        return func(spectrum, region=region)
+
+    # List of regions
+    elif isinstance(region, list):
+        return [func(spectrum, region=reg) for reg in region]
+
+
+def gaussian_sigma_width(spectrum, region=None):
+    """
+    Estimate the width of the spectrum using a second-moment analysis, scaled
```
eteq: Maybe split this into a short initial sentence? E.g. "Estimate the width of the spectrum (or region) using a second-moment analysis.  This estimate is scaled to ..."

This is useful mainly because the sphinx docs will use the first sentence to show a summary of the function.

(and similar for the FWHM)

docs/basic_analysis.rst
drdavella: Yes, this was an oversight.

specutils/analysis/width.py
drdavella: Yup, another oversight.

eteq:
Tests passed - LGTM!

PR #284, opened 2018-08-21
Add unit test for multidimensional flux

drdavella:
This adds a test that confirms the bug fix merged in #283.


specutils/tests/test_spectrum1d.py
```diff
@@ -21,6 +21,16 @@ def test_create_from_arrays():
     assert spec.flux.size == 50
 
 
+def test_create_from_multidimensional_arrays():
+
```
eteq: Can you note here that this is also a regression test for the bug fixed in #283 ?  Helpful to have that context for anyone who comes across it later.

PR #296, opened 2018-09-05
Fix failing rtd builds

nmearl:
Failing builds in read the docs seem to be due to a missing `rtd-pip-requirements` file that was removed (by accident?) some time ago. This PR adds the file back.


eteq:
(canceled travis because it wasn't going to test this anyway)

Issue #297, opened 2018-09-05
Warnings on creation of `Spectrum1D` objects


eteq:
A lot of the natural ways to create `Spectrum1D` objects lead to warnings like:
```
WARNING:root:GWCS does not store rest frequency information. Please define the rest value explicitly in the `Spectrum1D` object.
WARNING:root:GWCS does not store rest wavelength information. Please define the rest value explicitly in the `Spectrum1D` object.
```
while this is understandable since rest information is needed for situations where the user might want to actually do some of the relevant conversions, most of the time a user leaving it off probably just doesn't *need* that information for their use case.  So it would be better to not show those warnings, and instead just raise an error (with a specific enough error message that they know what's they need to do to fix it) when they try to do something that requires rest information.

PR #302, opened 2018-09-06
Reduce verbosity in gwcs creation

nmearl:
Demotes status messages about rest wavelength and frequencies to `debug` when generating GWCS objects on spectrum initialization.


nmearl:
Addresses #297.

nmearl:
@eteq [There already is a test](https://github.com/astropy/specutils/blob/master/specutils/tests/test_spectrum1d.py#L87) that checks for failure when attempting to call `velocity` (which uses the `rest_wavelength`) on a `Spectrum1D` that doesn't not have it defined.

PR #306, opened 2018-09-06
Should the default fitter be SimplexLSQFitter or LevMarLSQFitter?

eteq:
The fitting added in #210 defaults to `SimplexLSQFitter`.  I didn't want to block #210 by having this discssion there but since it's in... I'm inclined to say we should change the defaults to `LevMarLSQFitter` because that's what astronomers are often used to... Not a great reason, but possibly still the relevant one.

Thoughts, @brechmos-stsci @hcferguson @keflavich @crawfordsm ?


keflavich:
:+1: to defaulting to LevMar.  I don't know what Simplex does and have never used it, and I assume (without any data to support my assumption) that applies to other astronomers.

keflavich:
Based on benchmarks, we should probably be defaulting to LM:
http://docs.mantidproject.org/v3.7.2/concepts/FittingMinimizers.html
LM outperforms the other algorithms in all of their test cases.  I don't know enough about the test problems to say whether they're similar to spectral fit cases, but I'd be surprised if spectral fitting was a huge outlier.

hcferguson:
I'm happy defaulting to LM. Jerry Kriss's SPECFIT, often used in IRAF, recommended a Marquardt-type algorithm from Numerical recipes.  http://stsdas.stsci.edu/cgi-bin/gethelp.cgi?specfit .

eteq:
I can probably get to this in another ~hour, @brechmos-stsci, but if you can get to it sooner and have free cycles, by all means!

PR #308, opened 2018-09-07
Fix photon_flux units

eteq:
I realized the `photon_flux` expression was wrong - really the thing that makes sense here is a photon flux density - i.e. photon flux per unit spectral axis.  The only reason it seemed to work before was because Hz^-1 happens to conveniently be seconds, which made some things cancel out... but that's sweeping a lot under the rug.

Can someone who's willing to think about spectral units double-check my thinking here? (maybe @keflavich or @crawfordsm?)

### DIFF
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -133,9 +133,13 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
     @property
     def photon_flux(self):
         """
-        The flux, in units of photons per cm^2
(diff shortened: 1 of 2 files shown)


eteq:
needed a rebase - good to merge as soon as the tests pass

eteq:
The test failures cannot possible be related because it's in a file I didn't touch... although oddly they aren't occurring in master?  I think I'm going to merge this since it seems like a red herring, but if the tests start failing in master we know I'm the one to blame...

Issue #196, opened 2018-05-18
Wavelength unit conversion


brechmos-stsci:
Need the ability to convert wavelength units. This should leverage astropy units etc.

(will be used by specviz)

nmearl:
@robel

robelgeda:
Nothing new to add here. This feature is already implemented. This ticket can be closed.

brechmos-stsci:
I might have missed them, but are there tests for the wavelength unit conversions?  If not, that would be a great thing to add.

nmearl:
[Yes, there are](https://github.com/astropy/specutils/blob/master/specutils/tests/test_spectrum1d.py#L54)

robelgeda:
I have made a mistake here @brechmos-stsci and @nmearl. It would be nice to add `to_spectral_axes` just like there is a `to_flux` function for flux unit conversions. I was under the impression that the  [`with_spectral_unit`](https://github.com/astropy/specutils/blob/master/specutils/spectra/spectrum_mixin.py#L171) function satisfied this ticket. This is not correct because `with_spectral_unit` returns a new `Spectrum1D` object instead of converting the units of the spectral unit in question.

eteq:
@robelgeda - can you clarify the motivation behind a `to_spectral_axis`?  Can't the user just do `spectrum.spectral_axis.to(u.<whatever>)`?

(Same question for `to_flux`, now that I think about it...)

robelgeda:
I am not sure actually, @nmearl what are your thoughts?

nmearl:
@eteq `spectrum.spectral_axis.to` requires that the units the spectrum has on its instance are correct and convertible. @robelgeda's proposal is to have methods that allow a user to _overwrite_ the current unit, without consideration of conversion. E.g. in the case of incorrect units in a fits header being read into a new `Spectrum1D` object.

eteq:
Aha, now that I understand I am against this.  It's a lot more work to try to keep things validated when the user might change the unit at any time without any checks.  It's also heavily overloading the word `to` here, which in all other contexts means "give me a new object with where I *convert* the unit".  Users will *certainly* be confused by mixing those.

Instead to address this use case I'd say you'd do something like this:
```
new_spec = Spectrum1D(flux=old_spec.flux.value*newunit, spectral_axis=old_spec.spectral_axis)
```
That's more explicit anyway.

hcferguson:
[...]

PR #309, opened 2018-09-07
To flux no inplace

eteq:
#196 made me realize a serious problem with `to_flux`: it's an in-place operation.  `to` in the astropy/units context always means "return a new converted object".  So `to_flux` is a *very* confusing thing to subject users to.

So to rectify it this PR gives two options:
1. (the first two commits): rename to `set_flux_unit` - that makes it clearer that you are doing an inplace operation on the spectrum object
2. (the third commit): changes the method to instead yield a *new* `Spectrum1D` object.  

The second one is my preference, as I think it's much safer - in-place changes always cause unexpected trouble, because it's another place you have to verify that things make sense. This implementation still has that problem to some extent, but at least it doesn't paint us into the corner of definitely allowing in-place changes in the public API.

cc @robelgeda @nmearl (note this is time-sensitive because we don't want `to_flux` to go into 0.4 and then get pulled again in 0.5)

### DIFF
--- a/specutils/spectra/spectrum_mixin.py
+++ b/specutils/spectra/spectrum_mixin.py
@@ -83,9 +83,10 @@ class OneDSpectrumMixin(object):
         """
         return u.Quantity(self.data, unit=self.unit, copy=False)
 
-    def to_flux(self, unit, equivalencies=None, suppress_conversion=False):
+    def set_flux_unit(self, unit, equivalencies=None, suppress_conversion=False):
         """
-        Converts the flux data to the specified unit.
+        Converts the flux data to the specified unit.  This is an in-place
+        change to the object.
 
         Parameters
         ----------
--- a/specutils/spectra/spectrum_mixin.py
+++ b/specutils/spectra/spectrum_mixin.py
@@ -1,4 +1,5 @@
 import logging
+from copy import deepcopy
 
 import numpy as np
 from astropy.wcs import WCSSUB_SPECTRAL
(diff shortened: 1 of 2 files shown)


hcferguson:
I have the same niggling worry as on #196. I think I agree with @eteq, but am slightly worried about the SpecViz use case, where at least conceptually one is changing the *view* of a spectrum without changing the data.

hcferguson:
@eteq see my comment in #196...I still agree with you that we should be creating a new Spectrum1D object when converting units.

robelgeda:
Here are two use cases we thought of when we wrote that function:
1) As discussed, the user wants to convert the units.
2) The units in the input parameters or fits file was wrong and the user wants to set the units without converting anything.
That function is intended to provide a convenient way to address these use cases. 

My opinion is to make it available but think of good function names as @eteq has done. Maybe refactor the function into 3 functions:
1) Overrides the units without changing. 
2) Convert current units in place.
3) Something that converts and returns a new instance like astropy `to`. 

Perhaps make the first two functions hidden or "private" functions to make it clear it meant for devs.

nmearl:
I don't _think_ returning a new `Spectrum1D` object should be an issue in the new SpecViz, since there's now an intermediary object between the spectrum and the Qt item used in the internal data model.

I agree that, barring the SpecViz considerations, we should seek to make the user experience consistent with astropy precedent.

eteq:
Alright, since this is passing and approved, I'm going to go ahead and merge this so that something is in place for 0.4.  I'll make a new issue (referencing this one) where we can decide if we want to add the two private functions @robelgeda suggests in https://github.com/astropy/specutils/pull/309#issuecomment-419458266

PR #311, opened 2018-09-07
Better message if uncertainty is not defined for Spectrum1D

### LINKED ISSUE #282: Single region SNR calculation always expects uncertainties
[This line](https://github.com/astropy/specutils/blob/1014a877af178a255ca2b0ceae9e6ab96a451433/specutils/analysis/snr.py#L77) assumes that the `Spectrum1D.uncertainty` is not `None`.

brechmos-stsci:
There was not a check, previously, to see if the uncertainty is defined. If not, then raise a specific exception that states it is required.

Fixes #282

### DIFF
--- a/specutils/analysis/snr.py
+++ b/specutils/analysis/snr.py
@@ -30,6 +30,9 @@ def snr(spectrum, region=None):
 
     """
 
+    if spectrum.uncertainty is None:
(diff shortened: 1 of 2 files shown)


brechmos-stsci:
@eteq This can be reviewed and merged, if satisfactory.

PR #313, opened 2018-09-10
Front documentation updates

### LINKED ISSUE #294: Write some words in the docs about all the different "kinds" of spectra
This was inspired by discussion in #273 - we need some words about all the different "kinds" of spectrum and spectrum collection objects.  This should just be in the docs, so I will write some words about that.

eteq:
This provides some updates to the index page of the documentation.  In particular, it highlights the various "cases" of spectra we cover here (along with more contextual words about why users might care).

Note that this will not pass its doc build because it assumes `SpectrumCollection` exists, which is waiting on #273.  Once that's in this should be set.

Closes #294


docs/index.rst
```diff
@@ -1,35 +1,84 @@
...
-maintaines a few other features including unit support, custom data loading,
-arithmetic operation support, and multi-dimensional flux data support.
 
-.. note:: support for collections of spectra with distinct spectral axes
-          information is slotted for a future release.
+
+Types of Spectral Representations
+---------------------------------
+
+The core principal of ``specutils`` is to work as a toolbox.  That is, it aims
```
hcferguson: principal -> principle

docs/index.rst
```diff
@@ -1,35 +1,84 @@
...
+to provide the pieces needed to build particular spectroscopic workflows
+without imposing a specific *required* set of algorithms or approaches to
+spectroscopic analysis.  To that end, it aims to represent several different
+types of ways one might wish to represent sets of spectroscopic data in Python.
+These objects contains logic to handle multi-dimensional flux data, spectral
+axes in various forms (wavelenth, frequency, energy, velocity, etc.), convenient
+and unobtrusive wcs support, and uncertainty handling. This core containers also
```
hcferguson: "This core containers" -> "The core containers"

docs/index.rst
```diff
@@ -1,35 +1,84 @@
...
+handle units, a framework for reading and writing from various file formats,
+arithmetic operation support, and a variety of analysis and manipulation tools
+that work on them.
+
+
+The core data objects  are primarily distinguished by the different ways of
+storing the fluxand the spectral axis . These cases are detailed below, along
```
hcferguson: fluxand -> flux and

docs/index.rst
```diff
@@ -1,35 +1,84 @@
...
+   represented by the `~specutils.Spectrum1D` object with a 1-dimensional
+   ``flux``.
+2. A set of fluxes that can be represented in an array-like form of shape
+   ``m (x ...) x n``,  with a spectral axis strictly of length ``n`` (and a
+   matched WCS). In ``specutils`` this is represented by the
+   `~specutils.Spectrum1D` object where ``len(flux.shape) > 1`` . In this sense
+   the "1D" refers to the spectral axis, *not* the flux.
+3. A set of fluxes  of shape ``m (x ...) x n``, and a set of spectral axes that
+   are the same shape. This is distinguished from the above cases because there
+   are as many spectral axes as their are spectra.  In this sense it is a
```
hcferguson: their are -> there are

hcferguson:
I tried to make a diagram to capture what you have said in words about the contents of Spectrum1D and SpectrumCollection. I'm not sure I've got this right though....In the case of SpectrumCollection is there 1 WCS or are there N WCSs? And is there one set of meta information or N sets?
[specutils_classes.pdf](https://github.com/astropy/specutils/files/2372066/specutils_classes.pdf)

nmearl:
@hcferguson A `SpectrumCollection` with one WCS would, I believe, require resampling (this is the crux of the previous conversations where I made the assumption that _that_ was desired). The type-3 `SpectrumCollection` makes no such assumptions, and assumes all WCS are separate, and includes the restriction that all spectra in the collection be the same shape (and thus e.g. `spec_collection.flux[:, 0]` doesn't return all flux values at the same dispersion value, just a slice at that _pixel_ unit).

hcferguson:
@nearl Makes sense. What kind of object is the collection of WCS's? A list? 

And does the same go for meta? A list of meta objects one per spectral_axis?

hcferguson:
@nmearl @eteq I've updated my figure, but I'm not certain I've got the ordering of the dimensions right. Can you check, and if you like the figure include it in the docs? The original is in ppt, so that's attached as well.

[specutils_classes.pdf](https://github.com/astropy/specutils/files/2375952/specutils_classes.pdf)
[specutils_classes.pptx](https://github.com/astropy/specutils/files/2375953/specutils_classes.pptx)

nmearl:
@hcferguson The shape should be `(M, N)`. E.g.

```python
# I want to create a multi-dimensional array of 5 length-10 arrays
In [32]: a = np.random.sample(size=(5, 10))

# Shape is reported as MxN
In [33]: a.shape
Out[33]: (5, 10)

# The first index returns an N-sized array
In [34]: a[0].size
Out[34]: 10
```

hcferguson:
Okay...updated the figure to reflect that ordering.
[specutils_classes.pdf](https://github.com/astropy/specutils/files/2376421/specutils_classes.pdf)
[specutils_classes.pptx](https://github.com/astropy/specutils/files/2376422/specutils_classes.pptx)

eteq:
Note: this is *almost* ready, but needs a bit more TLC re: the recent changes in `analysis`, and I'd like to add a short "getting started" section

eteq:
Ok I think this now has all the things I was hoping to see.  However, it has revealed several bugs: #335 #334, and some issues I still need to resolve with the top-level links (it appears someone turned off/never turned on the machinery the helpers provides that allows sphinx links like `specutils.Spectrum1D` to be auto-resolved to `specutils.spectra.Spectrum1D` or similar... not sure why or how - @nmearl do you have any idea?).

eteq:
Note that if someone wants to review the *content* though (@hcferguson ?) you can have a look at  http://eteq.github.io/specutils/ . The code isn't quite working yet, but the goal is to get the code to match what's in the docs examples.

hcferguson:
Looks good. Just a few minor comments in the annotations at https://hyp.is/go?url=http%3A%2F%2Feteq.github.io%2Fspecutils%2F.

eteq:
The failure looks to be a red herring that's just about a missing download message in a docstring.  So I think this is all set!

PR #324, opened 2018-09-13
WIP: Fix equivalent width and implement line flux

### LINKED ISSUE #268: Revamp equivalent_width to use spectral region

### LINKED ISSUE #314: Basic implementation of flux/eqw
In some out-of-band discussion, we decided for the v0.4 release to have a "quick-and-dirty" EQW and line-flux implementation that does not do fractional-pixel interpolation correctly.  Then in 0.5 we'll try to do the more correct approach.

So this issue is for 0.4: to update `equivalent_width` and add a `flux` (or maybe `line_flux`?  or `region_flux`?) to exist in `analysis` and have a quick-and-dirty implementation that doesn't do the sub-pixel stuff.  That is, just sum up the pixels inside the region and divide by the delta-spectral axis for those pixels. (And then `equivalent_width` should be just based on that, with an added `continuum` keyword which is assumed to be 1)

cc @drdavella @brechmos-stsci

drdavella:
This PR needs more work, including documentation.

It resolves #268, and resolves #314.

**NOTE**: This PR is based on #281 and should not be merged until that PR is merged and this is rebased.


drdavella:
The equivalent width computation no longer passes with the preexisting unit tests. I'm not sure if this is a problem with the test or with the implementation of the function, but if @eteq has a chance for another look, it would be appreciated!

specutils/analysis/flux.py
```diff
@@ -0,0 +1,87 @@
...
+
+def line_flux(spectrum, region=None):
+    """
+    Computes the line flux over a spectrum. Applies a region if given.
+
+    Parameters
+    ----------
+    spectrum : Spectrum1D
+        The spectrum object over which the line flux will be calculated.
+
```
eteq: missing ``region`` keyword.  Also, shouldn't it be ``regions`` if it can be multiple regions?

specutils/analysis/flux.py
```diff
@@ -0,0 +1,87 @@
...
+from astropy.units.quantity import Quantity
+from .utils import computation_wrapper
+
+
+__all__ = ['line_flux', 'equivalent_width']
+
+
+def line_flux(spectrum, region=None):
+    """
+    Computes the line flux over a spectrum. Applies a region if given.
```
eteq: Maybe change to "Computes the flux in a spectrum or area of a spectrum. 
 Applies to the whole spectrum by default, but can be limited to a specific feature (like a spectral line) if a `region` is given."

specutils/analysis/flux.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Parameters
+    ----------
+    spectrum : Spectrum1D
+        The spectrum object over which the line flux will be calculated.
+
+    Returns
+    -------
+    flux : `~astropy.units.Quantity`
+        line flux result
```
eteq: -> "flux in the provided spectrum (or regions)"?

specutils/analysis/flux.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    return computation_wrapper(_compute_line_flux, spectrum, region)
+
+
+def equivalent_width(spectrum, continuum=1, region=None):
+    """
+    Does a naive equivalent width measures on the spectrum object.
```
eteq: Instead maybe "The equivalent width of a region of the spectrum".

EQW's don't *really* make sense if there's no sub-region. Nominally it's a well-defined object but no one really uses it that way.

specutils/analysis/flux.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Parameters
+    ----------
+    spectrum : Spectrum1D
+        The spectrum object overwhich the equivalent width will be calculated.
+
```
eteq: missing "regions" again

specutils/analysis/flux.py
```diff
@@ -0,0 +1,87 @@
...
+    continuum : `~astropy.units.Quantity`, optional
+        Constant continuum value
+
+    Returns
+    -------
+    ew : `~astropy.units.Quantity`
+        Equivalent width calculation.
```
eteq: + "in units of whatever the `spectral_axis` is"

specutils/analysis/flux.py
```diff
@@ -0,0 +1,87 @@
...
+
+    TODO:  what frame of reference do you want the spectral_axis to be in ???
```
eteq: I don't think this is supposed to be here?

specutils/analysis/flux.py
```diff
@@ -0,0 +1,87 @@
...
+    spectral_axis = calc_spectrum.spectral_axis
+    dx = spectral_axis[-1] - spectral_axis[0]
+
+
+    line_flux = _compute_line_flux(spectrum, region)
+
+    # Calculate equivalent width
+    ew =  dx - (line_flux / continuum)
+
+    return ew
```
eteq: Maybe this should actually be ``ew.to(calc_spectrum.flux.unit)``?  Unlike the flux case it's pretty likely that's what the user actually wants.

drdavella: I assume you meant `calc_spectrum.spectral_axis_unit`, but I agree.

PR #325, opened 2018-09-13
change SpectralRegion.from_center to be a classmethod

eteq:
I just noticed `SpectralRegion.from_center` is a staticmethod.  This is a good example for where `classmethod` is better to use than `staticmethod`, because if the user makes a subclass of `SpectralRegion`, they'd want `from_center` to yield their subclass, not `SpectralRegion`.  This (minor) PR makes that possible.

Tagging @brechmos-stsci for review since I think he's the originator of this.

This is pretty trivial so I'm tentatively scheduling it for the imminent 0.4 release.

### DIFF
--- a/specutils/spectra/spectral_region.py
+++ b/specutils/spectra/spectral_region.py
@@ -29,8 +29,8 @@ class SpectralRegion:
 
     """
 
-    @staticmethod
-    def from_center(center=None, width=None):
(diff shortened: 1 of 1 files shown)


brechmos-stsci:
I will just do this as part of https://github.com/astropy/specutils/pull/289

PR #329, opened 2018-09-18
Module and import cleanup in analysis subpackage

eteq:
Now that all the other PRs are in I did a few quick cleanup tasks in the `analysis` package.  Most notably:

* Imports are now uniformly the ``import *`` style.  I think this is good because it means there's only one place to specify what's in the public API: the ``__all__`` of the module.  As it stood before it required updating both that *and* the ``__init__.py``.
* All modules now have docstrings.
* All modules have names that are not the same as their functions.  I find things like ``from snr import snr`` often end up confusing users what they're actually importing so its best avoided if possible.

cc @drdavella @brechmos-stsci


brechmos-stsci:
@eteq I pulled your branch down, took a look at it, ran the tests etc.  Looks good!

PR #332, opened 2018-09-19
Use subok=True when broadcasting Quantity

drdavella:
A very minor change to the recently added analysis functions that allows `astropy.units.Quantity` to be broadcasted properly, and with less redundancy.

Not sure if this can be squeezed into the 0.4 release, but it's fairly inconsequential.


eteq:
Might be good to add a test case for the broadcasting... but we should probably do that more generally *anyway* as a task for 0.5...

Issue #333, opened 2018-09-19
Make fit_lines (and fit_continuum) output unitful models on non-unit input models


eteq:
This is tied to one of the examples I'm trying to do in #313, but is not formally a blocker as there is a work-around. 

I tried to do the following:
```
cont = fit_generic_continuum(spectrum)
spectrum / cont(spectrum.spectral_axis)
```
and surprisingly I got a unit-related failure.

The core problem is that if the input model to `fit_*` is unitless, the output is unitless.  That at first seems sensible except it ignores the fact that implicitly the fitting machinery is interpreting as though it *does* have units (which match the spectrum), and then strips them again at the end.

@brechmos-stsci hashed this out out-of-band and agree the way to go is to treat the "no units in the input model" case as meaning "use the units that are in the spectrum".

brechmos-stsci:
We wanted it for the continuum fitting but it seems like the Chebyshev1D model does not allow units/quantities for the parameters at all. 

From http://docs.astropy.org/en/stable/api/astropy.modeling.polynomial.Chebyshev1D.html#astropy.modeling.polynomial.Chebyshev1D:
```
This model does not support the use of units/quantities, because each term in the sum of Chebyshev polynomials is a polynomial in x - since the coefficients within each Chebyshev polynomial are fixed, we can't use quantities for x since the units would not be compatible. For example, the third Chebyshev polynomial (T2) is 2x^2-1, but if x was specified with units, 2x^2 and -1 would have incompatible units.
```

So, I am not sure this is possible.

@astrofrog  - Can you either confirm or deny this?

@eteq

astrofrog:
I can confirm that ``Chebyshev1D`` won't work with units

PR #338, opened 2018-09-20
Added in spectrum_from_model

### LINKED ISSUE #337: Create utility method to create a spectrum from a model and spectrum
This is a simple method that is needed given a model (with or without units) and a spectrum, then an output spectrum is created based on the dispersion axis of the input spectrum and the model.

The primary use case could be for building a continuum *spectrum* from a continuum model.

@eteq

brechmos-stsci:
This is a bit of a helper/utility method that creates a Spectrum1D object from a model given an input spectrum. The output Spectrum1D object will have all the same values as the input Spectrum1D except the flux will be based on the input model.

Fixes #337

### DIFF
--- a/specutils/manipulation/utils.py
+++ b/specutils/manipulation/utils.py
@@ -77,7 +77,7 @@ def excise_regions(spectrum, regions, exciser=linear_exciser):
 
     regions : list of `~specutils.SpectralRegion`
         Each element of the list is a `~specutils.SpectralRegion`. The flux
-        between these wavelengths will be "cut out" using the `exciser`
+        between these wavelengths will be "cut out" using the ``exciser``
         method.
 
     exciser: method
--- a/specutils/manipulation/utils.py
+++ b/specutils/manipulation/utils.py
@@ -2,7 +2,7 @@ import numpy as np
 
 from ..spectra import Spectrum1D, SpectralRegion
 
-__all__ = ['excise_regions', 'linear_exciser']
+__all__ = ['excise_regions', 'linear_exciser', 'spectrum_from_model']
 
 
 def linear_exciser(spectrum, region):
@@ -166,7 +166,7 @@ def spectrum_from_model(model_input, spectrum):
         The input model or compound model from which flux is calculated.
 
     spectrum : `~specutils.Spectrum1D`
-        The `~specutils.Spectrum1D` object to which the smoothing will be applied.
+        The `~specutils.Spectrum1D` object to use as the model template.
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


specutils/manipulation/utils.py
```diff
@@ -152,3 +152,42 @@ def excise_region(spectrum, region, exciser=linear_exciser):
     #
 
     return exciser(spectrum, region)
+
+
+def spectrum_from_model(model_input, spectrum):
```
eteq: This needs to be added to `__all__` if it's intended for the public API

specutils/manipulation/utils.py
```diff
@@ -152,3 +152,42 @@ def excise_region(spectrum, region, exciser=linear_exciser):
...
+    other parameters for the output `~specutils.Spectrum1D` object
+    will be the same as the input `~specutils.Spectrum1D` object.
+
+    Parameters
+    ----------
+    model : `~astropy.modeling.Model`
+        The input model or compound model from which flux is calculated.
+
+    spectrum : `~specutils.Spectrum1D`
+        The `~specutils.Spectrum1D` object to which the smoothing will be applied.
```
eteq: Copy-and-paste error?  Probably should be "The `~specutils.Spectrum1D` object to use as the model template" or something like that.

specutils/manipulation/utils.py
```diff
@@ -152,3 +152,42 @@ def excise_region(spectrum, region, exciser=linear_exciser):
...
+        flux = model_input(spectrum.spectral_axis)
+
+    # If the input model does not have units, then assume it is in
+    # the same units as the input spectrum.
+    else:
+        flux = model_input(spectrum.spectral_axis.value)*spectrum.flux.unit
+
+    return Spectrum1D(flux=flux,
+                      spectral_axis=spectrum.spectral_axis,
+                      uncertainty=spectrum.uncertainty,
```
eteq: I don't think it makes sense to copy over the uncertainty here - the uncertainty is *not* the same for the model (in one interpretation it's zeros, and in another it's undefined).

(Should also note this in the docstring)

eteq:
A bit more context: this provides an alternative solution to #333 from having the model itself know what to do.  We may want this *anyway*, but its proximate reason is that.

specutils/manipulation/utils.py
brechmos-stsci: 👍

specutils/manipulation/utils.py
brechmos-stsci: 👍

specutils/manipulation/utils.py
brechmos-stsci: 👍

eteq:
Since this isn't critical functionality and we're very close, I'm pushing this to 0.5

nmearl:
@brechmos Can you rebase this?

PR #339, opened 2018-09-21
convert absolute imports to relative

eteq:
I noticed several absolute imports floating around specutils.  Generally speaking its best to be consistent and always use relative imports inside a package where possible, so this fixes those.

@brechmos-stsci, I think these are mostly you, so can you have a look and make sure these weren't in there for some specific reason I'm missing?


brechmos-stsci:
All fine. I had copied it from somewhere else :/

PR #342, opened 2018-09-21
Using a model-with-units wrapper (UnitedModel)

brechmos-stsci:
This was done as some models are not able to hold and use quantities and units.


eteq:
After some polling it sounds like other (@nmearl @crawfordsm) liked `QuantityModel`, which on reflection I prefer, so going with that.

PR #345, opened 2018-09-21
Add region test

eteq:
Not urgent because it works in master, but this both adds a test that you can make a SpectralRegion from a multi-dimensional quantity (39b81bc) and does some minor cleanup to the test file here that I noticed on the way


nmearl:
@eteq Can you rebase and push?

PR #346, opened 2018-09-21
noise_region_uncertainty test fails

eteq:
This PR adds some tests for noise_region_uncertainty (which I guess was accidentally deleted at some point?) which are failing...


eteq:
I've renamed this issue to what the *real* problem is.  Also #313 make it clear that pixels *on their own* work, which I'll make the command decision to be enough for the 0.4 release.  So triaging this to 0.5.

PR #348, opened 2018-09-21
Fixed up a couple of tests from eteq

brechmos-stsci:
@eteq This is modified version of your PR. Please modify, delete as you feel necessary.


nmearl:
Same tests as #346?

eteq:
Yep it was a fixup of my tests, which is why it's the same as #346.  I just updated this against master now thought and I think it should be good if the tests pass.  Closes #346 if so.

Issue #343, opened 2018-09-21
Make it clear how to cite specutils


eteq:
Probably just want to use the Zenodo record for now, but we should say this in the docs and the release announcements.

cc @nmearl @keflavich @crawfordsm

eteq:
That is, I'd say we write a few sentences in the docs saying "please cite the Zenodo DOI if used in a paper", and similar words in the release announcement

stscicrawford:
If it isn't set up already, we should set up the create zenodo DOI with the release.  For right now, this should be the default citation for the package.   

In the not so longer term, we should determine what is the minimal functionality for the 1.0 release and then write a paper for a more formal citation for specutils.

If specutils is already being used in papers (and I believe it might be), then we should also consider an entry on ASCL.

eteq:
I'm leaving this open since @stscicrawford raises good ideas for future directions, but the needs for -.4 are met by #349

nmearl:
As mentioned above, specutils is being published on Zenodo and doi entries are being generated. However, we should still consider the case for a paper/entry on ASCL.

pllim:
There is https://github.com/astropy/specutils/blob/main/specutils/CITATION

PR #349, opened 2018-09-21
add citation entries

eteq:
This is partially a solution to #343, although @stscicrawford rightly suggests a bit more than that is called for down the road.  But it should suffice until/if there is a true journal paper to cite.

### DIFF
--- a/specutils/CITATION
+++ b/specutils/CITATION
@@ -1,0 +1,9 @@
+If you use Specutils for work/research presented in a publication (whether
+directly, or as a dependency to another package), please cite the Zenodo DOI
+for the appopriate version of Specutils.  The versions (and their BibTeX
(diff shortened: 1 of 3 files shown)


specutils/CITATION
```diff
@@ -0,0 +1,9 @@
+If you use Specutils for work/research presented in a publication (whether
+directly, or as a dependency to another package), please cite the Zenodo DOI
+for the appopriate version of Specutils.  The versions (and their BibTeX
+entries) can be found at:
+
+https://doi.org/10.5281/zenodo.1421356
+
+We also encourage the acknowledgement of Astropy, which is a dependency of
+specutils.  See
```
stscicrawford: See http://www.astropy.org/acknowledging.html

Issue #344, opened 2018-09-21
SpectrumCollection creation problem


skendrew:
I am trying to build a SpectrumCollection from a list of Spectrum1D objects, using the .from_spectra method, i.e.:

```
x = [Spectrum1D(flux=.....), Spectrum1D(flux=....)]
xc = SpectrumCollection.from_spectra(x)

```
The spectra are from a library of model spectra, they have spectral_axis, flux and meta but no uncertainties. spectral_axis and flux are given as Quantities. I'm getting an error that seems to point to a problem with creating a quantity from something that doesn't exist....? Error message is below.

```
xc = SpectrumCollection.from_spectra(x)
WARNING:root:Not all spectra have associated uncertainties, skipping uncertainties.
WARNING:root:Not all spectra have associated masks, skipping masks.
WARNING:root:No unit associated with uncertainty, assumingflux units of 'W / (Angstrom cm2)'.

TypeError                                 Traceback (most recent call last)
<ipython-input-48-ec642f1bf9cf> in <module>()
----> 1 xc = SpectrumCollection.from_spectra(x)

~/anaconda/envs/py36/lib/python3.7/site-packages/specutils-0.4.dev1210-py3.7.egg/specutils/spectra/spectrum_collection.py in from_spectra(cls, spectra)
    158 
    159         return cls(flux=flux, spectral_axis=spectral_axis,
--> 160                    uncertainty=uncertainty, wcs=wcs, mask=mask, meta=meta)
    161 
    162     @property

~/anaconda/envs/py36/lib/python3.7/site-packages/specutils-0.4.dev1210-py3.7.egg/specutils/spectra/spectrum_collection.py in __init__(self, flux, spectral_axis, wcs, uncertainty, mask, meta)
     75                 logging.warning("No unit associated with uncertainty, assuming"
     76                                 "flux units of '{}'.".format(flux.unit))
---> 77                 uncertainty = u.Quantity(uncertainty, unit=flux.unit)
     78 
     79             uncertainty = StdDevUncertainty(uncertainty)

~/anaconda/envs/py36/lib/python3.7/site-packages/astropy/units/quantity.py in __new__(cls, value, unit, dtype, copy, order, subok, ndmin)
    375                  isinstance(value.item(() if value.ndim == 0 else 0),
    376                             numbers.Number))):
--> 377             raise TypeError("The value must be a valid Python or "
    378                             "Numpy numeric type.")
    379 

TypeError: The value must be a valid Python or Numpy numeric type.

[...]

PR #352, opened 2018-09-21
Allow spectra collection creation without optional arguments

nmearl:
A fix for #344


eteq:
The failing test is unrelated... (not a major concern but will make an issue about it)

PR #356, opened 2018-09-28
Documentation: rework markup for an example

DougBurke:
I'm not sure if my proposed solution is how you want to mark it up, but it looked odd with the example text in the code block.


nmearl:
Thanks for catching this, @DougBurke. Certainly your changes would improve readability.

PR #358, opened 2018-10-03
Documentation fixes

larrybradley:
This PR fixes the errant "`specutils()`" on the main documentation page (it appears this was used for cross referencing?).  I fixed this, keeping the cross referencing intact.

specutils_docs

While testing the above, I discovered a few more documentation errors that I fixed.  The docs now build without (doc) warnings.


nmearl:
Thanks!

Issue #357, opened 2018-10-02
Fixed information lost in line fitting


nmearl:
Model fitting does not respect parameters that are marked as `fixed` in input models.

```python
n [45]: x = np.linspace(0., 10., 200)
    ...: y = 3 * np.exp(-0.5 * (x- 6.3)**2 / 0.8**2)
    ...: y += np.random.normal(0., 0.2, x.shape)
    ...: spectrum = Spectrum1D(flux=y*u.Jy, spectral_axis=x*u.um)
    ...:
    ...: # Fit the spectrum and calculate the fitted flux values (``y_fit``)
    ...: g_init = models.Gaussian1D(amplitude=3.*u.Jy, mean=6.1*u.um, stddev=1.*u.um)
    ...: g_fit = fit_lines(spectrum, g_init)

In [46]: g_fit
Out[46]: <Gaussian1D(amplitude=3.02092784 Jy, mean=6.28847159 um, stddev=0.79294043 um)>

In [47]: g_init = models.Gaussian1D(amplitude=3.*u.Jy, mean=6.1*u.um, stddev=1.*u.um, fixed={'mean': True})

In [48]: g_fit = fit_lines(spectrum, g_init)

In [49]: g_fit
Out[49]: <Gaussian1D(amplitude=3.02092784 Jy, mean=6.28847159 um, stddev=0.79294043 um)>
```

PR #360, opened 2018-10-18
Copied over fixed parameters to new models

brechmos-stsci:
The fixed parameter attribute of the model was not copied over when stripping and adding units. This is now fixed and a test was added based on the code in the issue below.

Fixes https://github.com/astropy/specutils/issues/357

### DIFF
--- a/specutils/fitting/fitmodels.py
+++ b/specutils/fitting/fitmodels.py
@@ -211,6 +211,19 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
 
     return fit_model
 
+def _convert(q, dispersion_unit, dispersion, flux_unit):
+    #
+    # Convert the quantity to the spectrum's units, and then we will use
+    # the *value* of it in the new unitless-model.
+    #
+
+    if q.unit.is_equivalent(dispersion_unit, equivalencies=u.equivalencies.spectral()):
+        quantity = q.to(dispersion_unit, equivalencies=u.equivalencies.spectral())
+
+    elif q.unit.is_equivalent(flux_unit, equivalencies=u.equivalencies.spectral_density(dispersion)):
+        quantity = q.to(flux_unit, equivalencies=u.equivalencies.spectral_density(dispersion))
+
+    return quantity
 
 def _convert_and_dequantify(poss_quantity, dispersion_unit, dispersion, flux_unit):
     """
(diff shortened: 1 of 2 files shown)


nmearl:
@brechmos-stsci Thanks for looking into this. I think we should also be sure that things like the `tied` and `bounds` parameters are all preserved.

brechmos-stsci:
Are there any others besides fixed, tied and bounds?  I'll add to this PR.

nmearl:
I believe that those three are the only things that influence the parameter behavior in model fitting.

brechmos-stsci:
Yeesh, this is going to get more complex.  It looks like I am going to have to do the same thing to the bounds (remove units on bounds, do fit, add units back on bounds).

eteq:
Quick thought about this: in the future (possibly even the upcoming astropy) there'll be options for more model attributes like `priors`, possibly some metadata, etc.  In principal that also needs to be transferred over.

I haven't looked close enough yet to know the details, but ideally this copying machinery would copy over *everything*.  Is that possible using e.g. `deepcopy` or similar?

brechmos-stsci:
I need to remember, but I think the problem is that once we do a `deepcopy` it is not possible to set some parameters as they are attributes and don't have setters. So, I don't **think** it is that easy.

brechmos-stsci:
Two options, either `_constraints` or loop through `constraints` and use `setattr` (per @eteq)

brechmos-stsci:
Ok, for this PR, I moved the convert-and-strip-units to a separate function as it was needed for the bounds constraint part. This is now in a function `_convert_and_dequantify`. As well, after a discussion with @eteq, I am copying the whole `model._constraints` dict so that things like `tied`, `bounds` etc and any future things in there will be copied to the new models.  I added some tests to confirm that these are being copied over correctly.

So.... this should be ready for review and merge. This is kinda necessary as technically the fitting isn't working properly (without `fixed` etc). 

@nmearl @eteq

specutils/fitting/fitmodels.py
```diff
@@ -212,6 +212,44 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
...
+    dispersion or flux units and then strip the units.
+
+    If the ``poss_quantity`` is None, or a number, we just return that...
+
+    Note: This method can be removed along with most of the others here
+          when astropy.fitting will fit models that contain units.
+
+    """
+
+    if poss_quantity is None or isinstance(poss_quantity, (float, int)):
```
eteq: You might want to add an `isinstance(poss_quantity, Quantity)` clause here. It's *possible* that in the future someone (or a future version of `astropy,modeling` would see Quantity's coming in directly...

brechmos-stsci: Uh, I don't quite understand. For that line we are specifically checking for *not* Quantities. is that what you meant?

specutils/fitting/fitmodels.py
```diff
@@ -212,6 +212,44 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
...
+        return poss_quantity
+
+    if hasattr(poss_quantity, 'quantity') and poss_quantity.quantity is not None:
```
eteq: I think my comment above might have better been here.  I think I meant a leading 
```
if isinstance(poss_quantity, Quantity):
    q = poss_quantity
elif ...
```

brechmos-stsci: I made a small change to include that check.

Issue #361, opened 2018-10-19
Line finding needed


brechmos-stsci:
@eteq and I talked about the need for a line finding method that would probably be needed before the actual fitting.  After talking with @eteq there are two methods that might be needed:

1. find lines - all this will do is find the dispersion locations that might appear to have a line
2. estimate parameters - this will take a list of lines (dispersion locations, possibly from 1.) and do a fast estimate of the parameters before calling the slower, iterative fitting routine

brechmos-stsci:
@eteq For the first one (find lines, meaning just dispersion points) will it be important to distinguish absorption lines from emission lines?  

So, for your line here:
![image](https://user-images.githubusercontent.com/18348847/47217558-c3d64000-d376-11e8-87a5-5256f6f4a201.png)

would it be better to return two lists, one with absorption and one with emission lines?

brechmos-stsci:
@eteq and I discussed and he suggested that qtable (from astropy) might be a better way to do it. It will be more exentisble for future things but remain reasonably simple.  So, I'll update the PR to output a quantity-table.

PR #362, opened 2018-10-19
Line Finding and Parameter Estimation

brechmos-stsci:
This is a first pass at line finding from within a spectrum. Two methods were added. 

A near-future push should include fast parameter estimation.

Fixes https://github.com/astropy/specutils/issues/361

### DIFF
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -50,6 +50,8 @@ Given we start with a spectrum:
    >>> spectrum = Spectrum1D(flux=y*u.Jy, spectral_axis=x*u.um)
 
 .. plot::
+    :include-source:
+    :align: center
 
    >>> from matplotlib import pyplot as plt
    >>> import numpy as np
@@ -67,7 +69,6 @@ Given we start with a spectrum:
    >>> x = np.linspace(0, 10, 200)
    >>> y = g1(x) + g2(x) + g3(x) + np.random.normal(0., 0.2, x.shape)
    >>> spectrum = Spectrum1D(flux=y*u.Jy, spectral_axis=x*u.um)
-   >>> plt.figure(1)
    >>> plt.plot(spectrum.spectral_axis, spectrum.flux)
    >>> plt.xlabel('Spectral Axis ({})'.format(spectrum.spectral_axis.unit))
    >>> plt.ylabel('Flux Axis({})'.format(spectrum.flux.unit))
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -170,11 +170,7 @@ An example where the parameter estimators are defined:
 
    >>> # Estimate parameter MexicanHat1D
    >>> mh = models.MexicanHat1D()
-   >>> estimators = {
-   >>>   'amplitude': lambda s: max(s.flux),
-   >>>   'x_0': lambda s: centroid(s, region=None),
-   >>>   'stddev': lambda s: fwhm(s)
-   >>> }
+   >>> estimators = { 'amplitude': lambda s: max(s.flux), 'x_0': lambda s: centroid(s, region=None), 'stddev': lambda s: fwhm(s) }
    >>> mh._constraints['parameter_estimator'] = estimators
 
    >>> g_init = estimate_line_parameters(spectrum, mh) #doctest:+SKIP
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -29,37 +29,17 @@ on calculating the derivative and then thresholding based on it.  Both technique
 return an `~astropy.table.QTable` that contains columns ``line_center``,
 ``line_type`` and ``line_center_index``.
 
-Given we start with a spectrum:
-
-.. code-block:: python
-
-   >>> import numpy as np
-   >>> from astropy.modeling import models
-   >>> import astropy.units as u
-   >>> from specutils import Spectrum1D, SpectralRegion
-   >>> from specutils.manipulation import noise_region_uncertainty
-
-   >>> # Define the spectrum
-   >>> np.random.seed(42)
-   >>> g1 = models.Gaussian1D(1, 4.6, 0.2)
-   >>> g2 = models.Gaussian1D(2.5, 5.5, 0.1)
-   >>> g3 = models.Gaussian1D(-1.7, 8.2, 0.1)
-   >>> x = np.linspace(0, 10, 200)
-   >>> y = g1(x) + g2(x) + g3(x) + np.random.normal(0., 0.2, x.shape)
-   >>> spectrum = Spectrum1D(flux=y*u.Jy, spectral_axis=x*u.um)
+We start with a synthetic spectrum:
 
 .. plot::
-    :include-source:
-    :align: center
+   :include-source:
+   :align: center
 
-   >>> from matplotlib import pyplot as plt
    >>> import numpy as np
    >>> from astropy.modeling import models
    >>> import astropy.units as u
    >>> from specutils import Spectrum1D, SpectralRegion
-   >>> from specutils.manipulation import noise_region_uncertainty
 
-   >>> # Define the spectrum
    >>> np.random.seed(42)
    >>> g1 = models.Gaussian1D(1, 4.6, 0.2)
    >>> g2 = models.Gaussian1D(2.5, 5.5, 0.1)
@@ -67,18 +47,21 @@ Given we start with a spectrum:
    >>> x = np.linspace(0, 10, 200)
    >>> y = g1(x) + g2(x) + g3(x) + np.random.normal(0., 0.2, x.shape)
    >>> spectrum = Spectrum1D(flux=y*u.Jy, spectral_axis=x*u.um)
+
+   >>> from matplotlib import pyplot as plt
    >>> plt.plot(spectrum.spectral_axis, spectrum.flux) # doctest: +IGNORE_OUTPUT
    >>> plt.xlabel('Spectral Axis ({})'.format(spectrum.spectral_axis.unit)) # doctest: +IGNORE_OUTPUT
    >>> plt.ylabel('Flux Axis({})'.format(spectrum.flux.unit)) # doctest: +IGNORE_OUTPUT
    >>> plt.grid('on') # doctest: +IGNORE_OUTPUT
 
-An example using the `~specutils.fitting.find_lines_threshold` where the uncertainty
-is not predefined in the spectrum but is added by calling the
-`~specutils.manipulation.noise_region_uncertainty` method:
+While we know the true uncertainty here, this is often not the case with real
+data.  Therefore, since `~specutils.fitting.find_lines_threshold` requires an
+uncertainty, we will produce an estimate of the uncertainty by calling the
+`~specutils.manipulation.noise_region_uncertainty` function:
 
 .. code-block:: python
 
-   >>> # Define a noise region for adding the uncertainty
+   >>> from specutils.manipulation import noise_region_uncertainty
    >>> noise_region = SpectralRegion(0*u.um, 3*u.um)
    >>> spectrum = noise_region_uncertainty(spectrum, noise_region)
 
@@ -178,13 +161,14 @@ Then estimate the line  parameters it it for a Gaussian line profile::
           1.1845669151078486 4.57517271067525 0.3015075376884422
 
 
-If an `~astropy.modeling.Model` is used that does not have the predefined parameter
-estimators, or if one wants to use different parameter estimators then one can create
-a dictionary where the key is the parameter name and the value is a lambda
-function that operates on a spectrum. For example if one wants to estimate the line
-parameters of a `~astropy.modeling.functional_models.MexicanHat1D` one can
+If an `~astropy.modeling.Model` is used that does not have the predefined
+parameter estimators, or if one wants to use different parameter estimators then
+one can create a dictionary where the key is the parameter name and the value is
+a function that operates on a spectrum (lambda functions are very useful for
+this purpose). For example if one wants to estimate the line parameters of a
+line fit for a `~astropy.modeling.functional_models.MexicanHat1D` one can
 define the ``estimators`` dictionary and attach in the model's ``_constraints``
-dictionary.
+dictionary:
 
 .. code-block:: python
 
@@ -196,7 +180,6 @@ dictionary.
    >>> sub_region = SpectralRegion(4*u.um, 5*u.um)
    >>> sub_spectrum = extract_region(spectrum, sub_region)
 
-   >>> # Estimate parameter MexicanHat1D
    >>> mh = models.MexicanHat1D()
    >>> estimators = { 'amplitude': lambda s: max(s.flux), 'x_0': lambda s: centroid(s, region=None), 'stddev': lambda s: fwhm(s) }
    >>> mh._constraints['parameter_estimator'] = estimators
@@ -213,9 +196,10 @@ dictionary.
        2.4220683957581444 3.6045476935889367   1.0
 
 .. warning::
-   Be aware the use of ``_constraints`` may change in
-   future versions of astropy or specutils, as this is a work around for a known
-   limitation in `astropy.modeling`.
+   Be aware the use of ``_constraints`` to store the estimators may change in
+   future versions of astropy or specutils to something more natural (i.e., not
+   a "private" attribute), as this is a workaround for a known limitation in
+   `astropy.modeling`.
 
 Model (Line) Fitting
 --------------------
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -46,8 +46,8 @@ An example using the `~specutils.fitting.find_lines_threshold`:
    >>> g2 = models.Gaussian1D(2.5, 5.5, 0.1)
    >>> g3 = models.Gaussian1D(-1.7, 8.2, 0.1)
    >>> x = np.linspace(0, 10, 200)
-   >>> y_double = g1(x) + g2(x) + g3(x) + np.random.normal(0., 0.2, x.shape)
-   >>> spectrum = Spectrum1D(flux=y_double*u.Jy, spectral_axis=x_double*u.um)
+   >>> y = g1(x) + g2(x) + g3(x) + np.random.normal(0., 0.2, x.shape)
+   >>> spectrum = Spectrum1D(flux=y*u.Jy, spectral_axis=x*u.um)
 
    >>> # Define a noise region for adding the uncertainty
    >>> noise_region = SpectralRegion(0*u.um, 3*u.um)
@@ -88,8 +88,8 @@ An example using the `~specutils.fitting.find_lines_derivative`:
    >>> g2 = models.Gaussian1D(2.5, 5.5, 0.1)
    >>> g3 = models.Gaussian1D(-1.7, 8.2, 0.1)
    >>> x = np.linspace(0, 10, 200)
-   >>> y_double = g1(x) + g2(x) + g3(x) + np.random.normal(0., 0.2, x.shape)
-   >>> spectrum = Spectrum1D(flux=y_double*u.Jy, spectral_axis=x_double*u.um)
+   >>> y = g1(x) + g2(x) + g3(x) + np.random.normal(0., 0.2, x.shape)
+   >>> spectrum = Spectrum1D(flux=y*u.Jy, spectral_axis=x*u.um)
 
    >>> # Define a noise region for adding the uncertainty
    >>> noise_region = SpectralRegion(0*u.um, 3*u.um)
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -117,8 +117,9 @@ Parameter Estimation
 Given a spectrum with a set of lines, the `~specutils.fitting.estimate_line_parameters`
 can be called to estimate the `~astropy.modeling.Model` parameters given a spectrum.
 
-For the `~astropy.modeling.models.Gaussian1D`, `~astropy.modeling.models.Voigt1D`, and
-`~astropy.modeling.models.Lorentz1D` models, there are predefined estimators for each 
+For the `~astropy.modeling.functional_models.Gaussian1D`, 
+`~astropy.modeling.functional_models.Voigt1D`, and
+`~astropy.modeling.functional_models.Lorentz1D` models, there are predefined estimators for each 
 of the parameters. For all other models one must define the estimators (see example below).
 
 For example:
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -18,18 +18,18 @@ components if desired).
 Line Finding
 ------------
 
-There are two methods implemented in order to find emission and/or absorption
+There are two techniques implemented in order to find emission and/or absorption
 lines in a `~specutils.Spectrum1D` spectrum.
 
-The first method is `~specutils.fitting.find_lines_threshold` that will
+The first technique is `~specutils.fitting.find_lines_threshold` that will
 find lines by thresholding the flux based on a factor applied to the
-spectrum uncertainty.  The second method is
+spectrum uncertainty.  The second technique is
 `~specutils.fitting.find_lines_derivative` that will find the lines based
-on calculating the derivative and then thresholding based on it.  Both methods 
+on calculating the derivative and then thresholding based on it.  Both techniques 
 return an `~astropy.table.QTable` that contains columns ``line_center``, 
 ``line_type`` and ``line_center_index``.
 
-An example using the `~specutils.fitting.find_lines_threshold`:
+Given we start with a spectrum:
 
 .. code-block:: python
 
@@ -49,11 +49,41 @@ An example using the `~specutils.fitting.find_lines_threshold`:
    >>> y = g1(x) + g2(x) + g3(x) + np.random.normal(0., 0.2, x.shape)
    >>> spectrum = Spectrum1D(flux=y*u.Jy, spectral_axis=x*u.um)
 
+.. plot::
+
+   >>> from matplotlib import pyplot as plt
+   >>> import numpy as np
+   >>> from astropy.modeling import models
+   >>> import astropy.units as u
+   >>> from specutils import Spectrum1D, SpectralRegion
+   >>> from specutils.manipulation import noise_region_uncertainty
+   >>> from specutils.fitting import find_lines_threshold
+
+   >>> # Define the spectrum
+   >>> np.random.seed(42)
+   >>> g1 = models.Gaussian1D(1, 4.6, 0.2)
+   >>> g2 = models.Gaussian1D(2.5, 5.5, 0.1)
+   >>> g3 = models.Gaussian1D(-1.7, 8.2, 0.1)
+   >>> x = np.linspace(0, 10, 200)
+   >>> y = g1(x) + g2(x) + g3(x) + np.random.normal(0., 0.2, x.shape)
+   >>> spectrum = Spectrum1D(flux=y*u.Jy, spectral_axis=x*u.um)
+   >>> plt.figure(1)
+   >>> plt.plot(spectrum.spectral_axis, spectrum.flux)
+   >>> plt.xlabel('Spectral Axis ({})'.format(spectrum.spectral_axis.unit))
+   >>> plt.ylabel('Flux Axis({})'.format(spectrum.flux.unit))
+   >>> plt.grid('on')
+
+An example using the `~specutils.fitting.find_lines_threshold` where the uncertainty
+is not predefined in the spectrum but is added by calling the 
+`~specutils.manipulation.noise_region_uncertainty` method:
+
+.. code-block:: python
+
    >>> # Define a noise region for adding the uncertainty
    >>> noise_region = SpectralRegion(0*u.um, 3*u.um)
    >>> spectrum = noise_region_uncertainty(spectrum, noise_region)
 
-   >>> # Derivative method
+   >>> # Derivative technique
    >>> lines = find_lines_threshold(spectrum, noise_factor=3)
 
    >>> emission_lines = lines[lines['line_type'] == 'emission'] #doctest:+SKIP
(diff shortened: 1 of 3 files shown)


brechmos-stsci:
@eteq This can be reviewed. Docs for line finding and parameter estimation have been added in too.

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+    else:
+        raise Exception('No method to estimate parameters')
+
+    return model
+
+def _consecutive(data, stepsize=1):
+    return np.split(data, np.where(np.diff(data) != stepsize)[0]+1)
+
+
+def find_lines_threshold(spectrum, sigma=3):
```
eteq: Somewhere in this docstring (and probably somewhere else?) you should note that this only works on continuum-subtracted spectra.

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+
+    #
+    # Create the QTable to return the lines
+    #
+
+    qtable = QTable()
+    qtable['lines'] = list(itertools.chain(*[spectrum.spectral_axis.value[emission_inds],
+                                             spectrum.spectral_axis.value[absorption_inds]]))*spectrum.spectral_axis.unit
+    qtable['line_type'] = ['emission']*len(emission_inds) + ['absorption']*len(absorption_inds)
+    qtable['index'] = list(itertools.chain(*[emission_inds, absorption_inds]))
```
eteq: perhaps the units here should be `u.pixel`?

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+        absorption_inds = [inds[np.argmin(spectrum.flux.value[inds])] for inds in line_inds_grouped]
+    else:
+        absorption_inds = []
+
+    #
+    # Create the QTable to return the lines
+    #
+
+    qtable = QTable()
+    qtable['lines'] = list(itertools.chain(*[spectrum.spectral_axis.value[emission_inds],
```
eteq: why is the `.value` here?  Don't we *want* the units?

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+    spectrum : Spectrum1D
+        The spectrum object over which the equivalent width will be calculated.
+
+    sigma : float
+        Threshold on line finding.
+
+    Returns
+    -------
+    qtable: `~astropy.table.QTable`
+        Table of emission and absorption lines. Spectral axis values, line type and indices stored.
```
eteq: Same idea of documenting this and the name scheme as the above function

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+    qtable['line_type'] = ['emission']*len(emission_inds) + ['absorption']*len(absorption_inds)
+    qtable['index'] = list(itertools.chain(*[emission_inds, absorption_inds]))
+
+    return qtable
+
+
+def find_lines_derivative(spectrum, sigma):
+    """
+    Find the emission and absorption lines in a spectrum. The method
+    here is a very simple threshold based on the standard deviation
```
eteq: This looks like the docstring for the above function instead of this one?

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+    """
+
+    # Take the derivative to find the zero crossings which correspond to
+    # the peaks (positive or negative)
+    kernel = [1, 0, -1]
+    dY = convolve(spectrum.flux, kernel, 'valid')
+
+    # Use sign flipping to determine direction of change
+    S = np.sign(dY)
+    ddS = convolve(S, kernel, 'valid')
```
eteq: this is clever! (no action needed :wink:)

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+        emission_inds = []
+
+    #
+    # Absorption lines
+    #
+
+    # Find all the indices that appear to be part of a -ve peak
+    candidates = np.where(dY < 0)[0] + (len(kernel) - 1)
+    line_inds = sorted(set(candidates).intersection(np.where(ddS == 2)[0] + 1))
+    line_inds = np.array(line_inds)[spectrum.flux[line_inds] < -sigma*spectrum.flux.unit]
```
eteq: Why is this called `sigma`? It doesn't seem like its related to the standard deviation... isn't it just a flux threshold (and if so, maybe it should be `flux_threshold`)?  And if so, it should be provided as a `Quantity` so that users can give whatever units they prefer (as long as they are compatible with the `spectrum.flux`)

specutils/tests/test_fitting.py
```diff
@@ -42,7 +46,41 @@ def double_peak():
...
+
+
+def test_find_lines_derivative():
+
+    # Create the spectrum to fit
+    x_double, y_double = double_peak_absorption_and_emission()
+    spectrum = Spectrum1D(flux=y_double*u.Jy, spectral_axis=x_double*u.um)
+
+    # Threshold method
+    lines = find_lines_threshold(spectrum, sigma=0.75)
```
eteq: the test is called derivative but this is for threshold?  I'd say you should either split these up or re-name the function.

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+#   a Quantity.
+
+parameter_estimators = {
+    'Gaussian1D': {
+        'amplitude': lambda s: max(s.flux),
+        'mean': lambda s: centroid(s, region=None),
+        'stddev': lambda s: fwhm(s) 
+    },
+    'Lorentzian1D': {
+        'amplitude': None,
```
eteq: Am I right that you're thinking to do a follow-on to add these?

specutils/tests/test_fitting.py
```diff
@@ -51,6 +89,27 @@ def test_single_peak_fit():
...
+    g_init = estimate_parameters(s_single, models.Gaussian1D())
+
+    assert np.isclose(g_init.amplitude.value, 3.354169257846847)
+    assert np.isclose(g_init.mean.value, 6.218588636687762)
+    assert np.isclose(g_init.stddev.value, 1.608040201005025)
+
+    assert g_init.amplitude.unit == u.Jy
+    assert g_init.mean.unit == u.um
+    assert g_init.stddev.unit == u.um
+
```
eteq: probably want to test the "non-supported model" case too, right?

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+        'x_0': None,
+        'fwhm':None 
+    },
+    'Voigt1D': {
+        'x_0': None,
+        'amplitude_L': None,
+        'fwhm_L': None,
+        'fwhm_G': None 
+    },
+    'Const1D': {
```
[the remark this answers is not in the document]
brechmos-stsci: Hmmm... maybe I am confused then. If the spectrum has a DC offset (for whatever reason) and one wants to fit a line on top of the DC offset, can one specify an offset in the Gaussian1D (for example) model? I didn't see something like that. If not, then how does one specify a fit like this?

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+        'amplitude': None 
+    }
+}
+
+def _set_parameter_estimators(model):
+    if model.__class__.__name__ in parameter_estimators:
+        model._constraints['parameter_estimator'] = parameter_estimators[model.__class__.__name__]
+    return model
+
+def estimate_parameters(spectrum, model):
```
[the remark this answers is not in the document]
brechmos-stsci: changed and added

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+# Define the initial estimators
+#   This is the default methods to use to estimate astropy model
+#   parameters. This is based on only a small subset of the astropy
+#   models but it was determined that this is a decent start as most
+#   fitting will probably use one of these.
+#
+#   Each method list must take a Spectrum1D object and should return
+#   a Quantity.
+
+parameter_estimators = {
```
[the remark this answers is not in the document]
brechmos-stsci: `-` makes sense

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+def find_lines_threshold(spectrum, sigma=3):
+    """
+    Find the emission and absorption lines in a spectrum. The method
+    here is a very simple threshold based on the standard deviation
+    of the noise.
+
+    Parameters
+    ----------
+    spectrum : Spectrum1D
+        The spectrum object over which the equivalent width will be calculated.
```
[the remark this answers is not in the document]
brechmos-stsci: changed

specutils/fitting/fitmodels.py
brechmos-stsci: good point. done

specutils/tests/test_fitting.py
brechmos-stsci: I split them up

specutils/fitting/fitmodels.py
brechmos-stsci: Oh, weird, I had added those. Huh, don't know what happened.  Will add (again)

brechmos-stsci: added estimators for each of them.

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+
+    sigma : float
+        Threshold on line finding.
+
+    Returns
+    -------
+    qtable: `~astropy.table.QTable`
+        Table of emission and absorption lines. Spectral axis values, line type and indices stored.
```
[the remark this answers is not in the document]
brechmos-stsci: I changed to `line_center`, `line_type`, and `line_center_index`.  I think those are more clear.

specutils/fitting/fitmodels.py
brechmos-stsci: added note to the docstring

specutils/fitting/fitmodels.py
brechmos-stsci: The units are appended at the end (next line). As I remember the itertools.chain or `*[list]` didn't like the quantities and wanted only floating point numbers.  So, it *does* have units, just had to be multiplied on (again) after the itertools part.

specutils/fitting/fitmodels.py
brechmos-stsci: Updated return  info in docstrings

specutils/fitting/fitmodels.py
brechmos-stsci: bad ctl-c, ctl-v form. fixed now.

specutils/fitting/fitmodels.py
brechmos-stsci: I would suggest that it should be unitless. The primary use case would be for doing indexing into a numpy array or list so probably easiest to just leave it as a list of numbers. If you feel strongly otherwise I can change it.

specutils/tests/test_fitting.py
brechmos-stsci: added one for MexicanHat1D and added more tests for the other "supported models"

specutils/fitting/fitmodels.py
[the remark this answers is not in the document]
eteq: Ah,  what I imagine as the flow is:

1. Get the continuum model
2. subtract the continuum
3. run the parameter estimators
4a. fit the lines to the subtracted spectrum
OR
4b .fit the lines+continuum to the original spectrum

But because the analysis tools operate on sub/normed spectrum, I think it's more consistent to assume the same for the line-estimator. 

We may want to later *add* the combined-estimator, but I think that's an add-on not v1

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,214 @@
...
+
+def _consecutive(data, stepsize=1):
+    return np.split(data, np.where(np.diff(data) != stepsize)[0]+1)
+
+
+def find_lines_threshold(spectrum, sigma=3):
+    """
+    Find the emission and absorption lines in a spectrum. The method
+    here is a very simple threshold based on the standard deviation
+    of the noise.
```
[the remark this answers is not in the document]
brechmos-stsci: @eteq If we use the `estimate_noise` idea then they are going to have to define the region over which the uncertainty is calculated. I could change the call so that they pass in a `SpectralRegion` that defines the noise region, though we would also probably want a factor on the noise estimate too (for thresholding). I am happy to do that if that is what you are thinking

docs/fitting.rst
```diff
@@ -15,6 +15,177 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+      4.522613065326634  emission                90
+      5.477386934673367  emission               109
+
+   >>> absorption_lines = lines[lines['line_type'] == 'absorption'] #doctest:+SKIP
+         line_center    line_type  line_center_index
+              um                                    
+      ----------------- ---------- -----------------
+      8.190954773869347 absorption               163
+
+
```
eteq:
```suggestion
While it might be surprising that these tables do not contain more information
about the lines, this is because the "toolbox" philosophy of `specutils` aims to
keep such functionality in separate distinct functions.  See :doc:`analysis` for
functions that can be used to fill out common line measurements more
completely.

```

docs/fitting.rst
```diff
@@ -15,6 +15,177 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+Parameter Estimation
+--------------------
+
+Given a spectrum with a set of lines, the `~specutils.fitting.estimate_line_parameters`
+can be called to estimate the `~astropy.modeling.Model` parameters given a spectrum.
+
+For the `~astropy.modeling.functional_models.Gaussian1D`, 
+`~astropy.modeling.functional_models.Voigt1D`, and
+`~astropy.modeling.functional_models.Lorentz1D` models, there are predefined estimators for each 
+of the parameters. For all other models one must define the estimators (see example below).
```
eteq:
```suggestion
of the parameters. For all other models one must define the estimators (see example below).
Note that in many (most?) cases where another model is needed, it may be better to create 
your own template models tailored to your specific spectra and skip this function entirely.
```

docs/fitting.rst
```diff
@@ -15,6 +15,177 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+
+.. code-block:: python
+
+   >>> import numpy as np
+   >>> from astropy.modeling import models
+   >>> import astropy.units as u
+   >>> from specutils import Spectrum1D
+   >>> from specutils.fitting import estimate_line_parameters
+
+   >>> # Define the spectrum
```
eteq: Why not use the same spectrum as above?  I think it might be clearer to the user if this section all uses one example spectrum.

docs/fitting.rst
```diff
@@ -15,6 +15,177 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+   >>> from specutils.fitting import find_lines_derivative
+
+   >>> # Define the spectrum
+   >>> np.random.seed(42)
+   >>> g1 = models.Gaussian1D(1, 4.6, 0.2)
+   >>> g2 = models.Gaussian1D(2.5, 5.5, 0.1)
+   >>> g3 = models.Gaussian1D(-1.7, 8.2, 0.1)
+   >>> x = np.linspace(0, 10, 200)
+   >>> y = g1(x) + g2(x) + g3(x) + np.random.normal(0., 0.2, x.shape)
+   >>> spectrum = Spectrum1D(flux=y*u.Jy, spectral_axis=x*u.um)
```
eteq: Maybe break this up into two code blocks, and have the first one be a matplotlib ``plot`` directive showing the spectrum so the user can actually see the spectrum that yields this table? (You can just copy-and-paste the idioms I used in getting started or analysis if you like)

docs/fitting.rst
```diff
@@ -15,6 +15,177 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+      Outputs: ('y',)
+      Model set size: 1
+      Parameters:
+              amplitude            mean             stddev     
+                  Jy                um                um       
+          ----------------- ----------------- -----------------
+          3.354169257846847 6.218588636687762 1.608040201005025
+
+
+An example where the parameter estimators are defined:
```
eteq: Maybe add a few more words here?  I.e., something about how you have to attach the estimators as `_constraints`, and that this is a behavior that may change in future versions of astropy and/or specutils (since it's a pretty ugly hack on some level).

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,246 @@
...
+
+
+def _consecutive(data, stepsize=1):
+    return np.split(data, np.where(np.diff(data) != stepsize)[0]+1)
+
+
+def find_lines_threshold(spectrum, noise_factor=1):
+    """
+    Find the emission and absorption lines in a spectrum. The method
+    here is a very simple threshold based on the standard deviation
```
eteq: this is no longer based on the stdev, right? Shouldn't it be something like "base on deviations significantly larger than the spectrum's uncertainty?

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,246 @@
...
+        ``noise_factor`` multiplied by noise estimate from ``noise_region``, used for
+        thresholding.
+
+    Returns
+    -------
+    qtable: `~astropy.table.QTable`
+        Table of emission and absorption lines. Line center (``line_center``),
+        line type (``line_type``) and index of line center (``line_center_index``) 
+        are stored for each line.
+    """
```
eteq: Maybe add something to either ``Notes`` or ``Examples`` that shows how you can combine this with `noise_region_uncertainty` (in one or two lines) to do this measurement even if the spectrum starts out without an uncertainty?

Could also go in `fitting.rst` - I think either would work.

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,246 @@
...
+    qtable['line_type'] = ['emission']*len(emission_inds) + ['absorption']*len(absorption_inds)
+    qtable['line_center_index'] = list(itertools.chain(*[emission_inds, absorption_inds]))
+
+    return qtable
+
+
+def find_lines_derivative(spectrum, flux_threshold):
+    """
+    Find the emission and absorption lines in a spectrum. The method
+    here is a based on finding the zero crossings in the derivative
```
eteq: (typo)
```suggestion
    here is based on finding the zero crossings in the derivative 
```

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,246 @@
...
+    here is a based on finding the zero crossings in the derivative 
+    of the spectrum.
+
+    Parameters
+    ----------
+    spectrum : Spectrum1D
+        The spectrum object over which the equivalent width will be calculated.
+
+    flux_threshold : float or `~astropy.units.Quantity`
+        Threshold on line finding.
```
eteq: I think this needs a bit more information.  Maybe something like:
```suggestion
        The threshold a pixel must be above to be considered part of a line. If
        a float, will assume the same units as ``spectrum.flux``.
```

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,246 @@
...
+        flux_threshold = float(flux_threshold) * spectrum.flux.unit
+
+    #
+    # Emmision lines
+    #
+
+    # Find all the indices that appear to be part of a +ve peak
+    candidates = np.where(dY > 0)[0] + (len(kernel) - 1)
+    line_inds = sorted(set(candidates).intersection(np.where(ddS == -2)[0] + 1))
+    line_inds = np.array(line_inds)[spectrum.flux[line_inds] > flux_threshold]
```
eteq: This is confusing to me.  Doesn't this have the effect of turning what *was* a "derivative" based method into a flux threshold-based finding method?

More broadly, doesn't this also mean that the spectrum has to be continuum-subtracted, because if it isn't, the `flux_threshold` includes some component from the continuum?

docs/fitting.rst
brechmos-stsci: will add

docs/fitting.rst
brechmos-stsci: added

specutils/fitting/fitmodels.py
brechmos-stsci: removed

specutils/fitting/fitmodels.py
brechmos-stsci: added

docs/fitting.rst
```diff
@@ -15,6 +15,177 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
 yielding a single composite model result (which can be split back into its
 components if desired).
 
+Line Finding
+------------
+
+There are two methods implemented in order to find emission and/or absorption
+lines in a `~specutils.Spectrum1D` spectrum.
+
+The first method is `~specutils.fitting.find_lines_threshold` that will
```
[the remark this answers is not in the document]
brechmos-stsci: changed all method to technique

docs/fitting.rst
brechmos-stsci: done. line finding and parameter estimation now all uses the same one.  The line fitting section below does not and I think that is fine as there are different types of fitting shown below.

docs/fitting.rst
brechmos-stsci: as I understand, this is now done as part of the previous comment

docs/fitting.rst
brechmos-stsci: added

specutils/fitting/fitmodels.py
brechmos-stsci: changed the description

specutils/fitting/fitmodels.py
brechmos-stsci: it actually is already in fitting.rst

specutils/fitting/fitmodels.py
brechmos-stsci: Well, the last line is a **filtering** step as it contains the `spectrum.flux[**line_inds**] > flux_threshold` and not `spectrum.flux > flux_threshold`.  One could have a strict derivative based method if one has `flux_threshold = 0`.

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,246 @@
...
+    if isinstance(flux_threshold, (int, float)):
+        flux_threshold = float(flux_threshold) * spectrum.flux.unit
+
+    #
+    # Emmision lines
+    #
+
+    # Find all the indices that appear to be part of a +ve peak
+    candidates = np.where(dY > 0)[0] + (len(kernel) - 1)
+    line_inds = sorted(set(candidates).intersection(np.where(ddS == -2)[0] + 1))
```
eteq: One other thing I just noticed: shouldn't these all be numpy set operations instead of python operations?

specutils/fitting/fitmodels.py
eteq: Ok, I gotcha now.  I think it still might need continuum-subtraction to be useful for many spectra *but* maybe not all now that you say this.

So then change this to:
1) Say in this argument's docstring that this is a threshold *above and beyond* the derivative-searching step 
2) Have an option for it to be `None`, which means "don't do the flux filtering at all". I realize this will yield lots of spurious lines, so maybe have it also add "center_flux" to the output table, which would just be the flux of that pixel (which you're already getting "for free" anyway). If that last part seems too painful don't worry about it though.

specutils/fitting/fitmodels.py
```diff
@@ -1,20 +1,246 @@
...
+    must be defined on the spectrum. To add the uncertainty, one could use
+    `~specutils.manipulation.noise_region_uncertainty` to add the uncertainty.
+
+    Parameters
+    ----------
+    spectrum : `~specutils.Spectrum1D`
+        The spectrum object in which the lines will be found.
+
+    noise_factor : float
+        ``noise_factor`` multiplied by noise estimate from ``noise_region``, used for
```
eteq: This doesn't look right... there's no ``noise_region`` anymore. 

```suggestion
        ``noise_factor`` multiplied by the spectrum's``uncertainty``, used for
```

eteq:
@brechmos-stsci - I think I get the derivative thing now and just had one only minor suggestion there.  But I think maybe you didn't push up your last round of changes?  I see resolved discussions but no new commits?

specutils/fitting/fitmodels.py
brechmos-stsci: I added a few comments to the documentation that should be sufficient for now.

specutils/fitting/fitmodels.py
brechmos-stsci: Ah, numpy has set operations?

specutils/fitting/fitmodels.py
brechmos-stsci: changed

docs/fitting.rst
```diff
@@ -15,6 +15,197 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+find lines by thresholding the flux based on a factor applied to the
+spectrum uncertainty.  The second technique is
+`~specutils.fitting.find_lines_derivative` that will find the lines based
+on calculating the derivative and then thresholding based on it.  Both techniques 
+return an `~astropy.table.QTable` that contains columns ``line_center``, 
+``line_type`` and ``line_center_index``.
+
+Given we start with a spectrum:
+
+.. code-block:: python
```
eteq: I don't think the code-block is necessary given the ``:include-source:`` below?

docs/fitting.rst
```diff
@@ -15,6 +15,197 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+
+For the `~astropy.modeling.functional_models.Gaussian1D`, 
+`~astropy.modeling.functional_models.Voigt1D`, and
+`~astropy.modeling.functional_models.Lorentz1D` models, there are predefined estimators for each 
+of the parameters. For all other models one must define the estimators (see example below).
+Note that in many (most?) cases where another model is needed, it may be better to create
+your own template models tailored to your specific spectra and skip this function entirely.
+
+
+For example, based on the spectrum defined above:
```
eteq:
```suggestion
For example, based on the spectrum defined above we can first select a region
in the spectrum corresponding to a line:
```

docs/fitting.rst
```diff
@@ -15,6 +15,197 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+.. code-block:: python
+
+   >>> from specutils import SpectralRegion
+   >>> from specutils.fitting import estimate_line_parameters
+   >>> from specutils.manipulation import extract_region
+
+   >>> sub_region = SpectralRegion(4*u.um, 5*u.um)
+   >>> sub_spectrum = extract_region(spectrum, sub_region)
+
+   >>> # Estimate parameter Gaussian1D
```
eteq:
```suggestion
Then estimate the line  parameters it it for a Gaussian line profile::
```

docs/fitting.rst
```diff
@@ -15,6 +15,197 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+   >>> g_init = estimate_line_parameters(sub_spectrum, models.Gaussian1D()) #doctest:+SKIP
```
eteq: Why skip the doctest here?  Definitely we want to check this so the user sees the right thing!
```suggestion
   >>> g_init = estimate_line_parameters(sub_spectrum, models.Gaussian1D())
   >>> g_init
```

docs/fitting.rst
```diff
@@ -15,6 +15,197 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+                  Jy                um                um
+          ------------------ ---------------- ------------------
+          1.1845669151078486 4.57517271067525 0.3015075376884422
+
+
+If an `~astropy.modeling.Model` is used that does not have the predefined parameter
+estimators, or if one wants to use different parameter estimators then one can create
+a dictionary where the key is the parameter name and the value is a lambda
+function that operates on a spectrum. For example if one wants to estimate the line
+parmaeters of a `~astropy.modeling.functional_models.MexicanHat1D` one can
```
eteq: (typo)
```suggestion
parameters of a `~astropy.modeling.functional_models.MexicanHat1D` one can
```

docs/fitting.rst
```diff
@@ -15,6 +15,197 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+define the ``estimators`` dictionary and attach in the model's ``_constraints`` 
+dictionary.  (This behavior may change in future versions of astropy or specutils).
```
eteq:
```suggestion
dictionary.  (Although be aware the use of ``_constraints`` may change in
future versions of astropy or specutils, as this is a work around for a known
limitation in `astropy.modeling`).
```

specutils/fitting/fitmodels.py
eteq: Yep, they're quite close to this but return arrays instead of set objects.  Can do this later though.

specutils/fitting/fitmodels.py
eteq: (also, for anyone reading this in the future: the next set of commits ended up doing the "skip if it's None" part)

docs/fitting.rst
```diff
@@ -15,6 +15,192 @@ initial guesses.  This model is then actually fit to the spectrum's ``flux``,
...
+.. code-block:: python
+
+   >>> from specutils import SpectralRegion
+   >>> from specutils.fitting import estimate_line_parameters
+   >>> from specutils.manipulation import extract_region
+
+   >>> sub_region = SpectralRegion(4*u.um, 5*u.um)
+   >>> sub_spectrum = extract_region(spectrum, sub_region)
+
+Then estimate the line  parameters it it for a Gaussian line profile::
```
neutrinoceros, line 150: @eteq just found this typo, but I don't know how to fix it (it it)

Issue #371, opened 2018-10-30
Ensure it's possible to create an "empty" Spectrum1D


eteq:
Right now creating a `Spectrum1D` with empty flux works, but `spectral_axis` does not (will create a PR with an example of this shortly).  This should be possible - after all, `np.array([])` and `u.Quantity([], u.Jy)` work as expected.  And to prove it's not just an idle exercise, see #367 for a case where it is likely necessary.

I would say the general policy should be we assume that "empty" `Spectrum1D` objects should be possible to create and using the basic accessors, but not necessarily expect that analysis tools or the like should be essentially "best effort" to work on them (since most results would be undefined anyway).

brechmos-stsci:
Summary... One *can* create an empty `Spectrum1D` object:
```
In [3]: tt = Spectrum1D(flux=[]*u.Jy, spectral_axis=[]*u.um)                                                                                                                                                     
WARNING:root:GWCS does not store rest frequency information. Please define the rest value explicitly in the `Spectrum1D` object.                                                                                 
WARNING:root:GWCS does not store rest wavelength information. Please define the rest value explicitly in the `Spectrum1D` object.                                                                                
                                                                                                                                                                                                                 
In [4]: tt.flux                                                                                                                                                                                                  
Out[4]: <Quantity [] Jy>                                                                                                                                                                                         
                                                                                                                                                                                                                 
In [5]: tt.spectral_axis                                                                                                                                                                                         
---------------------------------------------------------------------------                                                                                                                                      
ValueError                                Traceback (most recent call last)                                                                                                                                       
<ipython-input-5-e0f8841da2c3> in <module>                                                                                                                                                                        
----> 1 tt.spectral_axis                                                                                                                                                                                          
```                                                             
So the error is just in requesting the `spectral_axis` attribute if it is empty.

So, I will look into this error specifically.

brechmos-stsci:
@eteq At some level it would be even nicer to be able to do `>> tt = Spectrum1D()`. It would be easier to stick with the above (where you have to specify the empty-flux-with-units and empty-spectral_axis-with-units).

brechmos-stsci:
Huh, didn't realize this, you can create a `Spectrum1D` object with two different length `flux` and `spectral_axis`:
```
In [5]: tt = Spectrum1D(flux=[]*u.Jy, spectral_axis=[1,3,4]*u.um)
WARNING:root:GWCS does not store rest frequency information. Please define the rest value explicitly in the `Spectrum1D` object.
[...]

PR #372, opened 2018-10-30
add test for empty spectrum

eteq:
The PR is a companion to #371 - it's a test demonstrating that the bug in question exists.  Whoever fixes it should take this branch and build a fix on top of it - if we do that (and it's not me) it'll be awesome because we can call it an example of "distributed test-driven development".  Bonus trendiness points if you can figure out a way to use a blockchain somehow in the process.

(Note that this should not be merged as-is, as it is currently a test that will break master...)


eteq:
replaced by #373

Issue #367, opened 2018-10-29
extract_region sometimes returns None


eteq:
As detailed further in spacetelescope/specviz#471, `extract_region` returns None for some cases where a spectral wcs is provided.  This causes crashes when (understandably) user code assumes extract_region always returns a region.

@brechmos-stsci is already looking into this based on the discussion in spacetelescope/specviz#471, so assigning him here.

nmearl:
This seems to be due to the fact that if the `all_world2pix` transform fails, the `extract_region` appends a `None` to the spectrum list, and then return that. In the case that any region extraction fails, it ought to just return the input spectrum along with a logged error message.

brechmos-stsci:
Ummm..  but then what would you do if `extract_regions` is called and *one* of the SpectralRegions in the input list fails?  Would you want to return an error on `extract_regions` too?  I think it is incumbent on the user of `extract_region(s)` to check the output (ie None or not).  Though, having some type of error message might be helpful.

nmearl:
I think if a single extraction attempt fails (and the user has provided a list of potential extraction regions) _nothing_ should be appended to the list of regions, an error should be logged, and the extraction function should continue to the next region.

I suppose this is just a user experience thing. Either the `extract_regions` should _always_ return a `Spectrum1D` -- the original in the case that all extractions fail, or a combination of the regions in which the extraction succeeded -- or the `extract_regions` should return `None` when it reaches a failure.

I have a slight preference for the former, but can see where the latter would be desired (as it prevents users from assuming the operation succeeded if they neglect to look at the error message).

eteq:
I'm confused by this: what do we mean by "fail" here?  I would have thought the only possible "failure" is to ask for a region outside the spectrum, which should yield an "empty" spectrum or something like that.

hcferguson:
It seems okay for me for SpecUtils to raise an exception if the region goes outside the range of the spectrum. But then we should make it convenient for users to define one-sided regions that go "to the end" when they define them.

On the SpecViz side, we will need to include checking to make sure it is never passing something to SpecUtils that would go outside the range.

brechmos-stsci:
I want to be clear about what is there right now in the code. There are two methods:

1. `extract_region` which takes a `Spectrum1D` and a `SpectralRegion` and returns a `Spectrum1D` (of the extracted region.  Currently if this fails then None is returned (which is the discussion part).

[...]

PR #373, opened 2018-10-31
fix creation of empty spectrum, well, geting spectral_axis of empty spectrum

brechmos-stsci:
There was a need to create an empty spectrum, which works, but then when `empty_spectrum.spectral_axis` was called it threw errors and exceptions (though `empty_spectrum.flux` worked).  So, this PR fixes that issue.

Now, 
```
In [1]: from specutils import Spectrum1D

In [2]: from astropy import units as u

In [3]: tt = Spectrum1D(flux=[]*u.Jy, spectral_axis=[]*u.um)
WARNING:root:GWCS does not store rest frequency information. Please define the rest value explicitly in the `Spectrum1D` object.
WARNING:root:GWCS does not store rest wavelength information. Please define the rest value explicitly in the `Spectrum1D` object.

In [4]: tt.spectral_axis
Out[4]: <Quantity [] um>

In [5]: tt.flux
Out[5]: <Quantity [] Jy>
```

Fixes https://github.com/astropy/specutils/issues/371

### DIFF
--- a/specutils/spectra/spectrum_mixin.py
+++ b/specutils/spectra/spectrum_mixin.py
@@ -58,7 +58,13 @@ class OneDSpectrumMixin(object):
         """
         Returns a Quantity array with the values of the spectral axis.
         """
-        return self.wcs.pixel_to_world(np.arange(self.flux.shape[-1]))
+
+        if len(self.flux) > 0:
+            spectral_axis = self.wcs.pixel_to_world(np.arange(self.flux.shape[-1]))
+        else:
+            spectral_axis = []*self.wcs.spectral_axis_unit
+
+        return spectral_axis
 
     @property
     def spectral_axis_unit(self):
--- a/specutils/spectra/spectrum_mixin.py
+++ b/specutils/spectra/spectrum_mixin.py
@@ -63,10 +63,12 @@ class OneDSpectrumMixin(object):
             spectral_axis = self.wcs.pixel_to_world(np.arange(self.flux.shape[-1]))
         else:
             # After some discussion it was suggested to create the empty spectral
-            # axis this way to better use the WCS infrastructure. Changes in the mid
-            # term might necessitate a revisit of this.
-            dummy_spectrum=self.__class__(spectral_axis=[1,2]*self.wcs.spectral_axis_unit,
-                                          flux=[1,2]*self.flux.unit)
+            # axis this way to better use the WCS infrastructure. This is to prepare
+            # for a future where pixel_to_world might yield something more than
+            # just a raw Quantity, which is planned for the mid-term in astropy and
+            # possible gwcs.  Such changes might necessitate a revisit of this code.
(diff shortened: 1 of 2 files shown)


brechmos-stsci:
@eteq This PR is ready for review, too. This is needed for https://github.com/astropy/specutils/issues/367.

specutils/spectra/spectrum_mixin.py
```diff
@@ -58,7 +58,13 @@ def spectral_axis(self):
         """
         Returns a Quantity array with the values of the spectral axis.
         """
-        return self.wcs.pixel_to_world(np.arange(self.flux.shape[-1]))
+
+        if len(self.flux) > 0:
+            spectral_axis = self.wcs.pixel_to_world(np.arange(self.flux.shape[-1]))
+        else:
+            spectral_axis = []*self.wcs.spectral_axis_unit
```
eteq: In the future APE14 WCS world, I'm not sure this is going to be the same as what `pixel_to_world` should yield.  I wonder if it's safer to do something like this:
```
else:
    spectral_axis = self.wcs.pixel_to_world([0])[1:]
```

it's a bit of a hack but it seems like it might be more consistent with what `pixel_to_world` is doing otherwise?

brechmos-stsci: tbh, I would suggest we live this one as is and we can come back to change/check this another time.

eteq: Thinking about this again, the current implementation definitely will *not* work for some WCS's in the APE14 world, as there is not guarantee the output of `pixel_to_world` is a regular quantity object (although it *is* guaranteed to follow the indexing convention).  So I'm sticking with this being the safer way to do it.  My reasoning is basically that this is the sort of thing no one checks because it's a corner case. My reasoning is basically that this is the sort of thing no one checks because it's a corner case that will only show up for relatively unusual WCS's (which we aren't exhaustively testing for)

Although now that I've written that, I realize the alternate (possibly better?) solution is to specifically write the test to ensure the `spectral_axis` of the "empty spectrum" is of the same type as the output of the non-empty one, over all the tested input data sets.  That might be too much of a pain though?

brechmos-stsci: @eteq I am seeing exceptions, did you try it?

```
spectrum = Spectrum1D(spectral_axis=[]*u.um, flux=[]*u.Jy)
spectrum.wcs.pixel_to_world([0])[1:]
```
that fails with exceptions on bounding box etc etc. 

The first paragraph (and previous comment suggests changing the empty spectral_axis output. But the last paragraph suggests just changing the tests. Sorry, I am confused.  Should the code change or the tests. (Might be easier to discuss in person).

brechmos-stsci:
@eteq This now contains the new way of creating the empty spectrum axis to be returned to the user. Based on our discussion offline you suggested this might be more future proof as the return might contain more than just a `[]*u.spectral-axis-unit` kind of structure. (As I understood your reasoning :) )

eteq:
I added a bit more context in the comment, but it was just a comment, so merging now!

PR #376, opened 2018-11-01
Extend data loaders

nmearl:
Includes new data loaders from specviz. Also extends the data loader functionality to handle defining of filename extensions as well as priority.


docs/custom_loading.rst
```diff
@@ -94,6 +95,10 @@ ensure that the data file being loaded is compatible with the loader function.
         return Spectrum1D(flux=data, wcs=wcs, uncertainty=uncertainty, meta=meta)
 
 
+An ``extensions`` keyword can be provided. This allows for basic filename
```
eteq: This is for *creating* the extension part, but what about accessing it?  That is, if user code wants to ask things like "what are all the loaders for extension `fits`", that needs to be possible (and documented at least briefly here).

nmearl:
Alright. For tests, then, would you be able to setup a `specutils-data` repository to store test data?

As for author histories, we'll probably have to copy them from the v0.5 histories, since the current master histories have been "reset" essentially when we moved everything to an updated template and re-configured much of the loading functionality.

Agreed. The general fits loader can be removed. I think the documentation concerns go back to the what was talked about during the specviz sprint; mainly, that there needs to be a page describing the use of each loader and the expect input for correct parsing. Certainly though, docstrings and the like to allow users the ability to emulate the loaders should be a priority.

specutils/tests/conftest.py
```diff
@@ -1,11 +1,33 @@
...
+REMOTE_DATA_IDS = {
+    'cos_fuv': ('1481119', 'COS_FUV.fits')
+}
 
+REMOTE_DATA_NAMES = {
+    'COS_FUV.fits': '1481119'
+}
 
-@pytest.fixture(scope='module', params=REMOTE_DATA_NAMES)
+ACCESS_TOKEN = os.environ.get("ACCESS_TOKEN")
```
drdavella: Sorry for the nit, but can I suggest maybe `ZENODO_ACCESS_TOKEN` for the environment variable just to avoid an accidental collision.

specutils/tests/test_loaders.py
```diff
@@ -1,25 +1,55 @@
...
 
-from .conftest import remote_data
 from .. import Spectrum1D
+from .conftest import remote_data_path
 
 
-def test_spectrum1d_GMOSfits():
-    optical_fits_file = get_pkg_data_filename('data/L5g_0355+11_Cruz09.fits')
-    optical_spec_2 = Spectrum1D.read(optical_fits_file, format='wcs1d-fits')
+@pytest.mark.parametrize('remote_data_path',
```
drdavella: Just another nit but you could probably use `functools.partial` to reduce code duplication somewhat.

specutils/tests/conftest.py
```diff
@@ -4,30 +4,33 @@
...
-    'COS_FUV.fits': '1481119'
-}
-
 ACCESS_TOKEN = os.environ.get("ACCESS_TOKEN")
 
 
-@pytest.fixture(scope='module', params=REMOTE_DATA_IDS.values(), ids=REMOTE_DATA_IDS.keys())
-def remote_data(request):
+@pytest.fixture(scope='module')
+def remote_data_path(request):
```
drdavella: Maybe not for this PR, but if there's any chance we'd want to use this in `specviz` as well maybe this should be factored into a pytest plugin so it can be reused.

nmearl:
@drdavella @eteq any change either of you know of the failing issue on the LTS build. Alternatively, I was so tunnel visioned on using the Zenodo api, I neglected to see if I can just http the files... and, well, you can. So, yeah... 🦀

specutils/io/default_loaders/ascii.py
```diff
@@ -0,0 +1,114 @@
...
+    data: Spectrum1D
+        The data.
+    """
+    table = Table.read(file_name, format='ascii')
+
+    flux = None
+    dispersion = None
+    uncertainty = None
+
+    # If there are more than two columns, assume that the first is wavelength,
```
eteq: This seems very dangerous to me, as this is making assumptions that are very much not necessarily true and doing no validating.  This is the sort of reason why I was suggesting this "generic" reader not be in here...?

specutils/io/default_loaders/ascii.py
```diff
@@ -0,0 +1,114 @@
...
+    if name.lower().split('.')[-1] in ['txt', 'dat']:
+        return True
+
+    return False
+
+
+@data_loader(label="IPAC", identifier=ipac_identify, extensions=['txt', 'dat'])
+def ipac_loader(file_name, **kwargs):
+    """
+    Load spectrum from ASCII file
```
eteq: This looks like a docstring from above.  It's not just any ASCII file, but rather specifically an IPAC-style table

specutils/io/default_loaders/ascii.py
```diff
@@ -0,0 +1,114 @@
...
+    data: Spectrum1D
+        The data.
+    """
+    table = Table.read(file_name, format='ascii.ipac')
+
+    flux = None
+    dispersion = None
+    uncertainty = None
+
+    # If there are more than two columns, assume that the first is wavelength,
```
eteq: Again, an unsafe assumption... Many IPAC tables that carry spectrum-like things do not follow this rule.

specutils/io/default_loaders/ecsv.py
```diff
@@ -0,0 +1,45 @@
...
+    if name.lower().split('.')[-1] == 'ecsv':
+       return True
+
+    return False
+
+
+@data_loader(label="Spectrum ECSV", identifier=ecsv_identify, extensions=['ecsv'])
+def ecsv_spectrum_loader(file_name, **kwargs):
+    """
+    Load spectrum from ECSV file
```
eteq: I think this is much safer than the "generic ASCII" because it's unambiguous what the names of the table columns are.  *however*, the docstring needs to be much more specific.  It should say what column names it's expecting in the table.

specutils/io/default_loaders/ecsv.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Parameters
+    ----------
+    file_name: str
+        The path to the ECSV file
+
+    Returns
+    -------
+    data: Spectrum1D
+        The data.
```
eteq:
```suggestion
        The spectrum that is represented by the data in this table.
```

specutils/io/default_loaders/hst_cos.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Parameters
+    ----------
+    file_name: str
+        The path to the FITS file
+
+    Returns
+    -------
+    data: Spectrum1D
+        The data.
```
eteq:
```suggestion
        The spectrum that is represented by the data in this table.
```

specutils/io/default_loaders/hst_stis.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Parameters
+    ----------
+    file_name: str
+        The path to the FITS file
+
+    Returns
+    -------
+    data: Spectrum1D
+        The data.
```
eteq:
```suggestion
        The spectrum that is represented by the data in this table.
```

specutils/io/default_loaders/sdss.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Parameters
+    ----------
+    file_name: str
+        The path to the FITS file
+
+    Returns
+    -------
+    data: Spectrum1DRef
+        The data.
```
eteq:
```suggestion
        The spectrum that is represented by the data in this table.
```

specutils/io/default_loaders/hst_cos.py
```diff
@@ -0,0 +1,58 @@
...
+        if hdu[0].header['TELESCOP'] == 'HST' and hdu[0].header['INSTRUME'] == 'COS':
+           return True
+
+    return False
+
+
+@data_loader(label="HST/COS", identifier=cos_identify, extensions=['FITS', 'FIT', 'fits', 'fit'])
+def cos_spectrum_loader(file_name, **kwargs):
+    """
+    Load file from COS spectral data into a spectrum object.
```
eteq: This needs to be more specific.  There are lots of "COS files" - it should say exactly *which* COS file (I suspect it's the final result from the pipeline - e.g. what comes out of MAST, right?  If so, say that.)

specutils/io/default_loaders/hst_stis.py
```diff
@@ -0,0 +1,58 @@
...
+        if hdu[0].header['TELESCOP'] == 'HST' and hdu[0].header['INSTRUME'] == 'STIS':
+           return True
+
+    return False
+
+
+@data_loader(label="HST/STIS",identifier=stis_identify, extensions=['FITS', 'FIT', 'fits', 'fit'])
+def stis_spectrum_loader(file_name, **kwargs):
+    """
+    Load file from STIS spectral data into a spectrum object
```
eteq: As above: needs to be more specific so a user understands when it applies.

specutils/io/default_loaders/sdss.py
```diff
@@ -0,0 +1,129 @@
...
+    Registry.
+    """
+    return (isinstance(args[0], str) and
+            _spSpec_pattern.match(args[0]) is not None)
+
+
+@data_loader(label="SDSS-III/IV spec", identifier=spec_identify, extensions=['fits'])
+def spec_loader(file_name, **kwargs):
+    """
+    Loader for SDSS-III/IV spec files.
```
eteq:
```suggestion
    Loader for SDSS-III/IV optical spectrum "spec" files.
```
because technically APOGEE/MaNGA/etc are "SDSS spectrum files", but they use the name "optical spectrum" from what we're talking about here.

specutils/io/default_loaders/sdss.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Parameters
+    ----------
+    file_name: str
+        The path to the FITS file
+
+    Returns
+    -------
+    data: Spectrum1DRef
+        The data.
```
eteq:
```suggestion
        The spectrum that is represented by the data in this table.
```

specutils/io/default_loaders/tabular_fits.py
```diff
@@ -43,3 +45,14 @@ def tabular_fits(file_name, **kwargs):
         data = tab["flux"] * Unit("Jy")
```
eteq: this is definitely wrong! we could have a separate PR about this, but it is very much unsafe to assume all fluxes are in Jy!

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -23,15 +26,15 @@ def identify_wcs1d_fits(origin, *args, **kwargs):
...
 @data_loader("wcs1d-fits", identifier=identify_wcs1d_fits,
-             dtype=Spectrum1D)
-def wcs1d_fits(file_name, spectral_axis_unit=None, **kwargs):
+             dtype=Spectrum1D, extensions=['fits'])
+def wcs1d_fits_loader(file_name, spectral_axis_unit=None, **kwargs):
     """
-       Parameters
-       ----------
-       file_name : str
+    Parameters
```
eteq: this docstring needs an opening sentence.  Or more generally: what *is* this for?  That is, what *kind* of "WCS1D" are we talking about here?  I'm 95% sure think you mean WCS paper III spectral format, right? If so, awesome, but say that (with reference to the paper).

specutils/tests/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+
+@pytest.fixture(scope='module')
+def remote_data_path(request):
+    """
+    Remotely access the Zenodo deposition archive to retrieve the versioned
+    test data.
+    """
+    file_id, file_name = request.param.values()
+
+    f = requests.get("https://zenodo.org/record/{}/files/{}?download=1".format(
```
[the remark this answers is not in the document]
eteq: (Also, *very* minor nitpick: the variable shouldn't be `f` because it's not a file.  `res` is what I often use because it's the response to a request)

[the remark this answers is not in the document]
nmearl: I really like `requests` because it's the only package to implement a fully RESTful API, has a built-in json decoder, supports connection pooling, browser-style ssl verification, timeouts, thread safety, etc, etc. The list goes on. But I realize that that's not really necessary here, and the built-in library is probably sufficient.

specutils/io/default_loaders/ascii.py
nmearl: This was added from spacetelescope/specviz#503. The question would be then: do we add _no_ default loaders that make generic assumptions (at some point) about the user's data, or do we build in _some_ basic assumptions?

ASCII is probably the most difficult, but we could parse the file and see if the astropy ascii loader can detect _any_ unit information, and if not, then perhaps raise an exception about ambiguity; but if so, assume certain defined columns are flux/spectral_axis/uncertainty?

hcferguson: @eteq That was my suggestion to @SaOgaz, mostly motivated by having a way to make a guess that might not be terrible in the SpecViz context. But I'm okay with not having this loader as long as we have an wizard implemented in SpecViz that can help people with ascii or csv files, and we provide an example in the SpecUtils documentation of reading the file using astropy Table.read, adding units, and feeding it to Spectrum1D.

specutils/io/default_loaders/tabular_fits.py
hcferguson: @nmearl and @eteq I'd suggest folowing the model of generic_ecsv as a starting point for a generic_tabular_fits and a generic_ipac that call generic_table_from_spectrum to try to figure out which column is which based on the units. I can do this, but probably won't get to it today.

nmearl:
SDSS tests have been added.

> I'm not seeing any test for the WCS1D loader,

They are the first two tests in the test loaders file.

.travis.yml
```diff
@@ -32,11 +32,10 @@ env:
         - ASTROPY_VERSION=stable
         - MAIN_CMD='python setup.py'
         - SETUP_CMD='test'
-        - PIP_DEPENDENCIES='six gwcs scipy matplotlib'
+        - PIP_DEPENDENCIES='six gwcs scipy matplotlib requests'
         - EVENT_TYPE='pull_request push'
         - CONDA_DEPENDENCIES=''
         - CONDA_CHANNELS='astropy-ci-extras'
-        - secure: So72UFlPqF/qSLnyWiUzDAHP7zXUs7AU21/JmkWQvBzcpOpvXpsKXznk7OqXHT2g5egkTcVHFSzd4KGbBn6trggZ/UL3SWdXtRfywklZZtwG2hffL5I74TGOtXJzznbAGfAC7G2JMtCiB99mysz3EvcjPYvgm6VX//RYombup2w=
```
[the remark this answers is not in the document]
eteq: for posterity: via Slack @nmearl confirmed this was a mistaken check-in but not anything really worth worrying about from a security/privacy point of view.

specutils/io/default_loaders/ascii.py
eteq: To follow-through on this: in some slack discussion the conclusion was that it makes sense to allow the *generic* case but make it clear that this is an intentionally-limited "generic" that is roughly speaking the thing that the specviz wizard produces, but *no more* complicated than that.  That is, not put a lot of logic in the generic readers to figure out weird cases, but rather tell users that if it's not a simple column-to-spectrum1d attribute mapping, they should write a custom reader.  So sort of a middle ground (yay compromise!)

specutils/io/default_loaders/ascii.py
```diff
@@ -0,0 +1,129 @@
...
+        The path to the ASCII file.
+    column_mapping : dict
+        A dictionary describing the relation between the ASCII file columns
+        and the arguments of the `Spectrum1D` class, along with unit
+        information. The dictionary keys should be the ASCII file column names
+        while the values should be a two-tuple where the first element is the
+        associated `Spectrum1D` keyword argument, and the second element is the
+        unit for the ASCII file column::
+
+            column_mapping = {'FLUX': ('flux': 'Jy')}
```
eteq:
```suggestion
            column_mapping = {'FLUX': ('flux', 'Jy')}
```

docs/custom_loading.rst
```diff
@@ -77,7 +77,8 @@ ensure that the data file being loaded is compatible with the loader function.
                 os.path.splitext(args[0].lower())[1] == '.fits')
 
 
-    @data_loader("my-format", identifier=identify_generic_fits)
+    @data_loader("my-format", identifier=identify_generic_fits,
```
eteq: This is still missing a discussion of the "generic" vs "custom" distinction we were discussing.

specutils/io/default_loaders/wcs_fits.py
eteq: This docstring still does not tell the user what this loader is for - i.e., it doesn't address the questions I asked above in this comment.

nmearl: This is strange. I _did_ update this, and you can see the change [here](https://github.com/nmearl/specutils/blob/extend-data-loaders/specutils/io/default_loaders/wcs_fits.py#L30). It doesn't seem to be showing the github diff...

specutils/io/default_loaders/generic_cube.py
```diff
@@ -25,7 +27,7 @@ def identify_generic_fits(origin, *args, **kwargs):
             fits.getheader(args[0])['NAXIS'] == 3)
 
 
-@data_loader("cubetest1", identifier=identify_generic_fits)
+@data_loader("Cube", identifier=identify_generic_fits, extensions=['fits'])
```
eteq: This should probably be adjusted to use `column_mapping`?  or if that's not obvious how to do, perhaps this can just do what I've suggested here to just "hide" it.  Otherwise we are accidentally turning what is obviously a "test" into something that appears finished (by naming it "Cube"), and this is definitely *not* sufficient for the huge world of spectral cubes.

```suggestion
# not yet ready because it's not generic enough and does not use 
#@data_loader("Cube", identifier=identify_generic_fits, extensions=['fits'])
```

docs/custom_loading.rst
eteq: Deferred to a follow-on PR

specutils/io/default_loaders/tabular_fits.py
```diff
@@ -0,0 +1,96 @@
...
+        The path to the FITS file
+    column_mapping : dict
+        A dictionary describing the relation between the FITS file columns
+        and the arguments of the `Spectrum1D` class, along with unit
+        information. The dictionary keys should be the FITS file column names
+        while the values should be a two-tuple where the first element is the
+        associated `Spectrum1D` keyword argument, and the second element is the
+        unit for the ASCII file column::
+
+            column_mapping = {'FLUX': ('flux': 'Jy')}
```
eteq:
```suggestion
            column_mapping = {'FLUX': ('flux', 'Jy')}
```

specutils/io/default_loaders/tabular_fits.py
```diff
@@ -0,0 +1,96 @@
...
+    Load spectrum from a FITS file.
+
+    Parameters
+    ----------
+    file_name: str
+        The path to the FITS file
+    column_mapping : dict
+        A dictionary describing the relation between the FITS file columns
+        and the arguments of the `Spectrum1D` class, along with unit
+        information. The dictionary keys should be the FITS file column names
```
eteq: fits tables sometimes have units.  As with ECSV the way this should work is allow `{'FLUX': 'flux'}` to mean "use whatever units are present".  I would also argue that `{'FLUX': ('flux', someunit)}` should either just fail if untis are in the table, or try to auto-convert

specutils/io/default_loaders/wcs_fits.py
eteq: Right, I see it in the diff but what I'm saying is that it does not answer the above information. That is, "in FITS-files with defined WCS" is not sufficient because it's not clear what *kind* of WCS you mean.

That said, I don't want this to be a big blocker, so it's fine to defer this for later as long as you make an issue for it though, @nmearl .

specutils/io/default_loaders/generic_ecsv_reader.py
```diff
@@ -6,21 +6,33 @@
...
-        )
+            os.path.splitext(args[0].lower())[1] == '.ecsv')
+
 
 @data_loader("generic-ecsv", identifier=identify_ecsv, dtype=Spectrum1D)
 def generic_ecsv(file_name, **kwargs):
     """
     Read a spectrum from an ECSV file, using generic_spectrum_from_table_loader()
     to try to figure out which column is which.
+
```
eteq: It's a bit counter-intuitive that this doesn't have a `column_mapping` (I had a completely different comment I deleted after seeing this...).  So I suggest the following to reflect the difference for this function
```suggestion
    The ECSV columns must have units, as `generic_spectrum_from_table_loader`
    depends on this to determine the meaning of the columns.  For manual 
    control over the column to spectrum mapping, use the ASCII loader.

```

specutils/io/default_loaders/generic_ecsv_reader.py
```diff
@@ -6,21 +6,33 @@
...
+    file_name: str
+        The path to the ECSV file
+
+    Returns
+    -------
+    data: Spectrum1D
+        The spectrum that is represented by the data in this table.
     """
-    logging.info("Spectrum file looks like generic-ecsv")
+    logging.info("Spectrum file looks like 'generic-ecsv'.")
```
eteq: shouldn't this be "ECSV" instead of "generic-ecsv" given that the other readers have been renamed something like "ASCII"?

specutils/io/default_loaders/tabular_fits.py
nmearl: Okay. Currently, if the table contains unit information for a column, that unit information is used over anything defined in the column mapping. We can extend to auto converting in the future, but we'd have to parse physical type information so that we can pass in the correct equivalencies, so I think the fall back is fine.

specutils/io/default_loaders/generic_ecsv_reader.py
nmearl: I think it's better just to allow a column mapper here?

specutils/io/default_loaders/generic_ecsv_reader.py
nmearl: I'm not sure when these logging statements go popularized, but I don't think we should be including them in the loaders.

specutils/io/default_loaders/generic_ecsv_reader.py
hcferguson: Allowing a column mapper is fine. But the functionality of having it try to figure out the columns is still useful. One just needs to be clear about the logic if this loader is doing both. Meaning that the column mapper over-rides the attempt to identify the columns based on their units. 

Whatever is done here -- it seems like it should be identical for IPAC and tabular FITS, which are both tabular data that have support for units. So if we are going to use generic_table_from_spectrum if no column mapper is provided here, we should do the same in those loaders.

nmearl: @hcferguson well put. That's actually what's been implemented, and why I think we should continue exactly what you describe in this loader as well. As you've highlighted: all such tabular-style loaders will attempted to use a column mapper if provided, but will fallback to generating the spectrum from the table in cases where it's not defined.

specutils/io/default_loaders/generic_ecsv_reader.py
nmearl: Logging statements removed.

specutils/io/default_loaders/generic_cube.py
nmearl: Noted, removing from circulation for now.

specutils/io/default_loaders/wcs_fits.py
nmearl: I'm actually unsure of what you're really getting at here. Do you mean _kind_ as in "the assumptions about the structure of the WCS in terms of columns/hdus/etc", or do you mean kind as in "it's a FITS WCS"? The latter is plainly stated by the docstring, but the former would require the user to know what it's looking for in the fits wcs.

specutils/io/default_loaders/tabular_fits.py
eteq: Wait, so if the user gives a specific unit but the table conflicts it silently does what the table says? That's a classic UX mistake.  Can we maybe just raise an exception in that case and plan to do the better thing later as you say?

nmearl: We can do it the opposite way: prefer the column mapping over the table, but from your first comment

>  I would also argue that {'FLUX': ('flux', someunit)} should either just fail if untis are in the table

sounded like you preferred that the table take precedence over the column mapping. We can raise an exception if _both_ the table and the column mapping define a unit, or log something above preferring one over the other.

nmearl: Physical type information parsing has been added. If a user has defined a unit in the column mapping _and_ a unit is defined in the table already, specutils will attempt to convert the quantity array from the table to the unit defined in the column mapping.

eteq: I meant want I *really* wanted is something like this: `spec1dkwargs[mapping_value[0]] = table[mapping_key].to(mapping_value[1], whatevertherightequivalency)`.  That is, if the table has units, assume the file knows what its units are, but that the user is asking to *convert* them to whatever the explicitly asked for in `Spectrum1D`.  And if the value is just `'flux'`, trust that the table is correct.

But I'm fine with just raising an exception for now if figuring out `whatevertherightequivalency` is going to be too hard to do right now.

hcferguson: I think I agree with @eteq on the desired behavior, and that it's okay to just raise an exception for this release. 

A related question: Are we checking for and/or enforcing consistency between the units for the flux and the uncertainty? If nothing for this is coded yet, I'd suggest issuing a warning if they are in different units.

nmearl: @eteq Yes, this is what's already implemented. Although, I don't think the form of the column mapping dictionary should change if no unit is desired. I.e. it should be `{'FLUX': ('flux', None)}`.

hcferguson: @nmearl If someone were to feed flux in erg cm-2 s-1 AA-1 and uncertainties in Jy would calculations like SNR currently give the right answer? If not, perhaps the check should go into Spectrum1D itself, when the object is created?

nmearl: @hcferguson I think the best solution would be to be sure that all statistical functions are performed within an equivalency context manager. I don't know that a user who explicitly wants their flux and error in different units would appreciate either being told no, or having one or the other converted implicitly. We should just do the conversions at calculation time, for the same of the calculation, and leave the Spectrum1D in whatever state the user wants.

hcferguson: @nmearl Agreed. But what is it doing now?

nmearl:
@eteq @hcferguson I'm not sure we'll be able to do this before the release tomorrow. With the work that will probably need to done when the other PRs are merged, it seems this might need to be pushed to the next release.

nmearl:
@eteq whenever you get the chance. 😅

specutils/io/default_loaders/wcs_fits.py
eteq: I meant more the former, and that's precisely the point: this docstring has to be written for a user who has a file and is saying "is the appropriate for my file?"  And there's not enough information here for the user to know that.  That is, there are fits files out in the wild that are tabular but *also* have a WCS in the header.  This is not meant for that use case, because it assumes the `flux` data are the data in the "D" of the HDU.  It also assumes that hdu 0 is the only HDU of interest, etc.

I'll make a PR against your branch to demonstrate what I mean.  But the broader point is that once we make these "public", these docstrings have to have sufficient detail that the reader can understand what kind of file is needed without looking at the code itself (which is what I had to do).

eteq: (See nmearl/specutils#22)

specutils/io/parsing_utils.py
```diff
@@ -4,10 +4,77 @@
...
+        if tab_unit and cm_unit is not None:
+            # If the table unit is defined, retrieve the quantity array for
+            # the column
+            kwarg_val = u.Quantity(table[col_name], tab_unit)
+
+            # Attempt to convert the table unit to the user-defined unit.
+            logging.debug("Attempting auto-convert of table unit '%s' to "
+                          "user-provided unit '%s'.", tab_unit, cm_unit)
+
+            if cm_unit.physical_type in ['length', 'frequency']:
```
eteq: (minor performance gain)
```suggestion
            if cm_unit.physical_type in ('length', 'frequency'):
```

specutils/io/default_loaders/tabular_fits.py
eteq: I think this is a feature not a bug - that the `analysis` functions (which are *not* all statistical), don't make any strong assumptions nor try to guess equivalencies.  I think that's desirable because in *most* cases it does the right thing - e.g. equivalent widths end up with `spectral_axis` units *regardless* of the `flux` units.  

But this is pretty far-afield of this PR, so I'd suggest the discussion be moved to some other issue if any changes need to happen.

eteq: And I agree that this seems to do exactly what I said I wanted in https://github.com/astropy/specutils/pull/376#discussion_r233978718.  Not sure why I thought otherwise (perhaps I had an out-of-date view when I did part of this review?  Would explain some of the other rather confusing comments...)

Issue #374, opened 2018-10-31
argmax Bug in _compute_single_fwhm


robelgeda:
in [`_compute_single_fwhm`](https://github.com/astropy/specutils/blob/2dcbdafed18305e8233e1bd8667388b3b3e46181/specutils/analysis/width.py#L151) : 
If `flux` is all 0s and we do `argmax = np.argmax(flux)`,  `argmax` will be 0. The code then does `left = flux[:argmax] <= halfval` which means the slicing `flux[:argmax]` results in 0 elements. This causes a crash going down the code.

CC @brechmos-stsci

PR #377, opened 2018-11-01
fixed bug for highest flux at start or end of spectrum

brechmos-stsci:
There was an issue with a spectrum where the highest flux was at the beginning of the spectrum. This meant the `left` was empty which caused some issues. This should now be fixed in this code.

Fixes https://github.com/astropy/specutils/issues/374

### DIFF
--- a/specutils/tests/test_analysis.py
+++ b/specutils/tests/test_analysis.py
@@ -461,18 +461,18 @@ def test_fwhm():
 
 
     # Highest point at the first point
-    frequencies = np.linspace(1, 10, 100) * u.um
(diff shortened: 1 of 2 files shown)


specutils/tests/test_analysis.py
```diff
@@ -460,6 +460,23 @@ def test_fwhm():
     assert quantity_allclose(result, expected, atol=0.01*u.GHz)
 
 
+    # Highest point at the first point
+    frequencies = np.linspace(1, 10, 100) * u.um
```
eteq: (minor) this isn't actually a frequency - it's a wavelength, so the variable name is a bit confusing

PR #381, opened 2018-11-02
Add data loader for JWST spectra

drdavella:
This PR adds a new loader for JWST spectral data products.

It also adds a new lightweight container class called `SpectrumList`.

The essential problem is this: we need the JWST data loader to be able to return multiple spectra since any given JWST data product will contain an arbitrary number of spectra.

However, `SpectrumCollection` is not sufficient since it expects all spectra to have axes of the same length. This is not true for JWST data products since different slits may have slightly different dispersions.

I don't think that `SpectrumList` needs to be any more than a dumb container for `Spectrum1D` objects. It also plugs in to Astropy's IO machinery, which makes using the loader even more convenient.

### DIFF
--- a/docs/custom_loading.rst
+++ b/docs/custom_loading.rst
@@ -124,7 +124,7 @@ can be utilized by referencing its name in the ``read`` method of the
 .. _multiple_spectra:
 
 Loading Multiple Spectra
-========================
+^^^^^^^^^^^^^^^^^^^^^^^^
 
 It is possible to create a loader that reads multiple spectra from the same
 file. For the general case where none of the spectra are assumed to be the same
@@ -134,9 +134,9 @@ custom JWST data loader as an example:
 .. literalinclude:: ../specutils/io/default_loaders/jwst_reader.py
     :language: python
 
-Note that by default, any loader that uses `dtype=Spectrum1D` will also
+Note that by default, any loader that uses ``dtype=Spectrum1D`` will also
 automatically add a reader for `~specutils.SpectrumList`. This enables client
-code to call `~specutils.SpectrumList.read` in all cases if it can't make
+code to call ``specutils.SpectrumList.read`` in all cases if it can't make
 assumptions about whether a loader returns one or many `~specutils.Spectrum1D`
 objects.
 
--- a/docs/custom_loading.rst
+++ b/docs/custom_loading.rst
@@ -136,9 +136,10 @@ custom JWST data loader as an example:
 
 Note that by default, any loader that uses ``dtype=Spectrum1D`` will also
 automatically add a reader for `~specutils.SpectrumList`. This enables user
-code to call ``specutils.SpectrumList.read`` in all cases if it can't make
+code to call `specutils.SpectrumList.read` in all cases if it can't make
 assumptions about whether a loader returns one or many `~specutils.Spectrum1D`
-objects.
+objects. This method is available since `~specutils.SpectrumList` makes use of
+the Astropy IO registry.
 
 
 Creating a Custom Writer
--- a/docs/custom_loading.rst
+++ b/docs/custom_loading.rst
@@ -121,6 +121,25 @@ can be utilized by referencing its name in the ``read`` method of the
 
     spec = Spectrum1D.read("path/to/data", format="my-format")
 
+.. _multiple_spectra:
+
+Loading Multiple Spectra
+========================
+
+It is possible to create a loader that reads multiple spectra from the same
+file. For the general case where none of the spectra are assumed to be the same
+length, the loader should return a `~specutils.SpectrumList`. Consider the
+custom JWST data loader as an example:
+
+.. literalinclude:: ../specutils/io/default_loaders/jwst_reader.py
+    :language: python
+
+Note that by default, any loader that uses `dtype=Spectrum1D` will also
+automatically add a reader for `~specutils.SpectrumList`. This enables client
+code to call `~specutils.SpectrumList.read` in all cases if it can't make
+assumptions about whether a loader returns one or many `~specutils.Spectrum1D`
+objects.
+
 
 Creating a Custom Writer
 ------------------------
--- a/docs/custom_loading.rst
+++ b/docs/custom_loading.rst
@@ -136,10 +136,11 @@ custom JWST data loader as an example:
 
 Note that by default, any loader that uses ``dtype=Spectrum1D`` will also
 automatically add a reader for `~specutils.SpectrumList`. This enables user
-code to call `specutils.SpectrumList.read` in all cases if it can't make
-assumptions about whether a loader returns one or many `~specutils.Spectrum1D`
-objects. This method is available since `~specutils.SpectrumList` makes use of
-the Astropy IO registry.
+code to call `specutils.SpectrumList.read <astropy.nddata.NDIOMixin.read>` in
+all cases if it can't make assumptions about whether a loader returns one or
+many `~specutils.Spectrum1D` objects. This method is available since
+`~specutils.SpectrumList` makes use of the Astropy IO registry (see
+`astropy.io.registry.read`).
 
 
 Creating a Custom Writer
--- a/docs/custom_loading.rst
+++ b/docs/custom_loading.rst
@@ -135,7 +135,7 @@ custom JWST data loader as an example:
     :language: python
 
 Note that by default, any loader that uses ``dtype=Spectrum1D`` will also
-automatically add a reader for `~specutils.SpectrumList`. This enables client
+automatically add a reader for `~specutils.SpectrumList`. This enables user
 code to call ``specutils.SpectrumList.read`` in all cases if it can't make
 assumptions about whether a loader returns one or many `~specutils.Spectrum1D`
 objects.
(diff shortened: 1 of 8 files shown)


drdavella:
For now I just used FITS to read the data file, but eventually I think it would make to use ASDF instead (which also knows how to read JWST data). I just didn't want to introduce a new dependency on `specutils` at the moment.

eteq:
One thought re: `SpectrumList` - why not just literally a list of `Spectrum1D` objects?  The original thinking with `SpectrumCollection` vs `Spectrum1D` was that that is the recommended solution (although perhaps that should be written somewhere?)

drdavella:
@eteq that occurred to me too, but creating a new class allows us to make use of the Astropy IO registry infrastructure: e.g., we can do the following:

```python
data = SpectrumList.read(my_file, format='JWST')
```

This would not be possible if the loader just returns a plain old `list`.

eteq:
Ohh, gotcha.  Hmm, I see the reasoning then, @drdavella.  But I think a few words of this are then called for in the specutils docs.  Basically what you have in the docstring, but in this page: https://specutils.readthedocs.io/en/latest/types_of_spectra.html (item 4 I suppose).

drdavella:
@eteq that makes sense.

I'd actually like to propose a further change here: I think any loader that is defined with `dtype=Spectrum1D` should also implicitly create another loader for `dtype=SpectrumList`. This could easily be handled by the internals of `specviz.io.registers`. These new loaders would return a `SpectrumList` containing a single `Spectrum1D` object.

This would **greatly** simplify the implementation of data loading in `specviz` since it would provide a fully general interface for **any** number of spectra contained within a single data file. Rather than `specviz` having to determine whether to load a file using `Spectrum1D.read` or `SpectrumList.read`, it could simply always use the latter.

This use case is going to be extremely important for JWST data products which are guaranteed to contain an arbitrary number of spectra.

eteq:
>  think any loader that is defined with dtype=Spectrum1D should also implicitly create another loader for dtype=SpectrumList

I think this makes sense, *but* it might be better for it to be an opt-in rather than opt-out. Recall that `Spectrum1D` objects can be "lots of spectra with a shared spectral_axis", and in that case we don't want people to use `SpectrumList` for that case since it's a big performance penalty.

So, I think that makes three "categories" of loaders for multiple-spec-in-one-file:

1. `Spectrum1D`-only for anything that has a shared dispersion axis
2. `SpectrumCollection` if they have the same number of pixels but not same dispersion
3. `Spectrum1D` + `SpectrumList` (what @drdavella is proposing in https://github.com/astropy/specutils/pull/381#issuecomment-435551899) for those that are not the above

drdavella:
I agree with making it opt-in, but the question is how to do that since the loaders get instantiated at `specutils` import time.

drdavella:
Ah unless you mean passing an argument to the `data_loader` decorator. That probably makes sense, but all of the current default loaders would have to set it to `True`.

specutils/io/registers.py
```diff
@@ -34,6 +34,16 @@ def decorator(func):
 
         logging.debug("Successfully loaded reader \"{}\".".format(label))
 
+        # Automatically register a SpectrumList reader for any data_loader that
+        # reads Spectrum1D objects. TODO: it's possible that this
+        # functionality should be opt-in rather than automatic.
+        if dtype == Spectrum1D:
```
nmearl: This should probably be an `isinstance` call, not a value equivalency check.

drdavella: `isinstance` applies to instances, not types themselves 😉

drdavella: Maybe it should be using `is` instead of `==` though.

nmearl: Haha, my brain turned off for a moment. I meant `is`, not `isinstance` -- my fingers took control. Anyway, it's the same premise as checking for `None` -- `is` checks for object identity, while `==` checks value.

drdavella:
This should probably wait for #376 to be merged.

docs/custom_loading.rst
```diff
@@ -121,6 +121,25 @@ can be utilized by referencing its name in the ``read`` method of the
...
+It is possible to create a loader that reads multiple spectra from the same
+file. For the general case where none of the spectra are assumed to be the same
+length, the loader should return a `~specutils.SpectrumList`. Consider the
+custom JWST data loader as an example:
+
+.. literalinclude:: ../specutils/io/default_loaders/jwst_reader.py
+    :language: python
+
+Note that by default, any loader that uses ``dtype=Spectrum1D`` will also
+automatically add a reader for `~specutils.SpectrumList`. This enables client
```
eteq: (minor wording suggestion)
```suggestion
automatically add a reader for `~specutils.SpectrumList`. This enables user
```

docs/types_of_spectra.rst
```diff
@@ -42,7 +44,11 @@ with their corresponding ``specutils`` representations:
    not rectangular), this case does not have a specific representation in
    ``specutils``.  Instead, this case should be dealt with by making lists (or
    numpy object-arrays) of `~specutils.Spectrum1D` objects, and iterating over
-   them.
+   them. Specutils does provide a `SpectrumList` class which is
```
eteq: (minor)
After seeing this in the actual docs page I found the paragraph was getting a bit awkwardly long, so maybe just a stand-alone paragraph is better?  (Although my comment on the line below might lead you to changing this enough to negate this comment.)
```suggestion
   them. 
   
   Specutils does provide a `SpectrumList` class which is
```

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -0,0 +1,65 @@
...
+from ..registers import data_loader
+
+
+def identify_jwst_fits(origin, *args, **kwargs):
+    """
+    Check whether the given file is a JWST spectral data product.
+
+    This check is fairly simple. It expects FITS files that contain an ASDF
+    header (which is not used here, but indicates a JWST data product). It then
+    looks for at least one EXTRACT1D header, which contains spectral data.
```
eteq: I know I'm being a bit paranoid/overly future-proofing here, so you can feel free to ignore this if there's not an obvious answer... but perhaps it should actually look for some processing-specific metadata?  Like something that actually says "JWST" somewhere, or a list of instruments or something?

That is, suppose some of the JWST pipelines are re-used by some other mission. They might end up with `EXTRACT1D`, and it's counter-intuitive that the JWST reader would work on them.

docs/types_of_spectra.rst
```diff
@@ -42,7 +44,11 @@ with their corresponding ``specutils`` representations:
...
+   primarily intended for enabling data loaders that need to read multiple
+   heterogenous spectra (see :ref:`multiple_spectra`). Users should rarely need
```
eteq: I think you should add something here to indicate in what way it's different from a `list` (i.e., almost not at all).  I think you might be able to just get away with something like "Specutils does provide a SpectrumList class, but it is gives no user functionality beyond a list.  Rather, it enables data loaders..."

That is, right now when I read this trying to pretend to be a poor user, I was more confused because it leads me to think "well, if it's useful to them for loading, maybe it'll be useful to me - it must have more stuff after all!"

docs/custom_loading.rst
```diff
@@ -121,6 +121,25 @@ can be utilized by referencing its name in the ``read`` method of the
...
+code to call ``specutils.SpectrumList.read`` in all cases if it can't make
```
eteq: (minor)
```suggestion
code to call ``specutils.SpectrumList.read()`` in all cases if it can't make
```
or alternatively, I think there's some magic rst syntax that would let you link to `NDIOMixin.read`?

specutils/spectra/spectrum_list.py
```diff
@@ -0,0 +1,13 @@
...
+
+__all__ = ['SpectrumList']
+
+
+class SpectrumList(list, NDIOMixin):
+    """
+    A list that is used to hold a list of Spectrum1D objects
+
+    The primary purpose of this class is to allow loaders to return a list of
+    heterogenous spectra that do have a spectral axis of the same length.
```
eteq: As above, might be worth specifically saying that it is a list with the `NDIOMixin` added in.  It's obvious when I look at the code, but it's not as clear when one sees the docs (or if one were looking at the docstring  via `help(SpectrumList)` or similar.

docs/custom_loading.rst
drdavella: I'm not sure this is appropriate since `read` actually takes parameters, so it's not an accurate reflection of usage.

specutils/spectra/spectrum_list.py
drdavella: I think I accomplished this in a more generic way without referring to implementation details.

docs/custom_loading.rst
eteq: That's true, *but* it's a common convention in sphinx docs that means "this is a fuction/method".  See for example https://docs.python.org/3/tutorial/modules.html#the-dir-function .  Normally this happens automatically when you do ``:func:`whatever` ``, and I'm just thinking you can "mock that up" here.  But if you get the magical linking working it's a moot poi nt

eteq:
Nice trick with the `astropy.io.registry.read`!

PR #384, opened 2018-11-06
Region extract error vs empty spectrum

### LINKED ISSUE #367: extract_region sometimes returns None

brechmos-stsci:
This changes the functionality so that a SpectralRegion that is out of bounds of the Spectrum1D will return an empty Spectrum1D object (as opposed to None).  Any other exceptions are raised and passed through so the callers will need to catch the exceptions and deal with them appropriately.

Fixes #367 
Fixes https://github.com/spacetelescope/specviz/issues/471

### DIFF
--- a/specutils/tests/test_spectrum1d.py
+++ b/specutils/tests/test_spectrum1d.py
@@ -168,7 +168,7 @@ def test_wcs_transformations():
     assert isinstance(pix_axis, np.ndarray)
     assert isinstance(disp_axis, u.Quantity)
 
-    assert np.allclose(spec.wcs.world_to_pixel([7000*u.AA]), [461.2])
+    assert np.allclose(spec.wcs.world_to_pixel([7000*u.AA, 700*u.nm]), [461.2, 461.2])
 
 def test_create_explicit_fitswcs():
     my_wcs = fitswcs.WCS(header={'CDELT1': 1, 'CRVAL1': 6562.8, 'CUNIT1': 'Angstrom',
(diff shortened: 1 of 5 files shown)


brechmos-stsci:
This PR did double duty (not the best) and I *think* fixes an issue with the FITS WCS, but that file should be looked at very warily. It could be that it fixes a an issue but possibly in the wrong way. I am happy to back out that change if it is not the right place / method of fix.

specutils/wcs/adapters/fitswcs_adapter.py
```diff
@@ -45,7 +45,9 @@ def world_to_pixel(self, world_array):
         """
         Method for performing the world to pixel transformations.
         """
-        return self.axes.spectral.all_world2pix(world_array, 0)[0]
+        with u.set_enabled_equivalencies(u.spectral()):
+            world_array=u.Quantity(world_array, unit=self.spectral_axis_unit)
```
eteq: (PEP8 nitpick)

```suggestion
            world_array = u.Quantity(world_array, unit=self.spectral_axis_unit)
```

specutils/tests/test_spectrum1d.py
```diff
@@ -167,6 +168,8 @@ def test_wcs_transformations():
     assert isinstance(pix_axis, np.ndarray)
     assert isinstance(disp_axis, u.Quantity)
 
+    assert np.allclose(spec.wcs.world_to_pixel([7000*u.AA]), [461.2])
```
eteq: shouldn't this be `[7000]*u.AA` instead of `[7000*u.AA]` ?  Or if this is intentionally supposed to be the latter, maybe it should instead be something like `[7000*u.AA, 700*u.nm]` and then make sure those give the same pixel coordinate?

brechmos-stsci:
@eteq This is now fixed up and should be ready for approval.

PR #385, opened 2018-11-06
Fixed ck for empty, added test for issue

brechmos-stsci:
So, silly me. I did a wrong check for the non-empty condition. I used `len` which doesn't change when you do `len(a <=5)`, for example. It should be `sum(a<=5)`.  So, this PR is to fix that issue that came up and to add a test based on how this was found (flat spectrum).

### DIFF
--- a/specutils/analysis/width.py
+++ b/specutils/analysis/width.py
@@ -149,13 +149,13 @@ def _compute_single_fwhm(flux, spectral_axis):
     right = flux[argmax+1:] <= halfval
 
     # Highest signal at the first point
-    if sum(left) == 0:
+    if np.sum(left) == 0:
         l_idx = 0
     else:
         l_idx = np.where(left == True)[0][-1]
 
     # Highest signal at the last point
-    if sum(right) == 0:
+    if np.sum(right) == 0:
         r_idx = len(flux)-1
     else:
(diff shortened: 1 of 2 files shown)


specutils/analysis/width.py
```diff
@@ -149,13 +149,13 @@ def _compute_single_fwhm(flux, spectral_axis):
     right = flux[argmax+1:] <= halfval
 
     # Highest signal at the first point
-    if len(left) == 0:
+    if sum(left) == 0:
```
eteq:
```suggestion
    if np.sum(left) == 0:
```

Here (and below) I think you want to use `np.sum` - it's usually more performant for boolean arrays (and in principle this array is ~as big as the spectrum, right?

specutils/analysis/width.py
```diff
@@ -149,13 +149,13 @@ def _compute_single_fwhm(flux, spectral_axis):
...
         l_idx = 0
     else:
         l_idx = np.where(left == True)[0][-1]
 
     # Highest signal at the last point
-    if len(right) == 0:
+    if sum(right) == 0:
```
eteq:
```suggestion
    if np.sum(right) == 0:
```

brechmos-stsci: argh, will change

Issue #386, opened 2018-11-06
Add `ignore_units` parameter to the `fit_models` method


brechmos-stsci:
There has been some discussion that `fit_models` should be able to fit a Spectrum1D object to a model or compound model that does not have units.

So, the original idea is that we should add a parameter `ignore_units` on the `fit_models` method to tell the fitting to ignore the units on the model input as the initial guess. Then, if one says `ignore_units=True` then we just ignore the units on the model and implicitly the initial guess value for the parameters would then be "in the same unit as the spectrum to fit" (as the Spectrum1D object is required to have units).

But, maybe it is easier to just say that the input model can have units or not. If the input unit does NOT have units, then it is assumed the parameters have the same units as the Spectrum1D values. At some level this might make more sense, though less explicit.

@eteq @robelgeda

robelgeda:
This looks good. This is the best way to enable the fitting until the units machinery is in place.

brechmos-stsci:
There is discussion on the associated PR.  The `ignore_units` will be implicit depending on whether the input model has units or not. Rather than explicit with this parameter.

PR #387, opened 2018-11-06
Allow unit-less input models

brechmos-stsci:
This PR adds a parameter called `ignore_units` to the `fit_lines()` method.  This might be desired if one does not want to add units to the parameters of the input model but still wants to do fitting.

If the user has `ignore_units=True` in the `fit_lines()` method, then the model or compound model they are going to fit may have or may not have units. If it does have units then they will be stripped (but no conversion to spectrum units).  The output, fitted model will be unitless if `ignore_units=True`. It will be incumbant on the user to add units if they want them for some reason.

If the units are not going to be used on the model, then it is assumed all model parameters will effectively have the same units as what is on the spectrum that is going to be fit.

Fixes https://github.com/astropy/specutils/issues/386


brechmos-stsci:
@eteq This is ready for your review. I have added tests and @robelgeda grabbed this branch and tested it in his current version of specviz.  So, please review...

specutils/fitting/fitmodels.py
```diff
@@ -188,29 +236,75 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
...
-    model_unitless, dispersion_unitless, flux_unitless = _strip_units_from_model(model, spectrum)
+    model_unitless, dispersion_unitless, flux_unitless = _strip_units_from_model(model, spectrum, convert=not ignore_units)
 
     #
     # Do the fitting of spectrum to the model.
     #
 
-    fit_model_unitless = fitter(model_unitless, dispersion_unitless, flux_unitless)
+    fit_model_unitless = fitter(model_unitless, dispersion_unitless, flux_unitless,
+                                maxiter=maxiter, acc=acc, epsilon=epsilon)
```
eteq: in the future it is possible/likely that fitters have other keywords.  Maybe this is better as a `**kwargs` in `fit_lines` or at least a dictionary that gets passed in as `fitter(model_unitless, dispersion_unitless, flux_unitless, **fitkwargs)`?

specutils/tests/test_fitting.py
```diff
@@ -309,3 +310,77 @@ def test_fixed_parameters():
...
+    #
+
+    # Create the spectrum
+    x_single, y_single = single_peak()
+    s_single = Spectrum1D(flux=y_single*u.Jy, spectral_axis=x_single*u.um)
+
+    # Fit the spectrum
+    g_init = models.Gaussian1D(amplitude=3, mean=6.1, stddev=1.)
+    g_fit = fit_lines(s_single, g_init, ignore_units=True)
+    y_single_fit = g_fit(x_single)
```
eteq: Oops! I hit the wrong button on my review and didn't include this inline comment.  This is the one I meant to leave...:

```suggestion
    y_single_fit = g_fit(s_single.spectral_axis)
    assert y_single_fit.unit == s_single.flux.unit
```

The above should be the goal.  That is, even if the *parameters* have no units, the input to the model should be in the spectral axis units of the spectrum, and the output should be the flux.

specutils/fitting/fitmodels.py
brechmos-stsci: Agreed, it would have been really nice if astropy allowed these parameters to be defined at fitter instantiation. E.g., `fitter = LevMarLSQFitter(maxiter=200)`

brechmos-stsci:
@eteq Ok, tests passed and I **think** this PR is ready for another review. I have to admit I am getting less and less happy with all the dancing around we have to do in order to appease the astropy fitting machinery gods, as they are unhappy with units.

cc @robelgeda

specutils/fitting/fitmodels.py
```diff
@@ -188,29 +239,75 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
...
     return fit_model
 
+def _combined_region_data(spec):
+    if isinstance(spec, list):
+
+        # Merge sub-spec spectral_axis and flux values.
+        x = np.array([sv for subspec in spec if subspec is not None 
+                         for sv in subspec.spectral_axis.value])
+        y = np.array([sv for subspec in spec if subspec is not None 
+                         for sv in subspec.spectral_axis.flux])
```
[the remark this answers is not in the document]
brechmos-stsci: THANKS!

specutils/tests/test_fitting.py
```diff
@@ -309,3 +310,54 @@ def test_fixed_parameters():
...
+    y_single_fit = g_fit(x_single*u.um)
+
+    # Comparing every 10th value.
+    y_single_fit_expected = np.array([3.69669474e-13, 3.57992454e-11, 2.36719426e-09, 1.06879318e-07,
+                                      3.29498310e-06, 6.93605383e-05, 9.96945607e-04, 9.78431032e-03,
+                                      6.55675141e-02, 3.00017760e-01, 9.37356842e-01, 1.99969007e+00,
+                                      2.91286375e+00, 2.89719280e+00, 1.96758892e+00, 9.12412206e-01,
+                                      2.88900005e-01, 6.24602556e-02, 9.22061121e-03, 9.29427266e-04])
+
+    assert np.allclose(y_single_fit.value[::10], y_single_fit_expected, atol=1e-5)
```
eteq:
```suggestion
    assert np.allclose(y_single_fit.value[::10], y_single_fit_expected, atol=1e-5)
    assert y_single_fit.unit == s_single.flux.unit
```
I.e., need to make sure the *output* has the right units too. Could put that in the double_fit case also, although it should be the same machinery.

specutils/fitting/fitmodels.py
eteq: right, but I mean *now* it should still be ``kwargs`` in this specific call instead of explicitly using `maxiter`, `acc` and `epsilon`, since the fitter call's kwargs could change in the future.

specutils/fitting/fitmodels.py
```diff
@@ -37,6 +42,19 @@ def fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
         Regions of the spectrum to use in the fitting. If None, then the
         whole spectrum will be used in the fitting.
 
+    maxiter : int
```
eteq: Remove these three docstring entries and instead say 
```
additional keyword arguments are passed directly into the call to the ``fitter``
```
since there's no guarantees these specific arguments are available on the fitters now.

specutils/fitting/fitmodels.py
```diff
@@ -137,6 +164,10 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
         Regions of the spectrum to use in the fitting. If None, then the
         whole spectrum will be used in the fitting.
 
+    ignore_units : bool
+        If True, then ignore any units on the input model.
```
eteq:
```suggestion
        If True, then ignore any units on the input model parameters.
```

eteq:
LGTM now as long as the tests pass!

brechmos-stsci:
@eteq and I talked. He is good if I merge because the tests passed.

PR #389, opened 2018-11-07
Generic table loader

hcferguson:
This has a generic loader for tables `generic_spectrum_from_table` that tries to intuit which column is which based on units. It also includes a new generic_ecsv loader that uses that routine, and includes tests for both routines.


specutils/io/generic_spectrum_from_table.py
```diff
@@ -0,0 +1,134 @@
...
+import astropy.units as u
+from astropy.nddata import StdDevUncertainty
+from astropy.utils.exceptions import AstropyUserWarning
+import warnings
+
+from specutils.spectra import Spectrum1D
+
+def generic_spectrum_from_table(table, **kwargs):
+    """
+    Load spectrum an Astropy table.
```
nmearl: Minor: "Load spectrum _as_ an Astropy table."?

specutils/io/generic_spectrum_from_table.py
```diff
@@ -0,0 +1,134 @@
...
+        for c in columns_to_search:
+            try:
+                table[c].to("AA",equivalencies=u.spectral())
+                found_column = c
+                break
+            except:
+                continue
+
+        # If no success there, check for other possible length units
+        if found_column == None:
```
nmearl: `if found_column is None` -- want identity check, not equivalency check.

hcferguson:
Thanks. I've made those changes.

PR #390, opened 2018-11-08
FITS WCS deepcopy fix

nmearl:
This fixes a deep copying issue in which the adapter class was interfering with the underlying fits wcs copy handling.


robelgeda:
I can confirm this works. Thanks Nick!

PR #398, opened 2018-11-21
Remove stored test loader data; update loader test

nmearl:
Removes the stored test data file (that already exists in the Zenodo archive), and updates the `SpectrumList` test to use the remote file.


nmearl:
@drdavella fixed failures.

drdavella:
@nmearl looks good to me still, but my review is mostly meaningless here 😛.

nmearl:
Yeah, I realize now -- but I do appreciate you taking a look! 😉

specutils/tests/test_loaders.py
```diff
@@ -15,15 +15,11 @@
 from astropy.utils.data import get_pkg_data_filename
 
 from .. import Spectrum1D
-from .conftest import remote_data_path
+from .conftest import remote_data_path, remote_access
```
drdavella: Doesn't `pytest` automatically make these available within the test module namespace without an import?

nmearl: You are correct! Except for the remote_access function. That needs to be imported.

hamogu:
I'm clearly late to the party, but I don't understand while this file has been removed. The discussion on #223 concluded that test data should live in the repro and be made so small that size is not an issue (in the case of this file, keep the hdus[0].header, but make huus[0].data so tat it has < 100 numbers in it). That way, offline testing it not a problem.
(@eteq since he wrote the conclusion in #223).

PR #401, opened 2018-11-26
Allow tests that perform remote data access to be disabled

### LINKED ISSUE #399: 0.6: Tests fail when run without internet access
The specutils Debian package is built and build-time tested on a system without internet access. With version 0.6, a number of tests now fail or error; basically all from `specutils/tests/test_loaders.py`:

* `test_spectrum1d_GMOSfits[remote_data_path0]` (setup error)
* `test_specific_spec_axis_unit[remote_data_path0]` (setup error)
* `test_hst_cos[remote_data_path0/1]` (setup error)
* `test_hst_stis[remote_data_path0/1/2]` (setup error)
* `test_sdss_spec`
* `test_sdss_spspec`

Would it be possible to mark these tests so that they don't get executed in a disconnected environment?

drdavella:
This makes use of the [pytest-remotedata](https://github.com/astropy/pytest-remotedata) plugin in order to control tests that require access to remote data. By default all tests will run, including those that access remote data. To disable this, pass `--remote-data=none` to the `pytest` command.

This means that `pytest-remotedata` is now a test dependency.

This resolves #399. cc @olebole

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ env:
         - SETUP_CMD='test'
         - PIP_DEPENDENCIES='gwcs scipy matplotlib'
         - EVENT_TYPE='pull_request push'
-        - CONDA_DEPENDENCIES=''
+        - CONDA_DEPENDENCIES='pytest-remotedata'
         - CONDA_CHANNELS='astropy-ci-extras'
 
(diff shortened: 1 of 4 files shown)


drdavella:
@olebole do you have a chance to test this again?

PR #409, opened 2018-11-29
Intial changelog

eteq:
An initial cut at the changelog.  This means we need to start adding changelog entries for each PR.

We may want to experiment with [towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/quickstart.html), but to get something in for 0.5.1 this format should do the trick.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,0 +1,27 @@
+
+0.5.1 (unreleased)
+------------------
+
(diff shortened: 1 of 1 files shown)


SaOgaz:
Ended up cherry picking these changes due to some timing issues.  Closing this guy since those changes were committed.

Issue #415, opened 2018-12-03
`spectrum.flux /= x` raises error, but still changes value


hamogu:
Te code below follows the docs for continuum fitting (https://specutils.readthedocs.io/en/latest/fitting.html#continuum-fitting ) except that I also try to normalize a spectrum by the continuum:
```
import numpy as np
import matplotlib.pyplot as plt

from astropy.modeling import models
from astropy import units as u

from specutils.spectra import Spectrum1D, SpectralRegion
from specutils.fitting import fit_generic_continuum

np.random.seed(0)
x = np.linspace(0., 10., 200)
y = 3 * np.exp(-0.5 * (x - 6.3)**2 / 0.1**2)
y += np.random.normal(0., 0.2, x.shape)

y_continuum = 3.2 * np.exp(-0.5 * (x - 5.6)**2 / 4.8**2)
y += y_continuum

spectrum = Spectrum1D(flux=y*u.Jy, spectral_axis=x*u.um)

g1_fit = fit_generic_continuum(spectrum)

y_continuum_fitted = g1_fit(x*u.um)

plt.plot(x, y)
plt.plot(x, y_continuum_fitted)
plt.title('Continuum Fitting')
plt.grid(True)
```

So far, so good. 
Now I can print the flux values in the spectrum:
```
spectrum.flux
[1.9730844, 1.7201267, 1.8557246, …, 2.3072277, 2.5550603, 2.3695634]Jy
```
I now try to normalize the spectrum by deviding the flux by the fitted continuum values. I nthe docs, I can't find a description how to do that, so just do the naive thing:
```
spectrum.flux /= y_continuum_fitted
```
However, that raises an error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-90-c95d71413366> in <module>()
----> 1 spectrum.flux /= y_continuum_fitted

AttributeError: can't set attribute
```

So, question 1: How am I supposed to do a continuum normalization? I see that there are a few other issues looking into that e.g. #231. 

However, even more surprising, despite raising a `ValueError` it apparently worked ?!?
```
spectrum.flux
[1.1513992, 0.99465468, 1.0632936, …, 1.2593426, 1.4306224, 1.3625425]Jy
```
How can a statement raise an error and still be successful?

nmearl:
[...]

PR #420, opened 2018-12-04
Explain how to make a continuum normalized spectrum

hamogu:
In the fitting section, the docs talk quite a bit about how continuum
normalized spectra are useful and how the tools help to make them.
However, it is not shown how to actually normalize the spectrum.
I opened #415 because I could not find the answer after experimenting.
This doc change adds a simple example.

### DIFF
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -666,6 +666,10 @@ convenience functions to perform exactly this task.  An example is shown below.
     plt.title('Continuum Fitting')
     plt.grid('on')
 
+    # construct normalized spectrum by dividing the old spectrum
+    # by the continuum
+    spec_normalized = spectrum / y_continuum_fitted
+
 
 Reference/API
 -------------
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -637,6 +637,7 @@ convenience functions to perform exactly this task.  An example is shown below.
 .. plot::
     :include-source:
     :align: center
+    :context: close-figs
 
     import numpy as np
     import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


docs/fitting.rst
```diff
@@ -666,6 +666,10 @@ convenience functions to perform exactly this task.  An example is shown below.
     plt.title('Continuum Fitting')
     plt.grid('on')
 
+    # construct normalized spectrum by deviding the old spectrum
```
brechmos-stsci: deviding -> dividing

hamogu: @brechmos-stsci : Thanks! Done and force-pushed (this does not warrant a new commit).

docs/fitting.rst
```diff
@@ -666,6 +666,10 @@ convenience functions to perform exactly this task.  An example is shown below.
     plt.title('Continuum Fitting')
     plt.grid('on')
 
+    # construct normalized spectrum by dividing the old spectrum
```
nmearl: This is a little buried with it being a comment in a code block. I think you should bring this out of the code block and demonstrate the results of the normalization with its own code block and plot.

hamogu: The problem is that you can't continue the code from one plot directive in another plot directive. If I regenerate a random spectrum a second time, it'll look different and we don't want the normalized spectrum to look different from the original one. Could set seed for random, but that's kind of not what we want. So, instead, I made it a second plot. the code is still in the same code block, but the fact that there is a second plot makes is a little more obvious to look there.

hamogu: Thank you for your advice. I thought I had a reasonable understanding of the plot directive given that I actually contributed a PR to matplotlib for one of the features, but apparently I was wrong. I will revise my doc change yet again.

hamogu: How about this version?
(Let me also note that there was nothing like this in the docs before. Even if my addition does not highlight it perfectly, I think it still makes the docs better.)

hamogu:
I've restarted the Travis test, but get the same error on the Python 3.5 test, which I am *absolutely* certain has nothing to do with this PR (timeout at Zenodo).

Could someone please review with PR?

hamogu:
@astropy/specutils-maintainers The failures in the Travis test is unrelated and intermittent (404 from Zenodo). Can someone please review this PR?

Issue #380, opened 2018-11-01
Tests failing on astropy-dev


eteq:
See e.g. https://travis-ci.org/astropy/specutils/jobs/449547249 .  This looks like it might be some change in nddata.  Unclear if it's an astropy bug or `specutils` doing something it shouldn't have been doing with `nddata`, though.  Will investigate a bit more.

eteq:
`git bisect` reveals this as the commit at which failures started: astropy/astropy@fe10c9b9de0501a58b9751c60b78e5eb1c5011b0 (part of astropy/astropy#6971).

@mwcraig - any insight here?  Is this a "should fail" situation, and if so does it indicate need of a deprecation period or something?

cc @astrofrog and @bsipocz as it might impact astropy release/RC timing.

mwcraig:
@eteq -- the change that caused this failure was the addition of code to nddata to check for consistency between the unit and the uncertainty (the code in this [try/except block](https://github.com/astropy/astropy/pull/6971/files#diff-764f908a19bc78e920fb21695a526fe4R156)).

The line in the test that failed here is one in which the uncertainty generated from the data (whose unit is `Jy`) by calculating the variance of the data (see [this line](https://github.com/astropy/specutils/blob/master/specutils/tests/test_spectrum1d.py#L302)). That is then assigned to a `StdDevUncertainty` [in `estimate_uncertainty`](https://github.com/astropy/specutils/blob/master/specutils/manipulation/estimate_uncertainty.py#L51). The units of the uncertainty are `Jy2`, of course, since they are calculated from the variance.

When the `Spectrum1D` object is create [here](https://github.com/astropy/specutils/blob/master/specutils/manipulation/estimate_uncertainty.py#L54), the new code in nddata checks  whether the data and uncertainty units are compatible given the type of uncertainty (if they both have units). That check fails because a `StdDevUncertainty` is expected to have a unit convertible to the unit of the data.

The specific fix for this test would be to use a `VarianceUncertainty`...but looking at  [`estimate_uncertainty`](https://github.com/astropy/specutils/blob/master/specutils/manipulation/estimate_uncertainty.py#L51) it appears a `StdDevUncertainty` is always used, regardless of what the actual content of the uncertainty is. That was almost certainly a necessary hack, since `StdDevUncertainty` was the only concrete kind of uncertainty for a long time, but its use in that way should probably be reconsidered now that there are a couple of other options for storing uncertainty and doing propagation of uncertainties.

[...]

PR #429, opened 2019-01-02
Uncertainty estimation fix

brechmos-stsci:
The uncertainty estimation was adding a StdDevUncertainty object to the Spectrum1D's uncertainty even which was subsequently causing issues when checking the units. This PR fixes that by checking the output units of the noise estimation and selecting either StdDevUncertainty, VarianceUncertainty or InverseVariance depending on the units output from the uncertainty flux calculation.

This is related to comments in https://github.com/astropy/specutils/issues/380#issuecomment-435243599

Fixes: https://github.com/astropy/specutils/issues/380


brechmos-stsci:
@eteq  or anyone, if you can think of a better way to do this feel free to pass along and I can implement. I **think** this makes sense doing it this way at least from a unit point-of-view. 

This PR is assuming that flux.unit**2 is a VarianceUncertainty type of thing and 1/(flux.unit**2) is an InverseVariance type of thing. The only way for this not to make sense is if there is a noise estimation method that outputs units that match, for example, flux.unit**2 but is not a VarianceUncertainty type of thing.

brechmos-stsci:
@eteq I added several tests including ones for type checks on the uncertainty of the spectrum after calculating it.

**Odd**, though, I did a push and don't see the tests kick off.  Can you ck? I don't have admin on this repo.

Issue #430, opened 2019-01-03
Loader for Subaru-PFS object files


gpernot:
Loader for upcoming Subaru-PFS pfsObjects spectrum files, as defined in
https://github.com/Subaru-PFS/datamodel/blob/master/datamodel.txt

PR #431, opened 2019-01-03
Added PFS object loader. Closes astropy/specutils#430

gpernot:


specutils/io/default_loaders/pfs_spec.py
```diff
@@ -0,0 +1,80 @@
...
+from astropy.units import Unit
+from astropy.nddata import StdDevUncertainty
+
+import numpy as np
+
+from specutils.io.registers import data_loader
+from specutils import Spectrum1D
+
+__all__ = ['spec_identify', 'spec_loader']
+
```
SaOgaz: It would be great to get a comment here with a description of what this RE matching is doing.  Doesn't have to be super de-tailed, maybe just a couple examples of matching strings.

specutils/io/default_loaders/pfs_spec.py
```diff
@@ -0,0 +1,80 @@
...
+        The path to the FITS file
+
+    Returns
+    -------
+    data: Spectrum1D
+        The spectrum that is represented by the data in this table.
+    """
+    m = _spec_pattern.match(os.path.basename(file_name))
+
+    hdulist = fits.open(file_name, **kwargs)
```
SaOgaz: Change this to a `with hdulist as fits.open()`?

eteq:
Thanks for addressing all of this, @gpernot - all looks good to me, so merging!

PR #433, opened 2019-01-04
Fixed named parameters `from_center` and added several tests

### LINKED ISSUE #432: SpectralRegion.from_center fails
Looks like there might be a left-over from an older version of `SpectralRegion` that is making SpectralRegion.from_center fail:
```
>>> SpectralRegion.from_center(center=6563*u.AA, width=10*u.AA)
TypeError                                 Traceback (most recent call last)
<ipython-input-12-2f3ee56034ee> in <module>()
      2 HI_freq_region = SpectralRegion(1.4*u.GHz, 1.45*u.GHz)
      3 pixel_region = SpectralRegion(10*u.pixel, 20*u.pixel)
----> 4 ha_region = SpectralRegion.from_center(center=6563*u.AA, width=10*u.AA)

~/miniconda3/envs/astro36/lib/python3.6/site-packages/specutils/spectra/spectral_region.py in from_center(cls, center, width)
     46         """
     47 
---> 48         return cls(lower=center - width, upper=center + width)
     49 
     50     def __init__(self, *args):

TypeError: __init__() got an unexpected keyword argument 'lower'
```

I think this also means that there must not be a test from `from_center`, which probably should be fixed (above example should do the trick nicely, actually).

brechmos-stsci:
The `from_center` was not updated with the interface to `SpectralRegion` changed.  This was now fixed and several tests added.

Fixes #432

### DIFF
--- a/specutils/spectra/spectral_region.py
+++ b/specutils/spectra/spectral_region.py
@@ -45,7 +45,10 @@ class SpectralRegion:
            The width of the spectral region.
         """
 
-        return cls(lower=center - width, upper=center + width)
+        if width.value <= 0:
+            raise ValueError('SpectralRegion width must be positive.')
+
+        return cls(center - width, center + width)
 
     def __init__(self, *args):
         """
--- a/specutils/tests/test_regions.py
+++ b/specutils/tests/test_regions.py
@@ -1,3 +1,4 @@
+import pytest
 import numpy as np
 
 import astropy.units as u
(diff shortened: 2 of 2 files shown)


brechmos-stsci:
Passing tests are dependent on https://github.com/astropy/specutils/pull/429

brechmos-stsci:
@eteq  You can review this one now.

PR #434, opened 2019-01-06
Fix Gaussian Parameter estimation

eteq:
While doing some tests with specutils I noticed that the Gaussian parameter estimator has a bug where the `FWHM` estimator function is interpreted as a sigma (i.e., it's off by a factor of ~2.3). While the fix could have just been to rescale the FWHM to a sigma, I think for the Gaussian it makes more sense to use the `gaussian_sigma_width` estimator, which is specifically meant for Gaussians, so seems reasonably applicable.  

This PR implements that.

### DIFF
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -157,8 +157,8 @@ Then estimate the line  parameters it it for a Gaussian line profile::
       Parameters:
               amplitude            mean             stddev
                   Jy                um                um
-          ------------------ ---------------- ------------------
-          1.1845669151078486 4.57517271067525 0.3199324375933905
+          ------------------ ---------------- -------------------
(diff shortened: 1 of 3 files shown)


eteq:
Travis is now failing because it's picking up Astropy 4.0.  I made a test branch that's being run in https://travis-ci.org/eteq/specutils/builds/632595647 that forces it to use the older astropy/gwcs, and if that passes this is good to go (as @nmearl mentioned looking at it out-of-band, I believe)

eteq:
Turns out that revealed a bug, but that's now fixed and the remeaining failures in https://travis-ci.org/eteq/specutils/builds/632598019 are unrelated.  So I'm going to merge this.

PR #443, opened 2019-02-05
Support serialization of Spectrum1D and SpectrumList to ASDF

### LINKED ISSUE #442: Add ASDF tags and schemas for Spectrum1D and SpectrumList
This will allow `Spectrum1D` and `SpectrumList` to be serialized to ASDF. It will also probably be useful to add a generic ASDF data loader and dumper.

drdavella:
This PR adds ASDF tag implementations and schemas in order to enable `Spectrum1D` and `SpectrumList` objects to be serialized to ASDF files. 

It turns out that the `asdf` package is already an indirect package of `specutils` since it's a dependency of `gwcs`. However, I added a bit of logic here to ensure that the test suite continues to pass even if `asdf` is not installed. Maybe the correct solution is simply to add `asdf` as an explicit dependency now, though?

This resolves #442.

### DIFF
--- a/specutils/io/asdf/types.py
+++ b/specutils/io/asdf/types.py
@@ -1,0 +1,22 @@
+from asdf.types import CustomType, ExtensionTypeMeta
+
+
+_specutils_types = set()
+
+
+class SpecutilsTypeMeta(ExtensionTypeMeta):
+    """
+    Keeps track of `SpecutilsType` subclasses that are created so that they can
+    be stored automatically by specutils extensions for ASDF.
+    """
+    def __new__(mcls, name, bases, attrs):
+        cls = super().__new__(mcls, name, bases, attrs)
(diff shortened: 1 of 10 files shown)


specutils/io/asdf/types.py
```diff
@@ -0,0 +1,15 @@
+from asdf.types import CustomType
+
+
+_specutils_types = set()
+
+
+class SpecvizType(CustomType):
```
nmearl: Why is this called `SpecvizType` and not `SpecutilsType`?

drdavella: Because 🤦‍♂️

drdavella:
Py35 job is failing because I used some Py36 functionality because for some reason I thought we only supported Py36 and later. Will address shortly.

nmearl:
Looks good.

PR #447, opened 2019-02-14
Fix bug in creation of default identifiers for data loaders

drdavella:
Prior to this PR, if no identifier or list of compatible extensions was provided to the `data_loader` decorator, then an invalid identifier would be created. This fixes that bug and also adds more informative output for these cases.

It's worth noting that `specviz` now expects all loaders to have identifiers. So if a user creates a custom loader that has neither an explicit identifier nor a list of compatible extensions, it will not be possible to load the associated files using `specviz`.

We may want to consider adding an optional argument to `data_loader` that automatically creates a dummy identifier that positively identifies all files.

Thanks @nluetzge for providing the original bug report.

### DIFF
--- a/specutils/io/registers.py
+++ b/specutils/io/registers.py
@@ -59,7 +59,7 @@ def data_loader(label, identifier=None, dtype=Spectrum1D, extensions=None,
                              "explicit identifier. Creating identifier using "
                              "list of compatible extensions".format(
                                  label, dtype.__name__))
-                id_func = lambda *args, **kwargs: any([args[0].endswith(x)
+                id_func = lambda *args, **kwargs: any([args[1].endswith(x)
                                                           for x in extensions])
             # Otherwise, create a dummy identifier
             else:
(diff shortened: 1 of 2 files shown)


nmearl:
I believe this should also solve spacetelescope/specviz#672.

specutils/io/registers.py
```diff
@@ -48,19 +48,29 @@ def wrapper(*args, **kwargs):
...
+                             "list of compatible extensions".format(
+                                 label, dtype.__name__))
+                id_func = lambda *args, **kwargs: any([args[0].endswith(x)
+                                                          for x in extensions])
+            # Otherwise, create a dummy identifier
+            else:
+                logging.warn("'{}' data loader provided for {} without "
+                             "explicit identifier or list of compatible "
+                             "extensions".format(label, dtype.__name__))
+                id_func = lambda *args, **kwargs: False
```
drdavella: I could honestly see this just as easily being `True`, in which case it will blindly identify *any* file as compatible. I'm not entirely thrilled with either option.

A third way would be to simply raise an error in the case that neither an identifier nor a list of compatible extensions was provided. This would mean custom identifiers would be required to provide one or the other, though.

nmearl: Currently, the specviz loader wizard does not create identifier functions. This means that, right now, specviz cannot use any custom loader created through the loader wizard. I'm inclined to say that if no explicit identifier function is given, and no extension information is provided, blindly accept the input, and catch it later when/if it fails.

PR #449, opened 2019-02-15
Provide default data units in JWST data loader

drdavella:
These defaults are based on the units assigned by the `extract1d` step of the JWST pipeline (see [here](https://github.com/spacetelescope/jwst/blob/839accc191fe32ab2bfb85fc5e128fff98d82d1b/jwst/extract_1d/extract.py#L2581-L2589)).

While any data product that runs through the `extract1d` step should have valid unit metadata, we have encountered a few files that do not, so this is a useful precaution.


nmearl:
Do we know that an error array will always be provided?

PR #451, opened 2019-02-19
Add better air to vacuum wavelength conversion

tiagopereira:
Following the discussion at https://github.com/astropy/astropy/issues/8427, added new methods to convert from vacuum to air wavelengths. The new approach adds a `method` argument on `vac_to_air`, and uses a new function to calculate the index of refraction of air for given wavelengths. This makes it easier to add the data and formulation directly from the sources, avoiding any ambiguity. 

I also added different ways to calculate the inverse transformation (`air_to_vac`). Because the index of refraction depends on vacuum wavelengths, simply inverting the relation will introduce a small error, which can be significant for some applications. The new approach allows for three options: `inversion` - `1 / n`, same as before), `iteration` - iteratively solve for an index of refraction such that it is consistent with the inverse transform, within a certain user-defined  precision, and `piskunov` - following an approximate analytical expression derived by Nikolai Piskunov and [used in VALD](http://www.astro.uu.se/valdwiki/Air-to-vacuum%20conversion).

It can be debated if the wavelength conversion routines should be kept in the relatively obscure `utils.wcs_utils` or be put somewhere where a user could actually find them.

Another point is whether Greisen et al. (2006) should be set as the default method. I've kept it this way to ensure backwards compatibility, but I have strong reservations against using it for astronomical applications. See the notebook https://github.com/weaverba137/pydl/blob/master/docs/notebooks/Vacuum%20Wavelength%20Conversions.ipynb by @weaverba137  for some additional discussion on the methods. Greisen et al. (2006) differs significantly from all the methods used in astronomy over nearly a century. From my own experience, it also works very poorly in the UV (comparing vacuum and measured air wavelengths), and the authors themselves write that the expression only applies at 'normal optical wavelengths'. A few more points regarding the discussion in the notebook:

* Neither the Ciddor (1996) or Greisen et al. (2006) formulae have singularities below 200 nm. This is probably the wcslib implementation, and because historically the convention in spectroscopy is to not convert to air wavelengths below 200 nm (because the Earth's absorption is so high that they are unlikely to be observed from the ground). 
* The IAU standard (resolution No. C15, Commission 44, XXI GA, 1991) is to use the formula from Oosterhoff (1957), which in turn is just Edlen (1953). Sometimes people mention the IAU standard as following Morton (1991), but this is again just Edlen (1953).
* Ciddor (1996) should really be regarded as the better method. It works well in UV, optical, and infrared.
* More discussion in an [APOGEE Technical Note by Carlos Allende Prieto](http://www.as.utexas.edu/~hebe/apogee/docs/air_vacuum.pdf).

### DIFF
--- a/specutils/utils/wcs_utils.py
+++ b/specutils/utils/wcs_utils.py
@@ -363,26 +363,177 @@ def cdelt_derivative(crval, cdelt, intype, outtype, linear=False, rest=None):
         raise ValueError("Invalid in/out frames")
 
 
-def air_to_vac(wavelength):
+def refraction_index(wavelength, method='Griesen2006', co2=None):
     """
-    Implements the air to vacuum wavelength conversion described in eqn 65 of
-    Griesen 2006
-    """
-    wlum = wavelength.to(u.um).value
-    return (1+1e-6*(287.6155+1.62887/wlum**2+0.01360/wlum**4)) * wavelength
+    Calculates the index of refraction of dry air at standard temperature
+    and pressure, at different wavelengths, using different methods.
 
-def vac_to_air(wavelength):
-    """
-    Griesen 2006 reports that the error in naively inverting Eqn 65 is less
-    than 10^-9 and therefore acceptable.  This is therefore eqn 67
-    """
-    wlum = wavelength.to(u.um).value
-    nl = (1+1e-6*(287.6155+1.62887/wlum**2+0.01360/wlum**4))
-    return wavelength/nl
+    Parameters
+    ----------
+    wavelength : `Quantity` object (number or sequence)
+        Vacuum wavelengths with an astropy.unit.
+    method : str, optional
+        Method used to convert wavelengths. Options are:
+        'Griesen2006' (default) - from Greisen et al. (2006, A&A 446, 747),
+            eqn. 65, standard used by International Unionof Geodesy and Geophysics
+        'Edlen1953' - from Edlen (1953, J. Opt. Soc. Am, 43, 339). Standard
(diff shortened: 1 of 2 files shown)


specutils/utils/wcs_utils.py
```diff
@@ -363,26 +363,177 @@ def cdelt_derivative(crval, cdelt, intype, outtype, linear=False, rest=None):
...
+    -------
+    vac_wavelength : `Quantity` object (number or sequence)
+        Vacuum wavelengths with the same unit as wavelength.
+    """
+    VALID_SCHEMES = ['inversion', 'iteration', 'piskunov']
+    assert isinstance(scheme, str), 'scheme must be a string'
+    scheme = scheme.lower()
+    if scheme == 'inversion':
+        refr = refraction_index(wavelength, method=method, co2=co2)
+        #return wavelength * refr
```
nmearl:
```suggestion

```

specutils/utils/wcs_utils.py
```diff
@@ -363,26 +363,177 @@ def cdelt_derivative(crval, cdelt, intype, outtype, linear=False, rest=None):
...
+    elif scheme == 'piskunov':
+        wlum = wavelength.to(u.angstrom).value
+        sigma2 = (1e4 / wlum)**2
+        refr = (8.336624212083e-5 + 2.408926869968e-2 / (130.1065924522 - sigma2) +
+                1.599740894897e-4 / (38.92568793293 - sigma2)) + 1
+        #return wavelength * refr
```
nmearl:
```suggestion
```

tiagopereira:
Thanks for the fixes, @nmearl!

PR #460, opened 2019-03-09
Ensure figures are closed between renders (fixes #459); fix fitswcs comparisons

### LINKED ISSUE #459: Introductory example in documentation fails in standard Python/iPython terminal due to units mismatch in plots
The "Getting started with specutils" example on the main page of the documentation fails when commands are typed (or copy & pasted) into a standard Python or iPython terminal session at the second plot command, with a UnitConversionError:

    UnitConversionError: 'Unit(dimensionless)' is not a scaled version of 'Unit("erg / (Angstrom cm2 s)")'

followed by another:

    UnitConversionError: '' (dimensionless) and 'erg / (Angstrom cm2 s)' (spectral flux density wav) are not convertible

This appears to be because the initial use of `quantity_support()` and the first plot command

    >>> lines = plt.step(spec.spectral_axis, spec.flux)

sets the current plotting context to expect x and y values with units of (or compatible with) `spec.spectral_axis` ("Angstrom") and `spec.flux` ("erg / (Angstrom cm2 s)"), respectively. Then the second plot command

    >>> lines = plt.step(cont_norm_spec.wavelength, cont_norm_spec.flux)

sends x and y values with units of "Angstrom" and "(dimensionless)". Since the latter cannot be converted into "erg / (Angstrom cm2 s)", there is a UnitConversionError if the same plotting context is being used.

I'm guessing the example is derived from a Jupyter notebook. Since each separate cell in a notebook creates a new plotting context, you shouldn't see this problem in a notebook. (The second figure in the example has no y-axis label, which I take as evidence that this is what's going on.) But if the example is typed into a standard Python/iPython session, the second plot command will go to the *same* plot window and context as the first command, leading to the UnitConversionError.

The simple solution is to either create a new plot window for the second command, or call `plt.clf()` before calling the second plot command.

nmearl:


CHANGES.rst
```diff
@@ -15,6 +15,15 @@ Bug Fixes
...
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+0.5.3
+-----
+
+Bug Fixes
+^^^^^^^^^
+
+- Fix comparison of FITSWCS objects in arithmetic operations.
+- Fix example documentation when run in python interpreter.
```
eteq:
```suggestion

- Fix example documentation when run in python interpreter.
```

Issue #359, opened 2018-10-07
linear to log spectral axis conversion


skendrew:
Is there the ability to convert a Spectrum1D object's spectral axis from linear to log space (and reverse), with flux conservation?  I have seen other issues (e.g. #255) discussing resampling/rebinning with or without flux conservation, but I haven't seen this particular case discussed. There are probably numerous implementations of this case out there (e.g. Michele Cappellari's ppxf package has a function called log_rebin - [online docs here](http://www-astro.physics.ox.ac.uk/~mxc/software/#ppxf)), would be great to have this integrated into specutils. Its a common step in a lot of analysis that involved radial velocities.

eteq:
Does not exist, but I agree it's a good idea, @skendrew!

In my experience there's not a "single agreed-upon" way to do this kind of rebinning, though.  Do you agree, @skendrew?  My default would be to assume the user usually wants the flux-preserving rebinning that's discussed in #255, but onto a log-spaced axis.  If that sounds good to you I think that's a good thing to try to target for the next major release.

skendrew:
this is definitely similar to the discussion in #255 so I have posted further comment there.

Issue #255, opened 2018-07-23
Resampling (flux-preserving and non-flux-preserving interpolation)


SaOgaz:
notecard (i.e., the title) first put in by @keflavich

eteq:
Note that this is intertwined with #254 because this is probably the "pre-step".

More broadly, my current thinking is that this should be *several* functions, each with the same API, but with different algorithms.  E.g.,

```python
newspec = resample_flux_preserving(spectrum, new_spectral_axis)
newspec = resample_density_preserving, new_spectral_axis)
newspec = resample_interpolation, new_spectral_axis)
```
``new_spectral_axis`` should be either a spectral axis *or* a wcs, but it may be easier to implement that as two separate stages (since a general WCS implementation might be really complicated to get right with some of these algorithms).

They should *also* do the right thing w.r.t. the ``uncertainty`` - but the user should have the power to control their definition of "the right thing".  My feeling is that initially there should be a ``weights`` (or possibly ``uncertainty``?) keyword that accepts either `None`, an array (shape matching `flux`), or the string ``'unc'`` (or ``ivar``?).  The last one would be the default, and means "use the inverse variance from the uncertainty as the weights".  In all cases the new spectrum would get a new uncertainty array based on whatever is right for the specific algorithm.

eteq:
@nmearl - I'm advocating against the "keyword controls the algorithm" approach. This tends to lock you into limitations like bot returning extra information or confusing users about exactly what they are getting. This is particularly true in this case once you consider understanties.

An alternative "meet in the middle" following #228 might be to use classes to dictate the algorithm? Then we have to decide which arguments go in the class and what goes in the function, but maybe that's not so hard?

nmearl:
I have to disagree with the function set approach: it feels like we're polluting the namespace with things that are conceptually the same (resampling), but vary in only their implementation (flux-conserving, etc), which, on top of the fact that you're proposing enforcing an API, smells more like class territory than function territory, and begs for algorithms to be implemented as classes inheriting from some ABC.

The approach I was developing toward was the astropy models approach, which I think is reasonable. You instantiate your needed resampling sub class with your desired, algorithm-specific parameters, then call the instance passing in your current spectrum with the expectation the resulting spectrum has been resampled.

[...]

Issue #254, opened 2018-07-23
Splicing multiple spectra together (e.g., adding spectra together, possibly accounting for overlap). May end up being the same as combining


SaOgaz:
Related issue: #224

SaOgaz:
Possible break out issue noted by @eteq:

"Combine spectra where they do not match. Requires resampling of some sort. May end up being the same as splicing.

stscijgbot:
This ticket is now being tracked at [DATJPP-122](https://jira.stsci.edu/browse/DATJPP-122)

SaOgaz:
Things to think about for this.  Another similar function in IRAF was scombine.  I've gotten some user feedback that the options she liked to use in scombine were the masking, rejection, iterative rejections, and automatic log or linear scaling for the output spectra.

stscijgbot:
This ticket is now being tracked at [JDAT-10](https://jira.stsci.edu/browse/JDAT-10)

SaOgaz:
We should discuss what we want to have the API.  There are three options I can think of, one of which came from @eteq but I don't remember where I saw it. I'm just going to call it splice for now, although it's likely we will want to rename it.

- As a method of a `Spectrum1D`:   `my_spec1D.splice()` (This was @eteq's suggestion at some point)

- As a method of a `SpectralCollection`: `my_spec_collec.splice()`

- As a new class `Splice`, where we hook up the `Splice.__call__` the same way we've done the resampling.

SaOgaz:
I think I favor the last option, having it be a separate function, the most.  I think it will be the most obvious to users, and we can always have the `Splice` class ask the user what resampling function they want to use, and any other options it maybe need.

hcferguson:
I think I agree with @SaOgaz. Seems to make sense given that it is combining multiple spectra.

eteq:
Some evolution from the discussion above: whatever I thought in the past I'm definitely against a method of `Spectrum1D` for this, as that's more in line with the decisions in #266. Of the listed options, I agree the last one is the best one on the basis of #266.

*however*, at this point it is not clear that `Splice` needs to be a distinct operation *at all*.  That is, it could instead be the following:

1. Use the resampling classes to resample the two spectra onto a shared grid
2. Add the resampled spectra together

That makes "splicing" just a two-line operation using already existing functionality. If it's really that simple for the known use cases, then we shouldn't have a class/function to just do those two lines, instead we should add a section in the specutils docs that show this combination is how you do splicing.

[...]

PR #461, opened 2019-04-10
initial bones of resample code, just have flux conserving and interpolation right now.

SaOgaz:
See related issues: #359, #254, #255, #266


SaOgaz:
Questions:

I discussed this a little bit with @eteq, the initial suggestion for weighting (and or uncertainity) was to have a weight keyword that could have one of three values, None (meaning no weighting), provided array (use provided array as weights) or 'uncer' (use the uncertainty defined in the input `Spectrum1D` as the weights). If the user choses 'uncer' but no uncertainty is in the `Spectrum1D`, we throw an error. Right now this is unimplemented.

How do we want to handle propagating the uncertainties.... the paper I pulled the algorithm from has a prescription for calculating the new associated uncertainties (it's just using normal uncertainty propogation).  I haven't looked into how easy it would be to adapt the current code to use any built in uncertainty handling, buy my initial guess is that it won't be too difficult, since the flux doesn't get multiplied through until the very end.  The thing I'm not sure about is, if someone is using the uncertainties for weights, how does that effect the output uncertainty?

Another uncertainty related question. if they are not defined in the input spectrum, do we leave them undefined?  I'm leaning towards yes.

hcferguson:
Anyone familiar with the [uncertainties package](https://pythonhosted.org/uncertainties/)?

eteq:
@SaOgaz 

I would say for *this* PR we just follow the propagation the paper suggests.  The key point is that we need to be sure it's the *right* kind of uncertainty, though.  That is, I think the paper is built around gaussian statistics, so you do that propagation if the uncertainty in the spectrum is either SD, variance, or ivar uncertainties (which is probably like 99% of the use cases anyway). One thing about this, though: if the keyword argument is `weights`, the "use the unceratinty" option should describing the weighting.  I.e., instead of `"uncer"` I'd say it should be `ivar_weighted`, which means "convert the uncertainty to an inverse variance and use that as the weights" (which is also how you'd implement it).

As for the question of what to do if the weights aren't uncertainties - aren't the weights just the `c_ij` from https://arxiv.org/abs/1705.05165 ?  In that case you should just be able to use the expression exactly as given in section 5 (the covariance matrix), regardless of whether the weights come from the uncertainties or somewhere else.

Related thought: since we are following that algorithm, perhaps the uncertainty should record  the whole covariance matrix not just the diagonals (that is, the ).  That would require making a new uncertainty class, but it would be easy to make one that can be converted to `VarianceUncertainty` or `StdDevUncertainty` by just taking the diagonal of the matrix?  This may well be better done as a follow-on PR, though.

@hcferguson 
> Anyone familiar with the uncertainties package?

Yes, several years back I had a GSoC student work on that to see how well it can be used with some of the Astropy machinery.  At the time the answer was "it's very hard because they've done some numpy stuff that conflicts deeply with a lot of other machinery". It may be useful as an eventual thing to complement the (fairly new) existing machinery we built in astropy to do uncertainty propogation (http://docs.astropy.org/en/stable/uncertainty/index.html).  But it's hard to just use as it stands.

eteq:
One other very minor comment to just make sure I don't forget: the resample functions should be pulled out into `specutils.manipulation` via `from .resample import *` in the `__init__.py` in manipulation.  We agreed we generally want only 2-level imports at the most for the user API.

eteq:
@SaOgaz - as we discussed out-of-band, I've added a test here to check for flux conservation.  Basically the test I added is checking that flux is conserved between an input and output spectrum in the way I understand us to be aiming for.  When I ran locally this didn't work... But we shall see what travis says.

SaOgaz:
@eteq, huh, I did that a bit locally (not sure why I didn't think to add a test for it), and it came out correctly, but maybe my calculations were off?

specutils/tests/test_resample.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    A helper function that computes the "size" of a bin given the bin centers
+    for testing the flux conservation
+    """
+    l_widths = (saxis[1] - saxis[0])
+    r_widths = (saxis[-1] - saxis[-2])
+    # if three bins 0,1,2; want width of central bin.  width is avg of 1/2 minus
+    # average of 0/1: (i1 + i2)/2 - (i0 + i1)/2 = (i2 - i0)/2
+    mid_widths = (saxis[2:] - saxis[:-2]) / 2
+
```
SaOgaz: @eteq it doesn't look like you're handling the bins on the edges with this... was that intentional? I'm including the edge bins in my code.

SaOgaz:
@eteq ahh, I think I might know why those tests are failing.  For your input data, the bin edges aren't aligned, just input center wavelengths.  i.e. the left edge of the non linespace wavelength axis is 3,500, the linespace left edge is 3666.66.  The way I implemented this is that the flux conservation will only truly apply when the bin *edges* line up. (although i'm trying to edit your test example to follow this case and running into problems, so cannot yet confirm that results come out right aside from the testing i did earlier).

SaOgaz:
side note, meant to add to init, just hadn't gotten around to it yet.

SaOgaz:
> I would say for _this_ PR we just follow the propagation the paper suggests. The key point is that we need to be sure it's the _right_ kind of uncertainty, though. That is, I think the paper is built around gaussian statistics, so you do that propagation if the uncertainty in the spectrum is either SD, variance, or ivar uncertainties (which is probably like 99% of the use cases anyway). One thing about this, though: if the keyword argument is `weights`, the "use the unceratinty" option should describing the weighting. I.e., instead of `"uncer"` I'd say it should be `ivar_weighted`, which means "convert the uncertainty to an inverse variance and use that as the weights" (which is also how you'd implement it).

Is this in the situation where someone isn't using one of the standard `NDUncertainty` subclasses? I would think that just checking the type and assuming we want to return the same type would be the most straightforward approach, in which case using a bland "uncer" to mean, use whatever uncertainty is there as a weight would be the best way to go.

specutils/manipulation/resample.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        return NotImplemented
+
+    @staticmethod
+    def _bin_edges(x):
+        """
+        Calculate the bin edge values of an input dispersion axis. Input values
+        are assumed to be the center of the bins.
+
+        todo: this should live in the main spectrum object, but we're still
```
nmearl: The main `Spectrum1D` object already contains a `bin_edges` property.

SaOgaz: I tried to use this and it didn't seem to return anything, will look into it again.

nmearl: It might be that you're looking to return the dispersion grid shifted for bin edges, the the `bin_edges` property also performed the `pixel_to_world` transformation after calculating the edges.

SaOgaz: This is what I get.... i have the latest version of the gwcs library...

```
In [5]: flux_val = np.array([1, 3, 7, 6, 20])                                                                                                                                   

In [6]: wave_val = np.array([2, 4, 12, 16, 20])                                                                                                                                 

In [7]: input_spectra = Spectrum1D(flux_val*u.mJy, wave_val* u.AA)                                                                                                              

In [8]: input_spectra.bin_edges                                                                                                                                                 
<ipython-input-8-c112932b4a4f> in <module>
----> 1 input_spectra.bin_edges

~/miniconda3/envs/specutilsdev/lib/python3.7/site-packages/astropy/utils/decorators.py in __get__(self, obj, owner)
    724                 return val
    725             else:
--> 726                 val = self.fget(obj)
    727                 obj.__dict__[self._key] = val
    728                 return val

~/specutils/specutils/spectra/spectrum1d.py in bin_edges(self)
    148     @lazyproperty
    149     def bin_edges(self):
--> 150         return self.wcs.bin_edges()
    151 
    152     @property

AttributeError: 'GWCSAdapter' object has no attribute 'bin_edges'
```

nmearl: Right, well, it seems the bin edges are only calculated for fits wcs objects. :man_shrugging:

eteq:
@SaOgaz 

> edit: sorry I'm getting a bit confused, I thought we were just using quadrature in the case of no, or separate weights (where separate weights would be including in the quadrature calculation) as described in the "calculation of associated uncertainties" section...

Ah right, I see your point here, and I missed some of the subtlety.  I think for the purposes of *this* PR, the "weights" are basically 1/(sigma^2) in the notation of the "Calculation of associated uncertainties" section.  That doesn't make a ton of sense really in the single-spectrum case, but it's relevant for the "list of spectra" case.  And in fact this is making me realize maybe we don't want to call this `weights` but rather `pixel_weights`. We might later add a `spectrum_weights` or similar, but given the fact that I think *usually* the user will want to use the uncertainty rather than the weights array, this makes sense as a way to keep those concepts separate.

> Is this in the situation where someone isn't using one of the standard NDUncertainty subclasses? I would think that just checking the type and assuming we want to return the same type would be the most straightforward approach, in which case using a bland "uncer" to mean, use whatever uncertainty is there as a weight would be the best way to go.

Yes, if there's an `NDUncertainty` option definitely you use that information.  My comment here was about the possibility of the user providing a "weights" array, in which case you'd need to say what *kind* of weight it was.  But given the above realization I had, I think that's no longer relevant.  Instead we just say "if weights are arrays they are interpreted as inverse variance" and then it's the user's business to know how to convert their weights into that.  So I think you can ignore this concern now.

eteq:
Just to note here - after some out-of-band discussion with @SaOgaz we realized the best way to push forward is to not have the weights *at all* for this PR.  Rather instead this implements the flux-conserving algorithm (which doesn't really have meaningful per-pixel weights as described above if you want it to be flux-conserving...), but *does* propogate the uncertainties self consistently.

The whole "should there be a separate `pixel_weight` argument" is still relevant, especially in the context of *other* resampling algorithms (e.g. interpolation), but we decided that should be dealt with *after* this is implemented so that we can get something in to build off of.

eteq:
(also the test failure appears to be unrealted to this PR, as it's also in master)

SaOgaz:
I've added two interpolation versions as well, linear and spline.

specutils/manipulation/resample.py
```diff
@@ -0,0 +1,293 @@
...
+        inside_edges = (x[1:] + x[:-1]) / 2
+        edges = np.insert(inside_edges, 0, 2 * x[0] - inside_edges[0])
+        edges = np.append(edges, 2 * x[-1] - inside_edges[-1])
+
+        return edges
+
+
+class FluxConservingResample(ResampleBase):
+    """
+    This resample algorithim conserves overall flux (as opposed to flux density).
```
eteq:
```suggestion
    This resample algorithim conserves integrated flux (as opposed to flux density).
```

I *think* this is technically more correct - that is, the algorithm conserves the integrated flux in a particular area, not the actual sum-of-fluxes.

specutils/manipulation/resample.py
```diff
@@ -0,0 +1,293 @@
...
+        # calculation, which is probably easiest. Matrix math algorithm is
+        # geometry based, so won't work to just let quantity math handle it.
+        resampled_spectrum = Spectrum1D(np.nan_to_num(out_flux),
+                                        np.array(fin_lamb) * orig_spectrum.wavelength.unit,
+                                        uncertainty=np.nan_to_num(out_uncertainty))
+
+        return resampled_spectrum
+
+
+class InterpolatedResample(ResampleBase):
```
SaOgaz: Per conversation with @eteq we're probably gonna make these all separate interpolator classes... we'll probably want some different class input parameters based on type of interpolation.

specutils/manipulation/resample.py
```diff
@@ -0,0 +1,293 @@
...
+        Return the resulting `~specutils.Spectrum1D` of the resampling.
+        """
+        return self.resample1D(orig_spectrum, fin_lamb)
+
+    def _resample_matrix(self, orig_lamb, fin_lamb):
+        """
+        Create a re-sampling matrix to be used in re-sampling spectra in a way
+        that conserves flux. This is adapted from
+        https://ui.adsabs.harvard.edu/abs/2017arXiv170505165C/references,
+        eprint arXiv:1705.05165. This code was heavily influenced by Nick Earl's
```
eteq: Can you reference either a PR # or a commit on github to cite this just so that there's history?

specutils/manipulation/resample.py
```diff
@@ -0,0 +1,293 @@
...
+        # to zero by checking edges, and applying generated mask
+        left_clip = np.where(fin_edges[:-1] - orig_edges[0] < 0, 0, 1)
+        right_clip = np.where(orig_edges[-1] - fin_edges[1:] < 0, 0, 1)
+        keep_overlapping_matrix = left_clip * right_clip
+
+        resamp_mat *= keep_overlapping_matrix[:, np.newaxis]
+
+        return resamp_mat
+
+    def resample1D(self, orig_spectrum, fin_lamb):
```
eteq:
```suggestion
    def resample1d(self, orig_spectrum, fin_lamb):
```
PEP8 is admittedly a bit ambiguous, but in general the preference I tend to see is "1d" over "1D" in places that are normally all-lowercase.

(Note that I didn't go through and find all of these - probably it's easier done as a global find-and-replace in your editor)

specutils/manipulation/resample.py
```diff
@@ -0,0 +1,293 @@
...
+        """
+        Create a re-sampling matrix to be used in re-sampling spectra in a way
+        that conserves flux. This is adapted from *this* paper (ref TBD). If
```
eteq: missing the citation

specutils/manipulation/resample.py
eteq: I think this is better just split into two separate classes - `LinearInterpolatedResample` and `SplineInterpolatedResample`.  I'm thinking this because the `SplineInterpolatedResample` should get some keywords to set things like the order of the spline, but those don't apply in the linear case, so instead of having some confusing rule like "this keyword is ignored when interp_type is linear but not spline" you can just do separate classes.

eteq: Oh, hah, collided with my comment below :wink:

specutils/manipulation/resample.py
SaOgaz: Sure, there's no PR though.  I was referencing his branch off his fork: https://github.com/nmearl/specutils/commit/0ff6ef169b5607a1e059cb7cc9488238191f3c39.  What's the best way to reference that?

SaOgaz:
Oh shoot, @eteq I accidentally over-wrote the change of yours I merged in, what did you change "This resample algorithim conserves overall flux (as opposed to flux density)" to?

stscicrawford:
Tagging @philhodge, @jdavies-st, and @stscirij  to take a look at how useful this functionality would be for the calibration software or if there is something that could be added in the future

specutils/manipulation/resample.py
```diff
@@ -0,0 +1,409 @@
...
+        inside_edges = (x[1:] + x[:-1]) / 2
+        edges = np.insert(inside_edges, 0, 2 * x[0] - inside_edges[0])
+        edges = np.append(edges, 2 * x[-1] - inside_edges[-1])
+
+        return edges
+
+
+class FluxConservingResample(ResampleBase):
+    """
+    This resample algorithim conserves overall integrated flux (as opposed to
```
eteq:
```suggestion
    This resampling algorithim conserves overall integrated flux (as opposed to
```

specutils/manipulation/resample.py
eteq: I think it's find just to give the URL directly to that SHA (e.g. https://github.com/nmearl/specutils/commit/0ff6ef169b5607a1e059cb7cc9488238191f3c39 ) - this is a private method so no need to be too picky about it

specutils/manipulation/resample.py
```diff
@@ -0,0 +1,409 @@
...
+        keep_overlapping_matrix = left_clip * right_clip
+
+        resamp_mat *= keep_overlapping_matrix[:, np.newaxis]
+
+        return resamp_mat
+
+    def resample1d(self, orig_spectrum, fin_lamb):
+        """
+        Create a re-sampling matrix to be used in re-sampling spectra in a way
+        that conserves flux. This is adapted from eprint arXiv:1705.05165. If
```
eteq: I think there's no need to reference the paper twice.  It's better to just say it at the top of your class (as you have already), so that this docstring is about the actual method rather than the algorithm.

specutils/manipulation/resample.py
```diff
@@ -0,0 +1,409 @@
...
+        resampled_spectrum = Spectrum1D(flux=out_flux,
+                                        spectral_axis=np.array(fin_lamb) * orig_spectrum.spectral_axis_unit,
+                                        uncertainty=out_uncertainty)
+
+        return resampled_spectrum
+
+
+class LinearInterpolatedResample(ResampleBase):
+    """
+    This resample algorithim uses a linear interpolator.
```
eteq:
```suggestion
    Resample a spectrum onto a new ``spectral_axis`` using linear interpolation.
```

specutils/manipulation/resample.py
```diff
@@ -0,0 +1,409 @@
...
+        fin_lamb : ndarray
+            The desired dispersion array.
+
+        Returns
+        -------
+        resample_flux : ndarray
+            The resampled flux array generated from the interpolation.
+
+        """
+        cubic_spline = CubicSpline(orig_dispersion, flux, extrapolate=False)
```
eteq: Can you clarify why you used this instead of a `scipy.interpolate.InterpolatedUnivariateSpline`?  I just did some tests now and it seems like `InterpolatedUnivariateSpline` is a bit faster (and has the added advantage that you can set the spline degree as a free parameter).

specutils/manipulation/resample.py
```diff
@@ -0,0 +1,409 @@
...
+            The original 1D spectrum.
+        fin_lamb : ndarray
+            The desired dispersion array.
+
+        Returns
+        -------
+        resample_spectrum : `~specutils.Spectrum1D`
+            An output spectrum containing the resampled `~specutils.Spectrum1D`
+        """
+        out_flux = self._interpolation(orig_spectrum.spectral_axis, orig_spectrum.flux,
```
eteq: I agree it's out-of-scope to implementation uncertainties right now, but perhaps you should add a warning if the input spectrum has an uncertainty?  I.e.:
```suggestion
        if orig_spectrum.uncertainty is not None:
            warn("Linear interpolation currently does not propogate uncertainties")
        out_flux = self._interpolation(orig_spectrum.spectral_axis, orig_spectrum.flux,
```

docs/manipulation.rst
```diff
@@ -102,6 +102,64 @@ Note: This method is not flux conserving.
...
+Here's a set of simple examples showing each of the three types of resampling:
+
+.. code-block:: python
+
+    >>> from specutils import Spectrum1D
+    >>> import astropy.units as u
+    >>> import numpy as np
+    >>> from specutils.manipulation import FluxConservingResample, LinearInterpolatedResample, SplineInterpolatedResample
+
+    >>> input_spec = Spectrum1D(spectral_axis=np.arange(1, 20, 2) * u.nm, flux=np.random.sample(10)*u.Jy)
```
eteq: This is not an especially realistic spectrum.  Two possibilities I'd like to suggest:
1) Use a "real" spectrum, like the one that comes from the "gettting started" on the front of the docs, and resample that.  Then instead of displaying the results as numbers, you'd show *plots* of the spectrum (perhaps zoomed in) to demonstrate the resampling in action.
2) Have the wavelength range be more like 400 to 700 nm.  There aren't very many spectra out there that are 10s of nanometers (but plenty that are in the visible range)

1 is a bit more useful, but 2 is a much easier change.  So one option might be to do 2 for now, and then in a follow-on PR perhaps I or someone else can try to implement 1.

specutils/manipulation/resample.py
SaOgaz: I probably just didn't look into it enough when I was putting this class together. I will switch it out and add the degree as a class parameter.

SaOgaz: So this gets tricky in that right now we decided offline to only do NaN fill on the boundaries for now. This is possible using CubicSpline very easy.  But it's not easy to do with UnivariateSpline.  There's no built in way, so we'd have to do a more complicated work around, which I'm not sure I should spend the time on?  Given the situation, I'm for keeping it as a CubicSpline so that the edge handling is the same for all resampling, and change it over to UnivariateSpline when there is more time.

eteq:
Alright I think we're all good here now.  Thanks for your work on this @SaOgaz !

PR #407, opened 2018-11-28
Extend list of hacks for reading non-standard complied 1d WCS fits files

hamogu:
In the current code, there is a hack to allow "LINEAR" as a spectral
axis. That's clearly not compliant with the FITS WCS standard as defined in
https://www.aanda.org/articles/aa/full/2006/05/aa3818-05/aa3818-05.html
and might just be motivated by the GMOS spectrum currently used as
example for this loader (@kelle ?).
This code change makes this list of keywords that will be treated as
spectral extensible and adds "Wavelength" (which is much closer to
"WAVE" as required in the standard than "LINEAR").

Tests and docs are still missing, but those are easy to add once we agree on this approach.

PR #462, opened 2019-04-11
Gwcs refactor

nmearl:
This PR removes the adapter classes from specutils and relies on the unified WCS API implemented in both fitswcs and gwcs.
A few problems remain:
- [x] This requires that the transformations in gwcs be unit-aware (spacetelescope/gwcs#210)
- [x] Requires a fix for the return arrays of fitswcs transformations (astropy/astropy#8559)


hamogu:
This overlaps with #407. I was going to implement the simple change suggested there now, but if you want to work that into this PR instead let me know.

javerbukh:
@nmearl it seems this PR was abandoned in favor of #407, should this remain open?

nmearl:
@javerbukh PR #407 is not related to this.

Changes to PR now include @nden's PR #490 and requires that GWCS merged PR spacetelescope/gwcs#261. 

This PR now fully address #553 by completely removing the wcs adapter infrastructure, relying solely on APE14, and refactoring model behavior with the changes in astropy 4.0.

eteq:
The remaining failures are either network problem (a single test is failing on all of them due to that) or the thing that will be fixed in #564 .  So this is good to go!

PR #471, opened 2019-06-19
WIP: Discussion PR on SpectrumCollections decorator

SaOgaz:
Hi all,

@eteq and I had an offline  discussion during my work on the resample code about how we want to work in analysis on spectrum collections. I advocated for a decorator.  I've done an example of what I had in mind here that runs on my resample class.  @eteq had an alternate suggestion that I'm now blanking on.

I like the decorator approach as it seems a little easier to follow for users trying to track what's going on with the code, but this PR is definitely meant for discussion!

### DIFF
--- a/specutils/manipulation/resample.py
+++ b/specutils/manipulation/resample.py
@@ -6,9 +6,9 @@ from astropy.units import Quantity
 from astropy.nddata import StdDevUncertainty, VarianceUncertainty, InverseVariance
 
 from ..spectra import Spectrum1D
+from ..spectra.spectrum_collection import runs_on_spectrum_collection
 
-__all__ = ['ResamplerBase', 'FluxConservingResampler',
-           'LinearInterpolatedResampler', 'SplineInterpolatedResampler']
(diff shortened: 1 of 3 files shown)


nmearl:
Am I missing it? I don't see a decorator example in the code?

SaOgaz:
@nmearl it would probably help if I included the commit, LOL

SaOgaz:
Oh, another note, @eteq suggested putting the decorator in the SpectrumCollections class as a static method.  I'm not sure I'm convinced of this.  I feel like it makes the decorator call kind of awkward.

nmearl:
Can you provide an example of using this decorator?

SaOgaz:
It's in there.... in the resample function: https://github.com/astropy/specutils/pull/471/files#diff-64489eca25a7793c6b98cd48fa905511R125

SaOgaz:
Feedback from @jkcarlberg, we need to add an example section on usage for these classes, so that when help() is called it shows you how to use.  Otherwise users will (understandably) assume they need to call `resample1D`.

pllim:
This PR is stale and author no longer works at STScI.

PR #472, opened 2019-06-24
Fix equivalent width and flux computation

tiagopereira:
The current computations of flux and equivalent width do not work for cases with varying spectral sampling, or when the spectral axis is decreasing. This PR uses trapezoidal integration (`numpy.trapz`) to allow for more general wavelength spacings, and ensures the sign of the computation matches the changes in wavelength.

Also return proper units for flux and equivalent width.

### DIFF
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -15,7 +15,7 @@ def line_flux(spectrum, regions=None):
     Computes the integrated flux in a spectrum or region of a spectrum.
 
     Applies to the whole spectrum by default, but can be limited to a specific
-    feature (like a spectral line) if a region is given.
+    feature (such as a spectral line) if a region is given.
 
     Parameters
     ----------
@@ -29,7 +29,7 @@ def line_flux(spectrum, regions=None):
     Returns
     -------
     flux : `~astropy.units.Quantity`
-        Flux in the provided spectrum (or regions). Unit isthe ``spectrum``'s'
+        Flux in the provided spectrum (or regions). Unit is the ``spectrum``'s'
         ``flux`` unit times ``spectral_axis`` unit.
 
     Notes
@@ -45,7 +45,7 @@ def equivalent_width(spectrum, continuum=1, regions=None):
     Computes the equivalent width of a region of the spectrum.
 
     Applies to the whole spectrum by default, but can be limited to a specific
-    feature (like a spectral line) if a region is given.
+    feature (such as a spectral line) if a region is given.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


nmearl:
Thanks @tiagopereira for the PR. A couple of main issues I'd like to address:

1. The base implementation in specutils is meant to be the most basic implementation of the equivalent width calculation i.e. the simple integral over `(Fc - F)/Fc`, avoiding approximations. We'd like to keep this for several reasons, and so I'd suggest you implement the trapezoidal method as an addition, optional implementation (which may become the default in the future). By this I mean offers a keyword argument in the function call such that if `method='trapezoidal'`), the calculation uses the numpy `trapz` implementation.
2. The current implementation *does* deal with varying spectral sampling; the `delta_lambda` is formed with the `np.diff` function which codifies the spacing between pixel bins. (Again, the fact that the simple implementation deals predominantly in pixel space is another reason why your implementation is a good option to have.)
3. You're absolutely right about the case where the spectral axis is decreasing.
4. The simple implementation should have been returning proper units for wavelength and flux, if you haven't been seeing this, please submit issue!

To re-iterate, we would like to keep the current implementation of the line flux and equivalent width calculations (even if it does not become the default in the future), and it'd be great if you could refactor your code a bit to offer the trapezoidal method as an alternative.

tiagopereira:
Hi @nmearl, 

The current approach seems to be a poor attempt of a trapezoidal integration. It will always be off for any irregular spacing. If you look in the code for `numpy.trapz`, you'll see that it is essentially `sum(diff(x) * (x[1:] + x[:-1]) / 2)`. So I don't really see a case for keeping the current implementation instead of doing it properly with trapezoidal.

nmearl:
There is no attempt at a trapezoidal integration; the pixel bins of a spectrum have a known value of flux so we don't need to deal with approximating the value of a pixel bin. Therefore, the sum of the flux values over the bin widths of the spectrum is the most brute-force way of calculating the equivalent width; i.e.

ew

This is the method implemented currently. With that said, trapezoidal approximation isn't the most "proper" implementation either, it's just a very easy one. Simpson's rule, Gaussian quadrature, Romberg's method, etc. are all viable alternatives depending on your tolerance of error. But those methods (including the trapezoidal rule) are all approximations while the base implementation is a direct translation of the formula for calculating the equivalent width (because we know `d_lambda`).

We certainly would like the user to be able to choose whichever implementation they prefer, but we were also attempting to avoid making that decision for them. Therefore, we opted to stay with the most basic implementation, with (in the future) optional alternatives. We did assume at some point that we'd have to deal with cases like where users might provide dispersion bins that are not equal to the pixel bins. In such cases, an approximation is certainly the better route.

tiagopereira:
Thanks for the clarification. I see the source of the confusion now. I don't usually look at spectra as "bins with a known value of flux" because many of the spectra I work with are synthetic spectra and calculated for a specific wavelength point, not a bin. Hence also the irregular spacing - to save computer time these are often calculated with higher resolution in the line cores and lower resolution close to the continua. So I don't have a problem with keeping the current method but adding an option for trapezoidal (I'm usually against Simpson as it can be tricky), just different ways to approach the problem.

On a related note, and I know I'm late to the party, as a stellar spectroscopist turned solar physicist I would oppose calling a quantity of spectra "flux", because flux is just intensity integrated over solid angle, and in solar spectra a lot of the times we deal with intensities and not fluxes. Maybe one day we can find a more neutral term.

Issue #242, opened 2018-07-11
Full width at zero intensity measure


brechmos-stsci:
Create the method to calculate the full width at zero intensity measure of a spectrum.

eteq:
Since @crawfordsm specifically brought this up, can he point us to what's a good assumption to make for this algorithm?  Is it "start at some center point and walk out in pixels until flux first falls below zero on either side"?  If, so then unlike most of the other "analysis" tools, this one I guess takes just *one* input: the "starting point" of the walk/line center.

eteq:
Not going to do for the upcoming release, so setting it to the next. Still needs some context from someone (@crawfordsm ?) who can say what this algorithm should actually be.

nmearl:
@camipacifici So, there are two ways to go about the calculation for this: one way is to assume the emission feature is Gaussian, and use that to calculate the width at the continuum. This is convenient because it ignores are potential noise in the spectral data. However, this also means we're precluding a more accurate measurement if, say, the feature resembles a Voigt with e.g. really wide wings.

The alternative is to perform some smoothing in order to minimize the impact of the noise on calculating the full width.

Is there a preference?

camipacifici:
I would rather not assume a specific function (I would naively think that assuming a function is equivalent to analytically derive the FWZI from the FWHM from gaussian_fwhm, so there is not really a point in calculating it this way).
Ideally, I would like the approach without fitting and the noise in the spectrum be reflected in the uncertainty of the FWZI calculation.
Calling @crawfordsm again, maybe you have a different science case in mind?

hcferguson:
I'm inclined to say that anyone who uses FWZI should use it at their own risk. What are the plans for continuum here? Require the user to deal with it separately? Or offer regions on either side with some sort of interpolation?

crawfordsm:
I would agree with @camipacifici and as @hcferguson points out, the difficult in this calculation is what to do with the continuum.   

The simple situation is when the user supplies a spectrum where the continuum has been normalized to 1.   Then an algorithm is needed to assess when the absorption feature reaches a value of 1 on either side of the line.   The error can be calculated from the noise of the continuum and the gradient of the line.  

The more complex situation is when the continuum still needs to be normalized.  I see two main options for how this might be done:
1. The user supplied regions on either side of the line that could be used to normalize the spectrum
2. The user supplies a model for the line and the continuum.  Both are fit to the line, and the continuum model is then used to normalize the spectrum.  

[...]

PR #483, opened 2019-07-11
Implement full width at zero intensity calculation

nmearl:
Address #242. This implements the calculation with the assumption that the continuum has been subtracted such that it sits at 0 and therefore the width is calculated from the closest point on either side of a maximum at which the flux value is equal to the continuum.

### DIFF
--- a/specutils/analysis/width.py
+++ b/specutils/analysis/width.py
@@ -5,9 +5,11 @@ spectral features.
 
 import numpy as np
 from astropy.stats.funcs import gaussian_sigma_to_fwhm
+from astropy.modeling.models import Gaussian1D
 from ..manipulation import extract_region
 from . import centroid
 from .utils import computation_wrapper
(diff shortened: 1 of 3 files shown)


specutils/analysis/width.py
```diff
@@ -105,6 +107,76 @@ def fwhm(spectrum, regions=None):
...
+    scipy peak-finding to determine the index of the highest flux value, and
+    then calculates width at that base of the feature.
+
+    Parameters
+    ----------
+    spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
+        The spectrum object over which the width will be calculated.
+
+    regions: `~specutils.utils.SpectralRegion` or list of `~specutils.utils.SpectralRegion`
+        Region within the spectrum to calculate the FWHM value. If regions is
```
SaOgaz: is it still considered the FWHM value, or is this a typo?

PR #484, opened 2019-07-14
Sub-pixel precision for calculate_fwhm

### LINKED ISSUE #378: fwhm could be sub-pixel precise
I just realized (from review of #377) that the FWHM function uses whole pixels to compute the FWHM.  This isn't necessarily wrong formally, but it could be better: the current code identifies the left/right edge pixels, but a linear interpolation across that pixel could achieve sub-pixel precision.  Since it's just two pixels this can probably be expressed as an average to get better performance (np.interp is relatively slow for small numbers of data points), although that might take a bit of care to make sure it does a sensible thing for multiple-flux `Spectrum1D` objects.

cc @drdavella @brechmos-stsci

danielnperera:
Improvement to the existing algorithm using linear interpolation to achieve sub-pixel precision.

- Updated algorithm
- Updated tests
- Updated documentation tests

Fixes #378

### DIFF
--- a/specutils/analysis/width.py
+++ b/specutils/analysis/width.py
@@ -226,8 +226,8 @@ def _compute_single_fwhm(flux, spectral_axis):
 
     argmax = flux_value.argmax()
     halfval = flux_value[argmax] / 2
-    left = flux_value[:argmax] < halfval
-    right = flux_value[argmax + 1:] < halfval
+    left = flux_value[:argmax] <= halfval
+    right = flux_value[argmax + 1:] <= halfval
 
     # Highest signal at the first point
     i0 = np.nonzero(left)[0]
@@ -248,7 +248,7 @@ def _compute_single_fwhm(flux, spectral_axis):
     if i1.size == 0:
         right_value = spectral_value[-1]
     else:
-        i1 = i1[0] + argmax + 1
+        i1 = i1[0] + argmax
         i0 = i1 - 1
         left_flux = flux_value[i0]
         left_spectral = spectral_value[i0]
--- a/specutils/analysis/width.py
+++ b/specutils/analysis/width.py
@@ -220,7 +220,7 @@ def _compute_single_fwhm(flux, spectral_axis):
     # The .value attribute is used here as the following algorithm does not
     # use any array operations and would otherwise introduce a relatively
     # significant overhead factor.  Two-point linear interpolation is used to
-    # acheive sub-pixel precision.
+    # achieve sub-pixel precision.
     flux_value = flux.value
     spectral_value = spectral_axis.value
 
--- a/specutils/analysis/width.py
+++ b/specutils/analysis/width.py
@@ -213,26 +213,51 @@ def _compute_gaussian_sigma_width(spectrum, regions=None):
 
 
 def _compute_single_fwhm(flux, spectral_axis):
+    """
+    This is a helper function for the above `fwhm()` method.
+    """
 
-    argmax = np.argmax(flux)
-    halfval = flux[argmax] / 2
+    # The .value attribute is used here as the following algorithm does not
+    # use any array operations and would otherwise introduce a relatively
+    # significant overhead factor.  Two-point linear interpolation is used to
+    # acheive sub-pixel precision.
+    flux_value = flux.value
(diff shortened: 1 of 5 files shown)


danielnperera:
@nmearl, @eteq suggested this issue as a first time contribution from me.  One of the automated tests (Python 3.5) is failing when it tries to access remote data.  I'm not sure what I could do to fix that.  Would you be able to take a look at this?

Thanks,
Dan

brechmos-stsci:
The tests are kinda flaky sometimes. I restarted it twice and the second time it passed :)

specutils/analysis/width.py
```diff
@@ -141,26 +141,51 @@ def _compute_gaussian_sigma_width(spectrum, regions=None):
...
+    # acheive sub-pixel precision.
+    flux_value = flux.value
+    spectral_value = spectral_axis.value
 
-    left = flux[:argmax] <= halfval
-    right = flux[argmax+1:] <= halfval
+    argmax = flux_value.argmax()
+    halfval = flux_value[argmax] / 2
+    left = flux_value[:argmax] < halfval
+    right = flux_value[argmax + 1:] < halfval
```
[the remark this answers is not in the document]
danielnperera: Hi, yes - I was excluding it because flux_value[argmax] == halfval, so there's no point in including it during the comparison operation.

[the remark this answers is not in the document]
danielnperera: sorry, I meant flux_value[argmax] = halfval * 2

nmearl:
This looks good, but you'll need to rebase for the latest changes in master.

danielnperera:
I'm having some difficulty again with the CI connecting to zenodo.org with 404 not found errors.  Is there a way to rerun the tests without closing and reopening the PR?

specutils/analysis/width.py
```diff
@@ -141,26 +213,51 @@ def _compute_gaussian_sigma_width(spectrum, regions=None):
...
 
 def _compute_single_fwhm(flux, spectral_axis):
+    """
+    This is a helper function for the above `fwhm()` method.
+    """
 
-    argmax = np.argmax(flux)
-    halfval = flux[argmax] / 2
+    # The .value attribute is used here as the following algorithm does not
+    # use any array operations and would otherwise introduce a relatively
```
dhomeier: Do you really mean "any array operations" here, or rather "Quantity", because `flux.value` _is_ a `np.ndarray` and you are using indexing operations etc.?

specutils/analysis/width.py
```diff
@@ -141,26 +213,51 @@ def _compute_gaussian_sigma_width(spectrum, regions=None):
...
+    # significant overhead factor.  Two-point linear interpolation is used to
+    # acheive sub-pixel precision.
```
dhomeier:
```suggestion
    # achieve sub-pixel precision.
```

specutils/tests/test_analysis.py
```diff
@@ -483,7 +483,11 @@ def test_fwhm():
 
     spectrum = Spectrum1D(spectral_axis=wavelengths, flux=flux)
     result = fwhm(spectrum)
-    assert result == 0.9090909090909092*u.um
+    # Note that this makes a little more sense than the previous version;
+    # since the maximum value occurs at wavelength=1, and the half-value of
+    # flux (0.5) occurs at exactly wavelength=2, the result should be
+    # exactly 1 (2 - 1).
+    assert result == 1.0*u.um
```
dhomeier: How about modifying that test (or adding another one) so that wavelength=2 is not actually on the grid, to test the interpolation? E.g.
`wavelengths = np.linspace(1, 10, 31) * u.um`

nmearl: @dhomeier can you open another issue to address expanding on the unit testing? I'd like to merge this PR now for a release and iterate on it over time.

dhomeier: Sure. Just wanted to note that the PR adds some functionality for which there is no test coverage now, but if it has time I can look into it sometime.

nmearl: @dhomeier, I've gone ahead and included a test for this interpolation case. No need to make a new issue.

Issue #455, opened 2019-03-04
Decide on idiom for recording redshift/velocity of spectra


eteq:
This issue actually touches on several topics and is more of a roll-up of some related issues.  But we do need to answer this to move forward.  The core question is how we want to store the concept of "the redshift of this spectrum". 

First, the specific related issues:
* #422 - proposes creating a `SpectralCoord` (one minor awkwardness: requires changes in `gwcs` at the same time as `specutils`).  This isn't immediately obvious related, until you realize it would have a `rest` attribute, which is implicitly specifying a "redshift".  So wherever the redshift is specified has to be synced up with that somehow.  I also thing it's not the way most astronomers are used to thinking, so we have to either spend a lot of effort training them in or "hide it" somehow for use cases that don't require its full power.
* #413 - roughly speaking, this is that it should be possible for `SpectralCoord`s to transform themselves.  But it's mixed together with the question of whether that should be done on the spectrum vs separately
* #258 - this is basically "we should be able to do the radio<->optical conversion" (which the above two would address) but also whether or not this should be stateful on the spectrum.
* #230 - If we interpret this issue as "be able to correct the spectrum" it's also tied in, in that it's basically "do the correction in a very precise manner when needed".  But an important point that reveals I think is that the user often needs to have a lot of control over how the velocity corrections are applied: some of the cutting-edge science depends on *very* careful velocity work, so we shouldn't hide it behind layers of classes that the average astronomer might not bother with/want to understand.

Now for the questions (which are sort of questions about both `Spectrum1D` *and* `SpectralCoord`):

1. Should the redshift be stored on the spectrum objects themselves?  
2. How do we handle redshifts for the *array* `Spectrum1D` cases? (potentially problematic mainly for the `SpectralCoord` contains the redshift b/c then the `SpectralCoord` is not separable from the spectrum)
3. How do we handle redshifts for the `SpectrumCollection` case? (probably easier for the `SpectralCoord` case, but should be not terribly inconsistent with 2 above)
4. Depending on the answers to the above: how do we go about "de-redshifting" (which is generally just re-mapping the `spectral_coord`, rather than the flux, although a valid option is "don't do this")

I have some ideas, but want to allow room for discussion without spoiling things...

cc @keflavich @crawfordsm @hcferguson @astrofrog @nden

eteq:
[...]

Issue #422, opened 2018-12-05
Implement SpectralCoord class


astrofrog:
*As discussed with @eteq and @keflavich*

In the APE 14 API, we currently use `SkyCoord` as the high-level object for celestial coordinates, but we only use `Quantity` for spectral coordinates, without any attributes that describe the frame of reference the spectral quantity is in, meaning that we can't easily convert between different spectral representations. The solution is to develop a `SpectralQuantity` that subclasses `Quantity` and adds metadata about the spectral frame as well as conversion functionality. This is best developed in specutils to start with for faster development, and we can incorporate it in to astropy core later.

``SpectralQuantity`` should have the following attributes:

* ``frame=``: this would be a ``BaseFrame`` object
* ``rest=``: a quantity giving the rest wavelength or frequency (assuming the same frame etc. as the parent quantity)
* ``type``: ``radio_velocity``, ``optical_velocity``, ``frequency``, ``energy``, ``wavenumber``, ``air_wavelength``, etc.

In addition, we should make sure ``SpectralQuantity(SpectralQuantity(...), …)`` should work (needed for APE 14) and convert the spectral coordinate, as should e.g. ``SpectralQuantity.to(..., frame=’...’, type=’...’)``.

Any of the attributes can be passed into the constructor as `None`, in which case it basically just behaves like a normal Quantity.  But *if* they are not None, `to` does smart things where possible and raises exceptions if anything would be inconsistent(i.e., ``to(u.um, type=’velocity’)``).

We also decided that information about bin edges doesn’t belong inside this class - instead either there could be a SpectralAxis class that subclasses ``SpectralQuantity``, or that contains e.g. a ``SpectralQuantity`` for the bin edges.

One item that still needs to be discussed is where to keep information about the direction of the observations - needed when transforming frames. One option is to include it as an attribute of the ``SpectalQuantity``, and another option is to pass it as an argument during the transformation.

This should be developed in specutils in a file that should be well isolated from the rest of the code and could easily be moved to the core package once we are happy with it.

nmearl:
This is... interesting. So we want to move information about the axis to the axis's `Quantity` representation? Wouldn't this defeat the purpose a mixin?

astrofrog:
[...]

PR #487, opened 2019-07-17
Template matching with resampling

### LINKED ISSUE #256: Template matching: you have n models (which may be spectra themselves), and want to know which is the best
original card posted by @eteq

javerbukh:
This PR uses #461 to resample the template spectrum based on the observed spectrum. In order to test this, I created the following notebook (TempSpec acts as a loader for ascii files). Resolves #256 

```
from specutils import SpectrumCollection
from specutils.analysis import template_match

class TempSpec():

    def __init__(self, filename):
        self.file_data_wave = []
        self.file_data_flux = []
        self.file_data_weight = []
        
        self.filename = filename
        self.load_file(filename)

    def load_file(self, filename):
        first_pass = True

        for line in open(filename, 'r'):
            if first_pass:
                first_pass = False
                continue
            item = line.rstrip()
            floats = list(map(float, item.split()))
            self.add_floats(floats)

    def add_floats(self, floats):
        self.file_data_wave.append(floats[0])
        self.file_data_flux.append(floats[1])
        if len(floats)>2:
            self.file_data_weight.append(floats[2])

filename = "legac_M01_v0.5_0206573.dat"
model_filename = "models/00001.dat"
model_filename2 = "models/00028.dat"

redshift = 0.702

legac = TempSpec(filename)
model = TempSpec(model_filename)
model2 = TempSpec(model_filename2)

model.file_data_wave = [x / (1 + redshift) for x in model.file_data_wave]
model2.file_data_wave = [x / (1 + redshift) for x in model2.file_data_wave]

    
lamb = legac.file_data_wave * u.AA # doctest: +REMOTE_DATA
flux = legac.file_data_flux * u.Unit('erg cm-2 s-1 AA-1') # doctest: +REMOTE_DATA
spec_legac = Spectrum1D(spectral_axis=lamb, flux=flux) # doctest: +REMOTE_DATA

lamb = model.file_data_wave * u.AA # doctest: +REMOTE_DATA
flux = model.file_data_flux * u.Unit('erg cm-2 s-1 AA-1') # doctest: +REMOTE_DATA
spec_model = Spectrum1D(spectral_axis=lamb, flux=flux) # doctest: +REMOTE_DATA

lamb = model2.file_data_wave * u.AA # doctest: +REMOTE_DATA
flux = model2.file_data_flux * u.Unit('erg cm-2 s-1 AA-1') # doctest: +REMOTE_DATA
spec_model2 = Spectrum1D(spectral_axis=lamb, flux=flux) # doctest: +REMOTE_DATA

x = template_match.template_match(spec_legac, [spec_model, spec_model2])
# Or to test one at a time: 
# x = template_match.template_match(spec_legac, spec_model)

print(x)
```


specutils/analysis/template_match.py
```diff
@@ -0,0 +1,36 @@
...
+def _template_match(observed_spectrum, template_spectrum):
+    # resample the template spectrum to match the wavelength of the observed spectrum
+    fluxc_resample = FluxConservingResample()
+    output_spectrum1D = fluxc_resample(template_spectrum, observed_spectrum.wavelength)
+
+    # calculate chi square
+    x2 = chisquare(observed_spectrum.flux, output_spectrum1D.flux)
+    return x2
+
+def template_match(observed_spectrum, collection):
```
brechmos-stsci: Sorry, when I wrote on the board I think I used `collection`. We might want to stay away from that name as `collections` (plural) is a python thing and people, like me, might get confused.  Maybe `templates` or `spectral_templates`?

specutils/analysis/template_match.py
```diff
@@ -0,0 +1,36 @@
+from ..spectra.spectrum1d import Spectrum1D
+from ..spectra.spectrum_collection import SpectrumCollection
+from ..manipulation import FluxConservingResample
+from scipy.stats import chisquare
+
+def _template_match(observed_spectrum, template_spectrum):
```
brechmos-stsci: I would add a methods documentation section here to describe parameters and returns

specutils/analysis/template_match.py
```diff
@@ -0,0 +1,36 @@
...
+    # resample the template spectrum to match the wavelength of the observed spectrum
+    fluxc_resample = FluxConservingResample()
+    output_spectrum1D = fluxc_resample(template_spectrum, observed_spectrum.wavelength)
```
brechmos-stsci: I would suggest change `output_spectrum1D` to `template_spectrum1D` just so it is clear that is the resampled template.

specutils/analysis/template_match.py
```diff
@@ -0,0 +1,36 @@
...
+
+    elif isinstance(collection, SpectrumCollection):
+        pass
+
+    # Loop through spectra in list and return spectrum with lowest chi square
+    # and its corresponding chi square
+    elif isinstance(collection, list):
+        min = None
+        smallest_chi_spec = None
+        for spectrum in collection:
```
brechmos-stsci: I would suggest `for template in spectral_templates` just so it is clear the variable(s) you are iterating over are the templates.

specutils/analysis/template_match.py
```diff
@@ -0,0 +1,36 @@
...
+    if isinstance(collection, Spectrum1D):
+        return Spectrum1D, _template_match(observed_spectrum, collection)
+
+    elif isinstance(collection, SpectrumCollection):
+        pass
+
+    # Loop through spectra in list and return spectrum with lowest chi square
+    # and its corresponding chi square
+    elif isinstance(collection, list):
+        min = None
```
brechmos-stsci: `min` is a built in function.  Maybe `chi2_min`?

specutils/analysis/template_match.py
```diff
@@ -0,0 +1,36 @@
...
+        smallest_chi_spec = None
+        for spectrum in collection:
+            chi = _template_match(observed_spectrum, spectrum)
```
brechmos-stsci: And "the last thought", I would definitely change the variable to `chi2` from `chi` as mathematically they are two different things and it could confuse people when they read the code. :)

javerbukh: Thank you for the "thoughts" =) implemented and ready for another review. Also, is there SpectrumCollection documentation you can point me to so I can implement that aspect as well?

specutils/tests/test_template_match.py
```diff
@@ -0,0 +1,151 @@
...
+import numpy as np
+from astropy.nddata import StdDevUncertainty
+
+from ..spectra.spectrum1d import Spectrum1D
+from ..spectra.spectrum_collection import SpectrumCollection
+from ..analysis import template_match
+from ..manipulation import FluxConservingResampler
+
+
+def _normalize(observed_spectrum, template_spectrum):
```
brechmos-stsci: Why is this method defined in the test file?

camipacifici:
Tested with a single template spectrum.
Code works and numbers are correct.

Problem 1: 
chi2_min should have no dimension and at the moment it is in Jy^2. I think the problem is in spectrum1D.uncertainty that is weirdly defined.
Problem 2:
smallest_chi_spec seems to be a class and not an instance. I cannot find a way to see and plot the wavelengths and fluxes.

Correction:
In the description of def _template_match, scipy.stats.chisquare is mentioned, but it is not used in the code (as we discussed, that function is not appropriate here).

Request:
smallest_chi_spec should be the resampled version, normalized, so basically `normalization*template_spectrum1D.flux`, 
where 
`template_spectrum1D = fluxc_resample(template_spectrum, observed_spectrum.wavelength)`.
This is because the user already has the input spectrum and would need the resampled and normalized version to be able to plot it on top of the observed spectrum and check that the fit is acceptable. This is needed both in case the template is a single spectrum and in case it is a list or SpectrumCollection. In case of a list or SpectrumCollection for the model templates, it would be useful to have also the index of the best fitting template so that the user can go back to their input and find the original template.

More tests I will run:
- try with lists and SpectrumCollection;
- use spectra at different resolutions;
- think about what resample function is more appropriate here (it depends on the units of the flux I think - whether they are per unit wavelengths or not).

specutils/analysis/template_match.py
```diff
@@ -0,0 +1,64 @@
...
+    num = np.sum((observed_spectrum.flux*template_spectrum.flux)/(observed_spectrum.uncertainty.array**2))
+    denom = np.sum((template_spectrum.flux/observed_spectrum.uncertainty.array)**2)
+    normalized = num/denom
+
+    return normalized
+
+def _template_match(observed_spectrum, template_spectrum):
+    """
+    Resample the template spectrum to match the wavelength of the observed spectrum.
+    Then, run scipy.stats.chisquare on the flux of the two spectra.
```
javerbukh: Remove this, no longer using scipy

eteq:
Two quick issue links that are relevant for this PR:
* #266 - in the end that issue says that algorithms like this should generally be implemented as *classes*, where the parameters to the algorithm are stored as keywords.  The motivation mainly being that way users can do things like:

```
matcher = TemplateMatcher(...)
matcher(spectrum1)
...
matcher(spectrum2)
```
and thereby be sure the *same* matching operation is applied to two different spectra, and not have to remember all the keywords they used.  This PR seems to be following a more function-based approach. Fortunately I think the fix is pretty easy: wrap up the functions in this PR to follow the idiom #266 describes

* #455 is related to this issue in that if the answer to #455 is "store redshifts on spectrum objects", then this function should be aware of that and do something sensible (like either shifting the template to match the object or vice versa).  But since #455 is not resolved, I don't think it should *block* this - instead I think the current PR doesn't do anything with the redshift, right?  If so then we can add this functionality in a follow-on when #455 and #422 are resolved.

specutils/analysis/template_match.py
```diff
@@ -0,0 +1,105 @@
...
+    Find what instance spectral_templates is and run _template_match accordingly. If two template_spectra have the same
+    chi2, the first template is returned
+
+    Parameters
+    ----------
+    observed_spectrum : :class:`~specutils.Spectrum1D`
+        the observed spectrum
+    spectral_templates : :class:`~specutils.Spectrum1D` or :class:`~specutils.SpectrumCollection` or `list`
+        the template spectra, which will be resampled and normalized and compared to the observed_spectrum, where the
+        smallest chi2 and normalized_template_spectrum will be returned
```
eteq: If you go with my duck-typing suggestion below, you would want to clarify  here that the input needs to be anything where iterating over it gives a *single* `Spectrum1D`.  That will then demonstrate to the user that if they give something like an `(n,m, nwavelength)` `SpectrumCollection`, that's their fault for giving an invalid input and they need to reshape it into `(n*m, nwavelength)`

specutils/analysis/template_match.py
```diff
@@ -0,0 +1,105 @@
+from ..spectra.spectrum1d import Spectrum1D
+from ..spectra.spectrum_collection import SpectrumCollection
+from ..manipulation import FluxConservingResampler
+import numpy as np
+
+def _normalize(observed_spectrum, template_spectrum):
```
eteq: (minor) maybe call this something slightly more specific like `_normalize_for_template_matching`?  Goes with my suggestion of generifying the file name a bit so that we share this with other related use cases.

specutils/analysis/template_match.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # Create normalized template spectrum, which will be returned with corresponding chi2
+    normalized_template_spectrum = Spectrum1D(spectral_axis=template_spectrum.spectral_axis,
+                                              flux=template_spectrum.flux*normalization)
+
+    return normalized_template_spectrum, chi2
+
+def template_match(observed_spectrum, spectral_templates):
+    """
+    Find what instance spectral_templates is and run _template_match accordingly. If two template_spectra have the same
```
eteq: It's confusing for the docstring to reference a private function because the users don't know about the private function.  Maybe re-phrase as something like 
```suggestion
    Find which spectral templates is the best fit to an observed spectrum by computing the chi-squared. If two template_spectra have the same
```

specutils/analysis/template_comparison.py
```diff
@@ -0,0 +1,138 @@
...
+
+        index = 0
+        try:
+            for spectrum in spectral_templates:
+                normalized_spectral_template, chi2 = _template_match(observed_spectrum, spectrum, resample_method)
+                if chi2_min is None or chi2 < chi2_min:
+                    chi2_min = chi2
+                    smallest_chi_spec = normalized_spectral_template
+                    smallest_chi_index = index
+                index+=1
```
eteq: (very minor PEP8)
```suggestion
                index += 1
```

docs/analysis.rst
```diff
@@ -164,6 +164,30 @@ Each of the width analysis functions are applied to this spectrum below:
...
+       parameter is iterable
+
+An example of how to do template matching in is:
+
+.. code-block:: python
+
+   >>> from specutils.analysis import template_comparison
+   >>> spec_axis = np.linspace(0, 50, 50) * u.AA
+   >>> spec = Spectrum1D(spectral_axis=spec_axis, flux=np.random.randn(50) * u.Jy, uncertainty=StdDevUncertainty(np.random.sample(50), unit='Jy'))
+   >>> spec1 = Spectrum1D(spectral_axis=spec_axis, flux=np.random.randn(50) * u.Jy, uncertainty=StdDevUncertainty(np.random.sample(50), unit='Jy'))
```
eteq: Two related things:
1. Instead of `spec1`, perhaps this should be `spectral_template`? And maybe it makes sense to have this be a vector spectrum1d so the user sees that, which I think is likely to be the more scientific-relevant use case? I.e.,
```suggestion
   >>> spec1 = Spectrum1D(spectral_axis=spec_axis, flux=np.random.randn(2, 50) * u.Jy, uncertainty=StdDevUncertainty(np.random.sample(2, 50), unit='Jy'))
```

Later I think we want to beef this up with a more scientifically-relevant example using actual astronomical test data, but we don't need to do that in this PR.

specutils/analysis/template_comparison.py
```diff
@@ -0,0 +1,138 @@
...
+        that will give a single :class:`~specutils.Spectrum1D` when iterated over.
+        The template spectra, which will be resampled and normalized and compared to the observed_spectrum, where the
+        smallest chi2 and normalized_template_spectrum will be returned
+
+    Returns
+    -------
+    normalized_template_spectrum : :class:`~specutils.Spectrum1D`
+        the template_spectrum that has been normalized
+    chi2 : `float`
+        the chi2 of the flux of the observed_spectrum and the flux of the normalized_template_spectrum
```
eteq: This needs to also now have `smallest_chi_index` listed here as the third return

specutils/analysis/template_comparison.py
```diff
@@ -0,0 +1,140 @@
...
+        try:
+            for spectrum in spectral_templates:
+                normalized_spectral_template, chi2 = _template_match(observed_spectrum, spectrum, resample_method)
+                if chi2_min is None or chi2 < chi2_min:
+                    chi2_min = chi2
+                    smallest_chi_spec = normalized_spectral_template
+                    smallest_chi_index = index
+                index += 1
+        except Exception as e:
+            print("Parameter spectral_templates is not iterable. The following error was fired: {}".format(e))
```
nmearl:
```suggestion
            logging.warning("Parameter spectral_templates is not iterable. The following error was fired: {}".format(e))
```

specutils/analysis/template_comparison.py
```diff
@@ -0,0 +1,140 @@
+from ..spectra.spectrum1d import Spectrum1D
+from ..spectra.spectrum_collection import SpectrumCollection
+from ..manipulation import FluxConservingResampler
+from ..manipulation import LinearInterpolatedResampler
+from ..manipulation import SplineInterpolatedResampler
+import numpy as np
```
nmearl:
```suggestion
import numpy as np
import logging
```

PR #490, opened 2019-07-20
fix api changes with new modeling

nden:
**DO NOT MERGE:** Requires astropy/astropy#8769 

The changes in this PR are necessary to work with the refactored modeling. They are also mostly fixing code implemented as work around issues with modeling. They are due to

- Model parameters now store their values and constraints (as opposed to storing these in the Model).
  The estimators are implemented now as parameter attributes.
- `Model.n_submodels` is now a property, not a method.

The changes here are not made to be backwards compatible. If backwards compatibility is important it can be added with a check for the astropy version.

One test still fails because the constraints are not copied over when a model without units is created. This can be fixed in specutils and should eventually be fixed in modeling as well but I wanted to explore whether this is actually still necessary. The reason it was done (AFAIK) is to fit models with units. Since such changes were already made in modeling it'd be good to know if there are any lingering issues related to fitting of models with units.

### DIFF
--- a/specutils/fitting/fitmodels.py
+++ b/specutils/fitting/fitmodels.py
@@ -5,7 +5,6 @@ import logging
 import numpy as np
 from scipy.signal import convolve
 
-import astropy
 import astropy.units as u
 from astropy.stats import sigma_clipped_stats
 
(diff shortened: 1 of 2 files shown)


nden:
Rebased though I see multiple test failures lcoally.

nden:
@eteq I rebased the PR. Some tests are failing, mostly related to slicing

https://travis-ci.org/astropy/specutils/jobs/615287073

Only the dev Travis job is relevant.

eteq:
Note these commits were subsumed into #462 so I am closing this now that that is merged

Issue #258, opened 2018-07-23
Where should Redshift/RV be stored on Spectrum objects, and how to transform between different forms - RV(opt)/RV(rad)


SaOgaz:
original card by @keflavich

stscijgbot:
This ticket is now being tracked at [JDAT-7](https://jira.stsci.edu/browse/JDAT-7)

ibusko:
To implement this without solving the idiom issue at #455 probably means we should do it in a top-level function in the `spectrum1d.py`module. We can later move it to wherever we decide the redshift/rv info should be stored. 
The conversion will certainly be using the already existing `Spectrum1D` parameters `rest_value` and `velocity_convention`. What are these exactly? Is `rest_value` a scalar or a 1-D array? Is `velocity_convention` just a way to tell the difference between frequency and wavelength definitions? I bet there is more than that. What `doppler_relativistic` means in this context?

nmearl:
`rest_value` is the wavelength used in the velocity conversions (e.g. wavelength -> velocity conversions will result in dramatically different dispersions depending on which wavelength is considered the rest wavelength).

> Is velocity_convention just a way to tell the difference between frequency and wavelength definitions?

`velocity_convention` defines which of the three velocity conventions to use in a dispersion conversion (radio, optical, relativistic). Each uses a different formula.

ibusko:
Aren't radio and optical essentially the same conversion? The differences in between the formulae of these two have to do with spectral axis units, but for the non-relativistic approximation, the formulae perform the same operation. Or am I missing something?

Maybe we need a clarification on the purpose of this translation. Is it meant to generate a velocity axis in velocity units, against which the spectrum data can be paired? Is it meant to generate a rest spectral axis in wavelength/frequency units? Or something else? 

This discussion would probably benefit from  work done under #455

keflavich:
No, the optical and radio conventions use different equations:
http://docs.astropy.org/en/stable/api/astropy.units.equivalencies.doppler_optical.html
http://docs.astropy.org/en/stable/api/astropy.units.equivalencies.doppler_radio.html
http://docs.astropy.org/en/stable/api/astropy.units.equivalencies.doppler_relativistic.html

nmearl:
> Aren't radio and optical essentially the same conversion?

No. The differences in the formulas should be obvious below:

Radio:
[...]

PR #491, opened 2019-07-24
Add redshift correction to velocity conversions.

ibusko:
This PR addresses in part issue #258. It might, or might not be the correct interpretation of what this issue was intended to address in the first place.  I understood the goal is to apply a known redshift correction to the `velocity` attribute of `Spectrum1D` instances. 

The implementation uses the simple low-z approximate definition of redshift. It remains to be seen if this is what is actually intended. And, if not, we need more specific requirements on how to interpret redshift values, and what to do with them.

### DIFF
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -33,9 +33,11 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
         Any quantity supported by the standard spectral equivalencies
         (wavelength, energy, frequency, wave number). Describes the rest value
         of the spectral axis for use with velocity conversions.
-    redshift : float
-        The redshift correction to be applied to the spectral axis for use
-        with velocity conversions.
+    redshift : float or `~astropy.units.Quantity`
+        The redshift/RV correction to be applied to the spectral axis for use
+        with velocity conversions. If float, the value is interpreted as the
+        redshift z. If an instance of `~astropy.units.Quantity` is passed,
+        it is interpreted as a radial velocity with the appropriate velocity units.
     uncertainty : `~astropy.nddata.NDUncertainty`
         Contains uncertainty information along with propagation rules for
         spectrum arithmetic. Can take a unit, but if none is given, will use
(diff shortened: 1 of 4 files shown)


eteq:
I updated this to reflect #455, as we want to try to get this into 0.6 (and @ibusko indicated he didn't have time to do that).

specutils/spectra/spectrum1d.py
```diff
@@ -165,6 +182,53 @@ def bin_edges(self):
...
+    def redshift(self):
+        """
+        The redshift(s) of the objects represented by this spectrum.  May be
+        scalar (if this spectrum's ``flux`` is 1D) or vector.  Note that
+        the concept of "redshift of a spectrum" can be ambiguous, so the
+        interpretation is set to some extent by either the user, or operations
+        (like template fitting) that set this attribute when they are run on
+        a spectrum.
+        """
+        return self._radial_velocity/cnst.c
```
```suggestion
        return self._radial_velocity/cnst.c
        
```

specutils/spectra/spectrum1d.py
```diff
@@ -165,6 +182,53 @@ def bin_edges(self):
...
+    @redshift.setter
+    def redshift(self, val):
+        if val is None:
+            self._radial_velocity = None
+        else:
+            self.radial_velocity = val * cnst.c
```
nmearl: This assumption is really only true for small redshifts. Should we use the stricter definition?

specutils/spectra/spectrum1d.py
```diff
@@ -165,6 +182,53 @@ def bin_edges(self):
...
+    def radial_velocity(self):
+        """
+        The radial velocity(s) of the objects represented by this spectrum.  May
+        be scalar (if this spectrum's ``flux`` is 1D) or vector.  Note that
+        the concept of "RV of a spectrum" can be ambiguous, so the
+        interpretation is set to some extent by either the user, or operations
+        (like template fitting) that set this attribute when they are run on
+        a spectrum.
+        """
+        return self._radial_velocity
```
```suggestion
        return self._radial_velocity
        
```

PR #499, opened 2019-07-31
Refactor mixins

nmearl:
@keflavich


keflavich:
`spectral-cube` still relies on having a `FITS WCS` object as `SpectralCube.wcs` for the other dimensions; for example, it uses things like the `celestial` convenience function.  However, `specutils` requires that `wcs` have a `with_spectral_unit` method defined, which will happen once APE14 is completed.  So, we need to wait on APE14.

Also, APE14 probably needs a `.spectral` method to complement the `.celestial` method, and maybe a `.temporal` and `.polarizational` too?

Issue #326, opened 2018-09-14
Spectrum1D constructor should check whether flux and uncertainty dimensions match


drdavella:

brechmos:
I just hit this one too. A simple script:
```
import numpy as np
import astropy.units as u
from specutils import Spectrum1D

wavelengths = np.arange(1, 10)
s1d = Spectrum1D(spectral_axis=wavelengths*u.um, flux=np.random.randn(10)*u.Jy)
```

Seems innocuous but is returning:
```
Out[42]: ---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/.pyenv/versions/3.7.2/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj)
    700                 type_pprinters=self.type_printers,
    701                 deferred_pprinters=self.deferred_printers)
--> 702             printer.pretty(obj)
    703             printer.flush()
    704             return stream.getvalue()

~/.pyenv/versions/3.7.2/lib/python3.7/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    400                         if cls is not object \
    401                                 and callable(cls.__dict__.get('__repr__')):
--> 402                             return _repr_pprint(obj, self, cycle)
    403
    404             return _default_pprint(obj, self, cycle)

~/.pyenv/versions/3.7.2/lib/python3.7/site-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    695     """A pprint that just redirects to the normal repr function."""
    696     # Find newlines and replace them with p.break_()
--> 697     output = repr(obj)
    698     for idx,output_line in enumerate(output.splitlines()):
    699         if idx:

~/pCloudDrive/Documents/work/stsci/datb/specutils/specutils/specutils/spectra/spectrum1d.py in __repr__(self)
    264     def __repr__(self):
    265         inner_str = "flux={}, spectral_axis={}".format(repr(self.flux),
--> 266                                                        repr(self.spectral_axis))
    267
    268         if self.uncertainty is not None:

/opt/venvs/venv-specutils/lib/python3.7/site-packages/astropy/utils/decorators.py in __get__(self, obj, owner)
    742                 return val
    743             else:
--> 744                 val = self.fget(obj)
    745                 obj.__dict__[self._key] = val
    746                 return val
...
ValueError: One of the requested xi is out of bounds in dimension 0
```
The error is not helpful.  

An exception should be raised on construction if the spectral_axis and flux axis lengths don't match.

PR #508, opened 2019-09-06
Raise exception when spectral and flux axes have different lengths

brechmos-stsci:
This will check the spectral_axis and flux axis lengths to make sure they are the same and will raise an exception if that is not the case. 

Fixes https://github.com/astropy/specutils/issues/326

### DIFF
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -60,6 +60,12 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
         if flux is None and 'data' in kwargs:
             flux = kwargs.pop('data')
 
+        # Check to make sure the wavelength length is the same in both
+        if flux is not None and spectral_axis is not None:
+            if not spectral_axis.shape[0] == flux.shape[-1]:
+                raise ValueError('Spectral axis ({}) and flux axis ({}) lengths must be the same'.format(
+                    spectral_axis.shape[0], flux.shape[-1]))
+
         # Ensure that the unit information codified in the quantity object is
         # the One True Unit.
         kwargs.setdefault('unit', flux.unit if isinstance(flux, u.Quantity)
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -91,7 +91,7 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
         # Check to make sure the wavelength length is the same in both
         if flux is not None and spectral_axis is not None:
             if not spectral_axis.shape[0] == flux.shape[-1]:
-                raise ValueError('Spectral axis ({}) and flux axis ({}) lengths must be the same'.format(
+                raise ValueError('Spectral axis ({}) and the last flux axis ({}) lengths must be the same'.format(
                     spectral_axis.shape[0], flux.shape[-1]))
 
         self._velocity_convention = velocity_convention
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -120,7 +120,7 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
 
         if hasattr(self, 'uncertainty') and self.uncertainty is not None:
             if not flux.shape == self.uncertainty.array.shape:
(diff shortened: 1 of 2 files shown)


specutils/tests/test_spectrum1d.py
```diff
@@ -45,6 +45,11 @@ def test_create_from_multidimensional_arrays():
     assert (spec.frequency == freqs).all()
     assert (spec.flux == flux).all()
 
+    # Mis-matched lengths should raise and exception
```
nmearl:
```suggestion
    # Mis-matched lengths should raise an exception
```

specutils/tests/test_spectrum1d.py
```diff
@@ -54,6 +59,11 @@ def test_create_from_quantities():
     assert spec.spectral_axis.unit == u.nm
     assert spec.spectral_axis.size == 49
 
+    # Mis-matched lengths should raise and exception
```
nmearl:
```suggestion
    # Mis-matched lengths should raise an exception
```

specutils/spectra/spectrum1d.py
```diff
@@ -60,6 +60,12 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
         if flux is None and 'data' in kwargs:
             flux = kwargs.pop('data')
 
+        # Check to make sure the wavelength length is the same in both
```
nmearl: This block should occur after the parsing of the spectral axis and flux values (the if/else block starting at line 77). This is because of the way the arithmetic works on nddata objects -- it's possible that `flux` is a single float value since operands get converted to the nddata subclass (i.e. `my_spec * 15`, `15` is converted to an instance of `Spectrum1D`).

dhomeier: I would suggest to complement this with a check for the length/shape of the `uncertainty` property, if set, e.g.
```
        if flux is not None and self.uncertainty is not None:
            if self.uncertainty.array.shape != flux.shape:
                raise ValueError('Flux ({}) and uncertainty ({}) shapes must match'.format(
                    flux.shape, self.uncertainty.array.shape))
```
(probably needs to be inserted after the final `super(Spectrum1D, self).__init__`). Don't know if the above catches all valid cases, or if, say, a single value for `uncertainty` (shape `()`) should also be accepted.

specutils/tests/test_spectrum1d.py
brechmos: fixed

specutils/spectra/spectrum1d.py
brechmos: hopefully the order is better now.
added in ck for uncertainty shape too

dhomeier:
Just a note on the docs/diagnostics: the error message now states "must be the same (or uncertainty must be a singleton", but a length-1 or scalar `uncertainty` is actually rejected.

specutils/spectra/spectrum1d.py
```diff
@@ -112,6 +118,12 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
             data=flux.value if isinstance(flux, u.Quantity) else flux,
             wcs=wcs, **kwargs)
 
+        if hasattr(self, 'uncertainty') and self.uncertainty is not None:
+            if not flux.shape == self.uncertainty.array.shape:
+                raise ValueError('Flux axis ({}) and uncertainty ({}) shapes must be the same (or uncertainty must be a singleton)'.format(
```
```suggestion
                raise ValueError('Flux axis ({}) and uncertainty ({}) shapes must be the same.'.format(
```

specutils/spectra/spectrum1d.py
```diff
@@ -88,6 +88,12 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
             raise LookupError("No WCS object or spectral axis information has "
                               "been given. Please provide one.")
 
+        # Check to make sure the wavelength length is the same in both
+        if flux is not None and spectral_axis is not None:
+            if not spectral_axis.shape[0] == flux.shape[-1]:
+                raise ValueError('Spectral axis ({}) and flux axis ({}) lengths must be the same'.format(
```
eteq: Maybe this should be 
```suggestion
                raise ValueError('Spectral axis ({}) and the last flux axis ({}) lengths must be the same'.format(
```
?  Because for a vector `Spectrum1D`, "flux axis" is potentially ambiguous

Issue #516, opened 2019-09-19
Ensure that existing manipulation and analysis functions use the `mask` where appropriate.


eteq:
This comes out of #509, which introduces a function that computes and assigns `mask`'s to `Spectrum1D` objects.  This led to a discussion where it was realized that a lot of the downstream analysis tasks that would follow on from #509 need to respect the `mask` for #509 to actually be practically useful - without that it seemed like a `flux_masked` or similar seemed necessary, which is not desirable since it essentially replicates the mask in `Spectrum1D` itself.

So the point of this issue is to make sure the currently-existing `analysis` and `manipulation` tools are mask-aware. At the time of writing this, I think the current status is:

* `specutils.analysis` functions: `centroid`, `equivalent_width`, `fwhm`, `gaussian_fwhm`, `gaussian_sigma_width`, `line_flux`, `snr` - all of these ignore the mask entirely, but I think it's a pretty easy thing to make them work: if the mask is `None`, do what it does now, otherwise, just do `self.flux[~self.mask]` where `self.flux` currently appears (and `self.spectral_axis` -> `self.spectral_axis[~self.mask]`, although they don't all use the spectral axis).
* line and continuum fitting (same underlying implementation. in `fit_lines`): I think this is *also* basically a one-liner in the implementation, doing essentially the above, but in the call to the fitter.
* parameter estimators in @brechmos-stsci may be able to advise here, but I *think* these are all eventually implemented via the `analysis` functions, right?  So then if we fix those then it automatically carries over to the estimators? 
* `extract_region`: already does the correct thing of just carrying through the mask.

The one question in my mind that applies to all of the above: do we want `use_mask` to be a keyword we add to *all* of these?  My initial inclination is no: we always use the mask, but if the user does not want that, they can just set the mask to None. Or alternatively, 

cc @brechmos-stsci @camipacifici @nmearl

eteq:
Oh and since I already looked at it, I'll make a PR for `fit_lines`.

hcferguson:
This looks sensible to me.

Regarding without_mask, it seems like there should be some way of doing this without copying any data. In the case where the user substitutes the mask with None, there is no copy. But it's inconvenient for the user to get the mask back. It would be nice to be able to toggle the mask on and off, which is what you are getting at, but do it by just temporarily changing the reference of the mask attribute. 

[...]

PR #519, opened 2019-09-19
Make line/continuum fitting work in the mask case (and other mask-related tests)

eteq:
This is the PR I promised in #516. It fixes the line-fitting machinery to ignore any masked parts of the spectrum when doing a fit.

Note this also includes a couple of tests I added in process of the above to ensure the masks behave correctly for arithmetic and spectral regions.

cc @brechmos-stsci @camipacifici @nmearl 

(This may particularly benefit from a review by @brechmos-stsci who did a fair amount of the fitting stuff)

### DIFF
--- a/specutils/fitting/fitmodels.py
+++ b/specutils/fitting/fitmodels.py
@@ -284,10 +284,12 @@ def fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
         Fitter instance to be used when fitting model to spectrum.
     exclude_regions : list of `~specutils.SpectralRegion`
         List of regions to exclude in the fitting.
-    weights : list or 'unc', optional
+    weights : array-like or 'unc', optional
         If 'unc', the unceratinties from the spectrum object are used to
-        to calculate the weights. If list/ndarray, represents the weights to
-        use in the fitting.
+        to calculate the weights. If array-like, represents the weights to
+        use in the fitting.  Not that if a mask is present on the spectrum, it
+        will be applied to the ``weights`` as it would be to the spectrum
+        itself.
     window : `~specutils.SpectralRegion` or list of `~specutils.SpectralRegion`
         Regions of the spectrum to use in the fitting. If None, then the
         whole spectrum will be used in the fitting.
@@ -399,11 +401,14 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
         # Assume that the weights argument is list-like
         weights = np.array(weights)
 
+    mask = spectrum.mask
+
     dispersion = spectrum.spectral_axis
     dispersion_unit = spectrum.spectral_axis.unit
 
     flux = spectrum.flux
     flux_unit = spectrum.flux.unit
+
 
     #
     # Determine the window if it is not None.  There
@@ -426,6 +431,9 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
         dispersion = dispersion[indices]
         flux = flux[indices]
 
+        if mask is not None:
+            mask = mask[indices]
+
         if weights is not None:
             weights = weights[indices]
 
@@ -438,10 +446,14 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
         dispersion = dispersion[indices]
         flux = flux[indices]
 
+        if mask is not None:
+            mask = mask[indices]
+
         if weights is not None:
             weights = weights[indices]
 
     elif window is not None and isinstance(window, SpectralRegion):
+        # TODO: fix this for weights and masks
         try:
             idx1, idx2 = window.bounds
             if idx1 == idx2:
@@ -478,6 +490,12 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
     #
     # Do the fitting of spectrum to the model.
     #
+    if mask is not None:
+        nmask = ~mask
+        dispersion_unitless = dispersion_unitless[nmask]
+        flux_unitless = flux_unitless[nmask]
+        if weights is not None:
+            weights = weights[nmask]
 
     fit_model_unitless = fitter(model_unitless, dispersion_unitless,
                                 flux_unitless, weights=weights, **kwargs)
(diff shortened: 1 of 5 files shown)


specutils/fitting/fitmodels.py
```diff
@@ -284,10 +284,12 @@ def fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
         Fitter instance to be used when fitting model to spectrum.
     exclude_regions : list of `~specutils.SpectralRegion`
         List of regions to exclude in the fitting.
-    weights : list or 'unc', optional
+    weights : array-like or 'unc', optional
         If 'unc', the unceratinties from the spectrum object are used to
-        to calculate the weights. If list/ndarray, represents the weights to
-        use in the fitting.
+        to calculate the weights. If array-like, represents the weights to
+        use in the fitting.  Not that if a mask is present on the spectrum, it
```
keflavich:
```suggestion
        use in the fitting.  Note that if a mask is present on the spectrum, it
```

specutils/fitting/fitmodels.py
```diff
@@ -418,40 +423,49 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
     #
 
     # In this case the window defines the area around the center of each model
+    window_indices = None
     if window is not None and isinstance(window, (float, int)):
         center = model.mean
-        indices = np.nonzero((spectrum.spectral_axis >= center-window) &
+        window_indices = np.nonzero((spectrum.spectral_axis >= center-window) &
```
keflavich: probably should change indentation of subsequent line?

specutils/fitting/fitmodels.py
```diff
@@ -418,40 +423,49 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
...
-                raise Exception("Bad selected region.")
-            extracted_regions = extract_region(spectrum, window)
-            dispersion, flux = _combined_region_data(extracted_regions)
-            dispersion = dispersion * dispersion_unit
-            flux = flux * flux_unit
-        except ValueError as e:
-            return
+        idx1, idx2 = window.bounds
+        if idx1 == idx2:
+            raise Exception("Bad selected region.")
```
keflavich: don't raise a base exception - this could be an `IndexError`?

specutils/fitting/fitmodels.py
```diff
@@ -418,40 +423,49 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
...
+
+        # HACK WARNING! This uses the extract machinery to create a set of
+        # indices by making an "index spectrum"
+        # TODO: really the spectral region machinery should have the power
+        # to create a mask, and we'd just use that...
+        index_spectrum = Spectrum1D(spectral_axis=spectrum.spectral_axis,
+                                    flux=np.arange(spectrum.flux.size).reshape(spectrum.flux.shape)*u.Jy)
```
keflavich: is the `Jy` unit addition needed?  That adds a small performance hit and is weird for what should be an integer array to begin with (on which note, should this be `arange(..., dtype='int')`?)

specutils/tests/spectral_examples.py
```diff
@@ -71,22 +76,29 @@ def __init__(self):
...
 
         self._flux_e4 = (self.base_flux + 400 * np.random.random(self.base_flux.shape)) * 1000000
         self._s1_AA_nJy_e4 = Spectrum1D(spectral_axis=self.wavelengths_AA * u.AA,
                                         flux=self._flux_e4 * u.nJy)
 
 
+        #
+        # Create one spectrum like 1 but with a mask
+        #
+        self._s1_um_mJy_e1_masked = copy(self._s1_um_mJy_e1)  # SHALLOW copy
```
keflavich: maybe comment why shallow is emphasized here?  It's true, but does it matter in context?

specutils/tests/test_fitting.py
```diff
@@ -595,3 +595,55 @@ def test_spectrum_from_model():
                               1.5319218, 1.06886794, 0.71101768, 0.45092638, 0.27264641])
 
     assert np.allclose(spectrum_gaussian.flux.value[::10], flux_expected, atol=1e-5)
+
+
+def test_masking():
+    """
+    Test fitting spectra with masks
+    """
+    x, y = double_peak()
```
keflavich: avoid single-character variable names.... (my preference, not an official style recommendation, afaik)

specutils/fitting/fitmodels.py
eteq: pre-existing before this PR, but I agree!

specutils/fitting/fitmodels.py
eteq: The reasoning for this is that `Spectrum1D` expects flux inputs to be quantities - i.e., you can't give something that's unitless.  The choice of Jy is arbitrary, as really any quantity will do, but it seemed safest to use an actual flux. But you're right it makes sense to keep it as an `int` so I can switch to the explicit `u.Quantity(arr, u.Jy, dtype=int)`

As the "HACK WARNING" indicates this is defintely a hack, though I think it's probably better just to try to do the "TODO" ASAP instead beyond optimizing this any more.

specutils/tests/test_fitting.py
eteq: I think it's also in several of the default linters, so it's quasi-official... regardless, good point will fix

eteq:
I think I've addressed everything, so this should be good to go once the tests pass if @keflavich or @nmearl are willing to approve/merge it.

nmearl:
Looks like there's a hold over from the recent `MexicanHat` model name switch.

eteq:
@nmearl  - Re the mexicanhat failure:   I don't think it's related to this PR. I was about to added it but I realized we should treat it separately because I think we may want to leave the "old" name in for the Astropy 3.x compatible version.  So I made a separate PR #564 to address that and propose this be merged separately.

nmearl:
Just to clarify: this PR would be in for the astropy 3.x release, but the PR #564 would be included in the astropy 4.x release?

eteq:
@nmearl - yep, exactly!

eteq:
(since the test failure is unrelated I think this is good to go, and a release blocker)

pllim:
This PR actually introduced a bug that is being fixed at https://github.com/astropy/specutils/pull/1164 . FYI

PR #524, opened 2019-09-24
Implement spectral coordinate object

### LINKED ISSUE #422: Implement SpectralCoord class

nmearl:
Resolves #422. See the [astropy PR](https://github.com/nmearl/astropy/pull/1) for further discussion (which should take place here now).

### DIFF
--- a/specutils/spectra/spectral_coordinate.py
+++ b/specutils/spectra/spectral_coordinate.py
@@ -5,7 +5,7 @@ from collections import namedtuple
 
 import astropy.units as u
 from astropy.constants import c
-from astropy.coordinates import SkyCoord, ICRS, Distance, GCRS, RadialDifferential, CartesianDifferential, SphericalDifferential
+from astropy.coordinates import SkyCoord, ICRS, Distance, GCRS, CartesianRepresentation, CartesianDifferential, SphericalDifferential, SphericalRepresentation
 from astropy.coordinates.baseframe import BaseCoordinateFrame, FrameMeta
 from astropy.utils.exceptions import AstropyUserWarning
 
@@ -18,6 +18,8 @@ DOPPLER_CONVENTIONS = {
 RV_RS_EQUIV = [(u.cm / u.s, u.Unit(''),
                 lambda x: np.sqrt((1 + x/c.cgs.value)/(1 - x/c.cgs.value)) - 1,
                 lambda x: ((x + 1) ** 2 - 1) / ((x + 1) ** 2 + 1) * c.cgs.value)]
+
+DEFAULT_DISTANCE = 1 * u.AU
 
 DopplerConversion = namedtuple('DopplerConversion', ['rest', 'convention'])
 
@@ -95,23 +97,16 @@ class SpectralCoord(u.Quantity):
 
             observer_icrs = observer.transform_to(ICRS)
 
-            target = ICRS(ra=observer_icrs.ra,
-                          dec=observer_icrs.dec,
-                          pm_ra_cosdec=observer_icrs.pm_ra_cosdec,
-                          pm_dec=observer_icrs.pm_dec,
-                          distance=1 * u.kpc)
-
-            d_pos = (target.cartesian.without_differentials() -
-                     observer_icrs.cartesian.without_differentials())
-
-            pos_hat = d_pos / (d_pos.norm() or 1)
+            d = observer_icrs.cartesian.norm()
+            drep = CartesianRepresentation([(d + DEFAULT_DISTANCE).to(d.unit),
+                                            0 * d.unit, 0 * d.unit])
 
             tot_rv = radial_velocity + observer_icrs.radial_velocity
-            target_velocity = target.velocity + tot_rv * pos_hat
 
-            target = target.realize_frame(
-                target.cartesian.with_differentials(
-                    CartesianDifferential(target_velocity.xyz)))
+            target = (observer_icrs.cartesian.without_differentials() + drep).with_differentials(
+                CartesianDifferential([tot_rv, 0 * tot_rv.unit, 0 * tot_rv.unit]))
+
+            target = observer_icrs.realize_frame(target)
 
         obj._observer = cls._validate_coordinate(observer) if observer is not None else None
         obj._target = cls._validate_coordinate(target) if target is not None else None
--- a/specutils/spectra/spectral_coordinate.py
+++ b/specutils/spectra/spectral_coordinate.py
@@ -97,17 +97,6 @@ class SpectralCoord(u.Quantity):
 
             target = SpectralCoord._target_from_observer(observer, radial_velocity)
 
-            d = observer_icrs.cartesian.norm()
-            drep = CartesianRepresentation([(DEFAULT_DISTANCE).to(d.unit),
-                                            0 * d.unit, 0 * d.unit])
-
-            tot_rv = radial_velocity + observer_icrs.radial_velocity
-
-            target = (observer_icrs.cartesian.without_differentials() + drep).with_differentials(
-                CartesianDifferential([tot_rv, 0 * tot_rv.unit, 0 * tot_rv.unit]))
-
-            target = observer_icrs.realize_frame(target)
-
         obj._observer = cls._validate_coordinate(observer) if observer is not None else None
         obj._target = cls._validate_coordinate(target) if target is not None else None
 
@@ -285,7 +274,9 @@ class SpectralCoord(u.Quantity):
 
         self._frames_state['target'] = value is not None
 
-        value = self._validate_coordinate(value)
+        value = self._validate_coordinate(value,
+                                          pair_frame=self.observer,
+                                          radial_velocity=self.radial_velocity)
 
         self._target = value
 
@@ -426,7 +417,12 @@ class SpectralCoord(u.Quantity):
         d_pos = (target_icrs.data.without_differentials() -
                  observer_icrs.data.without_differentials()).to_cartesian()
 
-        pos_hat = d_pos / (d_pos.norm() or 1)
+        dp_norm = d_pos.norm()
+
+        # Reset any that are 0 to 1 to avoid nans from 0/0
+        dp_norm.ravel()[dp_norm.ravel() == 0] = 1 * dp_norm.unit
+
+        pos_hat = d_pos / dp_norm
         d_vel = target_icrs.velocity - observer_icrs.velocity
 
         return np.sum(d_vel.d_xyz * pos_hat.xyz, axis=0)
--- a/specutils/spectra/spectral_coordinate.py
+++ b/specutils/spectra/spectral_coordinate.py
@@ -4,7 +4,7 @@ import numpy as np
 
 import astropy.units as u
 from astropy.constants import c
-from astropy.coordinates import SkyCoord, ICRS, Distance, GCRS
+from astropy.coordinates import SkyCoord, ICRS, Distance, GCRS, RadialDifferential, CartesianDifferential
 from astropy.coordinates.baseframe import BaseCoordinateFrame, FrameMeta
 from astropy.utils.compat import NUMPY_LT_1_14
 from astropy.utils.exceptions import AstropyUserWarning
@@ -68,8 +68,8 @@ class SpectralCoord(u.Quantity):
                 raise ValueError("Observer must be a sky coordinate or "
                                  "coordinate frame.")
 
-        obj._observer = observer.frame if hasattr(observer, 'frame') else observer
-        obj.target = target.frame if hasattr(observer, 'frame') else target
+        obj.observer = observer
+        obj.target = target
 
         # TODO: I'm not sure what this radial velocity represents
         obj._radial_velocity = radial_velocity
@@ -127,8 +127,11 @@ class SpectralCoord(u.Quantity):
 
             # If no distance value is defined on the frame, assume a default
             # distance of either zero (for an observer), or 1000 kpc for target
-            if not hasattr(coord, 'distance') or \
-                    not isinstance(coord.distance, Distance):
+            if (not hasattr(coord, 'distance') or
+                not isinstance(coord.distance, Distance)) and \
+                (not hasattr(coord, 'spherical') or
+                 not isinstance(coord.spherical.distance, Distance)):
+
                 auto_dist = 0 * u.AU if is_observer else 1000 * u.kpc
 
                 warnings.warn(
@@ -147,7 +150,10 @@ class SpectralCoord(u.Quantity):
                        frame_loc, u.Quantity([0, 0, 0], unit=u.km/u.s)),
                     AstropyUserWarning)
 
-                coord = SkyCoord(coord, radial_velocity=0 * u.km / u.s).frame
+                vel_to_add = CartesianDifferential(
+                    0 * u.km / u.s, 0 * u.km / u.s, 0 * u.km/u.s)
+                new_data = coord.data.to_cartesian().with_differentials(vel_to_add)
+                coord = coord.realize_frame(new_data)
 
         return coord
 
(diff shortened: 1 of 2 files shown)


specutils/spectra/spectral_coordinate.py
```diff
@@ -0,0 +1,433 @@
...
+        return self._velocity_convention
+
+    @velocity_convention.setter
+    def velocity_convention(self, value):
+        """
+        New velocity convention used for velocity space conversions.
+
+        Parameters
+        ----------
+        value
```
keflavich: todo: add docs here?  Or is the expectation that this docstring won't be visible?

nmearl:
@eteq @astrofrog Any further opinions on this approach?

astrofrog:
I ran into an issue with the following example:

```python
In [1]: from astropy import units as u 
   ...: from astropy.coordinates import SkyCoord                                                                                                       

In [2]: LSRD = SkyCoord(u=0 * u.km, v=0 * u.km, w=0 * u.km, 
   ...:                 U=9 * u.km / u.s, V=12 * u.km / u.s, W=7 * u.km / u.s, 
   ...:                 frame='galactic', representation_type='cartesian', differential_type='cartesian')                                              

In [3]: from specutils.spectra.spectral_coordinate import SpectralCoord                                                                                

In [4]: spectral_axis = SpectralCoord([3, 4, 5] * u.micron, observer=LSRD)                                                                             
<ipython-input-4-947d16e15fcb> in <module>
----> 1 spectral_axis = SpectralCoord([3, 4, 5] * u.micron, observer=LSRD)

~/Dropbox/Code/Astropy/specutils/specutils/spectra/spectral_coordinate.py in __new__(cls, value, unit, radial_velocity, redshift, rest, velocity_convention, observer, target, **kwargs)
     58         obj.velocity_convention = velocity_convention
     59 
---> 60         obj.observer = observer
     61         obj.target = target
     62 

~/Dropbox/Code/Astropy/specutils/specutils/spectra/spectral_coordinate.py in observer(self, value)
    145     @observer.setter
    146     def observer(self, value):
--> 147         value = self._validate_frame(value, True)
    148         self._observer = value
    149 

~/Dropbox/Code/Astropy/specutils/specutils/spectra/spectral_coordinate.py in _validate_frame(frame, is_observer)
    110             # If no distance value is defined on the frame, assume a default
    111             # distance of either zero (for an observer), or 1000 kpc for target
--> 112             if not isinstance(frame.distance, Distance):
    113                 auto_dist = 0 * u.AU if is_observer else 1000 * u.kpc
    114 

~/python/dev/lib/python3.7/site-packages/astropy/coordinates/baseframe.py in __getattr__(self, attr)
   1543             return val
   1544 
-> 1545         return self.__getattribute__(attr)  # Raise AttributeError.
   1546 
   1547     def __setattr__(self, attr, value):

AttributeError: 'Galactic' object has no attribute 'distance'
```

I haven't had a chance to investigate - am I doing something that shouldn't work?

astrofrog:
Another thing which isn't an issue as such but would be nice to have work: if we pass an ``EarthLocation`` to ``observer`` it would be nice to translate on-the-fly to a coordinate object:

```python
spectral_axis = SpectralCoord([3, 4, 5] * u.micron, observer=EarthLocation.of_site('alma'))
```

At the moment this gives:

```
ValueError: `(2225015.30883296, -5440016.41799762, -2481631.27428014) m` is not a subclass of `BaseCoordinateFrame` or `SkyCoord`.
```

astrofrog:
If I modify the example with ``EarthLocation`` I still get an error related to ``distance``:


```
In [5]: from astropy.coordinates import EarthLocation                                                                                                  

In [6]: alma = EarthLocation.of_site('ALMA').itrs                                                                                                      

In [7]: from specutils.spectra.spectral_coordinate import SpectralCoord                                                                                

In [8]: spectral_axis = SpectralCoord([3, 4, 5] * u.micron, observer=alma)                                                                             
<ipython-input-8-9017f7b95060> in <module>
----> 1 spectral_axis = SpectralCoord([3, 4, 5] * u.micron, observer=alma)

~/Dropbox/Code/Astropy/specutils/specutils/spectra/spectral_coordinate.py in __new__(cls, value, unit, radial_velocity, redshift, rest, velocity_convention, observer, target, **kwargs)
     58         obj.velocity_convention = velocity_convention
     59 
---> 60         obj.observer = observer
     61         obj.target = target
     62 

~/Dropbox/Code/Astropy/specutils/specutils/spectra/spectral_coordinate.py in observer(self, value)
    145     @observer.setter
    146     def observer(self, value):
--> 147         value = self._validate_frame(value, True)
    148         self._observer = value
    149 

~/Dropbox/Code/Astropy/specutils/specutils/spectra/spectral_coordinate.py in _validate_frame(frame, is_observer)
    110             # If no distance value is defined on the frame, assume a default
    111             # distance of either zero (for an observer), or 1000 kpc for target
--> 112             if not isinstance(frame.distance, Distance):
    113                 auto_dist = 0 * u.AU if is_observer else 1000 * u.kpc
    114 

~/python/dev/lib/python3.7/site-packages/astropy/coordinates/baseframe.py in __getattr__(self, attr)
   1543             return val
   1544 
-> 1545         return self.__getattribute__(attr)  # Raise AttributeError.
   1546 
   1547     def __setattr__(self, attr, value):

AttributeError: 'ITRS' object has no attribute 'distance'
```

nmearl:
> Another thing which isn't an issue as such but would be nice to have work: if we pass an EarthLocation to observer it would be nice to translate on-the-fly to a coordinate object:

@astrofrog I don't think we can support this, as currently to convert an `EarthLocation` to a `SkyCoord` requires that the user provide an observation time and frame: 

```python
alma = EarthLocation.of_site('alma')
obstime = time.Time('2018-12-13 9:00')

observer_gcrs = alma.get_gcrs(obstime)
```

nmearl:
Ah, well, I stand corrected: if the frame doesn't require it, like in your example (`ITRS`), there's no need for the observation time.

astrofrog:
@nmearl - thanks for the fixes! I'm running into an issue related to (I guess?) the repr for SpectralCoord:

Screenshot from 2019-10-20 13-44-51

This is using the same notebook as before, just with an extra ``spectral_axis`` cell. I think this should probably work since I didn't actually ask for any velocities?

astrofrog:
I got further this time! See

https://gist.github.com/astrofrog/b11eb8e0213b682658789b984f63cd9f

for the remaining issues. In short:

* repr often errors even if the info mentioned by the error shouldn't be needed
* I can't figure out how to set the observation time when creating an observer frame on Earth
* Adding zero velocities to observatories on the Earth's surface is tedious at the moment, and maybe EarthLocations, when converted to ITRS, should have zero velocities by default?
* I was having issues transforming SpectralCoord from an Earth-based frame to another frame

Thanks for continuing to work on this!

nmearl:
@astrofrog Thanks for looking this over! A few restrictions come to mind: it sounded like we shouldn't be in the business of transforming _frames_ (@eteq?). In the case where we want to change _observer_, we should really only allow `SkyCoordinates` to the `with_observer` method, or another observer-like object?

For your last line: `spectral_axis.with_observer(LSRD)`, I'm not sure what behavior you're trying to get? I can't even transform the `alma_with_vel` to the `LSRD` frame without astropy giving me an error:

```python
alma_lsrd = alma_with_vel.transform_to(LSRD)

NotImplementedError                       Traceback (most recent call last)
<ipython-input-30-d9e9fba3f3c1> in <module>
      1 print(LSRD.velocity)
----> 2 alma_lsrd = alma_with_vel.transform_to(LSRD)
      3 spectral_axis.with_observer(alma_lsrd)

~/anaconda3/envs/specutils_dev/lib/python3.7/site-packages/astropy/coordinates/baseframe.py in transform_to(self, new_frame)
   1161            hasattr(self, 'obstime') and hasattr(new_frame, 'obstime') and
   1162            np.any(self.obstime != new_frame.obstime)):
-> 1163             raise NotImplementedError('You cannot transform a frame that has '
   1164                                       'velocities to another frame at a '
   1165                                       'different obstime. If you think this '

NotImplementedError: You cannot transform a frame that has velocities to another frame at a different obstime. If you think this should (or should not) be possible, please comment at https://github.com/astropy/astropy/issues/6280
```

astrofrog:
Just to be clear I didn't mean changing the frame but indeed changing the observer via SkyCoord.

Regarding the last line:

```
alma_lsrd = alma_with_vel.transform_to(LSRD)
```

what I'm **really** trying to accomplish is to transform the velocities to an observer that is stationary compared to LSRD. I'm surprised by the error you showed when directly transforming the coordinates

In general, I think what I've realized by thinking about this some more is that I don't think ever really want to change the **actual** position of the observer, but it's more that we want to change to an observer at the same position that is stationary relative to the specified SkyCoord?

I think it might be productive to have a short voice chat about all this between the three of us to iron out some of these remaining issues?

astrofrog:
I've now played around with this more again since we last spoke. Latest notebook:

https://gist.github.com/astrofrog/11d925ab02282440709f63114221e650

The bottom line is that I still had issues with the case where the observer location changes and we take into account geometrical issues, but I managed to get the case where the observer location stays the same and the velocities change, although there were issues with the units on the result.

eteq:
I pushed up some tests with some specific since use cases in the last two commits above. There are some questions/negotiable items in the tests as comments, but they are probably best thought about once an implementation is in hand?

Also at the very end of them I do a comparison with `radial_velocity_correction`. That's a good standard used in very high precision RV work that I think we want as a long-term goal but is not necessarily strictly necessary now.But that function might contain some useful implementation hints, and has been validated on actual observed data!

nmearl:
It's been determined in offline discussions that the simplest implementation for now would be to use private temporary observer/target objects when the user does not explicitly define them upon initialization. These private objects can be used for some, but not all, calculations involving the spectral coordinate object -- mainly, certain transformations and methods (e.g. `_change_observer_to`, `in_observer_velocity_frame`) will not be allowed. The user can define a new observer or target only if none have been defined previously (otherwise, the user must create a new `SpectralCoord` object).

If a user does not define either a target or observer, two private frame objects will be created and populated with appropriate position and differential information such that radial velocity on the `SpectralCoord` object returns the value defined on initialization (if the user has provided an rv/redshift).

The target is created based on the observer information provided by the user (or a default observer that is hidden if no observer is provided). Here, the target is defined as having an offset of `1 AU` along the x axis of the observer in the cartesian representation. The value `radial_velocity` parameter used when instancing the `SpectralCoord` will be used to ensure that the regular radial velocity calculation machinery returns the same value (i.e. the radial velocity is now stored in the differential data for the observer/target pair).

rosteen:
Minor thing, but it looks like the input named "doppler_convention" in SpectralCoord class is the same thing as the input named "velocity_convention" in the Spectrum1D class. It seems like keeping the names consistent across classes would minimize potential confusion - should this variable name be changed here? I assume it would be more disruptive to update Spectrum1D since it's already in use.

nmearl:
That's a fair point! However, we originally _did_ use `velocity_convention` in the `SpectralCoord` object, but it caused confusion because `velocity_convention` is only useful as a means for transforming spectral axes, and is quite distinct from other usages of "velocity" to mean the velocity of the coordinate representing the observer / target. Likewise `rest` was ambiguous as well, because it wasn't clear that it only related to the spectral axis transformations.

After much debate, we decided with renaming `velocity_convention` to `doppler_convention`, and `rest` to `doppler_rest`.

We will most likely make these same changes in `Spectrum1D`, and use deprecation warnings for `velocity_convention` and `rest` to encourage users to update their code.

specutils/spectra/spectral_coordinate.py
```diff
@@ -0,0 +1,651 @@
...
+                raise ValueError(
+                    "Unrecognized doppler convention: {}.".format(
+                        doppler_convention))
+
+            self.doppler_convention = doppler_convention
+
+        equivs = u.spectral()
+
+        if self.doppler_rest is not None and self.doppler_convention is not None:
+            vel_equiv = DOPPLER_CONVENTIONS[self.doppler_convention](self.rest)
```
rosteen:
```suggestion
            vel_equiv = DOPPLER_CONVENTIONS[self.doppler_convention](self.doppler_rest)
```

specutils/spectra/spectral_coordinate.py
```diff
@@ -0,0 +1,651 @@
...
+
+DOPPLER_CONVENTIONS = {
+    'radio': u.doppler_radio,
+    'optical': u.doppler_optical,
+    'relativistic': u.doppler_relativistic
+}
+
+RV_RS_EQUIV = [(u.cm / u.s, u.Unit(''),
+                lambda x: np.sqrt((1 + x/c.cgs.value)/(1 - x/c.cgs.value)) - 1,
+                lambda x: ((x + 1) ** 2 - 1) / ((x + 1) ** 2 + 1) * c.cgs.value)]
```
eteq: This doesn't seem right to me (and I've coded this into the tests I pushed up) - the RV to redshift conversion is much simpler: v = c*z .  This looks like the relativistic doppler convention which is *not* the one used in the context of the normal definition of redshift.  So I think you just want this:


```suggestion
RV_RS_EQUIV = [(u.cm / u.s, u.Unit(''),
                lambda x: x/c.cgs.value,
                lambda x: x*c.cgs.value)]
```

specutils/spectra/spectral_coordinate.py
```diff
@@ -0,0 +1,651 @@
...
+        if target is None:
+            raise ValueError("No target has been specified; cannot calculate "
+                             "radial velocity.")
+
+        target_icrs = target.transform_to(ICRS)
+
+        d_pos = (target_icrs.data.without_differentials() -
+                 observer_icrs.data.without_differentials()).to_cartesian()
+
+        pos_hat = d_pos / (d_pos.norm() or 1)
```
eteq: This isn't right because there's no guarantee norm() yields a scalar - `d_pos` can be a complicated vector-y thing.

Also it yields a confusing warning when I just display this because numpy says this kind of usage is deprecated anyway.  I think instead you want something like this:
```suggestion
        dpnorm = d_pos.norm()
        # reset any that are 0 to 1 to avoid nans from 0/0.
        dpnorm.ravel()[dpnorm.ravel() == 0] = 1*dpnorm.unit

        pos_hat = d_pos / dpnorm
```

eteq:
(oh and note that the tests I pushed up also include  8b512a0fc689d9b6abb8e323f85672bdedd9a740 which I think fixes the mysterious failure due to not finding keck by switching to a builtin site that requires no download)

PR #527, opened 2019-10-02
Template redshifting using chi2

### LINKED ISSUE #488: Template fitting for redshift
This came from a discussion during our latest sprint (July 18th) review where we were discussing the template matching ticket (#256 ). This discussion focused on what happens before template matching. As I understand it: given a group of templates, the user will want to find the redshift that creates the lowest chi square, which is then useful for the template matching code.

javerbukh:
This is still a hacky version of what will eventually become the template redshifting solution found in #488.
The current issue I'm having is with testing the redshift solution. My current test goes from a `min_redshift` of .5 to a `max_redshift` of 5.5, with a `delta_redshift` of .25. The test prints the current redshift along with the corresponding chi2. Everything seems to work as intended, **except** at a redshift of 1, where the **chi2 is exactly the same** as the intended correct redshift of 3. @camipacifici if you could help me understand this result that would be much appreciated!

Resolves #488

### DIFF
--- a/docs/analysis.rst
+++ b/docs/analysis.rst
@@ -167,24 +167,33 @@ Each of the width analysis functions are applied to this spectrum below:
 Template Comparison
 -------------------
 
-The `specutils.analysis.template_comparison` script contains the
-`specutils.analysis.template_comparison.template_redshift` method and the
-`specutils.analysis.template_comparison.template_match` method.
-`specutils.analysis.template_comparison.template_redshift` finds the most accurate redshift to shift the template
-spectrum by to match the observed spectrum. `specutils.analysis.template_comparison.template_match` takes one
-observed spectrum and n template spectra (and optionally redshift parameters in order to shift the template
-spectrum to match the observed spectrum) and does the following process:
+The ~`specutils.analysis.template_comparison.template_match` function takes one
+observed spectrum and n template spectra and returns the best template that
+matches the observed spectrum via chi-square minimization.
 
-With one observed spectrum and n template spectra, the process to do template matching is:
+If the redshift is known, the user will redshift the template spectra
+appropriately and then run the
+~`specutils.analysis.template_comparison.template_match` function.
+This method will:
 
-    1. Move the templates as close as possible to the observed spectrum by doing redshifting, matching the resolution,
-       and matching the wavelength spacing
-    2. Then loop through all of the n template spectra (accounting for redshift if parameters are provided) and
-       run chi square on them using the observed spectrum as the expected frequency and the template spectrum as
-       the observed
-    3. Once you have a corresponding chi square for each template spectrum, return the lowest chi square and its
-       corresponding template spectrum (normalized) and the index of the template spectrum if the original template
-       parameter is iterable
+    1. Match the resolution and wavelength spacing of the observed spectrum
+    2. Compute the chi-square between the observed spectrum and each template
+    3. Return the lowest chi-square and its corresponding template spectrum,
+       normalized to the observed spectrum (and the index of the template
+       spectrum if the list of templates is iterable)
+
+If the redshift is unknown, the user specifies the expected window in redshift
+(min_redshift and max_redshift) and the redshift spacing (delta_redshift).
+The ~`specutils.analysis.template_comparison.template_match` function will then:
+
+    1. Move each template to the first term in the redshift grid (min_redshift)
+    2. Run steps 1 and 2 of the case with known redshift
+    3. Move to the next term in the redshift grid (min_redshift + delta_redshift)
+    4. Run steps 1 and 2 of the case with known redshift
+    5. Repeat the steps until max_redshift is reached
+    6. Return the best redshift, the lowest chi-square and its corresponding
+       template spectrum, redshifted and normalized to the observed spectrum
+       (and the index of the template spectrum if the list of templates is iterable)
 
 An example of how to do template matching with an unknown redshift is:
 
@@ -196,10 +205,11 @@ An example of how to do template matching with an unknown redshift is:
    >>> min_redshift = 1.0
    >>> max_redshift = 3.0
    >>> delta_redshift = .25
+   >>> resample_method = "flux_conserving"
 
    >>> observed_spectrum = Spectrum1D(spectral_axis=spec_axis*(1+observed_redshift), flux=np.random.randn(50) * u.Jy, uncertainty=StdDevUncertainty(np.random.sample(50), unit='Jy'))
    >>> spectral_template = Spectrum1D(spectral_axis=spec_axis, flux=np.random.randn(50) * u.Jy, uncertainty=StdDevUncertainty(np.random.sample(50), unit='Jy'))
-   >>> tm_result = template_comparison.template_match(observed_spectrum=observed_spectrum, spectral_templates=spectral_template, resample_method="flux_conserving", min_redshift=min_redshift, max_redshift=max_redshift, delta_redshift=delta_redshift) # doctest:+FLOAT_CMP
+   >>> tm_result = template_comparison.template_match(observed_spectrum=observed_spectrum, spectral_templates=spectral_template, resample_method=resample_method, min_redshift=min_redshift, max_redshift=max_redshift, delta_redshift=delta_redshift) # doctest:+FLOAT_CMP
 
 
 Reference/API
--- a/docs/analysis.rst
+++ b/docs/analysis.rst
@@ -167,25 +167,39 @@ Each of the width analysis functions are applied to this spectrum below:
 Template Comparison
 -------------------
 
+The `specutils.analysis.template_comparison` script contains the
+`specutils.analysis.template_comparison.template_redshift` method and the
+`specutils.analysis.template_comparison.template_match` method.
+`specutils.analysis.template_comparison.template_redshift` finds the most accurate redshift to shift the template
+spectrum by to match the observed spectrum. `specutils.analysis.template_comparison.template_match` takes one
+observed spectrum and n template spectra (and optionally redshift parameters in order to shift the template
+spectrum to match the observed spectrum) and does the following process:
+
 With one observed spectrum and n template spectra, the process to do template matching is:
 
     1. Move the templates as close as possible to the observed spectrum by doing redshifting, matching the resolution,
        and matching the wavelength spacing
-    2. Then loop through all of the n template spectra and run chi square on them using the observed spectrum as the
-       expected frequency and the template spectrum as the observed
+    2. Then loop through all of the n template spectra (accounting for redshift if parameters are provided) and
+       run chi square on them using the observed spectrum as the expected frequency and the template spectrum as
+       the observed
     3. Once you have a corresponding chi square for each template spectrum, return the lowest chi square and its
        corresponding template spectrum (normalized) and the index of the template spectrum if the original template
        parameter is iterable
 
-An example of how to do template matching in is:
+An example of how to do template matching with an unknown redshift is:
 
 .. code-block:: python
 
    >>> from specutils.analysis import template_comparison
    >>> spec_axis = np.linspace(0, 50, 50) * u.AA
-   >>> observed_spectrum = Spectrum1D(spectral_axis=spec_axis, flux=np.random.randn(50) * u.Jy, uncertainty=StdDevUncertainty(np.random.sample(50), unit='Jy'))
+   >>> observed_redshift = 2.0
+   >>> min_redshift = 1.0
+   >>> max_redshift = 3.0
+   >>> delta_redshift = .25
+
+   >>> observed_spectrum = Spectrum1D(spectral_axis=spec_axis*(1+observed_redshift), flux=np.random.randn(50) * u.Jy, uncertainty=StdDevUncertainty(np.random.sample(50), unit='Jy'))
    >>> spectral_template = Spectrum1D(spectral_axis=spec_axis, flux=np.random.randn(50) * u.Jy, uncertainty=StdDevUncertainty(np.random.sample(50), unit='Jy'))
-   >>> tm_result = template_comparison.template_match(observed_spectrum, spectral_template) # doctest:+FLOAT_CMP
+   >>> tm_result = template_comparison.template_match(observed_spectrum=observed_spectrum, spectral_templates=spectral_template, resample_method="flux_conserving", min_redshift=min_redshift, max_redshift=max_redshift, delta_redshift=delta_redshift) # doctest:+FLOAT_CMP
 
 
 Reference/API
--- a/docs/analysis.rst
+++ b/docs/analysis.rst
@@ -168,32 +168,32 @@ Template Comparison
 -------------------
 
 The ~`specutils.analysis.template_comparison.template_match` function takes an
-observed spectrum and n template spectra and returns the best template that
+observed spectrum and ``n`` template spectra and returns the best template that
 matches the observed spectrum via chi-square minimization.
 
-If the redshift is known, the user will redshift the template spectra
-appropriately and then run the
+If the redshift is known, the user can set that for the ``known_redshift`` parameter
+and then run the
 ~`specutils.analysis.template_comparison.template_match` function.
 This function will:
 
-    1. Match the resolution and wavelength spacing of the observed spectrum
-    2. Compute the chi-square between the observed spectrum and each template
+    1. Match the resolution and wavelength spacing of the observed spectrum.
+    2. Compute the chi-square between the observed spectrum and each template.
     3. Return the lowest chi-square and its corresponding template spectrum,
        normalized to the observed spectrum (and the index of the template
-       spectrum if the list of templates is iterable)
+       spectrum if the list of templates is iterable).
 
 If the redshift is unknown, the user specifies the expected window in redshift
-(min_redshift and max_redshift) and the redshift spacing (delta_redshift).
+(``min_redshift`` and ``max_redshift``) and the redshift spacing (``delta_redshift``).
 The ~`specutils.analysis.template_comparison.template_match` function will then:
 
-    1. Move each template to the first term in the redshift grid (min_redshift)
-    2. Run steps 1 and 2 of the case with known redshift
-    3. Move to the next term in the redshift grid (min_redshift + delta_redshift)
-    4. Run steps 1 and 2 of the case with known redshift
-    5. Repeat the steps until max_redshift is reached
+    1. Move each template to the first term in the redshift grid (``min_redshift``).
+    2. Run steps 1 and 2 of the case with known redshift.
+    3. Move to the next term in the redshift grid (``min_redshift + delta_redshift``).
+    4. Run steps 1 and 2 of the case with known redshift.
+    5. Repeat the steps until ``max_redshift`` is reached.
     6. Return the best redshift, the lowest chi-square and its corresponding
        template spectrum, redshifted and normalized to the observed spectrum
-       (and the index of the template spectrum if the list of templates is iterable)
+       (and the index of the template spectrum if the list of templates is iterable).
 
 An example of how to do template matching with an unknown redshift is:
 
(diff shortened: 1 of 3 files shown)


javerbukh:
What should happen in the code when the spectral axes of the observed and template spectra **do not** overlap? Currently a `nan` is returned. Is this correct or should something else happen?
@camipacifici @nmearl @eteq

camipacifici:
Do not overlap entirely or partially?
I think that the code should send an error/warning saying that at the "attempted" redshift, the template does not sample the observed wavelength range. I could imaging that this could happen because the "attempted" redshift is really far from reality, or the template spectrum does not cover a large enough wavelength range.

docs/analysis.rst
```diff
@@ -167,25 +167,39 @@ Each of the width analysis functions are applied to this spectrum below:
 Template Comparison
 -------------------
 
+The `specutils.analysis.template_comparison` script contains the
+`specutils.analysis.template_comparison.template_redshift` method and the
+`specutils.analysis.template_comparison.template_match` method.
+`specutils.analysis.template_comparison.template_redshift` finds the most accurate redshift to shift the template
+spectrum by to match the observed spectrum. `specutils.analysis.template_comparison.template_match` takes one
```
nmearl:
```suggestion
spectrum by to match the observed spectrum. ~`specutils.analysis.template_comparison.template_match` takes one
```

specutils/analysis/template_comparison.py
```diff
@@ -59,7 +59,7 @@ def _resample(resample_method):
     return None
 
 
-def _template_match(observed_spectrum, template_spectrum, resample_method):
+def _chi_sqaure_for_templates(observed_spectrum, template_spectrum, resample_method):
```
nmearl:
```suggestion
def _chi_square_for_templates(observed_spectrum, template_spectrum, resample_method):
```

specutils/analysis/template_comparison.py
```diff
@@ -138,7 +147,14 @@ def template_match(observed_spectrum, spectral_templates,
         The index of the spectrum with the smallest chi2 in spectral templates.
     """
     if hasattr(spectral_templates, 'flux') and len(spectral_templates.flux.shape) == 1:
-        normalized_spectral_template, chi2 = _template_match(
+
+        # Account for redshift if provided
+        if min_redshift and max_redshift and delta_redshift:
```
nmearl:
```suggestion
        if all(x is not None for x in (min_redshift, max_redshift, delta_redshift)):
```

In this case you're checking identity of the parameters, not the truthiness, so you should use the `is` operator.

specutils/analysis/template_comparison.py
```diff
@@ -138,7 +147,14 @@ def template_match(observed_spectrum, spectral_templates,
...
+            redshift, redshifted_spectrum = template_redshift(observed_spectrum, spectral_templates,
+                                                               min_redshift, max_redshift, delta_redshift)
+            spectral_templates = redshifted_spectrum
+
+        normalized_spectral_template, chi2 = _chi_sqaure_for_templates(
```
nmearl:
```suggestion
        normalized_spectral_template, chi2 = _chi_square_for_templates(
```

specutils/analysis/template_comparison.py
```diff
@@ -151,7 +167,14 @@ def template_match(observed_spectrum, spectral_templates,
...
     for index, spectrum in enumerate(spectral_templates):
-        normalized_spectral_template, chi2 = _template_match(
+
+        # Account for redshift if provided
+        if min_redshift and max_redshift and delta_redshift:
+            redshift, redshifted_spectrum = template_redshift(observed_spectrum, spectrum,
+                                                               min_redshift, max_redshift, delta_redshift)
+            spectrum = redshifted_spectrum
+
+        normalized_spectral_template, chi2 = _chi_sqaure_for_templates(
```
nmearl:
```suggestion
        normalized_spectral_template, chi2 = _chi_square_for_templates(
```

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +183,53 @@ def template_match(observed_spectrum, spectral_templates,
...
+
+    Returns
+    -------
+    final_redshift : `float`
+        The most accurate redshift for template_spectrum to match the observed_spectrum
+    redshifted_spectrum: :class:`~specutils.Spectrum1D`
+        A new Spectrum1D object which incorporates the template_spectrum with a spectral_axis
+        that has been redshifted using the final_redshift
+    """
+    if not (min_redshift and max_redshift and delta_redshift) or (min_redshift>max_redshift):
```
nmearl:
```suggestion
    if min_redshift > max_redshift:
```

There are no optional arguments in the call to this function, so it's safe to assume that `min_redshift`, `max_redshift`, and `delta_redshift` have values.

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +183,53 @@ def template_match(observed_spectrum, spectral_templates,
...
+        return
```
nmearl:
```suggestion
        raise ValueError("The `min_redshift` value must be greater than `max_redshift`.")
        return
```
If the above conditional fails, a `ValueError` should be raised before exiting the function.

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +183,53 @@ def template_match(observed_spectrum, spectral_templates,
...
+    chi2_min = None
+    final_redshift = None
+
+    # Loop which goes through available redshift options and finds the smallest chi2
+    while redshift <= max_redshift:
+
+        # Create new redshifted spectrum and run it through the chi2 method
+        redshifted_spectrum = Spectrum1D(spectral_axis=template_spectrum.spectral_axis*(1+redshift),
+                        flux=template_spectrum.flux)
+        normalized_spectral_template, chi2 = _chi_sqaure_for_templates(
```
nmearl:
```suggestion
        normalized_spectral_template, chi2 = _chi_square_for_templates(
```

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +183,53 @@ def template_match(observed_spectrum, spectral_templates,
...
+    """
+    if not (min_redshift and max_redshift and delta_redshift) or (min_redshift>max_redshift):
+        return
+
+    redshift = min_redshift
+    chi2_min = None
+    final_redshift = None
+
+    # Loop which goes through available redshift options and finds the smallest chi2
+    while redshift <= max_redshift:
```
nmearl:
```suggestion
    for redshift in np.arange(min_redshift, max_redshift, delta_redshift):
```

Personal preference here: `while` loops always make me queasy, so _if possible_ I try to use `for` loops whenever possible.

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +183,53 @@ def template_match(observed_spectrum, spectral_templates,
...
+        redshifted_spectrum = Spectrum1D(spectral_axis=template_spectrum.spectral_axis*(1+redshift),
+                        flux=template_spectrum.flux)
+        normalized_spectral_template, chi2 = _chi_sqaure_for_templates(
+                        observed_spectrum, redshifted_spectrum, "flux_conserving")
+
+        # Set new chi2_min if suitable replacement is found
+        if not np.isnan(chi2) and (chi2_min is None or chi2 < chi2_min):
+            chi2_min = chi2
+            final_redshift = redshift
+        redshift += delta_redshift
```
nmearl:
```suggestion
```

(Again, personal preference -- feel free to ignore!)

specutils/tests/test_template_comparison.py
```diff
@@ -11,6 +11,66 @@
 # TODO: Add some tests that are outliers: where the observed and template do not overlap (what happens?),
 # TODO: where there is minimal overlap (1 point)?
 
+def test_template_match_no_overlap():
+    """
+    Test template_match when both observed and template spectra have no overlap on the wavelength axis
+    """
+    print("no overlap test")
```
nmearl:
```suggestion
```

specutils/tests/test_template_comparison.py
```diff
@@ -11,6 +11,66 @@
...
+                             flux=spec1.flux * template_comparison._normalize_for_template_matching(spec, spec1))
+
+    # assert quantity_allclose(tm_result[0].flux, spec_result.flux, atol=0.01*u.Jy)
+    assert np.isnan(tm_result[1])
+
+def test_template_match_minimal_overlap():
+    """
+    Test template_match when both observed and template spectra have minimal overlap on the wavelength axis
+    """
+    print("minimal overlap test")
```
nmearl:
```suggestion
```

specutils/tests/test_template_comparison.py
```diff
@@ -11,6 +11,66 @@
...
+    # Seed np.random so that results are consistent
+    np.random.seed(42)
+
+    # Create test spectra
+    spec_axis = np.linspace(0, 50, 50) * u.AA
+    spec_axis_no_overlap = np.linspace(45, 95, 50) * u.AA
```
nmearl:
```suggestion
    spec_axis_min_overlap = np.linspace(45, 95, 50) * u.AA
```

specutils/tests/test_template_comparison.py
```diff
@@ -11,6 +11,66 @@
...
+
+    spec = Spectrum1D(spectral_axis=spec_axis,
+                      flux=np.random.randn(50) * u.Jy,
+                      uncertainty=StdDevUncertainty(np.random.sample(50), unit='Jy'))
+
+    spec1 = Spectrum1D(spectral_axis=spec_axis_no_overlap,
```
nmearl:
```suggestion
    spec1 = Spectrum1D(spectral_axis=spec_axis_min_overlap,
```

specutils/tests/test_template_comparison.py
```diff
@@ -11,6 +11,66 @@
...
+    # Get result from template_match
+    tm_result = template_comparison.template_match(spec, spec1)
+
+    # Create new spectrum for comparison
+    spec_result = Spectrum1D(spectral_axis=spec_axis,
+                             flux=spec1.flux * template_comparison._normalize_for_template_matching(spec, spec1))
+
+    # assert quantity_allclose(tm_result[0].flux, spec_result.flux, atol=0.01*u.Jy)
+    assert np.isnan(tm_result[1])
+
```
nmearl: Should there be another `assert` to ensure that the places where they do overlap have non-`NaN` values?

specutils/analysis/template_comparison.py
```diff
@@ -151,7 +167,14 @@ def template_match(observed_spectrum, spectral_templates,
     smallest_chi_spec = None
 
     for index, spectrum in enumerate(spectral_templates):
-        normalized_spectral_template, chi2 = _template_match(
+
+        # Account for redshift if provided
+        if min_redshift and max_redshift and delta_redshift:
```
nmearl: (See above.)

specutils/analysis/template_comparison.py
javerbukh: \*facepalm\*

specutils/tests/test_template_comparison.py
javerbukh: I'm checking now and it looks like with one element of overlap, there is still a full array of `NaN` values.

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +183,54 @@ def template_match(observed_spectrum, spectral_templates,
             smallest_chi_index = index
 
     return smallest_chi_spec, chi2_min, smallest_chi_index
+
+
+def template_redshift(observed_spectrum, template_spectrum, min_redshift, max_redshift, delta_redshift):
+    """
+    Find the most accurate redshift for template_spectrum to match observed_spectrum using chi2
```
eteq:
```suggestion
    Find the most best-fit redshift for template_spectrum to match observed_spectrum using chi-squared fitting.
```

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +183,54 @@ def template_match(observed_spectrum, spectral_templates,
...
+        The minimum redshift allowed
+    max_redshift : `float`
+        The maximum redshift allowed
+    delta_redshift : `float`
+        The amount the redshift will change between loops
+
+    Returns
+    -------
+    final_redshift : `float`
+        The most accurate redshift for template_spectrum to match the observed_spectrum
```
eteq:
```suggestion
        The most best-fit redshift for template_spectrum to match the observed_spectrum
```
To clairfy the reasoning behind this suggestion: "most accurate" has a subtly-different meaning here - "most accurate" implies "the redshift closest to the true redshift", which is not quite the same as "the redshift closest to the redshift of the best-fit template".

(Also, in my experience, this is the words scientists actually use for this... which is perhaps more improtant than anything else?  ¯\_(ツ)_/¯)

docs/analysis.rst
```diff
@@ -167,25 +167,49 @@ Each of the width analysis functions are applied to this spectrum below:
...
-    1. Move the templates as close as possible to the observed spectrum by doing redshifting, matching the resolution,
-       and matching the wavelength spacing
-    2. Then loop through all of the n template spectra and run chi square on them using the observed spectrum as the
-       expected frequency and the template spectrum as the observed
-    3. Once you have a corresponding chi square for each template spectrum, return the lowest chi square and its
-       corresponding template spectrum (normalized) and the index of the template spectrum if the original template
-       parameter is iterable
-
-An example of how to do template matching in is:
+The ~`specutils.analysis.template_comparison.template_match` function takes one
```
eteq:
```suggestion
The ~`specutils.analysis.template_comparison.template_match` function takes an
```
(because in the future it may not really be just one)

docs/analysis.rst
```diff
@@ -167,25 +167,49 @@ Each of the width analysis functions are applied to this spectrum below:
...
+observed spectrum and n template spectra and returns the best template that
+matches the observed spectrum via chi-square minimization.
+
+If the redshift is known, the user will redshift the template spectra
+appropriately and then run the
+~`specutils.analysis.template_comparison.template_match` function.
+This method will:
```
eteq:
```suggestion
This function will:
```

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +183,54 @@ def template_match(observed_spectrum, spectral_templates,
...
+
+    redshift = min_redshift
+    chi2_min = None
+    final_redshift = None
+
+    # Loop which goes through available redshift options and finds the smallest chi2
+    while redshift <= max_redshift:
+
+        # Create new redshifted spectrum and run it through the chi2 method
+        redshifted_spectrum = Spectrum1D(spectral_axis=template_spectrum.spectral_axis*(1+redshift),
```
eteq: This doesn't pass through the uncertainty or `meta` or anything else like that.  Perhaps instead you should use `Spectrum1D(old_spectrum, spectral_axis=...)`?  That seems to do copying-while-still-resetting-the-spectral_axis...

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +183,54 @@ def template_match(observed_spectrum, spectral_templates,
...
+    delta_redshift : `float`
+        The amount the redshift will change between loops
+
+    Returns
+    -------
+    final_redshift : `float`
+        The most accurate redshift for template_spectrum to match the observed_spectrum
+    redshifted_spectrum: :class:`~specutils.Spectrum1D`
+        A new Spectrum1D object which incorporates the template_spectrum with a spectral_axis
+        that has been redshifted using the final_redshift
```
eteq: probably this should also return the best-fit chi-square (or perhaps *all* of the chi-squareds? see my general comment on this)

camipacifici:
@eteq, you raised very valid big-picture points. I agree that these features will be very useful, but I would suggest to finish this as it was originally designed for now and treat your points as enhancements that will be implemented in the near future.
For example, I would envision returning the chi-square values for all models not just when changing the redshift, but also for the different templates in case the user wants to build probability density functions for other parameters associated with the templates.

Would you be ok with this?

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +185,55 @@ def template_match(observed_spectrum, spectral_templates,
...
+    Find the best-fit redshift for template_spectrum to match observed_spectrum using chi2
+
+    Parameters
+    ----------
+    observed_spectrum : :class:`~specutils.Spectrum1D`
+        The observed spectrum.
+    template_spectrum : :class:`~specutils.Spectrum1D`
+        The template spectrum, which will have it's redshift calculated
+    min_redshift : `float`
+        The minimum redshift allowed
```
nmearl:
```suggestion
        The minimum redshift allowed.
```

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +185,55 @@ def template_match(observed_spectrum, spectral_templates,
...
+    max_redshift : `float`
+        The maximum redshift allowed
```
nmearl:
```suggestion
        The maximum redshift allowed.
```

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +185,55 @@ def template_match(observed_spectrum, spectral_templates,
...
+    delta_redshift : `float`
+        The amount the redshift will change between loops
```
nmearl:
```suggestion
        The amount the redshift will change between loops.
```

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +185,55 @@ def template_match(observed_spectrum, spectral_templates,
...
+
+    Returns
+    -------
+    final_redshift : `float`
+        The best-fit redshift for template_spectrum to match the observed_spectrum
```
nmearl:
```suggestion
        The best-fit redshift for template_spectrum to match the observed_spectrum.
```

specutils/analysis/template_comparison.py
```diff
@@ -160,3 +185,55 @@ def template_match(observed_spectrum, spectral_templates,
...
+    redshifted_spectrum: :class:`~specutils.Spectrum1D`
+        A new Spectrum1D object which incorporates the template_spectrum with a spectral_axis
+        that has been redshifted using the final_redshift
```
nmearl:
```suggestion
        that has been redshifted using the final_redshift.
```

nmearl:
This looks good to me as is. I think another issue ought to be opened with @eteq's suggestions.

eteq:
Unfortunately it's too late if it just got merged, but I was *not* ok with merging this as-is. Several of these are *permanent* API decisions we are stuck with if we do a release with this PR.  That's why I marked it as "change requested" as opposed to "accept"ing the PR and making the suggestions as-is.

They are all relatively straightforward changes though, so perhaps the best path forward is to make implementing these a high-priority by treating them as blockers for the next specutils release?

eteq:
To be clear above - the reason to resist merging it is that we don't want to let it drop and accidentally do a release having not finished it off.  An alternate workflow that would have been find would have been merging but making sure there was an issue created before (or at the time of merging) tagged for the current/upcoming release that serves as the reminder to either make those changes or agree we aren't going to. (that makes it a blocker for release so that we don't just let it drop/forget about it)

hcferguson:
@eteq Agreed on creating the issue(s). Not sure why you see it as a blocker for the *next* release as opposed to a v1.0 (API stable) release, but that does seem like a good approach to making sure it doesn't get dropped. 

I'm a bit worried that the elaboration of the features and API is now so fragmented between issues and PRs that it's hard to follow (although maybe that's just me)? Should someone put together a consolidated description of what we are aiming at? Including a (updated?) py-in-the-sky notebook with use cases?

javerbukh:
@eteq I opened up an issue with your points: #549. I'm not sure if you mentioned those points in a another issue or PR - or in a tag up - but that is the first time I heard those suggestions and it was when the PR was essentially finished according to the original ticket and according to all discussion leading up to this point. We can discuss those points further in future sprints and implement them as soon as possible.

PR #478, opened 2019-06-28
WIP: first pass as expanding spectrum1d doc page to include spectrum_mixin

SaOgaz:
Original reason for making this PR is #475

Here's what the spectrum1D API doc looks like after adding :inherited-members:

Are there some parameters in `spectrum_mixin` that we should make private? This resulted in quite a lot of attributes listed, some of which seem like they might not be user focused?

Issue #475, opened 2019-06-26
bit of a problem with `spectrum.wavelength` in the docs


SaOgaz:
So I've accidentally run across what really boils down to a documentation problem.  I had been using `my_spec1D.wavelength` the way I should have used `my_spec1D.spectral_axis`, not realizing that the `wavelength` attribute was getting auto-converted to Angstroms.  The reason I didn't realize this is because multiple examples in the docs make calls to `wavelength` with no notes about how the units for that get auto-updated.  I imagine a lot of these doc examples should be using `spectral_axis` and just haven't been updated.

Conversely, there's no documentation of `spectral_axis` or `spectral_axis_unit` in the API docs for `Spectrum1D`: https://specutils.readthedocs.io/en/latest/api/specutils.Spectrum1D.html?highlight=wavelength.

Does this take make sense to everyone or am I still mis-understanding something about the intended usage of these attributes.

dhomeier:
Sounds related to #466, i.e. it's probably not just the documentation, but also argument checking that needs improvement.

nmearl:
Docs now use `spectral_axis` instead of `wavelength`, and more information is provided about when to use which and how they are handled.

PR #532, opened 2019-10-10
Doc Cleanup

### LINKED ISSUE #211: add specutils logo to docs header
In #200, the specutils docs header was changed to say "specutils" instead of package-template.  @bsipocz pointed out that it really ought to also have the logo.  This is a good idea (and should be straightforward given that #205 is in), but requires a bit of css modification.  

Really the package-template/theme should be updated to allow this, though. If that's easy it should be done at the same time, if it's hard it can be done later and just manually adjust the css in specutils a la what photutils already does.

robelgeda:
This PR cleans up some outdated outputs and addresses the following topics:

-  closes #211 Adding logo to the doc toolbar via css override
    - Logo is overridden using the PhotUtils method.
    - New [ticket](https://github.com/astropy/astropy-sphinx-theme/issues/16) to update astropy-sphinx-theme to allow users to override the doc logo.

*  Fixes #478 PR. This change includes the changes in #478 and that PR can be closed.
    -  Related to #475.


robelgeda:
Here is a sample of the docs (built on my branch):
https://rag-specut.readthedocs.io/en/docs/

robelgeda:
Tests pass locally. Not sure why Travis is failing.

docs/spectrum1d.rst
```diff
@@ -126,8 +126,8 @@ common spectral axis.
 
     >>> spec = Spectrum1D(spectral_axis=np.arange(5000, 5010)*u.AA, flux=np.random.sample((5, 10))*u.Jy)
     >>> spec_slice = spec[0] #doctest:+SKIP
-    >>> spec_slice.wavelength #doctest:+SKIP
-    <Quantity [0., 1., 2., 3., 4., 5., 6., 7., 8., 9.] Angstrom>
+    >>> spec_slice.spectral_axis #doctest:+SKIP
```
javerbukh: Lines 25, 26, and 100 also still make reference to `wavelength` which should probably be changed to `spectral_axis`

nmearl: Lines 25/26 can stay `wavelength` since we only want to express that `spectral_axis` should be used while _interacting_ with the `Spectrum1D` object.

PR #536, opened 2019-10-10
Splicing Docs

robelgeda:
Related to #254
This PR adds a  Splicing section to the docs.

### DIFF
--- a/docs/manipulation.rst
+++ b/docs/manipulation.rst
@@ -85,7 +85,7 @@ that takes the spectrum and an astropy 1D kernel.  So, one could also do:
 In this case, the ``spec1_bsmooth2`` result should be equivalent to the ``spec1_bsmooth`` in
 the section above (assuming the flux data of the input ``spec`` is the same).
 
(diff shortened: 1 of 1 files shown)


eteq:
@robelgeda - while reviewing this I realized I wanted to see the plot of the result of this example, so I tested it out locally (see the commit I pushed up for how I implemented adding a plot to the end of what you wrote).  In the process, though, I realized what was implemented was not actually a splice - splicing a spectrum means the two pieces of the spectrum should be non-overlapping, which was not the case in your original example.  It was a small modification to what you had to adjust the grids a bit to make it a splice though.  So I just did this in 2f9ac1b and pushed it up since my understanding is that you are not free to work on this for the next few weeks.  But I think with my modifications this is now ready assuming the tests pass.

robelgeda:
Thanks a lot @eteq!

eteq:
The failures are unrelated and due to my re-running the tests, but the important ones all passed, so merging!

Issue #535, opened 2019-10-10
Check for continuum subtracted data


brechmos-stsci:
Methods like `find_lines` needs to have data input that is already continuum subtracted otherwise the methods may not work.  
This method should then be added to `find_lines` methods and possibly others. 
Should this method throw an exception? Warning?

brechmos-stsci:
There are a couple ways to check this, some slow and some fast:

 1. fit a continuum and see if it is very near 0 (probably too slow)
 2. could do a sigma clipping algorithm and then take the median (?) of what is left and see if it is near 0 (again, might be too slow)
 3. could just do a simple operation like mean or median to see how close it is to 0

eteq:
Following a bit more discussion, here's a concrete proposal to start with:

1. Implement a function `guess_continuum_near_zero` that uses method 2 of @brechmos-stsci.
2. Implement a decorator `warn_if_continuum_near_zero` that can decorate analysis functions and generates a warning (with instructions on how to turn it off completely) that.
3. Do some profiling of the decorator on a test data set and `specutils.analysis.line_flux`, called repeatedly (perhaps ~10x times?). Use that to determine how much of a performance penalty there is vs not doing and use that to decide if method 3 or 1 might be more practical (or possibly other optimization tricks like caching the result in `meta`, although there be dragons there...)

eteq:
Oh, but one clarification re: "method 2": if there is an uncertainty on the spectrum, instead of using the median and some tolerance, use flux divided by a standard-deviation-like uncertainty, and get the median of *that*. (basically per-pixel SNR, which is further motivation for #523 I guess.)

brechmos-stsci:
And mask needs to be considered too.

brechmos-stsci:
Umm... so... if we have a call structure of `is_continuum_near_zero(spectrum, eps)`  and we assume the spectrum has no mask nor uncertainty. What should the `eps` be?  For example, if flux is in Jy, that is going to be very different than if flux is in micro-Jy.  From an astronomy point of view, how would you determine if the continuum is near 0?  if the median is less than 0.1 Jy?  (I don't think that makes sense...).  Thoughts?

(I think I like `is_continuum_near_zero` rather than `guess_continuum_near_zero`, but that is an aside)

eteq:
@brechmos-stsci ah, good point.  I think this is the options we want to consider for `eps`:
[...]

PR #538, opened 2019-10-16
Simple utility continuum baseline checker

brechmos-stsci:
This PR will create a simple spectrum checker to see if the flux baseline is near zero as this is required for several of the other methods (e.g., find_lines which assumes the flux data is continuum subtracted).

Fixes https://github.com/astropy/specutils/issues/535

### DIFF
--- a/specutils/__init__.py
+++ b/specutils/__init__.py
@@ -41,10 +41,10 @@ class Conf(_config.ConfigNamespace):
     Configuration parameters for specutils.
     """
 
-    always_check_continuum = _config.ConfigItem(
+    do_continuum_function_check = _config.ConfigItem(
         True,
         'Whether to check the spectrum baseline value is close'
-        'to zero. If it is not within ``eps`` then a warning is raised.'
+        'to zero. If it is not within ``threshold`` then a warning is raised.'
     )
 
 conf = Conf()
--- a/specutils/__init__.py
+++ b/specutils/__init__.py
@@ -7,6 +7,7 @@ Specutils: an astropy package for spectroscopy.
 # should keep this content at the top.
 # ----------------------------------------------------------------------------
 from ._astropy_init import *
+from astropy import config as _config
 # ----------------------------------------------------------------------------
 
 # Enforce Python version check during package import.
@@ -34,3 +35,17 @@ if not _ASTROPY_SETUP_:
     _load_user_io()
 
 __citation__ = 'https://doi.org/10.5281/zenodo.1421356'
+
+class Conf(_config.ConfigNamespace):
+    """
+    Configuration parameters for specutils.
+    """
+
+    always_check_continuum = _config.ConfigItem(
+        True,
+        'Whether to check the spectrum baseline value is close'
+        'to zero. If it is not within ``eps`` then a warning is raised.'
+    )
+
+conf = Conf()
+
--- a/specutils/__init__.py
+++ b/specutils/__init__.py
@@ -48,4 +48,3 @@ class Conf(_config.ConfigNamespace):
     )
 
 conf = Conf()
-
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -127,17 +127,17 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
 
 def is_continuum_below_threshold(spectrum, threshold=0.01):
     """
-    Determine if the baseline of this spectrum is less than the threshold. 
+    Determine if the baseline of this spectrum is less than the threshold.
     I.e., an estimate of whether or not the continuum has been subtracted.
 
     If the threshold is an `~astropy.units.Quantity` then compare the median
-    of the flux to the threshold.  
+    of the flux to the threshold.
 
-    If the threshold is a float then the spectrum's uncertainty will be 
+    If the threshold is a float then the spectrum's uncertainty will be
     used or an estimate of the uncertainty. If the uncertainty is present then the
     threshold is compared to the median of the flux divided by the
     uncertainty.  If the uncertainty is not present then the threshold
-    is compared to the median of the flux divided by the 
+    is compared to the median of the flux divided by the
     `~astropy.stats.median_absolute_deviation`.
 
     Parameters
@@ -147,7 +147,7 @@ def is_continuum_below_threshold(spectrum, threshold=0.01):
 
     threshold: float or `~astropy.units.Quantity`
         The tolerance on the quantification to confirm the continuum is
-        near zero. 
+        near zero.
 
     Returns
     -------
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -2,14 +2,20 @@
 A module for analysis tools focused on determining fluxes of spectral features.
 """
 
+from functools import wraps
+import warnings
+
 import numpy as np
+from .. import conf
 from ..manipulation import extract_region
 from .utils import computation_wrapper
 from astropy.stats import sigma_clip
 from astropy.stats import median_absolute_deviation
+from astropy.utils.exceptions import AstropyUserWarning
 
 
-__all__ = ['line_flux', 'equivalent_width', 'is_continuum_near_zero']
+__all__ = ['line_flux', 'equivalent_width', 'is_continuum_near_zero',
+           'warn_spectrum_continuum_subtracted']
 
 
 def line_flux(spectrum, regions=None):
@@ -124,10 +130,6 @@ def is_continuum_near_zero(spectrum, eps=0.01):
     Determine if the typical baseline of this spectrum is near zero. I.e., an
     estimate of whether or not the continuum has been subtracted.
 
-    The value is scaled to match the sigma/standard deviation parameter of a
-    standard Gaussian profile. This will be calculated over the regions, if
-    they are specified.
-
     Parameters
     ----------
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
@@ -142,34 +144,46 @@ def is_continuum_near_zero(spectrum, eps=0.01):
     is_near_zero: bool
         A True if the continuum of the spectrum is with eps, False otherwise.
 
-    Notes
-    -----
-    The spectrum should be continuum subtracted before being passed to this
-    function.
     """
+
+    flux = spectrum.flux
+    uncertainty = spectrum.uncertainty if hasattr(spectrum, 'uncertainty') else None
+
+    # Apply the mask if it exists.
+    if hasattr(spectrum, 'mask') and spectrum.mask is not None:
+        flux = flux[~spectrum.mask]
+        uncertainty = uncertainty[~spectrum.mask] if uncertainty else uncertainty
 
     # If the eps has units then the assumption is that we want
     # to compare the median of the flux regardless of the
     # existence of the uncertainty.
     if hasattr(eps, 'unit'):
-        return np.median(spectrum.flux) < eps
+        print('here with med {}'.format(spectrum.flux))
+        return np.median(flux) < eps
 
     # If eps does not have a unit, ie it is not a quantity, then
     # we are going to calculate based on the S/N if the uncertainty
     # exists.
-    if hasattr(spectrum, 'uncertainty') and spectrum.uncertainty:
-        return np.median(spectrum.flux / spectrum.uncertainty.quantity) < eps
+    if uncertainty:
+        return np.median(flux / uncertainty.quantity) < eps
     else:
-        raise Exception('Spectrum flux has units, eps does not, either include uncertainty or use units on eps')
-        #return np.median(spectrum.flux) / median_absolute_deviation(spectrum.flux) < eps
+        warnings.warn('Spectrum does not appear to be continuum subtracted.', AstropyUserWarning)
+        #return np.median(flux) / median_absolute_deviation(flux) < eps
 
-def warn_spectrum_continuum_subtracted(eps):
-    def real_decorator(function):
+def warn_spectrum_continuum_subtracted(eps=0.01, check=conf.always_check_continuum):
+    """
+    Dectorator for methods that should warn if the baseline
+    of the spectrum does not appear to be near zero.
+    """
+    def actual_decorator(function):
+        @wraps(function)
         def wrapper(*args, **kwargs):
-            spectrum = args[0]
-            if not is_continuum_near_zero(spectrum, eps):
-                raise Exception('Spectrum does not appear to be continuum subtracted.')
+            if check:
+                spectrum = args[0]
+                if not is_continuum_near_zero(spectrum, eps):
+                    warnings.warn('Spectrum does not appear to be continuum subtracted.',
+                                  AstropyUserWarning)
             result = function(*args, **kwargs)
             return result
         return wrapper
-    return real_decorator
+    return actual_decorator
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -158,7 +158,6 @@ def is_continuum_near_zero(spectrum, eps=0.01):
     # to compare the median of the flux regardless of the
     # existence of the uncertainty.
     if hasattr(eps, 'unit'):
-        print('here with med {}'.format(spectrum.flux))
         return np.median(flux) < eps
 
     # If eps does not have a unit, ie it is not a quantity, then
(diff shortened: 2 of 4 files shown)


specutils/analysis/flux.py
```diff
@@ -116,3 +118,57 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
     ew =  dx - (line_flux / continuum)
 
     return ew.to(calc_spectrum.spectral_axis.unit)
+
+def is_continuum_near_zero(spectrum, eps=0.01):
+    """
+    Determine if the baseline continuum of the spectrum is near zero.
```
eteq:
```suggestion
    Determine if the typical baseline of this spectrum is near zero. I.e., an
    estimate of whether or not the continuum has been subtracted.
```

specutils/analysis/flux.py
```diff
@@ -116,3 +118,57 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+        near zero.
+
+    Returns
+    -------
+    is_near_zero: bool
+        A True if the continuum of the spectrum is with eps, False otherwise.
+
+    Notes
+    -----
+    The spectrum should be continuum subtracted before being passed to this
```
eteq: I think this note is not necessary, since the point is to check if this is true or not.

specutils/analysis/flux.py
```diff
@@ -116,3 +118,57 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
     ew =  dx - (line_flux / continuum)
 
     return ew.to(calc_spectrum.spectral_axis.unit)
+
+def is_continuum_near_zero(spectrum, eps=0.01):
+    """
+    Determine if the baseline continuum of the spectrum is near zero.
+
+    The value is scaled to match the sigma/standard deviation parameter of a
```
eteq: Is this paragraph maybe copied from something else?  I don't really see how it applies to this function

specutils/analysis/flux.py
```diff
@@ -116,3 +118,57 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+    # we are going to calculate based on the S/N if the uncertainty
+    # exists.
+    if hasattr(spectrum, 'uncertainty') and spectrum.uncertainty:
+        return np.median(spectrum.flux / spectrum.uncertainty.quantity) < eps
+    else:
+        raise Exception('Spectrum flux has units, eps does not, either include uncertainty or use units on eps')
+        #return np.median(spectrum.flux) / median_absolute_deviation(spectrum.flux) < eps
+
+def warn_spectrum_continuum_subtracted(eps):
+    def real_decorator(function):
```
eteq: I think you want to use the `wraps`or `update_wrapper` functions here (https://docs.python.org/3.7/library/functools.html#functools.wraps) - those do things like properly  pass through the docstring in a way that makes sense.

specutils/analysis/flux.py
```diff
@@ -116,3 +118,57 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+    if hasattr(eps, 'unit'):
+        return np.median(spectrum.flux) < eps
+
+    # If eps does not have a unit, ie it is not a quantity, then
+    # we are going to calculate based on the S/N if the uncertainty
+    # exists.
+    if hasattr(spectrum, 'uncertainty') and spectrum.uncertainty:
+        return np.median(spectrum.flux / spectrum.uncertainty.quantity) < eps
+    else:
+        raise Exception('Spectrum flux has units, eps does not, either include uncertainty or use units on eps')
```
eteq: I think this `else` case should instead use the  median absolute deviation (in the "scaled to be like sigma" form - https://docs.astropy.org/en/stable/api/astropy.stats.mad_std.html ) as the "uncertainty" estimate, rather than raising an error.  Then the decorator can default to using that case with an `eps` of .01 or something like that, and we can count on this function always running on a spectrum.

(This is what makes it a heuristic for continuum subtraction - not all spectra that have their continuum subtracted really will have this property, but it's a reasonable guess for many spectra.)

brechmos-stsci: Right, but, this was my question in https://github.com/astropy/specutils/issues/535#issuecomment-542339293.

specutils/analysis/flux.py
brechmos-stsci: Yup... will remove.

specutils/analysis/flux.py
brechmos-stsci: k

specutils/analysis/flux.py
brechmos-stsci: k

brechmos-stsci:
> 1. I think the docstring needs to describe the flow of what happens given various values of eps and whether or not the uncertainty is present (but maybe you were waiting on a first pass?)

I was waiting for the first pass to be done before finishing up the documentation.

> 2. This function doesn't seem to be using the mask. Presumably the user does not want masked values to be included, so I'd say the thing to do is just s.flux[~s.mask] if a mask is present.

Right, I was waiting to make sure the rest was good, but I added it in now.

> 3. There needs to be some way to turn the decorator "off", which the warning message should explain. I suggest using the configuration item machinery (see http://docs.astropy.org/en/stable/config/#adding-new-configuration-items) with just a boolean config item like always_check_continuum that defaults to True but just causes the decorator to immediately return True if it's set to False. Then you tell the user to use the with version of the config item machinery to supress the check.

I added this in, but you or someone will have to check it to make sure it is implemented correctly .I looked at astropy to see how it was used there, and mimiced it.

> 4. Once the modification for the "no uncertainty and no unit" case is in, we should test this on some realistic spectrum. Maybe use the spectrum from the "getting started" example? The goal is to check whether or not the three cases actually yield sensible answers, and also do some timing tests relative to other analysis functions like a line flux measurement.

Sure.

specutils/analysis/flux.py
eteq: Oops, sorry - to finish out that conversation here, I don't really understand your point there.  In the case of your example linked there, that looks to have done exactly what was expected: the fluxes are close together, and that closeness indicates the "uncertainty per pixel" is probably low, so therefore the fact that the average offset from zero is larger than the scatter is *exactly* the sign of "continuum not subtracted" that we are looking for.

specutils/__init__.py
```diff
@@ -34,3 +35,16 @@ class UnsupportedPythonError(Exception):
     _load_user_io()
 
 __citation__ = 'https://doi.org/10.5281/zenodo.1421356'
+
+class Conf(_config.ConfigNamespace):
+    """
+    Configuration parameters for specutils.
+    """
+
+    always_check_continuum = _config.ConfigItem(
```
eteq: Instead of "always", perhaps `do_continuum_function_check` or something like that?  Only because "always" could be taken as "for all functions" and it's really just "selected functions".

specutils/analysis/flux.py
```diff
@@ -116,3 +124,65 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+    """
+    Determine if the typical baseline of this spectrum is near zero. I.e., an
+    estimate of whether or not the continuum has been subtracted.
+
+    Parameters
+    ----------
+    spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
+        The spectrum object over which the width will be calculated.
+
+    eps: float
```
eteq: `eps` -> `threshold`.  And also should say that it can be a flux-unit quantity or a float/dimensionless qunatity, and that determines if it's a flux threshold vs a "signal/noise" threshold

specutils/analysis/flux.py
```diff
@@ -116,3 +124,65 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+    if hasattr(eps, 'unit'):
+        return np.median(flux) < eps
+
+    # If eps does not have a unit, ie it is not a quantity, then
+    # we are going to calculate based on the S/N if the uncertainty
+    # exists.
+    if uncertainty:
+        return np.median(flux / uncertainty.quantity) < eps
+    else:
+        warnings.warn('Spectrum does not appear to be continuum subtracted.', AstropyUserWarning)
```
eteq: I'm confused why this warning is here?

specutils/analysis/flux.py
```diff
@@ -116,3 +124,65 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+    Dectorator for methods that should warn if the baseline
+    of the spectrum does not appear to be near zero.
+    """
+    def actual_decorator(function):
+        @wraps(function)
+        def wrapper(*args, **kwargs):
+            if check:
+                spectrum = args[0]
+                if not is_continuum_near_zero(spectrum, eps):
+                    warnings.warn('Spectrum does not appear to be continuum subtracted.',
```
eteq: This should also have some text that refers back to `always_check_continuum` and tells the user exactly what they need to do to turn it off (or links to a page in the docs that explains it).

eteq:
Oops, I think my review there was against an old version, but I didn't realize there was a new commit until I had already done the review.  So ignore anything that was fixed in 72446dc that's in my comments.

specutils/__init__.py
brechmos-stsci: changed

specutils/analysis/flux.py
brechmos-stsci: already done

specutils/analysis/flux.py
brechmos-stsci: I think you saw an old PR.

specutils/analysis/flux.py
brechmos-stsci: I added a short message with a link back to the astropy configuration section.

specutils/analysis/flux.py
```diff
@@ -116,3 +124,78 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+    of the spectrum does not appear to be below a threshold.
+
+    The ``check`` parameter is based on the
+    `astropy configuration system <http://docs.astropy.org/en/stable/config/#adding-new-configuration-items>`_.
+    Examples are on that page to show how to turn off this type of warning checking.
+    """
+    def actual_decorator(function):
+        @wraps(function)
+        def wrapper(*args, **kwargs):
+            if check:
```
eteq: Hmm. I might be confused but I think this isn't quite right - I think you want to do 
```suggestion
            if check():
```
because a configuration item can be changed at runtime.

eteq: Or alternatively,
```suggestion
            if check() if callable(check) else check:
```
which would let you pass in True/False in addition to a configuration item

specutils/fitting/fitmodels.py
```diff
@@ -97,6 +97,7 @@ def _consecutive(data, stepsize=1):
     return np.split(data, np.where(np.diff(data) != stepsize)[0]+1)
 
 
+@warn_continuum_below_threshold(threshold=0.01, check=True)
```
eteq: I think you don't want `check=True` here (or below), right?  Because then the `do_continuum_function_check` option gets ignored?

brechmos-stsci: hah, good point

specutils/analysis/flux.py
brechmos-stsci: Based on our conversation, this changed to an alternative form (that is correct :) ).

specutils/analysis/flux.py
```diff
@@ -116,3 +124,83 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
     ew =  dx - (line_flux / continuum)
 
     return ew.to(calc_spectrum.spectral_axis.unit)
+
+def is_continuum_below_threshold(spectrum, threshold=0.01):
+    """
+    Determine if the baseline of this spectrum is less than the threshold.
```
eteq:
```suggestion
    Determine if the baseline of this spectrum is less than a threshold.
```

specutils/analysis/flux.py
```diff
@@ -116,3 +124,83 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+    I.e., an estimate of whether or not the continuum has been subtracted.
+
+    If the threshold is an `~astropy.units.Quantity` then compare the median
```
eteq:
```suggestion
    If ``threshold`` is an `~astropy.units.Quantity` with flux units, this 
    directly compares the median of the spectrum to the threshold.
```

specutils/analysis/flux.py
```diff
@@ -116,3 +124,83 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+    of the flux to the threshold.
```
eteq: (goes with the above suggestion...)
```suggestion
```

specutils/analysis/flux.py
```diff
@@ -116,3 +124,83 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+
+    If the threshold is a float then the spectrum's uncertainty will be
```
eteq:
```suggestion
    If the threshold is a float or dimensionless quantity then the spectrum's uncertainty will be
```

specutils/analysis/flux.py
```diff
@@ -116,3 +124,83 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+    used or an estimate of the uncertainty. If the uncertainty is present then the
+    threshold is compared to the median of the flux divided by the
+    uncertainty.  If the uncertainty is not present then the threshold
+    is compared to the median of the flux divided by the
+    `~astropy.stats.median_absolute_deviation`.
```
eteq: If you take my suggestion below this should be
```suggestion
    `~astropy.stats.mad_std`.
```

specutils/analysis/flux.py
```diff
@@ -116,3 +124,83 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+    if hasattr(threshold, 'unit'):
+        return np.median(flux) < threshold
+
+    # If threshold does not have a unit, ie it is not a quantity, then
+    # we are going to calculate based on the S/N if the uncertainty
+    # exists.
+    if uncertainty:
+        return np.median(flux / uncertainty.quantity) < threshold
+    else:
+        return np.median(flux) / median_absolute_deviation(flux) < threshold
```
eteq: This should be `astropy.stats.mad_std` instead of `astropy.stats. median_absolute_deviation`.  The difference is just a scaling factor, but that's important for making it interpretable as "signal divided by a standard-deviation-like noise"

specutils/analysis/flux.py
```diff
@@ -116,3 +124,83 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+    # to compare the median of the flux regardless of the
+    # existence of the uncertainty.
+    if hasattr(threshold, 'unit'):
+        return np.median(flux) < threshold
+
+    # If threshold does not have a unit, ie it is not a quantity, then
+    # we are going to calculate based on the S/N if the uncertainty
+    # exists.
+    if uncertainty:
+        return np.median(flux / uncertainty.quantity) < threshold
```
eteq: I'm just realizing needs a check to ensure the uncertainty is `'std'`.  E.g.`` uncertainty.uncertainty_type != 'std'``, otherwise it's not predictable what kind of ucnertainty is being used.

In principle we should be able to convert everything to std-like, but since it's not obvious how to do that, perhaps it should check if it's a 'std'-style uncertainty, and if not it produces a warning that it doesn't know what to do with the uncertainty unless it's 'std', and is defaulting to the case like when there's no uncertainty?

specutils/analysis/flux.py
```diff
@@ -116,3 +124,83 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
+
+    # Apply the mask if it exists.
+    if hasattr(spectrum, 'mask') and spectrum.mask is not None:
+        flux = flux[~spectrum.mask]
+        uncertainty = uncertainty[~spectrum.mask] if uncertainty else uncertainty
+
+    # If the threshold has units then the assumption is that we want
+    # to compare the median of the flux regardless of the
+    # existence of the uncertainty.
+    if hasattr(threshold, 'unit'):
```
eteq: I think we want to treat *dimensionless* quantities as though they are floats.  So something like:
```suggestion
    if hasattr(threshold, 'unit') and not threshold.unit == u.dimensionless_unscaled:
```
?

Issue #531, opened 2019-10-07
Spectrum1D.read fails on reading its own files written as 'tabular-fits' or 'wcs-fits'


dhomeier:
Reading a `BINTABLE` spectrum written with
`spectrum.write('spectab.fits', format='tabular-fits')`
(after fixing the `TUNITn` issue, see #528) fails with a WCS error:
```
>>> spect = Spectrum1D.read('spectab.fits', format='tabular-fits')
WARNING: FITSFixedWarning: The WCS transformation has more axes (2) than the image it is associated with (0) [astropy.wcs.wcs]
WARNING:astropy:FITSFixedWarning: The WCS transformation has more axes (2) than the image it is associated with (0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/sw/lib/python3.7/site-packages/astropy/nddata/mixins/ndio.py", line 54, in __call__
    return registry.read(self._cls, *args, **kwargs)
  File "/sw/lib/python3.7/site-packages/astropy/io/registry.py", line 523, in read
    data = reader(*args, **kwargs)
  File "/Users/derek/lib/python3.7/site-packages/specutils/io/default_loaders/tabular_fits.py", line 69, in tabular_fits_loader
    return generic_spectrum_from_table(tab, wcs=wcs, **kwargs)
  File "/Users/derek/lib/python3.7/site-packages/specutils/io/parsing_utils.py", line 208, in generic_spectrum_from_table
    meta=table.meta, wcs=wcs)
  File "/Users/derek/lib/python3.7/site-packages/specutils/spectra/spectrum1d.py", line 94, in __init__
    wcs = WCSWrapper(wcs)
  File "/Users/derek/lib/python3.7/site-packages/specutils/wcs/wcs_wrapper.py", line 41, in __new__
    return adapter_class(wcs, *args, **kwargs)
  File "/Users/derek/lib/python3.7/site-packages/specutils/wcs/adapters/fitswcs_adapter.py", line 47, in __init__
    self.axes = self.axes._replace(spectral=self.wcs.sub([self.spec_axis + 1]))
  File "/Users/derek/lib/python3.7/site-packages/specutils/wcs/adapters/fitswcs_adapter.py", line 92, in spec_axis
    "Cannot find a spectral axis in the provided WCS."
astropy.wcs._wcs.InvalidSubimageSpecificationError: Cannot find a spectral axis in the provided WCS.Are your 'ctype's correct?
```

[...]

PR #539, opened 2019-10-16
Extend spectrum information written by `tabular_fits_writer`

dhomeier:
This is an attempt to address some of the issues reported in #531, notably adding the missing unit information to the written spectrum. It also adds the `uncertainty` column, if present, and indicates the type of `spectral_axis` in `TTYPE1`. The last commit adds writing of any other header metadata, taking the 2nd option discussed in
https://github.com/astropy/specutils/issues/531#issuecomment-540605959 (with a new kwarg `update_header`). It also passes other kwargs though to the table writer, allowing e.g. to force overwriting of existing files.
Added tests for all these functions in `test_loaders.py`

Files thus written still fail to be read in as `tabular-fits` on the `WCSWrapper` error, but since they are fully compliant (except for a missing checksum) FITS files with complete spectral information, this really seems to be more of a problem on the reader side.

### DIFF
--- a/specutils/io/default_loaders/tabular_fits.py
+++ b/specutils/io/default_loaders/tabular_fits.py
@@ -72,10 +72,24 @@ def tabular_fits_loader(file_name, column_mapping=None, **kwargs):
 
 
 @custom_writer("tabular-fits")
-def tabular_fits_writer(spectrum, file_name, **kwargs):
+def tabular_fits_writer(spectrum, file_name, update_header=False, **kwargs):
     flux = spectrum.flux
     disp = spectrum.spectral_axis
-    meta = spectrum.meta
+    header = spectrum.meta.get('header', fits.header.Header()).copy()
+
+    if update_header:
+        hdr_types = (str, int, float, complex, bool,
+                     np.floating, np.integer, np.complexfloating, np.bool_)
+        header.update([k for k in spectrum.meta.items() if
+                       isinstance(k[1], hdr_types)])
+
+    # Strip header of FITS reserved keywords (strangely, loop does not work!)
+    if 'NAXIS' in header:
+        header.remove('NAXIS')
+    if 'NAXIS1' in header:
+        header.remove('NAXIS1')
+    if 'NAXIS2' in header:
+        header.remove('NAXIS2')
 
     # Mapping of spectral_axis types to header TTYPE1
     dispname = disp.unit.physical_type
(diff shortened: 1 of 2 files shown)


specutils/io/default_loaders/tabular_fits.py
```diff
@@ -72,11 +72,37 @@ def tabular_fits_loader(file_name, column_mapping=None, **kwargs):
...
+    disp = spectrum.spectral_axis
+    header = spectrum.meta.get('header', fits.header.Header()).copy()
+
+    if update_header:
+        hdr_types = (str, int, float, complex, bool,
+                     np.floating, np.integer, np.complexfloating, np.bool_)
+        header.update([k for k in spectrum.meta.items() if
+                       isinstance(k[1], hdr_types)])
+
+    # Strip header of FITS reserved keywords (strangely, loop does not work!)
```
nmearl: Alternatively, you could do:

```python
for keyword in ['NAXIS', 'NAXIS1', 'NAXIS2']:
    header.remove(keyword, ignore_missing=True)
```

specutils/io/default_loaders/tabular_fits.py
```diff
@@ -72,11 +72,37 @@ def tabular_fits_loader(file_name, column_mapping=None, **kwargs):
...
+        header.remove('NAXIS2')
+
+    # Mapping of spectral_axis types to header TTYPE1
+    dispname = disp.unit.physical_type
+    if dispname == "length":
+        dispname = "wavelength"
+
+    columns = [disp, flux]
+    colnames = [dispname, "flux"]
+    # Include uncertainty - ToDo: uncertainty units?
```
nmearl: I think it's alright to not deal with uncertainty units because they should be assumed the same as the flux units.

specutils/io/default_loaders/tabular_fits.py
```diff
@@ -72,11 +72,37 @@ def tabular_fits_loader(file_name, column_mapping=None, **kwargs):
...
+    if spectrum.uncertainty is not None:
+        columns.append(spectrum.uncertainty.array * spectrum.uncertainty.unit)
```
nmearl:
```suggestion
        columns.append(spectrum.uncertainty.quantity)
```

dhomeier:
Thanks for the suggestions, I think they make for notably more concise code!
`CHANGES.rst` seems to be one release behind; no v0.7 section yet.

PR #545, opened 2019-10-23
ENH: make default_loaders use `is_fits` to identify FITS and support compressed FITS 

dhomeier:
Currently all `default_loaders` try to identify FITS files based on filenames and extensions only. This is arguably less robust than using information from the files themselves, e.g. will not recognise downloads assigned a random tempfile name. It also breaks the potential to use `astropy.io.fits`'s transparent decompression of gzip and bzip2 files.
This PR proposes to use instead the `is_fits` function provided by Astropy, which is first looking for the FITS signature in the file before falling back to all the possible filename suffix variants. It also adds checks to `tabular_fits` to reject alternative formats (notably HST and SDSS ones) to allow auto-detection of the latter.
The tests try to automatically detect the format where possible; a test for compressed .fits is added.

Points for discussion:

- The two SDSS formats are still identified based on the filename; one could probably extract sufficient information from the header instead.

- Waiting whether to add `is_fits` use to `custom_loading.rst` or replace the current code example.

- Is there a better way to call `gzip` and `bzip2` in the tests (perhaps one that might even work on Windows)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,7 +102,7 @@ matrix:
 
     # check on somewhat older versions LTS version
     - os: linux
-      env: PYTHON_VERSION=3.6 ASTROPY_VERSION=lts NUMPY_VERSION=1.16
+      env: PYTHON_VERSION=3.6 ASTROPY_VERSION=lts NUMPY_VERSION=1.17
 
     # Add a job that runs from cron only and tests against astropy dev and
     # numpy dev to give a change for early discovery of issues and feedback
--- a/specutils/io/default_loaders/apogee.py
+++ b/specutils/io/default_loaders/apogee.py
@@ -28,8 +28,8 @@ def apVisit_identify(origin, *args, **kwargs):
     Registry.
     """
     return (isinstance(args[0], str) and
-            args[0].lower().split('.')[-1] == 'fits' and
-            args[0].startswith('apVisit'))
+            args[0].startswith('apVisit') and
+            fits.connect.is_fits(origin, *args))
 
 
 def apStar_identify(origin, *args, **kwargs):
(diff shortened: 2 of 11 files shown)


nmearl:
@dhomeier can you resolve the conflicts?

dhomeier:
> @dhomeier can you resolve the conflicts?

If v0.7 is scheduled to still support Python 3.5, I will remove the f-strings, that should do it.

nmearl:
That's right: v0.7 will still support python 3.5. f-strings can be used in the next release that makes specutils compatible with astropy 4.x.

dhomeier:
@nmearl getting the same failure in the 3.6 test as #570 now - seems that travis has an almost 2 yr old pytest installation there. https://github.com/pytest-dev/pytest/commit/82cdc48

nmearl:
@dhomeier the errors in #570 (aside from the pep8 ones that have since been fixed) are actually due to an outdated astropy version in the ci helpers. The ci helpers hasn't been updated for the 4.0 LTS release, so it's testing against an incompatible version currently. We won't worry about that one for now, as I've run this locally and it works fine.

dhomeier:
OK, I was already puzzled how one would set a minimum pytest version requirement for travis...

I noticed that when borrowing the `with fits.open(args[0]) as hdulist:` construct from `hst_stis.py` for the new *spec_identify functions, I missed the final `return False`. But I double-checked that they properly reject non-FITS files nonetheless (AFAICT `with...`  anyway seems to return `False` following an exception in its `EXPRESSION`, but I have not fully penetrated its control flow...)

nmearl:
@dhomeier can you rebase?

dhomeier:
Done.
@nmearl wondering about the failures on 3.6, I found that I could reproduce the `Quantity` errors in `test_resample` locally with the combination of `ASTROPY_VERSION=lts` (== 4.0) and  `NUMPY_VERSION=1.16` (1.16.3). It seems Astropy 4.0 builds (actually Python 3.6 should not yet require `lts` anyway, right?) are better combined with at least numpy >= 1.17.

PR #556, opened 2019-12-03
update doc sidebar to be more useful

eteq:
This PR makes a simple change to the specutils docs - it makes the left sidebar more useful (IMHO).  Currently the starting page sidebar looks like this:
image
while the PR changes it to:
image

(the "interior" pages are unchanged, although that's now explicit rather than implicit)


eteq:
(pretty trivial and @keflavich approved so I went ahead and merged.  But if someone very much *dis-likes* it, it's easy to revert too)

Issue #553, opened 2019-11-20
Fix specutils for Astropy 4.0


eteq:
Currently Specutils does not work on the master branch of Astropy (soon to be the 4.0 release).  We need to fix that and should probably get out a release ASAP once it's working so that Astropy 4.0 and specutils can be used together.  Known problems include:

- [x] wcs adapter related issues (probably requires #462 ?)
- [x] a new version of gwcs needs to be released that uses the property version of `low_level_wcs` (@nden)
- [x] modeling changes (#490 +)

That may not be a complete list, but based on current failures (e.g. https://travis-ci.org/astropy/specutils/jobs/611582198) that seems to hit the big ones.

cc @nmearl, @dhomeier (who has been reporting some of this in Slack and other issues)

Note that it might be a good idea to combine this with #543 since hopefully that test will be working again once we address this.

pllim:
This is a blocker for astropy/astropy-workshop#88 and astropy/astropy-workshop#89

javerbukh:
I’m looking through the errors in the link you posted @eteq  and I’m not seeing what `wcs_adapter` problems there are. I see the `GWCSAdapter` mentioned frequently but nothing about the `wcs_adapter`. Can you point me in the right direction of what I should be looking for?

If it is not an error in the CI and we just need to remove the `wcs_adapter` then should we just rebase PR #462  and continue from there or start from scratch? @nmearl

nmearl:
@javerbukh There are quite a few errors related to the adapter:

`E           AttributeError: 'GWCSAdapter' object has no attribute 'low_level_wcs'

`E       ValueError: Slicing the WCS object with the slice 'slice(19, 29, None)' failed, if you want to slice the NDData object without the WCS, you can remove by setting `NDData.wcs = None` and then retry.`

`E           AttributeError: 'FITSWCSAdapter' object has no attribute 'low_level_wcs'`

It's probably possible to fix the adapters, but it's probably much less work to take them out and use the unified API instead (which is the end goal anyway). PR #462 is old, but it may be a good starting point, if it helps you.

nmearl:
Issue #462 addresses all of the above and passes all tests locally, but still requires an open PR in gwcs.

eteq:
#462 is now in so this is all set

PR #557, opened 2019-12-04
[WIP] Remove wcs adapters

javerbukh:
Working on removing the wcs adapters and using the unified API instead.

Resolves part of #553


eteq:
this now looks to be overtaken by the merge of #462.  Closing with that in mind but this can be reopened if there's some bits that weren't integrated

PR #558, opened 2019-12-05
With new modeling

robelgeda:


nden:
Is the intention to stop testing with anything else but dev versions? If so this should be based on master and dev versions of gwcs and asdf added. Note that there is already a Travis job configured to use dev versions. This is what needs to pass before my PR is merged. 
If the intention is to test only my PR, this already done in the dev Travis job (allowed to fail).

robelgeda:
This PR was a mistake, it was supposed to be against my fork not master.

PR #564, opened 2020-01-02
mexican hat to ricker

eteq:
This PR addresses a problem @nmearl pointed out in https://github.com/astropy/specutils/pull/519#issuecomment-569531856 - that the renaming of `MexicanHat1D` to `Ricker1D` breaks one of the docs links.  I'm treating this as a stand-alone PR because we will want to not have this in a potential last-compatible-with-Astropy 3.x release, since the change is in Astropy 4.0.

### DIFF
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -166,7 +166,7 @@ parameter estimators, or if one wants to use different parameter estimators then
 one can create a dictionary where the key is the parameter name and the value is
 a function that operates on a spectrum (lambda functions are very useful for
 this purpose). For example if one wants to estimate the line parameters of a
-line fit for a `~astropy.modeling.functional_models.MexicanHat1D` one can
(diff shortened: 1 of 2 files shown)


eteq:
Note this should go in *after* #462 since it's also for the Astropy 4.0 version

eteq:
This required a rebase over #462.  In the process I realized #462 had substantially changed how the section using this as an example actually works, so needed to update the surrounding text.

@nmearl you might want to take a look, although given it's just docs I might merge it to finalize a release.  We can always update this after-the-fact though given that it's mainly a docs thing

eteq:
The two failures are redding herrings - one is #565 and the other is some sort of internet issue for one specific test (it looks like it might be that one of the remote data sources is blocking travis?  but not related to this directly I think).

Issue #394, opened 2018-11-14
Document the default loaders


eteq:
#376 added several data loaders.  But none of the loaders are actually in the docs.  We need to fix that by adding them to the sphinx docs before a release.

At the same time, the docs on this need some discussion of what the "generic" readers mean.  I.e., that they are intentionally limited because you should write a custom loader instead of using the "generic" readers if you can't do a simple column-to-attribute mapping.

eteq:
Moving this to 0.6, which makes these loaders "private functionality" for v0.5.

eteq:
I'm going to add a bit more to what this should do (and ping this...) due to #573 - it's clear from that we need to specify when the tabular fits loader should be used vs the generic cube loader.

eteq:
Oh and to given an example of the sort of thing *I* at least had in mind here - see https://docs.astropy.org/en/stable/io/unified.html#built-in-table-readers-writers .  

That particular one is a *manually* generated table, which is not ideal because it's hard to maintain when new loaders are added.  But there are functions built into the I/O machinery that can print out what's present without any manual work required.  For example, if you do ``specutils.Spectrum1D.read.help()`` it prints an ready-to-go rst-formatted string that has a table with all the `Spectrum1D` readers and writers.  That function has an argument ``out`` that makes me think it was intended to be used to generate documentation.  I don't have an example off-hand of how that might be used, but I'm guessing searching around a bit in `astropy` will reveal an example?

While calling functions like that directly from the doc build is non-trivial, the easiest work around I know of is to do it at import-time in something that gets documented in sphinx.  To see a (somewhat complicated) example of this, you can compare https://github.com/astropy/astropy/blob/master/astropy/coordinates/builtin_frames/__init__.py to https://docs.astropy.org/en/latest/coordinates/#built-in-frame-classes - the fancy graph there is generated as part of the import of `builtin_frames`, and then sphinx just shows the docstring and doesn't have to run any special code.  We could do something similar here.

nmearl:
This is solved with #672.

PR #573, opened 2020-01-23
Fix default_loaders tabular-fits format and automatic recognition

### LINKED ISSUE #531: Spectrum1D.read fails on reading its own files written as 'tabular-fits' or 'wcs-fits'

dhomeier:
Trying to clean up with a couple of problems I encountered in `default_loaders`:

1. `spectrum_from_column_mapping` was failing due to use of wrong `equivalence` keyword, also did not accept unit names the way indicated in the docstring (promoted to `tabular_fits_loader`).

2. `tabular_fits_loader` without `column_mapping` keyword was still failing on any file I could find, because it unconditionally used even an (essentially empty) `WCS` from headers without any wcs info, instead of using dispersion information from the table.

3. I had botched the correct `is_fits` calling method in some loaders.

4. `muscles-sed` seemed not to properly handle current `Unit` and `Quantity` usage; no tests either.

5. Added header-based identification for the `apogee` formats + tests.

In `tabular_fits.py` I am making only a very basic check if the `wcs` can represent a 1D `spectral_axis` to fix #531; maybe a closer comparison with the shape of the flux data is in place.


specutils/io/default_loaders/tabular_fits.py
```diff
@@ -69,6 +68,10 @@ def tabular_fits_loader(file_name, column_mapping=None, hdu=1, **kwargs):
 
     tab = Table.read(file_name, format='fits', hdu=hdu)
 
+    # Minimal consistency check for wcs consistency with table data
+    if wcs.naxis != 1:
```
nmearl: It's not always guaranteed that the number of world coordinate axes in the fits file will be one (e.g. a data cube, or data with a time dimension). Perhaps it'd be better to check explicitly for a spectral axis?

specutils/io/parsing_utils.py
```diff
@@ -17,16 +17,17 @@ def spectrum_from_column_mapping(table, column_mapping, wcs=None):
     Parameters
     ----------
     table : :class:`~astropy.table.Table`
-        The table object returned from parsing the data file.
+        The table object (e.g. returned from `Table.read('data_file')`).
```
nmearl:
```suggestion
        The table object (e.g. returned from ``Table.read('data_file')``).
```

specutils/io/parsing_utils.py
```diff
@@ -17,16 +17,17 @@ def spectrum_from_column_mapping(table, column_mapping, wcs=None):
...
     column_mapping : dict
-        A dictionary describing the relation between the file columns
+        A dictionary describing the relation between the table columns
         and the arguments of the `Spectrum1D` class, along with unit
-        information. The dictionary keys should be the file column names
+        information. The dictionary keys should be the table column names
         while the values should be a two-tuple where the first element is the
         associated `Spectrum1D` keyword argument, and the second element is the
-        unit for the file column::
+        unit for the file column (or `None` to take unit from the table)::
```
nmearl:
```suggestion
        unit for the file column (or ``None`` to take unit from the table)::
```

specutils/tests/test_loaders.py
```diff
@@ -368,3 +376,56 @@ def test_tabular_fits_header(tmpdir):
     assert 'OBSDATE' not in hdulist[0].header
     assert 'OBSDATE' not in hdulist[1].header
     hdulist.close()
+
+
+@pytest.mark.remote_data
+def test_apstar_loader():
+    '''Test remote read and automatic recognition of apStar spec from URL.
```
nmearl:
```suggestion
    """Test remote read and automatic recognition of apStar spec from URL.
```

specutils/tests/test_loaders.py
```diff
@@ -368,3 +376,56 @@ def test_tabular_fits_header(tmpdir):
...
+    '''
```
nmearl:
```suggestion
    """
```

specutils/tests/test_loaders.py
```diff
@@ -368,3 +376,56 @@ def test_tabular_fits_header(tmpdir):
...
+    spec = Spectrum1D.read(apstar_url)
+
+    assert isinstance(spec, Spectrum1D)
+    assert spec.flux.size > 0
+    assert spec.uncertainty.array.min() >= 0.0
+
+
+@pytest.mark.remote_data
+def test_apvisit_loader():
+    '''Test remote read and automatic recognition of apvisit spec from URL.
```
nmearl:
```suggestion
    """Test remote read and automatic recognition of apvisit spec from URL.
```

specutils/tests/test_loaders.py
```diff
@@ -368,3 +376,56 @@ def test_tabular_fits_header(tmpdir):
...
+    '''
```
nmearl:
```suggestion
    """
```

specutils/tests/test_loaders.py
```diff
@@ -368,3 +376,56 @@ def test_tabular_fits_header(tmpdir):
...
+    spec = Spectrum1D.read(apvisit_url)
+
+    assert isinstance(spec, Spectrum1D)
+    assert spec.flux.size > 0
+    assert spec.uncertainty.array.min() >= 0.0
+
+
+@pytest.mark.remote_data
+def test_aspcapstar_loader():
+    '''Test remote read and automatic recognition of aspcapStar spec from URL.
```
nmearl:
```suggestion
   """Test remote read and automatic recognition of aspcapStar spec from URL.
```

specutils/tests/test_loaders.py
```diff
@@ -368,3 +376,56 @@ def test_tabular_fits_header(tmpdir):
...
+    '''
```
nmearl:
```suggestion
    """
```

specutils/tests/test_loaders.py
```diff
@@ -368,3 +376,56 @@ def test_tabular_fits_header(tmpdir):
...
+    spec = Spectrum1D.read(aspcap_url)
+
+    assert isinstance(spec, Spectrum1D)
+    assert spec.flux.size > 0
+    assert spec.uncertainty.array.min() >= 0.0
+
+
+@pytest.mark.remote_data
+def test_muscles_loader():
+    '''Test remote read and automatic recognition of muscles spec from URL.
```
nmearl:
```suggestion
    """Test remote read and automatic recognition of muscles spec from URL.
```

specutils/tests/test_loaders.py
```diff
@@ -368,3 +376,56 @@ def test_tabular_fits_header(tmpdir):
...
+    '''
```
nmearl:
```suggestion
    """
```

specutils/io/default_loaders/generic_cube.py
```diff
@@ -22,7 +22,7 @@
 # need to add `format="my-format"` in the `Spectrum1D.read` call.
 def identify_generic_fits(origin, *args, **kwargs):
     return (isinstance(args[0], str) and
-            fits.connect.is_fits(origin, *args) and
+            fits.connect.is_fits(origin, origin, *args) and
```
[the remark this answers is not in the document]
dhomeier: You're right, it actually expects a file object (and unfortunately does not work with `urlopen` due to the `seek` requirements) - thought that the `identify_*` functions were already called with the open file, but of course it is explicitly testing for a name string above...
These are all loaders for formats without tests or sample files for the format; now that I've been looking into the other loaders it always seems to be preferable to use the `with fits.open()` context, which gives you the signature check for free plus the hdulist for further investigation of the content. Will try to fix the remaining ones as a use case nonetheless.

specutils/io/default_loaders/muscles_sed.py
```diff
@@ -50,14 +47,12 @@ def muscles_sed(file_name, **kwargs):
...
         header = hdulist[0].header
 
-        tab = Table.read(hdulist)
+        tab = Table.read(hdulist[1])
 
         meta = {'header': header}
         uncertainty = StdDevUncertainty(tab["ERROR"])
-        data = tab["FLUX"]
-        wavelength = tab["WAVELENGTH"]
+        data = Quantity(tab["FLUX"])
```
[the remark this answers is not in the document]
dhomeier: The unit is already parsed with the `Table.read()`. I've added a check for this to the test.

specutils/io/default_loaders/tabular_fits.py
dhomeier: I was not sure to what extent `tabular-fits` does, or is supposed to, support >1D spectra (thought that `generic_cube` was for that case, but again I don't have any sample files to test). But if any,  I would rather expect it to support the simplest case of a spectral cube or time series with a common (single 1D) spectral axis. Other cases, where each flux column could have its own zero point or completely individual dispersion scale, or possible intermediate cases where some subset of the spectra shares a spectra axis, are IMO really better left to `generic_cube`; at least I have no idea how to correctly identify such cases where

1. the flux data have >1 dimension

2. the spectra in the flux cube do not have identical spectral axes and

3. these wavelength/frequency scales are not included in the `BINTABLE`, but must be constructed from a WCS defined in the header.

Perhaps @teuben who wrote the initial version of `generic_cube` has a better idea about what possible data formats could be encountered.

So I would tend to just check for
```
    if not (wcs.naxis == 1 and wcs.array_shape[0] == len(tab):
```
which could be extended to
```
    if not (wcs.naxis == 1 and wcs.array_shape[0] == len(tab) or
            wcs.array_shape == tab[fluxname].shape):
```
just to allow for the case with all individual spectral axes, but note that then the correct column for the flux would have to be identified from either `column_mapping` or analogously to `_find_spectral_column`.

specutils/tests/test_loaders.py
dhomeier: Change all existing docstrings to double quotes as well, I assume?

specutils/io/default_loaders/muscles_sed.py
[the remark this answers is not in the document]
nmearl: Interesting, I always assumed one would have to use `QTable.read` to get columns as quantity objects. Good to know.

specutils/tests/test_loaders.py
nmearl: Yep! 😁

specutils/io/default_loaders/muscles_sed.py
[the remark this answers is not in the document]
dhomeier: In fact the regular Table `Column` supports basic [quantity features](https://docs.astropy.org/en/stable/table/access_table.html#columns-with-units), but is still subtly different Therefore the cast to `Quantity` is still needed - alternatively one could indeed get it directly with `QTable.read`.

specutils/io/default_loaders/generic_cube.py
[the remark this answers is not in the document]
dhomeier: Switched the other loaders, which access the header for further investigation, to `with fits.open`; that leaves `subaru_pfs_spec` as only case directly calling `is_fits`, and that is tested now.

specutils/io/default_loaders/subaru_pfs_spec.py
```diff
@@ -28,33 +28,52 @@
 
 def spec_identify(origin, *args, **kwargs):
```
nmearl: This may be worth adding to the `/io/default_loaders/parsing_utils.py` file. If `_spec_pattern` is unique between loader definitions, maybe adding in another parameter so that it can passed to the function.

specutils/io/default_loaders/tabular_fits.py
nmearl: Hmm. This is tricky. Ideally, the wcs should _tell_ us the column axis type (e.g. `wcs.world_axis_names`, `wcs.world_axis_types`, etc). But obviously this requires that the fits standard be well followed in all data files (😆). Maybe @eteq has some insight.

I'm tempted to just say "yeah, you're right -- people should just using `generic_cube`), but if that's the case, we should raise an appropriate error when `wcs.naxis != ` to point users to `generic_cube` for those cases.

dhomeier: Problem is also I cannot even tell if `generic_cube` works, as I don't really know how a valid file should look like.

specutils/io/default_loaders/subaru_pfs_spec.py
dhomeier: To provide a more generic wrapper for `is_fits`, with `_spec_pattern` to be passed by the loader? That might be useful, although I would in general recommend to try to identify the spectrum types on file content rather than naming schemes. But it's useful to have various file type and path checks provided in one function, agreed.

Note that `is_fits` called with `fileobj=None` will identify any valid name pattern as a FITS file, whether `filepath` actually points to an existing file or not. For `_fits_identify_by_name` I have made it a requirement that the file exists, since the loaders will eventually access it anyway.

dhomeier:
@nmearl I have rebased to pull in the updates to the JWST reader from #579, and while at that added a more readable MUSCLES label. Wondering if the `tabular-fits` should not also be changed to something more descriptive like "Generic FITS Table", but there might be more existing code relying on the old label.

nmearl:
@dhomeier I don't want to hold up this PR anymore, but I was hoping to get some comments from Erik concerning the checking for the spectral axis. Would you mind opening a PR capturing the conversation about spectral axis checking and what how explicit we should be knowing that we have the `generic_cube` loader. Then, I think we can go ahead and merge this.

eteq:
On the question of checking for spectral axis: I think I like the implementation as it stands here, at least as a solid iteration.  I also interpreted `tabular-fits` as being "a table that maps to a single spectrum" And as the 😆 surrounding:

> But obviously this requires that the fits standard be well followed in all data files 

Indicates, I think there's not really a foolproof way to do this.

The one case that *might* be useful to test is a fits file where the flux column is a multi-dimensional array - I've never seen that in the wild, but that doesn't mean it doesn't exist, and it seems natural (and trivial to implement) that that would load into `Spectrum1D` as expected.

We might consider an additional reader down the line that can handle "multiple spectra that aren't cubes", but I'd say that's a follow-on PR and doesn't need to hold this up.

So @dhomeier, if you want to add a quick test of the above case, please do so, but if we'd rather think of this as a separate follow-on as well, I'm fine with merging as this stands.

eteq:
(Also, as I side note, some info about the guiding principles separating these loaders should be in the docs, but #394 is really the main issue for that)

dhomeier:
> The one case that _might_ be useful to test is a fits file where the flux column is a multi-dimensional array - I've never seen that in the wild, but that doesn't mean it doesn't exist, and it seems natural (and trivial to implement) that that would load into `Spectrum1D` as expected.
> 
Oh, I have actually produced a couple of those, which might still be out somewhere - the column is not flux but `Intensity` (for different angles), but one might expect that this can be loaded with `column_mapping`.
Turns out this fails in various places:

1. `_find_spectral_column` tries to transform the flux column to "Jy" creating an equivalency from `spectral_axis` - this could be easily overcome by broadcasting the latter to the flux shape.

2. `Spectrum1D` expects the *last* axis of the flux initialiser to match spectral_axis, while Astropy Table is loading it as a (NAXIS1, shape(dtype(flux)) array. This will have to be transposed first. Conversely the writer currently fails when given a spectrum with such a flux of shape (M, NAXIS1) because `Table` again requires the transposed format. All this would be a bit easier if `Spectrum1D` and `Table`/FITS weren't using orthogonal conventions for the order of axes.

> We might consider an additional reader down the line that can handle "multiple spectra that aren't cubes", but I'd say that's a follow-on PR and doesn't need to hold this up.
> 
> So @dhomeier, if you want to add a quick test of the above case, please do so, but if we'd rather think of this as a separate follow-on as well, I'm fine with merging as this stands.

Since it's not as quick to produce a simple example case and requires a number of changes to `parsing_utils`, `Spectrum1D.__init__` and the loaders, I concur that this is a PR of its own.
Also this still does not address the `wcs` verification in `tabular_fits`, which deals with a `spectral_axis `that is not part of the table, but encoded as a `WCS` object.

When looking at the `wcs1d-fits` examples, their WCS returns a spectral_axis of shape (0, NAXIS1), so maybe `tabular-fits` should also test for
`wcs.naxis == 1 and wcs.array_shape[-1] == len(tab)`
but then again in the `wcs1d-fits` files the flux is stored as NAXIS1x1 dim image, so I don't really know what one should expect for the combination TABLE + WCS, so this can probably wait for the follow-up PR.

I've also tested a number of improvements for the auto-detection of `wcs1d-fits`, but I'd defer them to their own PR as well.

dhomeier:
> (Also, as I side note, some info about the guiding principles separating these loaders should be in the docs, but #394 is really the main issue for that)

Perhaps also #584 for ranking the loaders by priority.

nmearl:
> Spectrum1D expects the last axis of the flux initialiser to match spectral_axis, while Astropy Table is loading it as a (NAXIS1, shape(dtype(flux)) array

Can you expand on this a bit? How are you accessing the data within the table? Specutils follows the numpy row-major formalism; it may be that this is just a quirk of data tables being preferentially column-major in general.

Otherwise, it sounds like this PR is good to go with the caveat of opening (as I see it) two new issues to be addressed in follow-up PRs:

1. Support broadcasting in `_find_spectral_column` for file loading.
2. Extend tabular-fits to include multiple spectra that aren't cubes.

Does this sound right @dhomeier, @eteq?

dhomeier:
> > Spectrum1D expects the last axis of the flux initialiser to match spectral_axis, while Astropy Table is loading it as a (NAXIS1, shape(dtype(flux)) array
> 
> Can you expand on this a bit? How are you accessing the data within the table? Specutils follows the numpy row-major formalism; it may be that this is just a quirk of data tables being preferentially column-major in general.

I think it's the consequence of `fits.io` loading it as a `recarray`, e.g. my file in question is structured like this (actual second column is intensity, but renamed `Flux` here to allow `generic_spectrum_from_table` to find it).
```
TTYPE1  = 'Wavelength'
TFORM1  = 'E       '
TUNIT1  = 'Angstrom'
TTYPE2  = 'Flux     '          / I_mu/<I>
TFORM2  = '20E     '
TUNIT2  = 'erg/s/cm2/Angstrom'
```
which is loaded as a `FITS_rec` of
```
>>> hdu.data.shape
(314931,)
>>> hdu.data.dtype
dtype((numpy.record, [('Wavelength', '>f4'), ('Flux', '>f4', (20,))]))
```
and `Table.read()` makes it a
```
<Table length=314931>
Wavelength      Flux [20]            
 Angstrom     erg / (Angstrom cm2 s)     
 float32      float32             
---------- --------------------------------
   3.0e+03  9.5420114e-07 ..  1.3871580e+00
```
which I would actually consider the more natural representation – `Wavelength` and `Flux` column having the same length (number of rows). So I'd tend to adapt `Spectrum1D.__init__` to handle this for tables in general, since you could not even construct a table from a length-314931 wavelength and a length-20 flux column.

dhomeier:
I guess that would require `Spectrum1D.__init__` to distinguish between Table/recarray columns and plain arrays, so the current handling of 2D arrays is not broken. Although the latter seems to have some flaws e.g. with slicing:
```
>>> sp1d = Spectrum1D(spectral_axis=np.arange(5)*apu.nm, flux=np.ones(5)*u.Jy)
>>> sp1d[:2]
<Spectrum1D(flux=<Quantity [1., 1.] Jy>, spectral_axis=<Quantity [0., 1.] nm>)>
>>> sp2d = specutils.Spectrum1D(spectral_axis=np.arange(5)*apu.nm, flux=np.ones((3, 5))*u.Jy)
>>> sp2d[:2]
<Spectrum1D(flux=<Quantity [[1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1.]] Jy>, spectral_axis=<Quantity [0., 1., 2., 3., 4.] nm>)>
```

nmearl:
> I guess that would require Spectrum1D.__init__ to distinguish between Table/recarray columns and plain arrays

I do not think we want to implement this handling in the intializer. The supported inputs are specifically `Quantity` objects, and we want to avoid the case where we're bloating the initializer on doing users' data transformations for them (e.g. requests for taking pandas dataframes, dask objects, etc), and we certainly do not want that burden. It should be on the user and the data loaders to change the flux and spectral axis values to the appropriate `Quantity` format.

> Although the latter seems to have some flaws e.g. with slicing:

I'm not sure I'm seeing what the issue is here? If the flux array is multi-dimensional, the slicing is intended to occur on the elements in the flux array, returning a new `Spectrum1D`.

dhomeier:
> I do not think we want to implement this handling in the intializer. The supported inputs are specifically `Quantity` objects, and we want to avoid the case where we're bloating the initializer by doing users' data transformations for them (e.g. I can image getting future requests for taking pandas dataframes, dask objects, etc), and we certainly do not want that burden. 

OK, then it can be handled in `parsing_utils`. Using tables or columns seemed a common way to initialise the spectrum (and IMO are a bit different from pandas frames etc. since they are an Astropy standard data format and already come with `Quantity` support), but you're right that the docstrings do not mention them in any way.
I think the `astropy.nddata.NDData` part should be removed then, since they are not accepted either. 
> 
> > Although the latter seems to have some flaws e.g. with slicing:
> 
> I'm not sure I'm seeing what the issue is here? If the flux array is multi-dimensional, the slicing is intended to occur on the elements in the flux array, returning a new `Spectrum1D`.

Slicing would straightforwardly extract a desired wavelength range (assuming you have found the corresponding indices). For the 2d spectrum I don't see how to get a comparable functionality, even if you checked for the different `ndim` and `shape`:
```
>>> sp2.shape
(3, 5)
>>> sp2[:,:2]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/derek/lib/python3.8/site-packages/specutils/spectra/spectrum1d.py", line 172, in __getitem__
    return self._copy(
  File "/Users/derek/lib/python3.8/site-packages/specutils/spectra/spectrum1d.py", line 199, in _copy
    return self.__class__(**alt_kwargs)
  File "/Users/derek/lib/python3.8/site-packages/specutils/spectra/spectrum1d.py", line 111, in __init__
    raise ValueError('Spectral axis ({}) and the last flux axis ({}) lengths must be the same'.format(
ValueError: Spectral axis (5) and the last flux axis (2) lengths must be the same
```

dhomeier:
I've now implemented this as described above, trying to write it as general as possible, but really only checked and added tests for `ndim=2`.
I haven't added a changelog entry as there is no section for 1.0 or 1.1 yet.

specutils/tests/test_loaders.py
```diff
@@ -321,10 +322,43 @@ def test_tabular_fits_writer(tmpdir, spectral_axis):
...
+                             spectrum.uncertainty.quantity)
+
+
+@pytest.mark.parametrize("spectral_axis",
+                         ['wavelength', 'frequency', 'energy', 'wavenumber'])
+def test_tabular_fits_2d(tmpdir, spectral_axis):
+    wlu = {'wavelength': u.AA, 'frequency': u.GHz, 'energy': u.eV,
+           'wavenumber': u.cm**-1}
+    # Create a small data set with 2D flux + uncertainty
+    disp = np.arange(1, 1.1, 0.01)*wlu[spectral_axis]
```
nmearl: Eventually we're going to want to ensure that any spectral axis in frequency space is strictly descending. I'm not sure we need to worry about that case right now, but something to think about.

```suggestion
    disp = np.arange(1, 1.1, 0.01)*wlu[spectral_axis]
    
    if spectral_axis == 'frequency':
        disp = disp[::-1]
```

nmearl:
> I haven't added a changelog entry as there is no section for 1.0 or 1.1 yet.

Feel free to add a new section for 1.1 and include a change log entry, otherwise we generally comb through the PRs and add any merged ones to the PR right before we do a release.

This looks good!

specutils/tests/test_loaders.py
dhomeier: > Eventually we're going to want to ensure that any spectral axis in frequency space is strictly descending. I'm not sure we need to worry about that case right now, but something to think about.

That's actually everything but wavelength, right? Yes, if there is functionality that depends on descending order in frequency/energy, that's probably good to keep as reminder.

nmearl:
@dhomeier can you rebase to get rid of that last merge commit?

dhomeier:
@nmearl sorry, had already updated locally, but I hope everything is in order now.

nmearl:
Thanks for your work on this @dhomeier!

dhomeier:
Thanks for the feedback @nmearl!

PR #579, opened 2020-01-31
Update SpectrumList loader for JWST data

jdavies-st:
- Allow the reader to read either `flux` or `surf_bright`
- Generalize the extraction of units to track column name, not column number
- Read in the uncertainty array
- Fix up some docstrings elsewhere
- Write tests, tests, tests

Note that `extract1d` spectral products in the pipeline currently do not have populated uncertainty arrays.  They are all zeros.  That will be changing in the next month or so.

### DIFF
--- a/specutils/io/default_loaders/jwst_reader.py
+++ b/specutils/io/default_loaders/jwst_reader.py
@@ -34,11 +34,51 @@ def identify_jwst_x1d_fits(origin, *args, **kwargs):
         return False
 
 
-@data_loader("JWST", identifier=identify_jwst_x1d_fits, dtype=SpectrumList,
+@data_loader("JWST x1d", identifier=identify_jwst_x1d_fits, dtype=Spectrum1D,
              extensions=['fits'])
-def jwst_loader(filename, **kwargs):
+def jwst_x1d_single_loader(filename, **kwargs):
     """
     Loader for JWST x1d 1-D spectral data in FITS format
+
+    Parameters
+    ----------
+    filename: str
+        The path to the FITS file
+
+    Returns
+    -------
+    Spectrum1D
+        The spectrum contained in the file.
+    """
+    spectrum_list = _jwst_x1d_loader(filename, **kwargs)
+    if len(spectrum_list) == 1:
+        return spectrum_list[0]
+    else:
+        raise RuntimeError(f"{filename} has {len(spectrum_list)} spectra. "
+            "Use the SpectrumList.read.")
+
+
+@data_loader("JWST x1d multi", identifier=identify_jwst_x1d_fits, dtype=SpectrumList,
+             extensions=['fits'])
+def jwst_x1d_multi_loader(filename, **kwargs):
+    """
+    Loader for JWST x1d 1-D spectral data in FITS format
+
+    Parameters
+    ----------
+    filename: str
+        The path to the FITS file
+
+    Returns
+    -------
+    SpectrumList
+        A list of the spectra that are contained in the file.
+    """
+    return _jwst_x1d_loader(filename, **kwargs)
+
+
+def _jwst_x1d_loader(filename, **kwargs):
+    """Implementation of loader for JWST x1d 1-D spectral data in FITS format
 
     Parameters
     ----------
--- a/specutils/io/default_loaders/jwst_reader.py
+++ b/specutils/io/default_loaders/jwst_reader.py
@@ -1,6 +1,8 @@
 import warnings
 
 import astropy.units as u
+from astropy.units import Quantity
+from astropy.table import Table
 from astropy.io import fits
 from astropy.nddata import StdDevUncertainty
 from astropy.utils.exceptions import AstropyUserWarning
@@ -120,8 +122,9 @@ def _jwst_x1d_loader(filename, **kwargs):
             if hdu.name != 'EXTRACT1D':
                 continue
 
-            wavelength_units = u.Unit(hdu.columns["wavelength"].unit)
-            wavelength = hdu.data["wavelength"] * wavelength_units
+            data = Table.read(hdu)
+
+            wavelength = Quantity(data["WAVELENGTH"])
 
             # Determine if FLUX or SURF_BRIGHT column should be returned
             # based on whether it is point or extended source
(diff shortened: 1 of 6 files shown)


jdavies-st:
Good point.  I believe `srctype` can sometimes be `"UNKNOWN"` as well, so this would be a good thing to check.

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -22,52 +26,78 @@ def identify_jwst_fits(origin, *args, **kwargs):
...
+            if not hdulist[0].header["TELESCOP"] == "JWST":
+                return False
         return True
     # This probably means we didn't have a FITS file
     except Exception:
         return False
 
 
-@data_loader("JWST", identifier=identify_jwst_fits, dtype=SpectrumList,
+@data_loader("JWST", identifier=identify_jwst_x1d_fits, dtype=SpectrumList,
```
eteq:
```suggestion
@data_loader("JWST_x1d", identifier=identify_jwst_x1d_fits, dtype=SpectrumList,
```

I think this is necessary if we want to have a whole family of JWST loaders, right? I.e., if we don't do this there's no way to use the name to specify *which* of the JWST loaders to use?

eteq:
A related thought to the above (but somewhat independent): should there *also* be a `Spectrum1D` that can be used for JWST data sets that have just one spectrum in them?  I don't know if this is actually an at all common use case, but it seems like it would be a trivial addition to this if there are any?

specutils/io/default_loaders/jwst_reader.py
jdavies-st: Yes, I think this is necessary as well.  Would this be a backwards incompatible change though?  Does that matter?

Perhaps this is where my limited knowledge of the infrastructure of the io registry and specifically in how it has been implemented in `specutils` works.

jdavies-st:
I've added a reader for single spectrum x1d files for the `Spectrum1D` class.  If passed multiple spectra, it raises RuntimeError.  I've also expanded the testing.  Hopefully I've responded to all review comments now.

specutils/io/default_loaders/tests/test_jwst_loader.py
```diff
@@ -43,3 +67,45 @@ def test_jwst_loader(tmpdir):
...
+    assert data.shape == (100,)
+
+
+def test_jwst_x1d_singel_loader_fail_on_multi(tmpdir, x1d_multi):
+    """Make sure Spectrum1D.read on JWST x1d with many spectra errors out"""
+    tmpfile = str(tmpdir.join('jwst.fits'))
+    x1d_multi.writeto(tmpfile)
+
+    with pytest.raises(RuntimeError, match="SpectrumList"):
+        Spectrum1D.read(tmpfile, format='JWST x1d')
```
nmearl:
```suggestion
        Spectrum1D.read(tmpfile, format='JWST x1d multi')
```

Should this be failing to load using the `SpectrumList` loader?

jdavies-st: No, this test to verify that if I use `Spectrum1D.read(file, format="JWST x1d")` on a JWST multi-spec file, then we get a failure telling me to use the `SpectrumList.read()` loader.  And I guess I could remove the format statement here as it is not needed and is probably is causing confusion for the reader.

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -22,52 +26,118 @@ def identify_jwst_fits(origin, *args, **kwargs):
...
+                # SRCTYPE is in primary header because there is only one spectrum
+                srctype = hdulist.header.get("srctype")
+
+            if srctype == "POINT":
+                flux_units = u.Unit(hdu.columns["flux"].unit)
+                flux = hdu.data["flux"] * flux_units
+
+                error_units = u.Unit(hdu.columns["error"].unit)
+                uncertainty = StdDevUncertainty(hdu.data["error"] * error_units)
+
```
dhomeier: Having just sneaked into a discussion on the pros and cons of using the `Table` API for loading hdu data, I am wondering if this could be streamlined to
```
            data = Table.read(hdu)

            if srctype == "POINT":
                flux = Quantity(data["flux"])
                uncertainty = StdDevUncertainty(data["error"])
...
```
and so on for the other `srctype`s – see e.g. for comparison [muscles_sed.py](https://github.com/astropy/specutils/blob/master/specutils/io/default_loaders/muscles_sed.py).

jdavies-st: That looks good!  I'll give it a try.  Thanks for the tip.

jdavies-st: That worked very well.  Thanks.  Added in a new commit.  And I added some units checking to the tests, something I had overlooked before.

dhomeier: Just remembered that `parsing_utils` has a function `spectrum_from_column_mapping` that in principle would even allow to simply specify this as dictionaries like
```
            if srctype == "POINT":
                column_mapping = {'wavelength': ('spectral_axis', None),
                                  'flux': ('flux', None),
                                  'error': ('uncertainty', None)}
            elif srctype == "EXTENDED":
                column_mapping = {'wavelength': ('spectral_axis', None),
                                  'surf_bright': ('flux', None),
                                  'sb_error': ('uncertainty', None)}
```
but this would require #573 to be merged to work without explicitly specifying the units - just for reference.

jdavies-st: I can add that simplification in a future PR.  👍

jdavies-st:
And finally I added some population of the `Spectrum1D.meta` attribute.  Basically concat together the PRIMARY FITS header and the per-EXTRACT1D extension header into the `meta` dict as `key:value` pairs.  This doesn't retain the header object comments, but perhaps this is preferable to having the whole FITS header as a value in a `header` dict key?

nmearl:
@jdavies-st I think the method of concatenating the header is fine, I'm not see any reason why it should cause issues now. But we should remind ourselves that we did it if we do run into programs. This looks great!

PR #580, opened 2020-02-03
Fix typo on L98 of docs/analysis.rst

kakirastern:
Change "equivalen" into "equivalent" on L98 to correct misspelling.


nmearl:
Thanks for the doc fix, @kakirastern!

PR #581, opened 2020-02-07
Fix unit information getter for spectral axis

nmearl:
Since the switch to the new wcs framework in astropy, `WCS` no longer has a `.unit` property.

### DIFF
--- a/specutils/spectra/spectrum_mixin.py
+++ b/specutils/spectra/spectrum_mixin.py
@@ -84,10 +84,7 @@ class OneDSpectrumMixin:
         """
         Returns the units of the spectral axis.
         """
-        if isinstance(self.wcs, HighLevelWCSWrapper):
-            return u.Unit(self.wcs.world_axis_units[0])
-
-        return self.wcs.unit[0]
+        return u.Unit(self.wcs.world_axis_units[0])
 
     @property
     def flux(self):


specutils/spectra/spectrum_mixin.py
```diff
@@ -84,10 +84,7 @@ def spectral_axis_unit(self):
         """
         Returns the units of the spectral axis.
         """
-        if isinstance(self.wcs, HighLevelWCSWrapper):
-            return u.Unit(self.wcs.world_axis_units[0])
-
-        return self.wcs.unit[0]
+        return u.Unit(self.wcs.world_axis_units[0])
```
eteq:
```suggestion
        if hasattr(self.wcs, 'world_axis_units'):
            return u.Unit(self.wcs.world_axis_units[0])
        else:
            return u.Unit(self.wcs.unit[0])
```

nmearl: I'm not sure about this. The `self.wcs.unit[0]` formalism is the pre-astropy 4.0 method specifically for FITS WCSs. But since the current specutils claims to be incompatible with anything less that astropy 4.0, and we have made it such that it can work with anything that follows the APE14 WCS paradigm, I'm really hesitant to build in fail safes that suggest we aren't really committed to that.

nmearl:
>  Instead it's just important that we be sure the world_axis_units invocation is tested somewhere. 

The `spectral_axis_unit` call on `Spectrum1D` objects is called in several places including in smoothing functions and resampling functions, which are caught by their associated tests.

specutils/spectra/spectrum_mixin.py
eteq: Point taken. Especially given #582.  Already, I retract my suggestion then.

Issue #586, opened 2020-02-18
Subspectrum Fit Error


elliesch:
I'm running a fit on a subspectrum, following the specutils tutorial [here](https://specutils.readthedocs.io/en/stable/fitting.html).

This is the code routine that I'm using:
```python
from astropy.modeling import models
from astropy import units as u
from specutils.spectra import Spectrum1D
from specutils.fitting import fit_lines

spectrum = Spectrum1D(flux=flux*u.Jy, spectral_axis=lambda*u.angstrom,
                      rest_value=6561.16*u.angstrom)
sub_region = SpectralRegion(6490*u.angstrom, 6635*u.angstrom)
sub_spectrum = extract_region(spectrum, sub_region)

g_init = models.Gaussian1D(amplitude=1*u.Jy, mean=6561.157*u.angstrom, 
                           stddev=4.0721*u.angstrom)
g_fit = fit_lines(sub_spectrum, g_init)
y_fit = g_fit(sub_spectrum.spectral_axis)
```

This code snippet worked on 12/5/19. Since then I have updated both astropy to version 4.0 and specutils to 7.0, but I'm not sure what versions I had on 12/5/19. When running it now I get the following error on `g_fit`:

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-86-cd94dcaa50ee> in <module>
     21 
     22 
---> 23 g_fit = fit_lines(sub_spectrum, g_init)
     24 y_fit = g_fit(sub_spectrum.spectral_axis)
     25 

~/opt/anaconda3/lib/python3.7/site-packages/specutils/fitting/fitmodels.py in fit_lines(spectrum, model, fitter, exclude_regions, weights, window, **kwargs)
    364         fit_model = _fit_lines(spectrum, model_guess, fitter,
    365                                exclude_regions, weights, model_window,
--> 366                                ignore_units, **kwargs)
    367         if model_guess.name is not None:
    368             fit_model.name = model_guess.name

~/opt/anaconda3/lib/python3.7/site-packages/specutils/fitting/fitmodels.py in _fit_lines(spectrum, model, fitter, exclude_regions, weights, window, ignore_units, **kwargs)
    485         spectral_axis=dispersion.value * dispersion_unit,
    486         wcs=input_spectrum.wcs,
--> 487         velocity_convention=input_spectrum.velocity_convention,
    488         rest_value=input_spectrum.rest_value)
    489 

[...]

PR #593, opened 2020-02-23
Fix for regression in issue #586

nmearl:
The initializer was bailing early when being passed unit-less numpy arrays assuming they are single valued arithmetic operands.

### DIFF
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -85,8 +85,9 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
         # In the case where the arithmetic operation is being performed with
         # a single float, int, or array object, just go ahead and ignore wcs
         # requirements
-        if isinstance(flux, float) or isinstance(flux, int) \
-                or not isinstance(flux, u.Quantity):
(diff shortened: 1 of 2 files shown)


specutils/spectra/spectrum1d.py
```diff
@@ -85,8 +85,9 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
         # In the case where the arithmetic operation is being performed with
         # a single float, int, or array object, just go ahead and ignore wcs
         # requirements
-        if isinstance(flux, float) or isinstance(flux, int) \
-                or not isinstance(flux, u.Quantity):
+        if np.ndim(flux) == 0 and (isinstance(flux, float)
+            or isinstance(flux, int)
+            or not isinstance(flux, u.Quantity)):
```
eteq:
```suggestion
        if not isinstance(flux, u.Quantity) and (np.ndim(flux) == 0 and
                                                (isinstance(flux, float)
                                                 or isinstance(flux, int))):
```
A minor adjustment, but I think it's probably better to *first* check the `Quantity` bit since that's a lot cheaper than `np.ndim` (which potentially does some conversions)

eteq:
LGTM now, will merge once tests pass!

PR #595, opened 2020-02-25
Spectrum1D loader for JWST s2d and s3d rectified spectra

jdavies-st:
This PR

- implements Spectrum1D and SpectrumList readers for JWST `_s2d.fits` and `_s3d.fits` files, the resampled slit image outputs from `resample_spec` and build IFU cubes from `cube_build`.

- adds `asdf>=2.5` as a package dependency.  This has always been a package dependency, but it was not in `install_requires` and was being fulfilled by the `gwcs` dependency which has `asdf` as a dependency.

- adds another matrix element to the TravisCI build that installs the latest stable version of `jwst` so as to allow the unit tests for these readers to get tested.  In the future this can be added as a test dependency as an optional install if we like.


jdavies-st:
Thanks @rosteen, I added detail to the docstrings.

jdavies-st:
I think this is ready for final review.  🤞

Issue #604, opened 2020-03-12
Hide or delete spectral_resolution method


eteq:
I just realized there's a `spectral_resolution` method of `Spectrum1D` which is quite old (~3 years) but has no implementation.  I think this is an artifact from when `Spectrum1D` was a concept API instead of actually being implemented.  This method should either be changed to ``raise NotImplementedError` instead of ``pass``, or perhaps just removed (or at least made private with a leading underscore) instead of left in the API seemingly to do something.

If I can find the time I might make a quick PR to do this a bit later.

rosteen:
According to the [python docs](https://docs.python.org/3/library/exceptions.html#NotImplementedError) , NotImplementedError "should not be used to indicate that an operator or method is not meant to be supported at all". My vote is to remove or make private for now if we might want it later.

hcferguson:
My recollection from the workshop a few years ago is that DESI was planning to carry around a line-spread function (LSF) in metadata; either that or a spectral-resolution value. It would be interesting to find out what they have implemented for their pipeline.

rosteen:
I created a PR that removes the function, which seemed like the cleanest fix for the 1.0 release while we look into e.g. the DESI implementation and decide on what to do in the future.

PR #606, opened 2020-03-13
Remove unimplemented spectral_resolution function from Spectrum1D

rosteen:
Resolves [issue 604](https://github.com/astropy/specutils/issues/604). It seemed to me that it would be cleanest for the 1.0 release to just delete the function; we can always reinstate it at a later time if desired.


nmearl:
Thanks @rosteen! Can you open an issue to discuss the necessity of such a method, so we can be sure the intent (if anyone remembers) is represented somewhere?

rosteen:
Done, see [Issue #607 ](https://github.com/astropy/specutils/issues/607)

PR #608, opened 2020-03-17
Add 6dFGS loader

aragilar:
This adds support for loading files from 6dFGS which were stored in the tabular variant.

Some questions:
 * What's the best way of uploading test data for the tests, currently there is no stable place for downloading the files (I can provide example files though)? I'll add tests once I know where I can put the data files.
 * I'm unsure where the tox configuration is stored—I'm currently running into problems with tests not finding Cython (even though it is installed).


dhomeier:
> * What's the best way of uploading test data for the tests, currently there is no stable place for downloading the files (I can provide example files though)? I'll add tests once I know where I can put the data files.

You mean you don't have a permanent download link for the files?
There are a number of test files uploaded to Zenodo for just these purposes (see `tests/conftest.py` and `tests/test_default_loaders.py` for more details). I don't know if it is possible to add files to one of the existing archives, or if you could simply upload them into your own archive and pass the location with the respective
`@remote_access([{'id': file_id, 'filename': file_name}])`

Else I guess for small files it would be acceptable to create e.g. a `io/default_loaders/tests/data` subdirectory analogously to the astropy test file setup, but I'd wait for one of the maintainers' opinion.

dhomeier:
> * I'm unsure where the tox configuration is stored—I'm currently running into problems with tests not finding Cython (even though it is installed).

Have you tried running `pytest` directly? The default options (like `--remote-data=any`) are already set in `setup.cfg`; I'd think if cython is in your standard path this should work.

aragilar:
I managed to run setup.py test within a venv where I installed all the deps, so I can now run the tests.

There's no permanent place yet. I'm currently working on getting a large number of older surveys ingested into Data Central, but it will take some time before there's a permanent and reliable host for the data.

specutils/io/default_loaders/sixdfgs_reader.py
```diff
@@ -0,0 +1,54 @@
...
+    file_name: str
+        The path to the FITS file
+
+    Returns
+    -------
+    data: Spectrum1D
+        The 6dF spectrum that is represented by the data in this table.
+    """
+    with fits.open(filename) as hdulist:
+        header = hdulist[0]
```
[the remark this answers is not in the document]
aragilar: Ah, yeah, typo on my part. :)

The original 6dFGS Final Data Release released the observations in two formats, the tabular one (which this loader loads), which only contained flux and wavelength (no uncertainty), and one where the reduced spectra (with sky/uncertainty information, both with the separate arms and combined) was included with 2MASS and SuperCOSMOS (I think?) postage stamps. Then GAMA came along and only used the combined and best version (so single HDU). And now the plan for Data Central is to split all the 6dFGS spectra into their own individual file (avoiding the need to keep downloading postage stamps, and make it easier to understand what data each object has).

I thought about including the header from the actual table HDU, but there's no metadata there (other than the table metadata), so rather than adding another key/dict pair, I left it out. Happy to include it for completeness sake though.

specutils/io/default_loaders/sixdfgs_reader.py
```diff
@@ -0,0 +1,54 @@
...
+        table = QTable(hdulist[1])
+        flux = table["FLUX"]
```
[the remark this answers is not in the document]
aragilar: Oh, cool, I didn't know I could do that. I'll match the other loaders for consistency then.

specutils/io/default_loaders/sixdfgs_reader.py
```diff
@@ -0,0 +1,54 @@
...
+        wavelength = table["WAVE"]
+
+        if flux.unit == COUNTS_PER_SECOND:
+            flux._unit = Unit("count/s")
+        meta = {"header": header}
+
+        return Spectrum1D(flux=flux, spectral_axis=wavelegth, meta=meta)
```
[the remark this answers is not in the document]
aragilar: Oops, thanks.

Yeah, the table variant didn't include uncertainty data unfortunately.

aragilar:
I'll look at uploading a few of the FITS files to zenodo, and then add some tests.

Do the apogee tests normally timeout (I'm noticing that they are the failing tests on travis).

specutils/io/default_loaders/sixdfgs_reader.py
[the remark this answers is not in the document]
dhomeier: If the table HDU header really contains no information of interest (you could also check `table.meta` below, which will only preserve non-trivial header cards), then there is probably no need to add it. #617 has raised a few more questions on how to organise the header info inside `meta`.

aragilar:
Sorry for not getting back to this earlier, I've added the test and fixed the other issues.

dhomeier:
Looks good; I agree that in your case the only `meta` information of interest is in the primary HDU's header.

Thanks for pointing to the `IMAGE`-format files, that data model turned out to be of interest as I am adding a WCS spectrum writer in #632, but have not implemented any functionality for storing `uncertainty`!
If it is of interest, the updated loader from that PR would also be able to read these spectra with `format='wcs1d-fits', hdu_idx=5` and so on, so in principle the uncertainties from there should be accessible as `StdDevUncertainty(np.sqrt(spectrum.flux.value[1]))`.

dhomeier:
Have you made any changes in the last commit? I got a notification some 3 days ago, but the commit it referred to seems to have already been deleted again.
Generally it looks like you have completely reset all previous commits before force-pushing a new version here. This makes it almost impossible to keep track of the changes as your PR evolved – I suggest to check the recommendations on [rebasing and squashing](http://docs.astropy.org/en/latest/development/workflow/development_workflow.html#rebase-but-only-if-asked).

aragilar:
Odd, I only rebased this at the same time as #633 (as I figured it'd have the same CI failure otherwise). The code hasn't changed since I added the tests.

dhomeier:
Strange – yes, the rebase after 8339d8a was in place to fix the `build_docs` failure, but I see a total of 6 force-pushes and only a single commit in the entire PR.

specutils/io/default_loaders/sixdfgs_reader.py
```diff
@@ -0,0 +1,63 @@
...
+    https://docs.datacentral.org.au/6dfgs/. Catalogues and spectra were
+    produced, with the spectra being provided as both fits tables and as fits
+    images. This loads the tabular variant of the spectra. Note that this does
+    not include uncertainties - uncertainties are only provided in the image
+    variants.
+
+    Parameters
+    ----------
+    file_name: str
+        The path to the FITS file
```
dhomeier, line 46:
```suggestion
    file_obj: str, file-like or HDUList
         FITS file name, object (provided from name by Astropy I/O Registry),
         or HDUList (as resulting from astropy.io.fits.open()).
```

specutils/io/default_loaders/sixdfgs_reader.py
```diff
@@ -0,0 +1,63 @@
...
+        header = hdulist[0].header
+        table = Table.read(hdulist)
+        flux = Quantity(table["FLUX"])
+        wavelength = Quantity(table["WAVE"])
+
+        if flux.unit == COUNTS_PER_SECOND:
+            flux._unit = Unit("count/s")
+        meta = {"header": header}
+
+        return Spectrum1D(flux=flux, spectral_axis=wavelength, meta=meta)
```
dhomeier, line 63:
```suggestion
    if isinstance(file_obj, fits.hdu.hdulist.HDUList):
        hdulist = file_obj
    else:
        hdulist = fits.open(file_obj, **kwargs)

    header = hdulist[0].header
    table = Table.read(hdulist)
    flux = Quantity(table["FLUX"])
    wavelength = Quantity(table["WAVE"])

    if flux.unit == COUNTS_PER_SECOND:
        flux._unit = Unit("count/s")
    meta = {"header": header}

    if not isinstance(file_obj, fits.hdu.hdulist.HDUList):
        hdulist.close()

    return Spectrum1D(flux=flux, spectral_axis=wavelength, meta=meta)
```

specutils/tests/test_loaders.py
```diff
@@ -490,3 +490,11 @@ def test_subaru_pfs_loader(tmpdir):
...
     assert spec.spectral_axis.unit == u.nm
     assert spec.flux.unit == u.nJy
+
+
+@remote_access([{'id': '3733958', 'filename': '1D-c0022498-344732.fits'}])
+def test_spectrum1d_6dfgs_tabular(remote_data_path):
+    spec = Spectrum1D.read(remote_data_path)
+
+    assert spec.spectral_axis.unit == u.Unit("Angstrom")
+    assert spec.flux.unit == u.Unit("count/s")
```
dhomeier, line 500:
```suggestion
    assert spec.flux.unit == u.Unit("count/s")

    # Read from HDUList object
    hdulist = fits.open(remote_data_path)
    spec = Spectrum1D.read(hdulist, format="6dFGS-tabular")
    assert isinstance(spec, Spectrum1D)
    assert spec.flux.unit == u.Unit("count/s")
    assert spec.flux.size > 0
    hdulist.close()
```

specutils/io/default_loaders/sixdfgs_reader.py
```diff
@@ -0,0 +1,63 @@
...
+        if primary_hdu.header["NAXIS"] != 0:
+            return False
+        for keyword in SIXDFGS_PRIMARY_HDU_KEYWORDS:
+            if keyword not in primary_hdu.header:
+                return False
+        return True
+
+
+@data_loader("6dFGS-tabular",
+             identifier=identify_6dfgs_tabular_fits,dtype=Spectrum1D,
```
dhomeier, line 28:
```suggestion [E231]
             identifier=identify_6dfgs_tabular_fits, dtype=Spectrum1D,
```
[E231]

specutils/io/default_loaders/sixdfgs_reader.py
```diff
@@ -0,0 +1,63 @@
...
+             extensions=["fit", "fits"])
+def sixdfgs_tabular_fits_loader(filename, **kwargs):
```
dhomeier, line 30:
```suggestion
             identifier=identify_6dfgs_tabular_fits, dtype=Spectrum1D,
             extensions=["fit", "fits"])
def sixdfgs_tabular_fits_loader(file_obj, **kwargs):
```

aragilar:
Cool, I've applied the suggestions (it's probably worth pulling the things from #637 into a common decorator, so that support for the different types doesn't require custom code each time).

aragilar:
I've added the other format variants that 6dFGS can come in, looks like the tests are again failing due to a reference issue in the docs, should I rebase this again?

dhomeier:
> I've added the other format variants that 6dFGS can come in, looks like the tests are again failing due to a reference issue in the docs, should I rebase this again?

I think we can ignore the `build_docs` failures for now; from #655 there is no **consistent** fix for this yet anyway, so no point in rebasing, thanks!

nmearl:
This looks pretty good to me -- is there anymore intended work on the loaders?

aragilar:
No, these should cover all the possible outputs I know to exist for 6dFGS.

Issue #496, opened 2019-07-29
Docs: what does `excise` do?


keflavich:
There are excise_region functions:
https://specutils.readthedocs.io/en/latest/api/specutils.manipulation.excise_regions.html#specutils.manipulation.excise_regions

that say they're using convolution.  I think the docs were incorrectly copy and pasted from something else?  I'd submit a PR, but I don't know the answer...

dhomeier:
I've come to similar conclusions when trying to retrace the procedure `fit_lines(..., exclude_regions=[...])` is initiating in #505.
What it does is certainly replacing specified regions of the spectrum with a fill-in function provided by `exciser` - with `linear_exciser` currently the only available option. The doc strings on convolution don't seem to have anything to do with this. There were some other questions that came to my mind:

1. Is there really a need for two separate functions `excise_regions` and `excise_region`, when the former has the sole function of repeatedly calling the latter? And if there is, does it really need a full duplication of the documentation etc., instead of just marking it as a thin wrapper to `excise_region`? But since the latter is not exposed outside of this module, I would think that `excise_regions` could just as well do the call to `exciser` for each `region` directly.

2. `linear_exciser` currently returns the `uncertainty`, if present, unmodified, which is obviously no longer correct for the replaced flux values in the excised regions. Since the new fluxes in each region between indices `s` and `e` only depend on the neighbouring spectral points as
`modified_flux[i] = ws*flux[s] + we*flux[e]`
with the weights given by the relative distances to `wavelengths[[s,e]]`, technically the corresponding uncertainties `spectrum.uncertainty[i]` are given by
`np.sqrt(ws*spectrum.uncertainty[s]**2 + we*spectrum.uncertainty[e]**2)`
but those will all be correlated for the entire region, and should thus certainly not be used as a `StdDevUncertainty`.

3. Doing the above interpolation for the `flux` values with `np.linspace` implicitly assumes constant sampling of `spectral_axis`.

4. `excise_region` requires and checks for a `SpectralRegion` as `region` argument, but raises an error `'The region parameter must be a 2-tuples of start and end wavelengths.'` otherwise (probably left over from earlier versions where it did expect wavelength tuples).

rosteen:
@nmearl I'm working on some code fixes/improvements in addition to correcting the docs for this. Should I make a new ticket for the new/improved functionality or can that live under this ticket?

[...]

PR #609, opened 2020-03-17
Specutils 496

rosteen:
Addresses [Issue #496](https://github.com/astropy/specutils/issues/496) and some points raised in the discussion there. A couple notes:

1) The default exciser function for excise_regions is now true_exciser, which simply deletes the array elements in the input spectral region rather than replacing the values with a linear ramp between the boundary points as before.

2) The linear_exciser function now handles regions with multiple subregions, but the behavior if those subregions overlap with each other might still be a little wonky. For now I added a warning telling the user they should probably use non-overlapping regions rather than a bunch of complex code to handle that edge case.


rosteen:
Note that test_region_extract.test_linear_excise_invert_from_spectrum was passing in previous builds because both linear_exciser and the test itself were incorrect in ways that canceled each other out to passing. The test has been fixed in this build to do what (I think) it was intended to do in the first place.

specutils/manipulation/utils.py
```diff
@@ -1,26 +1,92 @@
...
+       In the case that ``spectrum`` and ``region`` are not the correct types.
+
+    """
+
+    spectral_axis = spectrum.spectral_axis
+    excise_indices = None
+
+    for subregion in region:
+        # Find the indices of the spectral_axis array corresponding to the subregion
+        wavelengths_in = (spectral_axis >= region.lower) & (spectral_axis < region.upper)
```
nmearl: I know this isn't particular to this PR, but something we certainly should have caught when this was implemented. The spectral axis is not guaranteed to be in wavelength space, so I think we ought to avoid using the term "wavelengths".

```suggestion
        region_mask = (spectral_axis >= region.lower) & (spectral_axis < region.upper)
```

specutils/manipulation/utils.py
```diff
@@ -1,26 +1,92 @@
...
+        temp_indices = np.nonzero(wavelengths_in)[0]
```
nmearl:
```suggestion
        temp_indices = np.nonzero(region_mask)[0]
```

specutils/manipulation/utils.py
```diff
@@ -34,56 +100,74 @@ def linear_exciser(spectrum, region):
...
-    wavelengths_in = (wavelengths >= region.lower) & (wavelengths < region.upper)
-    inclusive_indices = np.nonzero(wavelengths_in)[0]
-
-    #
-    # Now set the flux values for these indices to be a
-    # linear range
-    #
-
-    s, e = max(inclusive_indices[0]-1, 0), min(inclusive_indices[-1]+1,
+    wavelengths = spectrum.spectral_axis.copy()
```
nmearl:
```suggestion
    spectral_axis = spectrum.spectral_axis.copy()
```

specutils/manipulation/utils.py
```diff
@@ -34,56 +100,74 @@ def linear_exciser(spectrum, region):
...
+
+        # Add the uncertainty of the two linear interpolation endpoints in
+        # quadrature and apply to the excised region.
+        if new_uncertainty is not None:
+            new_uncertainty[s:e] = np.sqrt(spectrum.uncertainty[s]**2 + spectrum.uncertainty[e]**2)
 
     # Return a new object with the regions excised.
-    return Spectrum1D(flux=modified_flux*spectrum.flux.unit,
+    return Spectrum1D(flux=modified_flux,
                       spectral_axis=wavelengths,
```
nmearl:
```suggestion
                      spectral_axis=spectral_axis,
```

specutils/manipulation/utils.py
```diff
@@ -34,56 +100,74 @@ def linear_exciser(spectrum, region):
...
+        # Raise a warning if the SpectralRegion has more than one subregion, since
+        # the handling for this is perhaps unexpected
+        warnings.warn("A SpectralRegion with multiple subregions was provided as "
+             "input. This may lead to undesired behavior with linear_exciser if "
+             "the subregions overlap.",
+             AstropyUserWarning)
+
+    for subregion in region:
+        # Find the indices of the spectral_axis array corresponding to the subregion
+        wavelengths_in = (wavelengths >= subregion.lower) & (wavelengths < subregion.upper)
```
nmearl:
```suggestion
        region_mask = (spectral_axis >= subregion.lower) & (spectral_axis < subregion.upper)
```

specutils/manipulation/utils.py
```diff
@@ -34,56 +100,74 @@ def linear_exciser(spectrum, region):
...
-def excise_regions(spectrum, regions, exciser=linear_exciser):
+def excise_regions(spectrum, regions, exciser=true_exciser):
     """
-    Apply a convolution based smoothing to the spectrum. The kernel must be one
-    of the 1D kernels defined in `astropy.convolution`.
-
-    This method can be used along but also is used by other specific methods below.
+    Method to replace the flux in the defined regions of the spectrum with
+    interpolated values produced by the function given in the ``exciser``
+    argument.
```
nmearl: This isn't particularly true in the case of a `true_exciser`. Perhaps something like

```suggestion
    Method to remove or replace the flux in the defined regions of the spectrum 
    depending on the function provided in the ``exciser`` argument.
```

specutils/manipulation/utils.py
```diff
@@ -34,56 +100,74 @@ def linear_exciser(spectrum, region):
...
     spectrum : `~specutils.Spectrum1D`
-        The `~specutils.Spectrum1D` object to which the smoothing will be applied.
+        The `~specutils.Spectrum1D` object to which the excision will be applied.
 
     regions : list of `~specutils.SpectralRegion`
         Each element of the list is a `~specutils.SpectralRegion`. The flux
-        between these wavelengths will be "cut out" using the ``exciser``
-        method.
+        between the lower and upper wavelength of each region will be "cut out"
+        and replaced with interpolated values using the ``exciser`` method.
```
nmearl:
```suggestion
        between the lower and upper spectral axis value of each region will be "cut out"
        and optionally replaced with interpolated values using the ``exciser`` method.
```

specutils/manipulation/utils.py
```diff
@@ -34,56 +100,74 @@ def linear_exciser(spectrum, region):
...
+        Note that non-overlapping regions should be provided as separate
+        `~specutils.SpectralRegion` objects in this list, not as sub-regions
+        in a single object in the list.
 
     exciser: method
```
nmearl:
```suggestion
    exciser : function
```

specutils/manipulation/utils.py
```diff
@@ -107,25 +191,28 @@ def excise_regions(spectrum, regions, exciser=linear_exciser):
...
-def excise_region(spectrum, region, exciser=linear_exciser):
+def excise_region(spectrum, region, exciser=true_exciser):
     """
-    Apply a convolution based smoothing to the spectrum. The kernel must be one
-    of the 1D kernels defined in `astropy.convolution`.
-
-    This method can be used along but also is used by other specific methods below.
+    Method to replace the flux in the defined region of the spectrum with
+    interpolated values produced by the function given in the ``exciser``
+    argument.
```
nmearl:
```suggestion
    Method to remove or replace the flux in the defined regions of the spectrum 
    depending on the function provided in the ``exciser`` argument.
```

specutils/tests/test_manipulation.py
```diff
@@ -9,8 +9,31 @@
...
 from ..spectra import Spectrum1D, SpectralRegion, SpectrumCollection
-from ..manipulation import snr_threshold
-
+from ..manipulation import snr_threshold, excise_regions, linear_exciser
+
+def test_true_exciser():
+    np.random.seed(84)
+    spectral_axis = np.linspace(5000,5100,num=100)*u.AA
+    flux = (np.random.randn(100) + 3) * u.Jy
+    spec = Spectrum1D(flux=flux, spectral_axis = spectral_axis)
```
nmearl:
```suggestion
    spec = Spectrum1D(flux=flux, spectral_axis=spectral_axis)
```

PR #610, opened 2020-03-18
Implement SpectralCoord in Spectrum1D

rosteen:


docs/spectrum1d.rst
```diff
@@ -29,6 +29,9 @@ create it explicitly from arrays or `~astropy.units.Quantity` objects:
     >>> ax.set_xlabel("Dispersion")  # doctest: +SKIP
     >>> ax.set_ylabel("Flux")  # doctest: +SKIP
 
+Note that the spectral_axis can also be provided as a :class:`~specutils.SpectralCoord` object, 
+and in fact will internally convert the spectral_axis to :class:`~specutils.SpectralCoord` if it
+is provided as an array or `~astropy.units.Quantity`.
```
nmearl: Let's make this stand out a little more. 😉 

```suggestion
.. note::
    Note that the ``spectral_axis`` can also be provided as a :class:`~specutils.SpectralCoord` object, 
    and in fact will internally convert the spectral_axis to :class:`~specutils.SpectralCoord` if it
    is provided as an array or `~astropy.units.Quantity`.
```

specutils/spectra/spectrum1d.py
```diff
@@ -72,6 +73,11 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
             elif flux.isscalar:
                 flux = u.Quantity([flux])
 
+        # Ensure that only one or neither of these parameters is set
+        if redshift is not None and radial_velocity is not None:
+            raise ValueError('cannot set both radial_velocity and redshift at '
+                             'the same time.')
```
```suggestion
            raise ValueError("Cannot set both radial_velocity and redshift at "
                             "the same time.")
```

specutils/spectra/spectrum1d.py
```diff
@@ -95,13 +101,29 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
         # parse a given wcs. This is put into a GWCS object to
         # then be used behind-the-scenes for all specutils operations.
         if spectral_axis is not None:
-            # Ensure that the spectral axis is an astropy quantity
+            # Ensure that the spectral axis is an astropy Quantity
             if not isinstance(spectral_axis, u.Quantity):
```
nmearl: We're going to need to check if the instance might also be a `SpectralCoord` (otherwise we'll bail here and never continue on).

```suggestion
            if not isinstance(spectral_axis, (u.Quantity, SpectralCoord)):
```

specutils/spectra/spectrum1d.py
```diff
@@ -95,13 +101,29 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
...
-                raise ValueError("Spectral axis must be a `Quantity` object.")
+                raise ValueError("Spectral axis must be a `Quantity` or `SpectralCoord` object.")
+
+            # If spectral axis is provided as an astropy Quantity, convert it
+            # to a specutils SpectralCoord object.
+            if not isinstance(spectral_axis, SpectralCoord):
```
nmearl: To be more inline with the comment and because it's more explicitly what we want to check (i.e. that it's a quantity).

```suggestion
            if isinstance(spectral_axis, u.Quantity):
```

specutils/spectra/spectrum1d.py
```diff
@@ -95,13 +101,29 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
...
+                        radial_velocity=radial_velocity, doppler_rest=rest_value,
+                        doppler_convention=velocity_convention)
+            # If a SpectralCoord object is provided, we assume it doesn't need
+            # information from other keywords added
+            else:
+                for a in [radial_velocity, redshift]:
+                    if a is not None:
+                        raise ValueError("Cannot separately set redshift or radial_velocity if "
+                                        "a SpectralCoord object is input to spectral_axis")
+                        break
```
nmearl: Raising an error with halt everything, so no need to break.

```suggestion
                        
```

specutils/spectra/spectrum1d.py
```diff
@@ -95,13 +101,29 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
...
+                self._spectral_coord = SpectralCoord(spectral_axis, redshift=redshift,
+                        radial_velocity=radial_velocity, doppler_rest=rest_value,
+                        doppler_convention=velocity_convention)
+            # If a SpectralCoord object is provided, we assume it doesn't need
+            # information from other keywords added
+            else:
+                for a in [radial_velocity, redshift]:
+                    if a is not None:
+                        raise ValueError("Cannot separately set redshift or radial_velocity if "
+                                        "a SpectralCoord object is input to spectral_axis")
```
```suggestion
                                         "a SpectralCoord object is input to spectral_axis")
```

specutils/spectra/spectrum1d.py
```diff
@@ -138,16 +160,16 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
...
         # requires that the flux be initialized
-
-        if redshift is None:
-            self.radial_velocity = radial_velocity
-        elif radial_velocity is None:
-            self.redshift = redshift
-        else:
-            raise ValueError('cannot set both radial_velocity and redshift at '
-                             'the same time.')
+        self.radial_velocity = self._spectral_coord.radial_velocity
```
nmearl: This shouldn't be necessary.

```suggestion
        
```

specutils/tests/test_spectrum1d.py
```diff
@@ -93,6 +93,15 @@ def test_create_implicit_wcs_with_spectral_unit():
     assert pix2world.size == 5
     assert isinstance(pix2world, np.ndarray)
 
+def test_create_with_spectral_coord():
+
+    spectral_coord = SpectralCoord(np.arange(5100, 5150)*u.AA, radial_velocity = u.Quantity(1000.0, "km/s"))
```
nmearl:
```suggestion
    spectral_coord = SpectralCoord(np.arange(5100, 5150)*u.AA, radial_velocity=u.Quantity(1000.0, "km/s"))
```

specutils/tests/test_spectrum1d.py
```diff
@@ -93,6 +93,15 @@ def test_create_implicit_wcs_with_spectral_unit():
...
+    flux = np.random.randn(50)*u.Jy
+    spec = Spectrum1D(spectral_axis = spectral_coord, flux = flux)
```
nmearl:
```suggestion
    spec = Spectrum1D(spectral_axis = spectral_coord, flux=flux)
```

specutils/spectra/spectrum1d.py
rosteen: Won't isinstance([a SpectralCoord], u.Quantity) return True, since SpectralCoord is a subclass of Quantity? Seemed to work when I was testing, and thus I thought this was cleaner. Plus (in regards to your next comment) we can then check to see if the Quantity is specifically a SpectralCoord, and create one if it's not.

nmearl: > Won't isinstance([a SpectralCoord], u.Quantity) return True, since SpectralCoord is a subclass of Quantity?

Ah, yep. Neglected to think about the fact that we're inheriting from `Quantity`. You can ignore this suggestion, then.

> Plus (in regards to your next comment) we can then check to see if the Quantity is specifically a SpectralCoord, and create one if it's not.

We've already checked that the spectral axis is _either_ a `Quantity` _or_ a `SpectralCoord` a few lines up, so as soon as we defined it as one, then it isn't the other, so we know to create one.

rosteen: Implemented the rest of the suggestions, can this conversation be resolved since we're leaving as-is?

nmearl: Yep!

specutils/spectra/spectrum1d.py
rosteen: Not implementing since, as mentioned above, checking on u.Quantity won't discriminate between Quantity and SpectralCoord

specutils/utils/wcs_utils.py
```diff
@@ -550,7 +550,14 @@ def gwcs_from_array(array):
     Create a new WCS from provided tabular data. This defaults to being
     a GWCS object.
     """
-    array = u.Quantity(array)
+    orig_array = u.Quantity(array)
+
+    # TODO: Input arrays must be strictly ascending. This is not always the
+    #  case for a spectral axis (e.g. when in frequency space). Thus, we
+    #  convert to wavelength to create the wcs.
```
jdavies-st: Maybe add the note here that the reason it has to be ascending is because the way that `astropy.modeling.tabular.Tabular1D` is coded up to use `scipy.interpolate`.  The `inverse` for `Tabular1D` already handles this problem, but the forward should as well.  See https://github.com/astropy/astropy/blob/d6a069ee3e5831b58a83ce3e0a5bbf6a7aeaef4f/astropy/modeling/tabular.py#L238-L248

PR #613, opened 2020-03-19
Change log update

rosteen:
Updated the change log for the 1.0 release. Will probably need a few formatting fixes, but I _think_ I got all the merged PR's.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,12 +1,18 @@
-1.0 (2020-03-27)
-----------------
+1.0
+---
 
 New Features
 ^^^^^^^^^^^^
 
-- SpectrumCollection objects now support len(). [#575]
+- Implement ``SpectralCoord`` object. [#524] 
 
-- Improved JWST data loaders. [#579]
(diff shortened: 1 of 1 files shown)


CHANGES.rst
```diff
@@ -1,3 +1,45 @@
...
+
+- Implement ``SpectralCoord`` object. [#524] 
+
+- Implement cross-correlation for finding redshift/radial velocity. [#544]
+
+- Improve FITS file identification in default_loaders. [#545]
+
+- Support ``len()`` for ``SpectrumCollection`` objects. [#575]
+
+- Improve JWST data loaders. [#579]
```
jdavies-st: Could you add #595 to this as well?  I think it will be merged before 1.0 goes out.

nmearl: I think this could be split into two entries:

```suggestion
- Improved 1D JWST loader and allow parsing into an ``SpectrumCollection`` object. [#579]

- Implemented 2D and 3D data loaders for JWST products. [#595] 
```

PR #614, opened 2020-03-20
Change implementation of .quantity to .view

eteq:
This makes a small change to how `SpectralCoord.quantity` is implemented to use the more standard `view` `ndarray` method.  

It may be we should completely dispense with `.quantity` in favor of just using ``self.view(u.Quantity)`` wherever it's needed...?

### DIFF
--- a/specutils/spectra/spectral_coordinate.py
+++ b/specutils/spectra/spectral_coordinate.py
@@ -262,8 +262,8 @@ class SpectralCoord(u.Quantity):
 
         Returns
         -------
-        quantity
-            This object viewed as a `Quantity`
+        `~astropy.units.Quantity`
+            This object viewed as a `~astropy.units.Quantity`.
 
         """
         return self.view(u.Quantity)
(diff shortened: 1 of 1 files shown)


specutils/spectra/spectral_coordinate.py
```diff
@@ -244,12 +244,15 @@ def _copy(self, **kwargs):
     def quantity(self):
         """
         Convert the ``SpectralCoord`` to a simple ``~astropy.units.Quantity``.
+        Equivalent to ``self.view(u.Quantity)``.
 
         Returns
         -------
+        quantity
```
nmearl:
```suggestion
        `~astropy.units.Quantity`
```

specutils/spectra/spectral_coordinate.py
```diff
@@ -244,12 +244,15 @@ def _copy(self, **kwargs):
...
+            This object viewed as a `Quantity`
```
nmearl:
```suggestion
            This object viewed as a `~astropy.units.Quantity`.
```

eteq:
(Discussion thread with @nmearl  suppressed by my accepting a suggestion) Point taken that this is a bit more convenient than `.view(u.Quantity)`, so we'll just leave it here

PR #618, opened 2020-03-23
Deprecate spectral_axis_unit property

### LINKED ISSUE #582: Deprecate and remove spectral_axis_unit
Currently the `spectral_axis_unit` property exists and sometimes causes trouble (see  #581).  It's not really clear that there's any reason for this property to exist.  We should instead transition users to just do `spectral_axis.unit`, since a `spectral_axis` already *is* 

This property is used in a few places in `specutils`, but it should be straightofrward to fix those.  But in case it's also in use "in the wild" we should probably deprecate it for at least a version or so before removing it entirely.

cc @nmearl

nmearl:
Resolves #582. Removes references to `spectral_axis_unit` in analysis and manipulation functions, and adds a deprecation warning to the property.

### DIFF
--- a/specutils/spectra/spectrum_mixin.py
+++ b/specutils/spectra/spectrum_mixin.py
@@ -65,7 +65,7 @@ class OneDSpectrumMixin:
         return self._spectral_axis
 
     @property
-    @deprecated('v1.0', alternative="spectral_axis.unit")
+    @deprecated('v1.1', alternative="spectral_axis.unit")
     def spectral_axis_unit(self):
         """
         Returns the units of the spectral axis.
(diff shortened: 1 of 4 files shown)


specutils/spectra/spectrum_mixin.py
```diff
@@ -65,6 +65,7 @@ def spectral_axis(self):
         return self._spectral_axis
 
     @property
+    @deprecated('v1.0', alternative="spectral_axis.unit")
```
eteq, line 68: Shouldn't this be either:
```suggestion
    @deprecated('v1.1', alternative="spectral_axis.unit")
```
or
```suggestion
    @deprecated('v1.0.1', alternative="spectral_axis.unit")
```

Since 1.0 is already out without the deprecation in there.

eteq:
There's something strange going on here in CI-land, but I just ran the tests locally and they all seem to be passing (including the docs which was failing before), so I'm going to go ahead and merge this.

Note that it's also updating the current version number, which @nmearl noticed separately while updating the deprecation version.

PR #619, opened 2020-03-23
Implement SpectralCoord in SpectrumCollection

### LINKED ISSUE #616: Integrate SpectralCoord with SpectralCollection
SpectralCoord has been implemented and is used in Spectrum1D, this task is to use it in SpectralCollection. Should be a fairly straight forward drop-in.

### LINKED ISSUE #622: Have SpectralCoord carry over observer/target/rv info when copied via initializer
As @rosteen pointed out in #616, `SpectralCoord` does not carry over its metadata when copying.  See the following:
```
>>> obscoo = specutils.SpectralCoord([1,2]*u.angstrom, observer=coordinates.EarthLocation.of_site('example_site').get_gcrs(obstime=Time('2018-1-1')))
>>> obscoo.observer
<GCRS Coordinate (obstime=2018-01-01 00:00:00.000, obsgeoloc=(0., 0., 0.) m, obsgeovel=(0., 0., 0.) m / s): (ra, dec, distance) in (deg, deg, m)
    (100.24714682, 51.31007018, 6365136.14550845)
 (pm_ra_cosdec, pm_dec, radial_velocity) in (mas / yr, mas / yr, km / s)
    (2.96840133e+11, 8.03090516e+08, -1.58045408e-08)>
>>> obscoo2 = specutils.SpectralCoord(obscoo)
>>> print(obscoo2.observer)
None
```
that should be fixed so that all the "metadata" from the original `SpectralCoord` gets copied over.

rosteen:
Now converts input `spectral_axis` to `SpectralCoord` if it isn't one already. Also checks that the `spectral_axis` parameters (radial_velocity, observer, etc) in the spectra provided to `from_spectra` match, otherwise throws an error.

EDIT by @eteq: this closes #616 (adding this note to the description will make it auto-close when we merge this PR)

EDIT2 by @rosteen: also closes #622

### DIFF
--- a/specutils/spectra/spectral_coordinate.py
+++ b/specutils/spectra/spectral_coordinate.py
@@ -69,6 +69,20 @@ class SpectralCoord(u.Quantity):
         if isinstance(value, u.Quantity) and unit is None:
             obj._unit = value.unit
 
+        # If we're initializing from an existing SpectralCoord, keep any
+        # parameters that aren't being overridden
+        if isinstance(value, SpectralCoord):
+            if observer is None:
+                observer = value.observer
+            if target is None:
+                target = value.target
+            if radial_velocity is None and redshift is None:
+                radial_velocity = value.radial_velocity
+            if doppler_rest is None:
+                doppler_rest = value.doppler_rest
+            if doppler_convention is None:
+                doppler_convention = value.doppler_convention
+
         # Store state about whether the observer and target were defined
         #  explicitly (True), or implicity from rv/redshift (False)
         obj._frames_state = dict(observer=observer is not None,
--- a/specutils/spectra/spectral_coordinate.py
+++ b/specutils/spectra/spectral_coordinate.py
@@ -243,7 +243,7 @@ class SpectralCoord(u.Quantity):
     @property
     def quantity(self):
         """
-        Convert the ``SpectralCoord`` to a simple `~astropy.units.Quantity`.
+        Convert the ``SpectralCoord`` to a `~astropy.units.Quantity`.
 
         Returns
         -------
--- a/specutils/spectra/spectral_coordinate.py
+++ b/specutils/spectra/spectral_coordinate.py
@@ -243,7 +243,7 @@ class SpectralCoord(u.Quantity):
     @property
     def quantity(self):
         """
-        Convert the ``SpectralCoord`` to a simple ``~astropy.units.Quantity``.
+        Convert the ``SpectralCoord`` to a simple `~astropy.units.Quantity`.
 
         Returns
         -------
--- a/specutils/spectra/spectrum_collection.py
+++ b/specutils/spectra/spectrum_collection.py
@@ -133,18 +133,20 @@ class SpectrumCollection:
         # Check that the spectral parameters are the same for each input
         # spectral_axis and create the multi-dimensional SpectralCoord
         sa = [x.spectral_axis for x in spectra]
-        sa_0 = spectra[0].spectral_axis
-        if not all(x.radial_velocity == sa_0.radial_velocity for x in sa) or \
-            not all(x.target == sa_0.target for x in sa) or \
-            not all(x.doppler_convention == sa_0.doppler_convention for
+        if not all(x.radial_velocity == sa[0].radial_velocity for x in sa) or \
+            not all(x.target == sa[0].target for x in sa) or \
+            not all(x.observer == sa[0].observer for x in sa) or \
+            not all(x.doppler_convention == sa[0].doppler_convention for
                     x in sa) or \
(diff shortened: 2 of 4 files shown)


specutils/spectra/spectral_coordinate.py
```diff
@@ -243,7 +243,7 @@ def _copy(self, **kwargs):
     @property
     def quantity(self):
         """
-        Convert the ``SpectralCoord`` to a simple ``~astropy.units.Quantity``.
+        Convert the ``SpectralCoord`` to a simple `~astropy.units.Quantity`.
```
eteq:
```suggestion
        Convert the ``SpectralCoord`` to a `~astropy.units.Quantity`.
```
(there's a guideline somewhere in the Astropy doc dev section to avoid words like "trivial" or "simple" because they are very context-dependent)

specutils/spectra/spectrum_collection.py
```diff
@@ -58,6 +59,10 @@ def __init__(self, flux, spectral_axis=None, wcs=None, uncertainty=None,
             if not (flux.shape == spectral_axis.shape):
                 raise ValueError("Shape of all data elements must be the same.")
 
+            # Convert to SpectralCoord if simple Quantity was input
+            if not isinstance(spectral_axis, SpectralCoord):
+                spectral_axis = SpectralCoord(spectral_axis)
```
eteq:
```suggestion
            spectral_axis = SpectralCoord(spectral_axis)
```
The ``if`` is unnecessary if #622 is fixed, and in fact this is the primary motivation for a "copy as-is" initializer like this.

So if you think you can roll in a fix to #622 easily here, @rosteen, you might do that and then you can accept the above suggestion.  But as an alternative, maybe just:
```suggestion
            #TODO: change this to just ``spectral_axis = SpectralCoord(spectral_axis)`` when #622 is fixed 
            if not isinstance(spectral_axis, SpectralCoord):
                spectral_axis = SpectralCoord(spectral_axis)
```

specutils/spectra/spectrum_collection.py
```diff
@@ -124,7 +129,24 @@ def from_spectra(cls, spectra):
...
+        sa = [x.spectral_axis for x in spectra]
+        if not all(x.radial_velocity == sa[0].radial_velocity for x in sa) or \
+            not all(x.target == sa[0].target for x in sa) or \
+            not all(x.observer == sa[0].observer for x in sa) or \
+            not all(x.doppler_convention == sa[0].doppler_convention for
+                    x in sa) or \
+            not all(x.doppler_rest == sa[0].doppler_rest for x in sa):
+                raise ValueError("All input spectral_axis SpectralCoord "
+                                 "objects must have the same parameters.")
+        spectral_axis = SpectralCoord(sa,
```
eteq: Two problems with the part from here down:
1. In principle it should be OK for an RV/observer/target to be multi-dimensional, matching the dimension of the spectral collection. But that requires a lot of work so probably we should just make an issue and do a follow-on PR?
2. The fact that the "first" one is used should be in the docs (could be just the docstring for now)

specutils/spectra/spectrum_collection.py
rosteen: I'll see if I can roll in a fix for 622 today, don't think it should be difficult

specutils/spectra/spectrum_collection.py
rosteen: I'll add a note to the docstring for now and make another issue for this.

specutils/spectra/spectrum_collection.py
nmearl: @rosteen if you're not able to get #622 implemented, I've also already started (but not finished) an implementation, so feel free to keep it as a separate issue that we can work on later if it's too time-consuming!

rosteen: I have it working, unless I'm missing something more complex that's required. Just coding a bit of test coverage before I push it here.

specutils/spectra/spectrum_collection.py
rosteen: I added a note to the docstrings that the spectral_axis parameters for each spectrum need to be identical. Given this (and the fact that an error is raised if they aren't) I don't think we need to specifically call out that the first one is used.

eteq: Hmm.  I have some misgivings about this, but they are all things that make this "less" restrictive, so best bet is to probably deal with this in a follow-on issue so as not to block this PR further.  I'll make a relevant issue.

eteq:
LGTM now, @rosteen, aside from some concerns that I think are better dealt with in a follow-on issue.  Merging!

PR #621, opened 2020-03-24
MNT: Set max-line-length for flake8

pllim:
This is an automated update made by the `batchpr` tool :robot: - feel free to close if it doesn't look good! You can report issues to @pllim.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -100,4 +100,6 @@ skip_install = true
 changedir = .
 description = check code style, e.g. with flake8
 deps = flake8
-commands = flake8 specutils --count --max-line-length=100 --select=E101,W191,W291,W292,W293,W391,E111,E112,E113,E502,E722,E901,E902
+commands =
+    -flake8 specutils -qq --statistics --select=E501,W505
+    flake8 specutils --count --select=E101,W191,W291,W292,W293,W391,E111,E112,E113,E502,E722,E901,E902
--- a/setup.cfg
+++ b/setup.cfg
@@ -85,3 +85,7 @@ exclude_lines =
     pragma: py{ignore_python_version}
     # Don't complain about IPython completion helper
     def _ipython_key_completions_
+
+[flake8]
+max-line-length = 100
+max-doc-length = 79


dhomeier:
@nmearl thoughts about adopting this? The `pycodestlye` test doesn't seem to mind about the line lengths, but it would be convenient not to have flake8 complaining while editing source files.

dhomeier:
@pllim , I just checked that the literal take of PEP 8 on this is
> For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters.

flake8 by default appears to use the same length for doc lines as in `max-line-length`, so to strictly honour the above, one would need to set in addition
```
max-doc-length = 72
```
which seems excessively restrictive (and it's perhaps not 100 % clear if the 72 characters should include leading tabs; for flake8 they obviously do). Looks like most of Astropy still follows the 79/80 character limit in docstrings, which might be a reasonable compromise.

dhomeier:
Actually
https://github.com/nmearl/specutils/blob/0e0752a12601d649faa2617916d98720f9f6ac07/tox.ini#L103
put a manual override of the `setup.cfg` defaults in place...

pllim:
What about now? 😸

dhomeier:
Actually irrelevant for the Code style check since it does not select for line length (still too many violations, I guess). 😹

For reference, in current master I am getting
```
miranda:13373> flake8 specutils --count --select=E501 | tail -1
190
miranda:13374> flake8 specutils --count --select=W505 | tail -1
1011
miranda:13375> flake8 specutils --max-doc-length=80 --count --select=W505 | tail -1
178
miranda:13376> flake8 specutils --max-doc-length=100 --count --select=W505 | tail -1
25
```
So even if it's not tested in CI, `max-doc-length=72` might indeed be unrealistically strict.

pllim:
So, you want 100?

dhomeier:
I am not sure what I really want. ;-)
But what matters is what is agreed upon in this project for doc strings. From @eteq's https://github.com/astropy/specutils/pull/724#discussion_r518399153 I'd take that to be 80.

pllim:
Re: 80 -- If that includes line break, should set to 79 😉 

Anyways, I can put whatever number you want there, just need a decision. Or feel free to edit this branch directly too.

dhomeier:
I guess so, PEP 8 also always mentions 79 (but also has 99 for the increased `max-line-length`).
It's not really up to me to decide, since I am only a contributor here, but if I end up being the only one to whom it matters, since my flycheck-plugin is picking up those settings, I vote for `max-doc-length=79`.

tox.ini
```diff
@@ -100,4 +100,4 @@ skip_install = true
 changedir = .
 description = check code style, e.g. with flake8
 deps = flake8
-commands = flake8 specutils --count --max-line-length=100 --select=E101,W191,W291,W292,W293,W391,E111,E112,E113,E502,E722,E901,E902
+commands = flake8 specutils --count --select=E101,W191,W291,W292,W293,W391,E111,E112,E113,E502,E722,E901,E902
```
dhomeier, line 103:
```suggestion
commands =
    flake8 specutils -qq --statistics --select=E501,W505
    flake8 specutils --count --select=E101,W191,W291,W292,W293,W391,E111,E112,E113,E502,E722,E901,E902
```
Assuming we don't want to _enforce_ line lengths anytime soon, this might be helpful to keep track of how much of the code conforms (or not).

pllim: Sure, but changing this won't help your editor; this is just for the CI. Your editor only picks up `setup.cfg`.

dhomeier:
Hmm, apparently flake8 exits with warning status when reporting nonzero violations, so the return value of the first call would have to be intercepted.

pllim:
Maybe `specutils` should allow flake8 to fail until the maintainers have time to clean up the warnings.

dhomeier:
Looks like it only takes a dash to ignore the exit value
```
    -flake8 specutils -qq --statistics --select=E501,W505
```
but it's just a suggestion anyway – astropy core does not check line lengths either.

pllim:
Sure, why not. 😄

pllim:
Wow, I forgot about this PR. Can we merge or should we just close it?

pllim:
```
Err:1 http://azure.archive.ubuntu.com/ubuntu focal-updates/main amd64 language-pack-de-base all 1:20.04+20210802
  Connection failed [IP: 52.147.219.192 80]
```

Why does this look so familiar, @astrofrog ?

Issue #625, opened 2020-03-26
Having problems using specutils to write Spectrum1D object to FITS file


kakirastern:
I have tried using the approach as suggested on https://specutils.readthedocs.io/en/stable/custom_loading.html#creating-a-custom-writer to save a `Spectrum1D` object to the FITS file format, but have been encountered various problems.

The `Spectrum1D` object concerned is an integrated spectrum obtained from some data cube of a well-known planetary nebula after splicing a two spectra (one on the red side and one on the blue side) into one in order to derive some optical spectrum for the scientific target. The given `Spectrum1D` object can be specified by its `flux` and `spectral_axis` as follows: 

``` python
>>> type(final_spec)
specutils.spectra.spectrum1d.Spectrum1D

>>> final_spec.flux.value
array([1.37377823e-15, 1.42879234e-15, 1.51984429e-15, ...,
       2.73236522e-17, 3.10544382e-17, 2.67768878e-42])

>>> final_spec.spectral_axis.value
array([3500.        , 3500.77431644, 3501.54863288, ..., 6999.12060362,
       6999.55980181, 6999.999     ])
```

At first I tried the plain old `write` method without specifying the supposedly optional `format` parameter. But I got the following error with traceback:

```
---------------------------------------------------------------------------
IORegistryError                           Traceback (most recent call last)
<ipython-input-120-2f4909935542> in <module>
----> 1 final_spec.write('/Users/krisstern/Desktop/junk.fits')

~/astropy/astropy/nddata/mixins/ndio.py in __call__(self, *args, **kwargs)
     97 
     98     def __call__(self, *args, **kwargs):
---> 99         registry.write(self._instance, *args, **kwargs)
    100 
    101 

~/astropy/astropy/io/registry.py in write(data, format, *args, **kwargs)
    561 
    562         format = _get_valid_format(
--> 563             'write', data.__class__, path, fileobj, args, kwargs)
    564 
    565     writer = get_writer(format, data.__class__)

~/astropy/astropy/io/registry.py in _get_valid_format(mode, cls, path, fileobj, args, kwargs)
    601                               " 'format' argument.\n"
    602                               "The available formats are:\n"
--> 603                               "{}".format(format_table_str))
    604     elif len(valid_formats) > 1:
    605         raise IORegistryError(

IORegistryError: Format could not be identified based on the file name or contents, please provide a 'format' argument.
The available formats are:
    Format    Read Write Auto-identify
------------- ---- ----- -------------
[...]

Issue #584, opened 2020-02-11
Priority system in loaders no longer works


nmearl:
Loaders are assigned priorities and they should be taken into consideration when conflicts arise amongst potential loaders when a format is not explicitly give.

dhomeier:
Those are passed as kwarg to `data_loader`, right? But where are they (supposed to be) assigned? I'd expect they might be written into the `@data_loader` decorators, but can't find one that has a `priority` value or seems to have had one in recent times...

nmearl:
The loader priorty is stored [here](https://github.com/astropy/specutils/blob/master/specutils/io/registers.py#L79), but it looks like we need to be more specific about using it (before, it seemed that the order in the registry was part of the determination for which loader to choose). It seems at some point, we took them out during a refactor.

I'm imagining that we'll want to find all loaders that meet the identifier criteria. If there is more than one, simply compare priorities and return the appropriate loader.

aragilar:
This is fixed by #757.

PR #632, opened 2020-03-31
Add WCS 1D (including multi-D flux) writer in default_loaders

dhomeier:
In #625 and related discussion on Slack the lack of a builtin `writer` for `wcs1d-fits` (writing the flux data as an image array) came up again.
This PR implements such an initial generic writer for saving a `Spectrum1D` to the (primary) `IMAGE_HDU` in a FITS file.

- Writing as `wcs1d-fits` only works for `Spectrum1D` objects that have a fully functional WCS (specifically having a `spectrum.wcs.to_fits` method; it's up to the user to provide this.

- The writer also supports on-the-fly change of flux units or dtype (e.g. to save space by setting `dtype='f4'`) via kwargs; the same options have been added to `tabular-fits`, but may need some discussion (this implementation adds separate kwargs `[fw]type` for the dtypes of flux and spectral_axis, and there are no tests yet).

- `wcs1d-fits` actually happily accepts 2D flux arrays, and the resulting files can be opened e.g. in `SAOImageDS9`, but `Spectrum1D` cannot read them back in, as there is no unambiguous way to handle the WCS-constructed spectral axis. Deferring this to a separate PR or issue on handling multi-D spectra.

- Also following the discussion on Slack I have added auto-identification functionality one `write` to `tabular-fits`; from what I understood this is of limited use, as it can basically just check if the file naming scheme conforms to the rules. For `wcs1d` it would be helpful to check for a working `wcs`, but since the spectrum object is not passed to the identifier, I see no way to implement this.

### DIFF
--- a/specutils/io/default_loaders/tabular_fits.py
+++ b/specutils/io/default_loaders/tabular_fits.py
@@ -18,8 +18,12 @@ __all__ = ['tabular_fits_loader', 'tabular_fits_writer']
 
 
 def identify_tabular_fits(origin, *args, **kwargs):
-    # check if file can be opened with this reader
     # args[0] = filename
+    # check if filename conforms to naming convention for writer
+    if origin == 'write':
+        return args[0].endswith(('.fits', '.fit'))
+
+    # check if file can be opened with this reader
     with fits.open(args[0]) as hdulist:
         # Test if fits has extension of type BinTable and check against
         # known keys of already defined specific formats
@@ -36,7 +40,7 @@ def identify_tabular_fits(origin, *args, **kwargs):
 
 
 @data_loader("tabular-fits", identifier=identify_tabular_fits,
-             dtype=Spectrum1D, extensions=['fits'])
+             dtype=Spectrum1D, extensions=['fits'], priority=6)
 def tabular_fits_loader(file_obj, column_mapping=None, hdu=1, **kwargs):
     """
     Load spectrum from a FITS file.
@@ -100,9 +104,15 @@ def tabular_fits_writer(spectrum, file_name, update_header=False, **kwargs):
         The path to the FITS file
     update_header: bool
         Update FITS header with all compatible entries in `spectrum.meta`
+    wunit : str or `~astropy.units.Unit`
+        Unit for the spectral axis (wavelength or frequency-like)
+    funit : str or `~astropy.units.Unit`
+        Unit for the flux (and associated uncertainty)
+    wtype : str or `~numpy.dtype`
+        Floating point type for storing spectral axis array
+    ftype : str or `~numpy.dtype`
+        Floating point type for storing flux array
     """
-    flux = spectrum.flux
-    disp = spectrum.spectral_axis
     header = spectrum.meta.get('header', fits.header.Header()).copy()
 
     if update_header:
@@ -115,16 +125,28 @@ def tabular_fits_writer(spectrum, file_name, update_header=False, **kwargs):
     for keyword in ['NAXIS', 'NAXIS1', 'NAXIS2']:
         header.remove(keyword, ignore_missing=True)
 
+    # Add dispersion array and unit
+    wtype = kwargs.pop('wtype', spectrum.spectral_axis.dtype)
+    wunit = u.Unit(kwargs.pop('wunit', spectrum.spectral_axis.unit))
+    disp = spectrum.spectral_axis.to(wunit, equivalencies=u.spectral())
+
     # Mapping of spectral_axis types to header TTYPE1
-    dispname = disp.unit.physical_type
+    dispname = wunit.physical_type
     if dispname == "length":
         dispname = "wavelength"
 
-    columns = [disp, flux]
+    # Add flux array and unit
+    ftype = kwargs.pop('ftype', spectrum.flux.dtype)
+    funit = u.Unit(kwargs.pop('funit', spectrum.flux.unit))
+    flux = spectrum.flux.to(funit, equivalencies=u.spectral_density(disp))
+
+    columns = [disp.astype(wtype), flux.astype(ftype)]
     colnames = [dispname, "flux"]
+
     # Include uncertainty - units to be inferred from spectrum.flux
     if spectrum.uncertainty is not None:
-        columns.append(spectrum.uncertainty.quantity)
+        unc = spectrum.uncertainty.quantity.to(funit, equivalencies=u.spectral_density(disp))
+        columns.append(unc.astype(ftype))
         colnames.append("uncertainty")
 
     # For 2D data transpose from row-major format
--- a/specutils/io/default_loaders/tabular_fits.py
+++ b/specutils/io/default_loaders/tabular_fits.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import _io
 
 import numpy as np
 
@@ -18,28 +19,41 @@ __all__ = ['tabular_fits_loader', 'tabular_fits_writer']
 
 def identify_tabular_fits(origin, *args, **kwargs):
     # args[0] = filename
-    # check if filename conforms to naming convention for writer
+    hdu = kwargs.get('hdu', 1)
+    # Check if filename conforms to naming convention and writer has not been
+    # asked to write to primary (IMAGE-only) HDU
     if origin == 'write':
-        return args[0].endswith(('.fits', '.fit'))
+        return (hdu > 0 and args[0].endswith(('.fits', '.fit')) and not
+                args[0].endswith(('wcs.fits', 'wcs1d.fits', 'wcs.fit')))
 
-    # check if file can be opened with this reader
-    with fits.open(args[0]) as hdulist:
-        # Test if fits has extension of type BinTable and check against
-        # known keys of already defined specific formats
-        return (len(hdulist) > 1 and
-                isinstance(hdulist[1], fits.BinTableHDU) and not
-                (fits.getheader(args[0]).get('TELESCOP') == 'MULTI' and
-                 fits.getheader(args[0]).get('HLSPACRN') == 'MUSCLES' and
-                 fits.getheader(args[0]).get('PROPOSID') == 13650) and not
-                (fits.getheader(args[0]).get('TELESCOP') == 'SDSS 2.5-M' and
-                 fits.getheader(args[0]).get('FIBERID') > 0) and not
-                (fits.getheader(args[0]).get('TELESCOP') == 'HST' and
-                 fits.getheader(args[0]).get('INSTRUME') in ('COS', 'STIS')) and not
-                fits.getheader(args[0]).get('TELESCOP') == 'JWST')
+    if isinstance(args[2], fits.hdu.hdulist.HDUList):
+        hdulist = args[2]
+    elif isinstance(args[2], _io.BufferedReader):
+        hdulist = fits.open(args[2])
+    else:  # if isinstance(args[2], _io.FileIO):
+        hdulist = fits.open(args[0], **kwargs)
+
+    # Test if fits has extension of type BinTable and check against
+    # known keys of already defined specific formats
+    is_tab = (len(hdulist) > 1 and
+              isinstance(hdulist[hdu], fits.BinTableHDU) and not
+              (hdulist[0].header.get('TELESCOP') == 'MULTI' and
+               hdulist[0].header.get('HLSPACRN') == 'MUSCLES' and
+               hdulist[0].header.get('PROPOSID') == 13650) and not
+              (hdulist[0].header.get('TELESCOP') == 'SDSS 2.5-M' and
+               hdulist[0].header.get('FIBERID') > 0) and not
+              (hdulist[0].header.get('TELESCOP') == 'HST' and
+               hdulist[0].header.get('INSTRUME') in ('COS', 'STIS')) and not
+              hdulist[0].header.get('TELESCOP') == 'JWST')
+
+    if not isinstance(args[2], (fits.hdu.hdulist.HDUList, _io.BufferedReader)):
+        hdulist.close()
+
+    return is_tab
 
 
 @data_loader("tabular-fits", identifier=identify_tabular_fits,
-             dtype=Spectrum1D, extensions=['fits'], priority=6)
+             dtype=Spectrum1D, extensions=['fits', 'fit'], priority=6)
 def tabular_fits_loader(file_obj, column_mapping=None, hdu=1, **kwargs):
     """
     Load spectrum from a FITS file.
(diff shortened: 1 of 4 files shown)


dhomeier:
Cannot really consistently reproduce the doc build failure (or the warning for that matter) locally:
```
Warning, treated as error:
1426 /home/travis/build/astropy/specutils/docs/api/specutils.SpectralCoord.rst:42:<autosummary>:1:py:obj reference target not found: Quantity
1427 Sphinx Documentation subprocess failed with return code 2
```
It initially showed up with Python3.7 + Sphinx 2.4.4, but not with Python3.8 + Sphinx 2.2.1.; after updating the latter to Sphinx 2.4.4 as well, the warning once popped up as well, but never since (always removing the `_build` and `api` subdirs in between)...

dhomeier:
Worked out a way how to load 2D spectra on a common spectral axis (i.e. a range of spectra in the 1. dimension with the 2. dimension the spectral axis).
Auto-identification for 2D is not enabled, since it creates ambiguous matches with most other wcs-based loaders (I tried to resolve this by assigning them different priorities, but that seems to be broken at a deeper level as indicated in #584).

dhomeier:
Worked out a way how to load 2D spectra on a common spectral axis (i.e. a range of spectra in the 1. dimension with the 2. dimension the spectral axis).
Auto-identification for 2D is not enabled, since it creates ambiguous matches with most other wcs-based loaders (I tried to resolve this by assigning them different priorities, but that seems to be broken at a deeper level as indicated in #584.

dhomeier:
Seems the way the reader handles additional dimensions could be extended _ad infinitum_; included cases up to `ndim`/`NAXIS` = 4 now with the tests, I don't know if there are general use cases or interest in higher dimensions.

dhomeier:
I have rebased since #637 introduced a number of non-trivial merge conflicts.

Main open questions remaining to be settled on my part are

1. Should the `flux` dimension `wcs1d_fits_loader` accepts remain limited to 4, any other number, or allow unlimited dimensions?

2. Are the naming conventions of the `unit` and `dtype` keywords for the writers sensible?

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -85,9 +90,80 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
...
+        message = "Saving uncertainties in wcs1d format is not yet supported!"
+        warnings.warn(message, AstropyUserWarning)
+
+    # Add flux array and unit
+    ftype = kwargs.pop('dtype', spectrum.flux.dtype)
+    funit = u.Unit(kwargs.pop('unit', spectrum.flux.unit))
+    flux = spectrum.flux.to(funit, equivalencies=u.spectral_density(wl))
+    hdulist[0].data = flux.value.astype(ftype)
+
+    if hasattr(funit, 'long_names') and len(funit.long_names) > 0:
```
nmearl, line 158: Just a clarification: does `long_names` give something different from `physical_type`? In my (limited testing), `physical_type` == `long_names[1]`.

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -85,9 +90,80 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
...
+                try:
+                    wcs = wcs.dropaxis(i)
+                except(_wcs.NonseparableSubimageCoordinateSystemError) as e:
+                    raise ValueError(f'WCS cannot be reduced to 1D: {e} {wcs}')
+
     return Spectrum1D(flux=data, wcs=wcs, meta=meta)
 
 
+@custom_writer("wcs1d-fits")
+def wcs1d_fits_writer(spectrum, file_name, update_header=False, **kwargs):
```
nmearl, line 106: I think we might want to rename this to something like `image-fits`/`image_fits_writer` (analogous to tabular) -- `wcs1d` should be implicit it because it's on a `Spectrum1D` writer, and I think we might want to express explicitly that it's writing to an images vs a table.

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -85,9 +90,80 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
...
+    try:
+        wcs = spectrum.wcs
+        hdulist = wcs.to_fits()
+        header = hdulist[0].header
+    except AttributeError as err:
+        raise ValueError(f'Only Spectrum1D objects with valid WCS can be written as wcs1d: {err}')
+
+    # Verify spectral axis constructed from WCS
+    wl = spectrum.spectral_axis
+    dwl = (wcs.all_pix2world(np.arange(len(wl)), 0) - wl.value) / wl.value
```
nmearl, line 133: `all_pix2world` seems to be a FITS-specific method -- I can easily imagine that users are going to want to write out their spectrum loaded from e.g. an ascii file (and therefore have a gwcs), so it seems this writer is only useful for `Spectrum1D` objects that are built with FITS wcs's?

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -85,9 +90,80 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
...
+        hdr_types = (str, int, float, complex, bool,
+                     np.floating, np.integer, np.complexfloating, np.bool_)
+        header.update([keyword for keyword in spectrum.meta.items() if
+                       (isinstance(keyword[1], hdr_types) and
+                        keyword[0] not in ('NAXIS', 'NAXIS1', 'NAXIS2'))])
+
+    # Cannot include uncertainty in IMAGE_HDU - maybe provide option to
+    # separately write this to BINARY_TBL extension later.
+    if spectrum.uncertainty is not None:
+        message = "Saving uncertainties in wcs1d format is not yet supported!"
```
nmearl, line 149: This seems a bit extraneous? Perhaps just place the message in the `warnings.warn` call?

specutils/io/default_loaders/wcs_fits.py
dhomeier: `long_names` can include any (alias) names the unit was given on initialisation; it rarely (if ever) includes `physical_type`. E.g.
```
>>> u.Jy.long_names
['Jansky', 'jansky']
>>> u.Jy.physical_type
'spectral flux density'
```
so `comment` would become `'[Jansky] spectral flux density'` for `BUNIT = Jy` (`u.Jy.to_string()`).

specutils/io/default_loaders/wcs_fits.py
dhomeier: Technically there are two characteristic properties – storing the flux spectrum as a 1D image, and encoding the spectral axis as `wcs.WCS`. But I agree the first one is probably more informative to the user, and this might make the documentation on the different loader types clearer or easier to write.

specutils/io/default_loaders/wcs_fits.py
dhomeier: More generally it is a method of `astropy.wcs`'s `WCS`, but that is based on the FITS model and constructed from a FITS header. This is already tested with the `wcs.to_fits()` call above, but the error message could perhaps state this more specifically (I am afraid it will still potentially confuse users why their `spectrum.wcs` does not qualify).
Generally the `gwcs.WCS`, at least in the use cases here, primarily seems to function as a container for an array of `spectral_axis` values, so it cannot be directly used here. For linear, equidistant spectral axes it should be relatively straightforward to construct an `astropy.wcs.WCS` that encodes this `spectral_axis`, but I don't think this should be part of a loader - it would better fit into `utils.wcs_utils`.

specutils/io/default_loaders/wcs_fits.py
dhomeier: Probably just adopted from some other example and/or to stay within 80 characters, but if #621 is adopted, that won't matter anymore.

specutils/io/default_loaders/wcs_fits.py
nmearl: Ah, sorry, I got my lines (of code) crossed: I meant the does `long_names` give something different from `to_string`; but was testing with units that seem to be inconsistent (e.g. Angstroms gives `['AA', 'angstrom']` for long names, and `angstrom` for `to_string()`, but centimeters gives `['centimeter']` and `cm` respectively).

dhomeier: Yes, `short_names` and `long_names` for `u.AA` don't seem to give terribly intuitive results...

specutils/io/default_loaders/wcs_fits.py
dhomeier: #676 brought up another issue with how and where to handle WCS properties:
One straightforward fix would not work for some of our example files which do not include a canonical wavelength unit definition at all, e.g.  `L5g_0355+11_Cruz09.fits`.

As I see the current situation, `Spectrum1D` no longer accepts `flux` or `spectral_axis` without units, and the `gwcs` also always has one, but using the `wcs.WCS` still allows the loader to sneak around this requirement.

So my proposal would be to add some additional checks for a spectral unit in this loader and assign a default unit, if none is found (which generally will be `Angstrom` for FITS spectra) – either in this PR or a follow-up. But the question remains if this verification should be left in general to the loader, or if `Spectrum1D` should not also check this if initialised with `wcs` and no `spectral_axis`.

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -1,36 +1,44 @@
...
+from ..registers import data_loader, custom_writer
 
 __all__ = ['wcs1d_fits_loader', 'wcs1d_fits_writer', 'non_linear_wcs1d_fits']
 
 
 def identify_wcs1d_fits(origin, *args, **kwargs):
-    # check if file can be opened with this reader
     # args[0] = filename
+    # check if filename conforms to naming convention for writer
+    if origin == 'write':
```
nmearl, line 22: We might want the identifiers to also check what type of file object `args[0]` is in the case that it might be an already existing `HDUList`.

specutils/io/default_loaders/wcs_fits.py
nmearl: > As I see the current situation, Spectrum1D no longer accepts flux or spectral_axis without units, and the gwcs also always has one, but using the wcs.WCS still allows the loader to sneak around this requirement

This isn't strictly true: if a fits wcs is provided, the initializer will still attempt to create a `SpectralAxis` from the wcs information and fall back to a length unit if unsuccessful:

```python
In [13]: my_wcs = fitswcs.WCS(header={'CDELT1': 1, 'CRVAL1': 6562.8,
    ...:                                  'CTYPE1': 'WAVE', 'RESTFRQ': 1400000000, 'CRPIX1': 25},
    ...:                          naxis=1)
    ...: spec = Spectrum1D(flux=[5,6,7] * u.Jy, wcs=my_wcs)
WARNING: FITSFixedWarning: 'datfix' made the change 'Set DATE-REF to '1858-11-17' from MJD-REF'. [astropy.wcs.wcs]
WARNING:astropy:FITSFixedWarning: 'datfix' made the change 'Set DATE-REF to '1858-11-17' from MJD-REF'.

In [14]: spec
Out[14]:
<Spectrum1D(flux=<Quantity [5., 6., 7.] Jy>, spectral_axis=<SpectralAxis
   (doppler_rest=1400000000.0 Hz
    doppler_convention=None)
  [6538.8, 6539.8, 6540.8] m>)>
```

dhomeier: > This isn't strictly true: if a fits wcs is provided, the initializer will still attempt to create a `SpectralAxis` from the wcs information and fall back to a length unit if unsuccessful:

If the WCS finds a `CTYPE` it can interpret as length (or frequency). Otherwise, without `CTYPE1` or set to something like `'LINEAR'` in our example spectrum above, the default unit becomes `pixel`. But that is an accepted unit for `spectral_axis` as well (and results in a separate error when trying to call e.g. `spec.with_spectral_unit('nm')`), so I think this can be separated from the solution for #676.
Meaning I'd leave it to the loader to determine the best default unit for such a WCS.

nmearl:
This looks great other than the comment on having the identifier functions accept hdulist objects.

dhomeier:
I had half forgotten this was still pending when updating the identifier functions in #708. I am adding the hdulist handling here as well, should be straightforward to merge if the new PR needs to be rebased.

specutils/io/default_loaders/tabular_fits.py
```diff
@@ -18,8 +17,12 @@
 
 
 def identify_tabular_fits(origin, *args, **kwargs):
-    # check if file can be opened with this reader
     # args[0] = filename
+    # check if filename conforms to naming convention for writer
+    if origin == 'write':
```
nmearl, line 22: @dhomeier would mind updating the file handling here as well?

dhomeier: OK, looking at that line I first thought that you wanted to see the generalised file handling in the writing case as well. I frankly did not see an obvious way to auto-detect by content the format of a file that is yet to be written. In the case of an file handle that is opened for updating, we might assume that the format should match the previous content. Or in fact, in this particular case an hdulist that is passed for writing to the _primary_ HDU would indeed imply a `wcs` format, since tables can only be written to extensions. But currently neither writer even accepts an hdulist as destination, both only work on file paths...

That rant aside, for the `tabular_fits_loader`, yes, will do – I thought that had already been added in master at some point. But the `'write'` case also needs to be updated anyway to remove ambiguity in the naming scheme.

specutils/io/default_loaders/wcs_fits.py
dhomeier: So once I have updated both identifiers it would be probably be a good moment to go ahead with the renaming. Are you more in favour of `FITS IMAGE`, `FITS BINTABLE`, `[FITS] IRAF`; or rather `IMAGE-FITS` etc. @nmearl?

nmearl:
Looks good, thanks @dhomeier!

dhomeier:
Thanks @nmearl – a bit of explanation on the `hdu` kwargs for the writers: the overall default for writing is still `tabular-fits`, as it is the most general and flexible format. However if `hdu=0` is specified, `wcs1d` becomes the default, since the primary HDU only accepts `IMAGE` data. Currently that kwarg has no other purpose in `tabular_fits_writer`; ideally it should support writing to a HDU of choice just as `wcs1d_fits_writer` now does, but since astropy's `Table.write` does not support any HDU selection, this will require some additional work at the `io.fits` level, so it's better left to a new PR. Then an updating mode (writing only to the specific HDU in an existing file) could also be tackled.

nmearl:
> ideally it should support writing to a HDU of choice just as wcs1d_fits_writer now does, but since astropy's Table.write does not support any HDU selection, this will require some additional work at the io.fits level, so it's better left to a new PR. Then an updating mode (writing only to the specific HDU in an existing file) could also be tackled.

Can you open an issue that points this out specifically, with the idea that it'll be tackled in a future PR (and link to this PR for context)?

PR #633, opened 2020-04-01
Add 2SLAQ-LRG loader

aragilar:
This adds a loader for the LRG subset of the 2dF-SDSS LRG/QSO survey.

There's no uncertainty data contained within the fits file, and the test uses Data Central as the source of the data.

As the data includes both the science spectrum and the sky spectrum, I extracted both. I did try to create a loader which only loaded the science spectrum, which is currently commented out, is that useful, or should I remove that code.

### DIFF
--- a/specutils/io/default_loaders/twoslaq_lrg.py
+++ b/specutils/io/default_loaders/twoslaq_lrg.py
@@ -1,0 +1,115 @@
+import astropy.io.fits as fits
+from astropy.units import Unit
+from astropy.wcs import WCS
+from specutils.io.registers import data_loader
+from specutils import Spectrum1D, SpectrumList
+
+
+def identify_2slaq_lrg(origin, *args, **kwargs):
+    """
+    Identify if the current file is a 2SLAQ-LRG file
+    """
+    file_obj = args[0]
+    if isinstance(file_obj, fits.hdu.hdulist.HDUList):
+        hdulist = file_obj
+    else:
+        hdulist = fits.open(file_obj, **kwargs)
+
+    if hdulist[0].header["MSTITLE"].startswith("2dF-SDSS LRG/QSO survey"):
+        # apparently the QSO part doesn't have MSTITLE, so we should be safe
+        # with just the above condition, but just in case, we know they have
+        # a different structure (LRG has one ext, QSO has more; LRG is 3d,
+        # QSO is 1d
+        if len(hdulist) == 1 and hdulist[0].data.ndim == 3:
+            if not isinstance(file_obj, fits.hdu.hdulist.HDUList):
+                hdulist.close()
+            return True
+
+    if not isinstance(file_obj, fits.hdu.hdulist.HDUList):
+        hdulist.close()
+    return False
+
+
+@data_loader("2SLAQ-LRG", identifier=identify_2slaq_lrg,dtype=SpectrumList,
+             extensions=["fit", "fits"])
+def twoslaq_lrg_fits_loader(file_obj, **kwargs):
+    """
+    Load a file from the LRG subset of the 2dF-SDSS LRG/QSO survey (2SLAQ-LRG)
+    file.
+
+    2SLAQ was one of a number of surveys that used the 2dF instrument on the
+    Anglo-Australian Telescope (AAT) at Siding Spring Observatory (SSO) near
+    Coonabarabran, Australia, and followed up the 2QZ survey. Further details
+    can be seen at http://www.physics.usyd.edu.au/2slaq/ or at
+    https://docs.datacentral.org.au/.
+
+    The LRG and QSO data appear to be in different formats, this only loads the
+    LRG version. As there is a science and sky spectrum, the `SpectrumList`
+    loader provides both, whereas the `Spectrum1D` loader only provides the
+    science.
+
+    Parameters
+    ----------
+    file_name: str
+        The path to the FITS file
+    Returns
+    -------
+    data: SpectrumList
+        The 2SLAQ-LRG spectrum that is represented by the data in this file.
+    """
+    if isinstance(file_obj, fits.hdu.hdulist.HDUList):
+        hdulist = file_obj
+    else:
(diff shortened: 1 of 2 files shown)


dhomeier:
> As the data includes both the science spectrum and the sky spectrum, I extracted both. I did try to create a loader which only loaded the science spectrum, which is currently commented out, is that useful, or should I remove that code.

Just throwing this into the discussion, I am currently testing loading of 2D spectra with common spectral axis in #632. The 2SLAQ-LRG data seem to be 3-dim, but you might be able to load `hdulist[0].data[0:2,0]` as a 2D collection as well, so `spectrum.flux[0]` would be the science spectrum and `spectrum.flux[1]` the sky spectrum (if it is possible to bring the WCS into the right shape with something like `wcs.dropaxis(2); wcs.dropaxis(1)`, which I honestly don't know).
Not sure if this is a more sensible solution than loading them into a `SpectrumList`; the discussion how to interpret 2D spectra is still in its early stages...

aragilar:
Apparently, the LRG variant is similar to what IRAF would output (if that helps).

I'm curious what you mean by 2D spectra? To me, it implies the data has a spacial nature, rather than simply jamming the information into one HDU (which means either you have issues with the WCS when you try viewing it in ds9 or other image viewer, or you have to manually read in the WCS, and ensure that what you get is meaningful). I've been working on two loaders for Data Central (which are tied to our ingestion setup - they will be open source in the future, but currently some of our best test cases are those where the survey is not publicly released yet, hopefully around June/July), and have classified them as either "Single-Split" or "Multiline-Single"—"Single-Split" have the data spread across HDUs in 1D arrays (one for science, one for uncertainty, one for sky, one for mask etc.), which is probably a better format, as you don't have issues with the WCS, or with units (BUNIT applies to all the data, which doesn't work when you mix it); whereas "Multiline-Single" have multiple rows of data in a single HDU, which makes recording the unit problematic (e.g. data taken by the GAMA survey have both flux-calibrated and non-flux-calibrated data in the same HDU, so the units for the flux calibration are in a comment), and end up with a conflict between having a 2D WCS when the data really is 1D.

I suspect the nicest way of handing this would be to have some conventions around additional keywords in `meta` (e.g. has this data been flux-calibrated or not, is this sky/other calibration etc.) or as boolean properties on the spectrum, and then be able to associate other spectra with a specific science spectra (whether that's similar to how we do uncertainties, or simply a link via `meta`).

aragilar:
It looks like the test is failing because Angstrom is being used, is there a suggested workaround?

dhomeier:
> It looks like the test is failing because Angstrom is being used, is there a suggested workaround?

The `build_docs` test? No, that's apparently a broken intersphinx link to the `Quantity` documentation; should have been fixed by 8339d8a, so a rebase ought to get you rid it.

dhomeier:
> I'm curious what you mean by 2D spectra? To me, it implies the data has a spacial nature, rather than simply jamming the information into one HDU (which means either you have issues with the WCS when you try viewing it in ds9 or other image viewer, or you have to manually read in the WCS, and ensure that what you get is meaningful)

I don't think there is a universal convention, and I certainly have not seen all the possible variants that might be out there. Sure 2D and cubes often represent spatial data along a slit or from an IFU, but combining any set of spectra to a 2D array seems a legitimate use as well - as long as you ensure consistent wavelength scale. Of course this might involve resampling spectra if you want sub-pixel precision, so it's not ideal but might just be convenient...
Anyway that is essentially what `wcs1d-fits` is covering, where the `1d` signifies a single `spectral_axis` scale. One example I've just come upon is the [6dFGS](http://www-wfau.roe.ac.uk/6dFGS/fits.html), where variance and background flux have simply been added as 2nd and 3rd column in `NAXIS2`. But that's certainly a format that requires some external information, e.g. no `BUNIT` and all the other shortcomings you mention above.

aragilar:
Cool, rebased.

aragilar:
Is this waiting on anything, can it be merged?

nmearl:
@aragilar sorry for the late reply; it needs a rebase.

aragilar:
Rebased

aragilar:
The tests seem to be failing on an unrelated test, apparently `specutils.tests.helpers` doesn't exist?

specutils/io/default_loaders/twoslaq_lrg.py
```diff
@@ -0,0 +1,99 @@
...
+    Parameters
+    ----------
+    file_name: str
+        The path to the FITS file
+    Returns
+    -------
+    data: SpectrumList
+        The 2SLAQ-LRG spectrum that is represented by the data in this file.
+    """
+    with fits.open(filename) as hdulist:
```
nmearl, line 50: This should be updated to support the case (as in other loaders) of providing an already existing hdulist or file objects. See the [cos loader](https://github.com/astropy/specutils/blob/master/specutils/io/default_loaders/hst_cos.py#L38) as an example.

nmearl, line 50: This should be updated to support the case (as in other loaders) of providing an already existing hdulist or file objects. See the [cos loader](https://github.com/astropy/specutils/blob/master/specutils/io/default_loaders/hst_cos.py#L38) as an example.

specutils/io/default_loaders/twoslaq_lrg.py
```diff
@@ -0,0 +1,105 @@
...
+from astropy.wcs import WCS
+from specutils.io.registers import data_loader
+from specutils import Spectrum1D, SpectrumList
+
+
+def identify_2slaq_lrg(origin, *args, **kwargs):
+    """
+    Identify if the current file is a 2SLAQ-LRG file
+    """
+    with fits.open(args[0]) as hdulist:
```
nmearl, line 12: You may want to do the same here, to make sure the identifier function works with [file objects](https://github.com/astropy/specutils/blob/master/specutils/io/default_loaders/sdss.py#L33).

aragilar: Sure, done.

PR #637, opened 2020-04-02
Make default loaders work with fits objects

### LINKED ISSUE #450: Make sdss loader (and other loaders?) work with file-like objects and fits objects
Right now the SDSS loaders (and I think many of the other loaders?) only accept file *names*.  E.g., `specutils.Spectrum1D.read('filename')` works but `specutils.Spectrum1D.read(open('filename'))` does not.  Most of the other loaders in astropy (and elsewhere) understand how to work directly with file-like objects, important for e.g. streaming files from the internet or other non-filesystem sources of spectral data.

Relatedly, to support use cases like the `astroquery.sdss` module where an `HDUList` is created, it would be nice for fits formats (like the SDSS spectral loaders) to *also* accept `HDUList`s instead of needing a file at all.

cc @drdavella @nmearl

rosteen:
Adds HDUList handling to the default_loaders where applicable, and adds examples to the docs of using Spectrum1D.read() on both HDUList and open file-like objects.

This closes #450

### DIFF
--- a/docs/spectrum1d.rst
+++ b/docs/spectrum1d.rst
@@ -66,6 +66,7 @@ Note that the same spectrum could be more conveniently downloaded via
 astroquery, if the user has that package installed:
 
 .. code-block:: python
+
      >>> from astroquery.sdss import SDSS # doctest: +SKIP
      >>> specs = SDSS.get_spectra(plate=751, mjd=52251, fiberID=160) # doctest: +SKIP
      >>> Spectrum1D.read(specs[0], format="SDSS-III/IV spec") # doctest: +SKIP
--- a/docs/spectrum1d.rst
+++ b/docs/spectrum1d.rst
@@ -53,6 +53,7 @@ internet. Note that in these cases, a format string corresponding to an
 existing loader must be supplied.
 
 .. code-block:: python
+    
     >>> from astroquery.sdss import SDSS
     >>> from specutils import Spectrum1D
     >>> import urllib
--- a/docs/spectrum1d.rst
+++ b/docs/spectrum1d.rst
@@ -54,11 +54,11 @@ format string corresponding to an existing loader must be supplied.
 
 .. code-block:: python
     
-    >>> from astroquery.sdss import SDSS
+    >>> from astroquery.sdss import SDSS # doctest: +SKIP
     >>> from specutils import Spectrum1D
     >>> import urllib
-    >>> specs = SDSS.get_spectra(plate=751, mjd=52251, fiberID=160)
-    >>> Spectrum1D.read(specs[0], format="SDSS-III/IV spec")
+    >>> specs = SDSS.get_spectra(plate=751, mjd=52251, fiberID=160) # doctest: +SKIP
+    >>> Spectrum1D.read(specs[0], format="SDSS-III/IV spec") # doctest: +SKIP
     <Spectrum1D(flux=<Quantity [30.596626,...]...>
     >>> specs = urllib.request.urlopen('https://data.sdss.org/sas/dr14/sdss/spectro/redux/26/spectra/0751/spec-0751-52251-0160.fits')
     >>> Spectrum1D.read(specs, format="SDSS-III/IV spec")
--- a/docs/spectrum1d.rst
+++ b/docs/spectrum1d.rst
@@ -47,9 +47,9 @@ encouraged to :doc:`create their own loader </custom_loading>`.
     >>> from specutils import Spectrum1D
     >>> spec1d = Spectrum1D.read("/path/to/file.fits")  # doctest: +SKIP
 
-The built-in default loaders can also read an existing 
-`astropy.io.fits.hdu.HDUList` object, or from an open file object as resulting 
-from e.g. streaming a file from the internet. Note that in these cases, a 
+Most of the built-in specutils default loaders can also read an existing 
+`astropy.io.fits.hdu.HDUList` object or an open file object (as resulting 
+from e.g. streaming a file from the internet). Note that in these cases, a 
 format string corresponding to an existing loader must be supplied.
 
 .. code-block:: python
--- a/docs/spectrum1d.rst
+++ b/docs/spectrum1d.rst
@@ -47,10 +47,10 @@ encouraged to :doc:`create their own loader </custom_loading>`.
     >>> from specutils import Spectrum1D
     >>> spec1d = Spectrum1D.read("/path/to/file.fits")  # doctest: +SKIP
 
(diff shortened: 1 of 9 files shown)


dhomeier:
LGTM, the tests seem to cover all relevant changes. Sorry #450 slipped under my radar when I submitted #545, otherwise I might have joined the discussion on handling file-like objects.

eteq:
Aha, that explains why the results I was seeing in #450 originally weren't consistent with now #545 already did part of it - thanks for catching that @dhomeier !

docs/spectrum1d.rst
```diff
@@ -47,7 +47,22 @@ encouraged to :doc:`create their own loader </custom_loading>`.
...
+.. code-block:: python
+    
+    >>> from astroquery.sdss import SDSS # doctest: +SKIP
+    >>> from specutils import Spectrum1D
+    >>> import urllib
+    >>> specs = SDSS.get_spectra(plate=751, mjd=52251, fiberID=160) # doctest: +SKIP
+    >>> Spectrum1D.read(specs[0], format="SDSS-III/IV spec") # doctest: +SKIP
+    <Spectrum1D(flux=<Quantity [30.596626,...]...>
+    >>> specs = urllib.request.urlopen('https://data.sdss.org/sas/dr14/sdss/spectro/redux/26/spectra/0751/spec-0751-52251-0160.fits') # doctest: +SKIP
+    >>> Spectrum1D.read(specs, format="SDSS-III/IV spec") # doctest: +SKIP
```
eteq, line 64: If you do this instead:

```suggestion
    >>> specs = urllib.request.urlopen('https://data.sdss.org/sas/dr14/sdss/spectro/redux/26/spectra/0751/spec-0751-52251-0160.fits') # doctest: +REMOTE_DATA
    >>> Spectrum1D.read(specs, format="SDSS-III/IV spec") # doctest: +REMOTE_DATA
```

then this part of the doctests will run but only when the --remote-data argument is given, which is good because then this doubles as a unit test (and also a check that the example in the docs still works!)

docs/spectrum1d.rst
```diff
@@ -47,7 +47,22 @@ encouraged to :doc:`create their own loader </custom_loading>`.
...
     >>> spec1d = Spectrum1D.read("/path/to/file.fits")  # doctest: +SKIP
 
+Most of the built-in specutils default loaders can also read an existing 
+`astropy.io.fits.HDUList` object or an open file object (as resulting 
+from e.g. streaming a file from the internet). Note that in these cases, a 
+format string corresponding to an existing loader must be supplied.
 
+.. code-block:: python
+    
+    >>> from astroquery.sdss import SDSS # doctest: +SKIP
```
eteq, line 57: It might be better to separate the `astroquery` and `urllib` into to separate blocks.  The complication being that `astroquery` requires the user to have `astroquery` installed.  So you might have this be a second example with some explanatory text like "or to download the same spectrum using `astroquery`, you could do:" so that the user knows that's a more convenient way to do the same thing?

specutils/tests/test_loaders.py
```diff
@@ -117,34 +122,64 @@ def test_hst_stis(remote_data_path):
...
+            hdulist = fits.open(tmp_file.name)
+            spec = Spectrum1D.read(hdulist, format="SDSS-III/IV spec")
+            assert isinstance(spec, Spectrum1D)
+            assert spec.flux.size > 0
+
 
 @pytest.mark.remote_data
 def test_sdss_spspec():
     sp_pattern = 'spSpec-51957-0273-016.fit.'
+    with urllib.request.urlopen('http://das.sdss.org/spectro/1d_26/0273/1d/spSpec-51957-0273-016.fit') as response:
```
eteq, line 160: Maybe worth adding a quick comment here clarifying that these two cases are for the file-like object and also separately "load a file from disk"?  It's not super-obvious on a quick read.

rosteen:
Resolved suggestions from @eteq - I thought they were all reasonable. Should be ready to merge.

specutils/tests/test_loaders.py
rosteen: My comments there were rather terse, added a few words to hopefully make the different cases more clear

docs/spectrum1d.rst
```diff
@@ -47,6 +47,29 @@ encouraged to :doc:`create their own loader </custom_loading>`.
...
+    >>> import urllib
+    >>> specs = urllib.request.urlopen('https://data.sdss.org/sas/dr14/sdss/spectro/redux/26/spectra/0751/spec-0751-52251-0160.fits') # doctest: +REMOTE_DATA
+    >>> Spectrum1D.read(specs, format="SDSS-III/IV spec") # doctest: +REMOTE_DATA
+    <Spectrum1D(flux=<Quantity [30.596626,...]...>
+
+Note that the same spectrum could be more conveniently downloaded via
+astroquery, if the user has that package installed:
+
+.. code-block:: python
+     >>> from astroquery.sdss import SDSS # doctest: +SKIP
```
eteq, line 69:
```suggestion

     >>> from astroquery.sdss import SDSS # doctest: +SKIP
```
Looks like a space is missing here, I think that's why the CI failed.

specutils/io/default_loaders/hst_cos.py
```diff
@@ -20,36 +20,45 @@ def cos_identify(origin, *args, **kwargs):
...
     -------
     data: Spectrum1D
         The spectrum that is represented by the data in this table.
     """
+    if isinstance(file_obj, fits.hdu.hdulist.HDUList):
+        close_hdulist = False
+        hdu = file_obj
+    else:
+        close_hdulist = True
+        hdu = fits.open(file_obj, **kwargs)
```
dhomeier, line 43: Would there be serious overhead associated with just calling
`isinstance(file_obj, fits.hdu.hdulist.HDUList)`
again before the call to `hdu.close()`?
That would obviate the need for an extra variable.

rosteen: I think both of these are good suggestions, implemented.

dhomeier:
Are the tests passing with `open_files=True`? It doesn't seem like you are ever calling `hdulist.close()`.

nmearl:
@rosteen when you get the chance, can you rebase this?

rosteen:
> Are the tests passing with `open_files=True`? It doesn't seem like you are ever calling `hdulist.close()`.

Not sure what you mean here @dhomeier , could you point me to a specific place you're looking? 

@nmearl will rebase shortly...

eteq:
Looks to me like everything has been addressed, so merging!

PR #640, opened 2020-04-03
Clean code of spurious references to SpectralCoord attributes

ibusko:
As part of the adaptation of code in analysis/manipulation functions to use SpectralCoord, we cleaned up the code of references that pointed directly to underlying attributes of the spectral_axis object. These references now point to the object itself, thus ensuring that existing underlying machinery will be used (if present). The exception are references to the .quantity attribute, a couple of which exist only at the point in the execution flow  where they are necessary.

The remaining part of this issue, namely, integrating bin edge functionality in SpectralCoord, will be handled in a separate PR.

### DIFF
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -97,7 +97,7 @@ def _compute_line_flux(spectrum, regions=None):
         calc_spectrum = spectrum
 
     # Average dispersion in the line region
-    avg_dx = np.abs(np.diff(calc_spectrum.spectral_axis.quantity))
+    avg_dx = (np.abs(np.diff(calc_spectrum.spectral_axis))).quantity
 
     line_flux = np.sum(calc_spectrum.flux[1:] * avg_dx)
 
@@ -115,10 +115,9 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
     if continuum == 1:
         continuum = 1*calc_spectrum.flux.unit
 
-    spectral_axis = calc_spectrum.spectral_axis.quantity
+    spectral_axis = calc_spectrum.spectral_axis
 
-    # Ensure that the
-    dx = np.abs(spectral_axis[-1] - spectral_axis[0])
+    dx = (np.abs(spectral_axis[-1] - spectral_axis[0])).quantity
 
     line_flux = _compute_line_flux(spectrum, regions)
 
@@ -178,7 +177,7 @@ def is_continuum_below_threshold(spectrum, threshold=0.01):
     # we are going to calculate based on the S/N if the uncertainty
     # exists.
     if uncertainty and uncertainty.uncertainty_type != 'std':
-        return np.median(flux / uncertainty.quantity) < threshold
+        return np.median(flux / uncertainty) < threshold
     else:
         return np.median(flux) / mad_std(flux) < threshold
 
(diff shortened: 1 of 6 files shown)


specutils/analysis/location.py
```diff
@@ -80,10 +81,10 @@ def _centroid_single_region(spectrum, region=None):
         calc_spectrum = spectrum
 
     flux = calc_spectrum.flux
-    dispersion = calc_spectrum.spectral_axis.quantity
+    dispersion = (calc_spectrum.spectral_axis).quantity
```
nmearl, line 84:
```suggestion
    dispersion = calc_spectrum.spectral_axis.quantity
```

specutils/analysis/location.py
```diff
@@ -80,10 +81,10 @@ def _centroid_single_region(spectrum, region=None):
...
 
     if len(flux.shape) > 1:
-        dispersion = np.tile(dispersion, [flux.shape[0], 1])
+        dispersion = (np.tile(dispersion, [flux.shape[0], 1]))
```
nmearl, line 87:
```suggestion
        dispersion = np.tile(dispersion, [flux.shape[0], 1])
```

specutils/analysis/flux.py
```diff
@@ -178,7 +177,8 @@ def is_continuum_below_threshold(spectrum, threshold=0.01):
     # we are going to calculate based on the S/N if the uncertainty
     # exists.
     if uncertainty and uncertainty.uncertainty_type != 'std':
-        return np.median(flux / uncertainty.quantity) < threshold
+        # return np.median(flux / uncertainty.quantity) < threshold
```
nmearl, line 180:
```suggestion
```

ibusko:
Looks like my squash actually worked in reverse. Now I have the squashed commits showing twice in the commit history.

specutils/analysis/flux.py
ibusko: Should we use uncertainty.quantity, or just uncertainty here? The tests have no combination that checks this line of code.

specutils/analysis/flux.py
```diff
@@ -178,7 +177,7 @@ def is_continuum_below_threshold(spectrum, threshold=0.01):
     # we are going to calculate based on the S/N if the uncertainty
     # exists.
     if uncertainty and uncertainty.uncertainty_type != 'std':
-        return np.median(flux / uncertainty.quantity) < threshold
+        return np.median(flux / uncertainty) < threshold
```
nmearl, line 180:
```suggestion
        return np.median(flux / uncertainty.quantity) < threshold
```

We certainly want this with `.quantity` -- `NDUncertainty` classes do not support arithmetic otherwise.

nmearl:
@ibusko can you rebase to remove your merge commit?

ibusko:
@nmearl interesting, it doesn't show on the edit list in `git rebase -i origin`. Isn't it created by the 'git push --force` ?

nmearl:
It doesn't look like you've rebased -- I still see duplicate commits.

ibusko:
@nmearl  I will try a rebase again. Last time, commits were duplicated instead of removed. Do you have any clue on why that would happen?

ibusko:
@nmearl if this grows out of control (as it is doing now), I can easily start over from a fresh fork. That way only the relevant commits will be in the PR. The downside is that our conversation here will be hidden in a closed PR. Although we can always add a reference to the old PR to the description.

nmearl:
@ibusko would you like me to try and rebase it? Otherwise, we can certainly just open an new PR and reference this one.

nmearl:
@ibusko would you like me to try and rebase it? Otherwise, we can certainly just open an new PR and reference this one.

ibusko:
@nmearl please, by all means, try and see what you get.

PR #641, opened 2020-04-05
Update custom loading doc to include spectral_axis in Spectrum1D object 

### LINKED ISSUE #625: Having problems using specutils to write Spectrum1D object to FITS file

kakirastern:
Fixes #625. 

To clarify an issue previously encountered when using custom loader defined as described in official docs to load FITS data to file. It appears that the `spectral_axis` is missing from the example, which could cause problems somewhere down the line, when the FITS file loaded this way is then subsequently read with a compatible custom reader as a `Spectrum1D` object then called, as the two essential requirements of any `Spectrum1D` object are the `flux` and the `spectral_axis`. It is in the hope that any confusion stemming from the example can be cleared.

### DIFF
--- a/docs/custom_loading.rst
+++ b/docs/custom_loading.rst
@@ -88,9 +88,10 @@ ensure that the data file being loaded is compatible with the loader function.
             meta = {'header': header}
             wcs = WCS(hdulist[0].header)
             uncertainty = StdDevUncertainty(tab["err"])
+            lamb = tab["spectral_axis"] * Unit("Angstrom")
             data = tab["flux"] * Unit("Jy")
 
-        return Spectrum1D(flux=data, wcs=wcs, uncertainty=uncertainty, meta=meta)
+        return Spectrum1D(flux=data, spectral_axis=lamb, wcs=wcs, uncertainty=uncertainty, meta=meta)
 
 
 An ``extensions`` keyword can be provided. This allows for basic filename
--- a/docs/custom_loading.rst
+++ b/docs/custom_loading.rst
@@ -94,6 +94,9 @@ ensure that the data file being loaded is compatible with the loader function.
         return Spectrum1D(flux=data, spectral_axis=lamb, wcs=wcs, uncertainty=uncertainty, meta=meta)
 
 
+The above is based on the ``Quantity`` formulation, as for any reasonably "generic" format ``Table.read()``
+would extract the units from the FITS header anyway.
+
 An ``extensions`` keyword can be provided. This allows for basic filename
 extension matching in the case that the ``identifier`` function is not
 provided.
@@ -165,6 +168,10 @@ This again will be done in a separate python file and placed in the user's
         tab = Table([disp, flux], names=("spectral_axis", "flux"), meta=meta)
 
         tab.write(file_name, format="fits")
+
+Please note that it is assumed that the axis (of the ``flux`` and ``spectral_axis``)
+units information is stored in the ``meta`` parameter as introduced in the definition
+of ``generic-fits`` as the ``custom_writer()`` above.
 
 The custom writer can be used by passing the name of the custom writer to the
 ``format`` argument of the ``write`` method on the
--- a/docs/custom_loading.rst
+++ b/docs/custom_loading.rst
@@ -159,8 +159,8 @@ This again will be done in a separate python file and placed in the user's
     from specutils.io.registers import custom_writer
 
 
-    @custom_writer("fits-writer")
-    def generic_fits(spectrum, file_name, **kwargs):
+    @custom_writer("my-format")
+    def my_format_writer(spectrum, file_name, **kwargs):
         flux = spectrum.flux
         disp = spectrum.spectral_axis
         meta = spectrum.meta
@@ -169,7 +169,7 @@ This again will be done in a separate python file and placed in the user's
 
         tab.write(file_name, format="fits")
 
-Please note that it is assumed that the axis (of the ``flux`` and ``spectral_axis``)
+Note that this will store the units information of the ``flux`` and ``spectral_axis`` as attributes of the ``Quantity`` columns in the ``Table``, which is saving them automatically to the FITS file.
 units information is stored in ``Table`` object as introduced in the definition
 of ``generic-fits`` as the ``custom_writer()`` above.
 
--- a/docs/custom_loading.rst
+++ b/docs/custom_loading.rst
@@ -94,8 +94,52 @@ ensure that the data file being loaded is compatible with the loader function.
         return Spectrum1D(flux=data, spectral_axis=lamb, wcs=wcs, uncertainty=uncertainty, meta=meta)
 
 
-The above is based on the ``Quantity`` formulation, as for any reasonably "generic" format ``Table.read()``
-would extract the units from the FITS header anyway.
+The above is based on the ``Quantity`` formulation, as for any reasonably "generic" format
+``Table.read()`` would extract the units from the FITS header anyway.
+
+Another approach is based on the ``Table`` formulation, as follows:
+
+.. code-block:: python
+
+    import os
+
+    from astropy.io import fits
+    from astropy.nddata import StdDevUncertainty
+    from astropy.table import Table
+    from astropy.units import Unit
+    from astropy.wcs import WCS
+
+    from specutils.io.registers import data_loader
+    from specutils import Spectrum1D
+
+
+    # Define an optional identifier. If made specific enough, this circumvents the
+    # need to add ``format="my-format"`` in the ``Spectrum1D.read`` call.
+    def identify_generic_fits(origin, *args, **kwargs):
+        return (isinstance(args[0], str) and
+                os.path.splitext(args[0].lower())[1] == '.fits')
+
+    @data_loader("my-format", identifier=identify_generic_fits,
+                 extensions=['fits'])
+    def generic_fits(file_name, **kwargs):
+        with fits.open(file_name, **kwargs) as hdulist:
+
+            tab = Table.read(file_name)
+
+            uncertainty = StdDevUncertainty(tab["err"])
+            lamb = tab["spectral_axis"] * Unit('')
+            data = tab["flux"] * Unit('')
+
+        return Spectrum1D(flux=data, spectral_axis=lamb, uncertainty=uncertainty)
+
+
+The above example makes the assumption that the units of the wavelength and the flux are
+stored in the table's ``spectral_axis`` and ``flux`` columns as ``Quantity`` objects.
+
+Also, ``specutils`` provides some generic functions for parsing data in cases
+where the loader needs to determine characteristics at load time. Please see
+the api for `~specutils.io.parsing_utils.generic_spectrum_from_table` and
+`~specutils.io.parsing_utils.spectrum_from_column_mapping`.
 
 An ``extensions`` keyword can be provided. This allows for basic filename
 extension matching in the case that the ``identifier`` function is not
@@ -169,9 +213,9 @@ This again will be done in a separate python file and placed in the user's
 
         tab.write(file_name, format="fits")
 
-Note that this will store the units information of the ``flux`` and ``spectral_axis`` as attributes of the ``Quantity`` columns in the ``Table``, which is saving them automatically to the FITS file.
-units information is stored in ``Table`` object as introduced in the definition
-of ``generic-fits`` as the ``custom_writer()`` above.
+Note that this will store the units information of the ``flux`` and ``spectral_axis``
+as attributes of the ``Quantity`` columns in the ``Table``, which is saving them
+automatically to the FITS file.
 
 The custom writer can be used by passing the name of the custom writer to the
 ``format`` argument of the ``write`` method on the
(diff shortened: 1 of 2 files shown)


nmearl:
This particular example in the documentation explicitly doesn't set the spectral axis because it's instead created from the `WCS` object. (Likewise your additions also assume that the input spectral axis is in `Angstrom`, but the units should be extracted from the WCS information.)

I'd instead add a paragraph beneath discussing the fact that if your WCS is malformed, or doesn't adhere to the the FITS standard, you can instead set the spectral axis explicitly as you do in your addition.

docs/custom_loading.rst
```diff
@@ -88,9 +88,10 @@ ensure that the data file being loaded is compatible with the loader function.
             meta = {'header': header}
             wcs = WCS(hdulist[0].header)
             uncertainty = StdDevUncertainty(tab["err"])
+            lamb = tab["spectral_axis"] * Unit("Angstrom")
```
dhomeier, line 91:
```suggestion
            lamb = Quantity(tab["spectral_axis"])
```

docs/custom_loading.rst
```diff
@@ -88,9 +88,10 @@ ensure that the data file being loaded is compatible with the loader function.
...
             data = tab["flux"] * Unit("Jy")
```
dhomeier, line 92:
```suggestion
            data = Quantity(tab["flux"])
```

docs/custom_loading.rst
dhomeier, line 91:
```suggestion
            lamb = Quantity(tab["spectral_axis"])
```

docs/custom_loading.rst
dhomeier, line 92:
```suggestion
            data = Quantity(tab["flux"])
```

kakirastern:
> Thanks for taking this on!
> I suggest to use the `Quantity` formulation, since for any reasonably "generic" format `Table.read()` would set the units from the FITS header anyway.
> 
> More generally it is perhaps important to note that there is not really one "generic" FITS format for spectra, but at least two fundamentally different types –
> 
> 1. Files with a _ Linear Wavelength Solution_ defined in the header, and the flux typically stored in `IMAGE` format in the corresponding `data` section, which is the format the `wcs1d-fits` reader described above will read.
> 2. Files storing flux and spectral axis as two data arrays in a table extension (which should in many cases be readable by `tabular-fits`).
> 
> This section of the docs describes a customised version for reading format 2, but for completeness I think the docs should (not necessarily to be resolved in this PR)
> 
> * A) Include a description of the builtin loader for format 2 as well
> * B) Possibly also include a description how to create a custom reader for format 1.
> 
> Even more generally I suggest to discuss if the documentation of the existing loaders should go into its own section to make it easier to find methods for reading and writing files that do not require writing custom extensions.

I originally intended to make the modifications "brief (and sweet)" so as to preserve the original structure and content organization. But on second thought maybe it would be a better idea to go with @dhomeier's suggestions and expand on the matter further in order to give a comprehensive treatment regarding the potential confusion that could arise given the current state of the documentation. Changes to this PR to follow.

dhomeier:
Seems my browser did not update properly on this PR when I posted my comments, so ours somewhat overlap.

> I'd instead add a paragraph beneath discussing the fact that if your WCS is malformed, or doesn't adhere to the the FITS standard, you can instead set the spectral axis explicitly as you do in your addition.

This of course assumes that there _is_ a table row with spectral axis information...
On the other hand the original version assumes the file does have a correctly defined WCS in the _primary HDU_ and a matching flux array in a _table extension_, which is a combination I have not seen in real life yet.

docs/custom_loading.rst
```diff
@@ -166,6 +169,10 @@ This again will be done in a separate python file and placed in the user's
 
         tab.write(file_name, format="fits")
 
+Please note that it is assumed that the axis (of the ``flux`` and ``spectral_axis``)
+units information is stored in the ``meta`` parameter as introduced in the definition
+of ``generic-fits`` as the ``custom_writer()`` above.
+
```
dhomeier, line 175: This is not correct. In the general case we should not make any assumptions on the information stored in the `meta` attribute of a `Spectrum1D` object. However the units _are_ guaranteed to be already part of `flux` and `spectral_axis`, because the are verified to be `Quantity`instances on initialisation. That's why just
```
        flux = spectrum.flux
        disp = spectrum.spectral_axis
```
should be used.

dhomeier:
We still need to come to an agreement on how to document the two cases `WCS` + `hdu.data` and `Table` – @nmearl I still favour providing one example for each variant.
I would also suggest to change the name in the examples from `generic_fits` to `my_format`, as this is the `data_loader` label given, and they are meant to be custom, not generic.

docs/custom_loading.rst
kakirastern: I have changed the wording to match your descriptions.

docs/custom_loading.rst
```diff
@@ -161,16 +161,16 @@ This again will be done in a separate python file and placed in the user's
 
     @custom_writer("fits-writer")
```
dhomeier, line 162:
```suggestion
    @custom_writer("my-format")
```

docs/custom_loading.rst
```diff
@@ -161,16 +161,16 @@ This again will be done in a separate python file and placed in the user's
...
     def generic_fits(spectrum, file_name, **kwargs):
```
dhomeier, line 163:
```suggestion
    def my_format_writer(spectrum, file_name, **kwargs):
```

docs/custom_loading.rst
```diff
@@ -161,16 +161,16 @@ This again will be done in a separate python file and placed in the user's
...
-        disp = spectrum.spectral_axis.value
+        flux = spectrum.flux
+        disp = spectrum.spectral_axis
         meta = spectrum.meta
 
         tab = Table([disp, flux], names=("spectral_axis", "flux"), meta=meta)
 
         tab.write(file_name, format="fits")
 
 Please note that it is assumed that the axis (of the ``flux`` and ``spectral_axis``)
```
dhomeier, line 172:
```suggestion
Note that this will store the units of the ``flux`` and ``spectral_axis`` as attributes of the ``Quantity`` columns in the ``Table``, which is saving them automatically to the FITS file.
```

docs/custom_loading.rst
```diff
@@ -161,16 +161,16 @@ This again will be done in a separate python file and placed in the user's
...
-units information is stored in the ``meta`` parameter as introduced in the definition
+units information is stored in ``Table`` object as introduced in the definition
```
dhomeier, line 173: Sorry, it was not clear to me what you meant by "introduced in the definition" here; anyway this is not assuming anything, since these two `Spectrum1D` attributes _must_ have `unit` properties, and a `Table` will preserve them as `Quantity` columns.
Also added my suggestions for consistent naming above (see my comment in the main discussion); I had not noticed that this section was using yet different names (and also duplicating the function definition of `generic_fits` if followed verbatim).

kakirastern: Sure, have incorporated your suggestion for consistent naming.

nmearl:
> We still need to come to an agreement on how to document the two cases WCS + hdu.data and Table – @nmearl I still favour providing one example for each variant.

I am definitely on board with two separate and obvious examples -- the more, the better.

> I would also suggest to change the name in the examples from generic_fits to my_format, as this is the data_loader label given, and they are meant to be custom, not generic.

*Very* good point, and it looks like @kakirastern has made that change.

kakirastern:
Sure, I will add examples for the the two cases `WCS + hdu.data` and `Table` in my PR to make a clear distinction of the two variants.

dhomeier:
@nmearl perhaps also to be discussed, should the convenience functions `generic_spectrum_from_table` and `spectrum_from_column_mapping` in `parsing_utils` be "officially" recommended for `table` loaders, i.e. should the example use these?

nmearl:
> should the convenience functions generic_spectrum_from_table and spectrum_from_column_mapping in parsing_utils be "officially" recommended for table loaders, i.e. should the example use these?

Mm, I'm not sure we should advertise these too much. They're meant to be used for the generic loaders that need to essentially dynamically determine the characteristics of the data. At the point that the user is creating a custom viewer, I think it's safe to assume in the documentation that the user has a specific data set they are trying to load and parse, and so these functions should not be necessary.

With that said, we should definitely have a blurb that's something like "`specutils` also provides some generic functions for parsing data in cases where the loader needs to determine characteristics at load time. Please see the api for `generic_spectrum_from_table` and `spectrum_from_column_mapping`."

dhomeier:
> With that said, we should definitely have a blurb that's something like "`specutils` also provides some generic functions for parsing data in cases where the loader needs to determine characteristics at load time. Please see the api for `generic_spectrum_from_table` and `spectrum_from_column_mapping`."

Good idea. `spectrum_from_column_mapping` as I understand it already expects a fairly precise description from the user, but then I don't see that much functionality there that could not be replicated with 3 or 4 lines using `Table` functions – except perhaps for the built-in unit conversion.

`generic_spectrum_from_table` also introduces the additional complication that it always wants to use a provided `wcs`, whether it is compatible with the `flux` array or not.

kakirastern:
@dhomeier Sorry about the delay... The PR is ready for another review

docs/custom_loading.rst
```diff
@@ -88,11 +88,59 @@ ensure that the data file being loaded is compatible with the loader function.
...
+
+            uncertainty = StdDevUncertainty(tab["err"])
+            lamb = tab["spectral_axis"] * Unit('')
+            data = tab["flux"] * Unit('')
+
+        return Spectrum1D(flux=data, spectral_axis=lamb, uncertainty=uncertainty)
+
+
+The above example makes the assumption that the units of the wavelength and the flux are
+stored in the table's ``spectral_axis`` and ``flux`` columns as ``Quantity`` objects.
```
nmearl, line 137: I'm confused by this statement, as `lamb` and `data` are both being made into `Quantity` objects using explicitly empty `Unit` instances. How are we assuming that the units of the wavelength and flux are in the columns?

eteq:
closed at request of original author (but others can of course take it up if they desire - last commit was SHA 4f12b65)

Issue #629, opened 2020-03-30
Have the FluxConservingResampler use SpectralCoord's bin edge functionality


eteq:
This is an issue that depends on #626 getting finished.  But once that's done, the `FluxConservingResampler` should use the same way of understanding the "edge" of a bin as what #626 implements. 

This is a bit circular in that my hunch is that #626 might be implemented by taking some of the machinery from `FluxConservingResampler`.  But either way, it's important this circles back and ensures that the resampler machinery ends up consistent with `SpectralCoord` (and the pre-existing tests pass!)

rosteen:
This was addressed but not closed, closing.

Issue #495, opened 2019-07-24
Replace `_calc_bin_edges` in resample


SaOgaz:
When #494 is fixed, the `_calc_bin_edges` function in the resample code should be replaced with the spectrum1D built in edge calculator.

nmearl:
Fixed in #645.

Issue #494, opened 2019-07-24
make GWCS-based Spectrum1D objects support `bin_edges`


SaOgaz:
When you instantiate a `Spectrum1D` with a wavelength array, and then try and call the `bin_edges` method, it gives you an error. I suspect after looking at the code that when you instantiate `Spectrum1D` this way, it doesn't even build the WCS object.

```
In [1]: from specutils import Spectrum1D                                                                                                        

In [2]: import numpy as np                                                                                                                      

In [3]: import astropy.units as u                                                                                                               

In [4]: freqs = np.arange(50) * u.GHz                                                                                                           

In [5]: flux = np.random.random((5, len(freqs))) * u.Jy                                                                                         

In [6]: spec = Spectrum1D(spectral_axis=freqs, flux=flux)                                                                                       

In [7]: spec.bin_edges                                                                                                                          
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-7-413064155989> in <module>
----> 1 spec.bin_edges

~/miniconda3/envs/specutilsdev/lib/python3.7/site-packages/astropy/utils/decorators.py in __get__(self, obj, owner)
    742                 return val
    743             else:
--> 744                 val = self.fget(obj)
    745                 obj.__dict__[self._key] = val
    746                 return val

~/specutils/specutils/spectra/spectrum1d.py in bin_edges(self)
    148     @lazyproperty
    149     def bin_edges(self):
--> 150         return self.wcs.bin_edges()
    151 
    152     @property



```

SaOgaz:
Whoops, forgot to tag @nmearl.

nmearl:
I should note that this has nothing to do with using the wavelength property, or initializing the spectral axis in wavelength space. This is because the gwcs adapter does not provide a bin edges calculation, unlike the fits wcs adapter.

eteq:
Hmm, so should we re-label this issue "make GWCS-based Spectrum1D objects support `bin_edges`", @nmearl?

nmearl:
Yeah, I think that's more appropriate.

nmearl:
This bug has been fixed in #645. Further discussion in other linked tickets.

PR #645, opened 2020-04-07
Implement SpectralAxis subclass of SpectralCoord with bin edge support

### LINKED ISSUE #626: Have `SpectralCoord` subclass and the spectrum objects consistently understand the concept of bin edges
For some background relevant for this issue: a spectrum can be thought of in two different ways:
1. A "signal", where the signal is sampled at different points along the spectral axis, but the `flux` value is just "a sample of the signal"
2. "Flux in boxes" - e.g., the amount of energy (or number of photons) falling in one spectral "bin". 
Often this is functinonally very similar, but not always - see the flux conserving resampler for why it's sometimes critical to think like option 2. The OP of #176 also has a clear example of a science case where this is necessary.

For option 2, this means it's sometimes important to define a spectrum using bin *edges* rather than the current approach where `spectral_axis` is always aligned to the `flux`.  The most natural way to do this is to have `spectrum.spectral_axis.bin_edges` or possible `spectrum.spectral_axis_edges`.  Either way, the current `SpectralCoord` needs to be connected into this idea.

Some known pieces of this that should work:
1. For a spectrum created as the current `Spectrum1D` where the flux and `spectral_axis` are aligned, the "inner" bin edges should be the half-way point between two neighboring `spectral_axis` values.  The "outer" edges should be whatever makes the first and last bins be symmetric (i.e. for the first bin this means the left edge should be d/2 from the 0th spectral axis element where d is spectral_axis[1]-spectral_axis[0]).
2. There should also be a way to create a `Spectrum1D` where the bins are *explicitly* given (that's the need in #176) where the `spectral_axis` is then auto-created based on the middle-points of the bins.
3. The WCS should be such that the left-most bin edge is `spec.wcs.pixel_to_world(-0.5)`, the next bin edge is `spec.wcs.pixel_to_world(0.5)`, the next is `spec.wcs.pixel_to_world(1.5)` and so on.  (although see #494 - this might be to tall an order to start with, and could then be added later... @nmearl might have some insight).

Note also this issue is partly a roll-up of #176, #494, #495, and *possibly* #489, in that all of those are involved in a more consistent treatment of bin edges, which implementing in `SpectralCoord` would allow.

rosteen:
This currently closes #626, #176, #494. I'm working on bug fixing for additional changes that will address #629 and #495, but wanted to get a PR with the core functionality out for other eyes sooner rather than later in case there are objections to my implementation. #494 was addressed simply by changing `spec.bin_edges` to return `spec.spectral_axis.bin_edges` rather than `spec.wcs.bin_edges()`.

I also plan to have docs and more test coverage added by end of day tomorrow.

### DIFF
--- a/specutils/spectra/spectral_axis.py
+++ b/specutils/spectra/spectral_axis.py
@@ -34,8 +34,11 @@ class SpectralAxis(SpectralCoord):
         return obj
 
     @static_method
-    def _edges_from_centers():
-        pass
+    def _edges_from_centers(centers):
+        a = np.insert(centers, 0, 2*centers[0]-centers[1])
+        b = centers.append(2*centers[-1]-centers[-2])
+        edges = (a + b) / 2
+        return edges
 
     @static_method
     def _centers_from_edges(edges):
--- a/specutils/spectra/spectral_axis.py
+++ b/specutils/spectra/spectral_axis.py
@@ -15,21 +15,36 @@ class SpectralAxis(SpectralCoord):
     """
     """
 
-    def __new__():
-        pass
+    def __new__(cls, value, unit=None, observer=None, target=None,
+                 radial_velocity=None, redshift=None, doppler_rest=None,
+                 doppler_convention=None, bin_specification="centers",
+                 **kwargs):
+
+        # Convert to bin centers if bin edges were given, since SpectralCoord
+        # only accepts centers
+        if bin_specification == "edges":
+            value = self._centers_from_edges(value)
+            bin_specification = "bin_centers"
+
+        obj = super().__new__(cls, value, unit=unit, observer=observer,
+                              target=target, radial_velocity=radial_velocity,
+                              redshift=redshift, doppler_rest=doppler_rest,
+                              doppler_convention=doppler_convention, subok=True,
+                              **kwargs)
+        return obj
 
     @static_method
     def _edges_from_centers():
         pass
 
     @static_method
-    def centers_from_edges():
-        pass
+    def _centers_from_edges(edges):
+        return (edges[1:] + edges[:-1]) / 2
 
     @property
     def bin_edges(self):
-        pass
+        return self._edges_from_centers(self.value)
 
     @property
     def bin_centers(self):
-        pass
+        return self.value
--- a/specutils/spectra/spectral_axis.py
+++ b/specutils/spectra/spectral_axis.py
@@ -34,6 +34,14 @@ class SpectralAxis(SpectralCoord):
 
         return obj
 
+    def __quantity_subclass__(self, unit):
+        """:wq
+
+        Overridden by subclasses to change what kind of view is
+        created based on the output unit of an operation.
+        """
+        return SpectralAxis, True
+
     @staticmethod
     def _edges_from_centers(centers):
         a = np.insert(centers, 0, 2*centers[0]-centers[1])
--- a/specutils/spectra/spectral_axis.py
+++ b/specutils/spectra/spectral_axis.py
@@ -53,7 +53,7 @@ class SpectralAxis(SpectralCoord):
         obj = super().__new__(cls, value, *args, **kwargs)
 
         if bin_specification == "edges":
-            obj.bin_edges = bin_edges
+            obj._bin_edges = bin_edges
 
         return obj
 
@@ -65,7 +65,7 @@ class SpectralAxis(SpectralCoord):
         return SpectralAxis, True
 
     @staticmethod
-    def _edges_from_centers(centers):
+    def _edges_from_centers(centers, unit):
         """
         Calculates interior bin edges based on the average of each pair of
         centers, with the two outer edges based on extrapolated centers added
@@ -74,7 +74,7 @@ class SpectralAxis(SpectralCoord):
         a = np.insert(centers, 0, 2*centers[0] - centers[1])
         b = np.append(centers, 2*centers[-1] - centers[-2])
         edges = (a + b) / 2
-        return edges
+        return edges*unit
 
     @staticmethod
     def _centers_from_edges(edges):
@@ -89,4 +89,7 @@ class SpectralAxis(SpectralCoord):
         Calculates bin edges if the spectral axis was created with centers
         specified.
         """
-        return self._edges_from_centers(self.value)
+        if hasattr(self, '_bin_edges'):
+            return self._bin_edges
+        else:
+            return self._edges_from_centers(self.value, self.unit)
--- a/specutils/spectra/spectral_axis.py
+++ b/specutils/spectra/spectral_axis.py
@@ -42,10 +42,7 @@ class SpectralAxis(SpectralCoord):
         are interpreted as bin edges or bin centers. Defaults to "centers".
     """
 
-    def __new__(cls, value, unit=None, observer=None, target=None,
-                 radial_velocity=None, redshift=None, doppler_rest=None,
-                 doppler_convention=None, bin_specification="centers",
-                 **kwargs):
+    def __new__(cls, value, *args, bin_specification="centers", **kwargs):
 
         # Convert to bin centers if bin edges were given, since SpectralCoord
         # only accepts centers
@@ -53,10 +50,7 @@ class SpectralAxis(SpectralCoord):
             bin_edges = value
             value = SpectralAxis._centers_from_edges(value)
 
-        obj = super().__new__(cls, value, unit=unit, observer=observer,
-                              target=target, radial_velocity=radial_velocity,
-                              redshift=redshift, doppler_rest=doppler_rest,
-                              doppler_convention=doppler_convention, **kwargs)
+        obj = super().__new__(cls, value, *args, **kwargs)
 
         if bin_specification == "edges":
             obj.bin_edges = bin_edges
--- a/specutils/spectra/spectral_axis.py
+++ b/specutils/spectra/spectral_axis.py
@@ -20,24 +20,7 @@ class SpectralAxis(SpectralCoord):
 
     Parameters
     ----------
-    value : ndarray or `~astropy.units.Quantity` or `SpectralCoord`
-        Spectral axis data values.
-    unit : str or `~astropy.units.Unit`
-        Unit for the given data.
-    observer : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`, optional
-        The coordinate (position and velocity) of the observer.
-    target : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`, optional
-        The coordinate (position and velocity) of the target.
-    radial_velocity : `~astropy.units.Quantity`, optional
-        The radial velocity of the target with respect to the observer.
-    redshift : float, optional
-        The redshift of the target with respect to the observer.
-    doppler_rest : `~astropy.units.Quantity`, optional
-        The rest value to use for velocity space transformations.
-    doppler_convention : str, optional
-        The convention to use when converting the spectral data to/from
-        velocity space.
-    bin_specificatoin: str, optional
+    bin_specification: str, optional
         Must be "edges" or "centers". Determines whether specified axis values
         are interpreted as bin edges or bin centers. Defaults to "centers".
     """
--- a/specutils/spectra/spectral_axis.py
+++ b/specutils/spectra/spectral_axis.py
@@ -14,6 +14,32 @@ __doctest_skip__ = ['SpectralAxis.*']
 
 class SpectralAxis(SpectralCoord):
     """
+    Coordinate object representing spectral values corresponding to a specific
+    spectrum. Overloads SpectralCoord with additional information (currently
+    only bin edges).
+
+    Parameters
+    ----------
+    value : ndarray or `~astropy.units.Quantity` or `SpectralCoord`
+        Spectral axis data values.
+    unit : str or `~astropy.units.Unit`
+        Unit for the given data.
+    observer : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`, optional
+        The coordinate (position and velocity) of observer.
+    target : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`, optional
+        The coordinate (position and velocity) of observer.
+    radial_velocity : `~astropy.units.Quantity`, optional
+        The radial velocity of the target with respect to the observer.
+    redshift : float, optional
+        The redshift of the target with respect to the observer.
+    doppler_rest : `~astropy.units.Quantity`, optional
+        The rest value to use for velocity space transformations.
+    doppler_convention : str, optional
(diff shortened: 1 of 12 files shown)


specutils/spectra/spectral_axis.py
```diff
@@ -0,0 +1,98 @@
...
+        # Convert to bin centers if bin edges were given, since SpectralCoord
+        # only accepts centers
+        if bin_specification == "edges":
+            bin_edges = value
+            value = SpectralAxis._centers_from_edges(value)
+
+        obj = super().__new__(cls, value, unit=unit, observer=observer,
+                              target=target, radial_velocity=radial_velocity,
+                              redshift=redshift, doppler_rest=doppler_rest,
+                              doppler_convention=doppler_convention, **kwargs)
```
nmearl, line 59: I think it's more appropriate to just keep the bundles init arguments buried in the `args` and `kwargs`, that way if we ever change something on the base `SpectralCoord` class, we won't need to also update the `SpectralAxis` class.

```suggestion
    def __new__(cls, value, *args, bin_specification="centers", **kwargs):

        # Convert to bin centers if bin edges were given, since SpectralCoord
        # only accepts centers
        if bin_specification == "edges":
            bin_edges = value
            value = SpectralAxis._centers_from_edges(value)

        obj = super().__new__(cls, value, *args, **kwargs)
```

specutils/spectra/spectral_axis.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        return (edges[1:] + edges[:-1]) / 2
+
+    @lazyproperty
+    def bin_edges(self):
+        """
+        Calculates bin edges if the spectral axis was created with centers
+        specified.
+        """
+        return self._edges_from_centers(self.value)
```
nmearl, line 98: A minor quibble. I realize that the user should be aware of what specification their axis is in, however, if they initialized the `SpectralAxis` with `bin_specification='edges'` (as is default), then this gives wrong answer. I think we may want to check the state of the bin specification and return just `self.value` it's already `edges`.

```suggestion
        if self._bin_edges == 'edges':
            return self.value
            
        return self._edges_from_centers(self.value)
```

specutils/spectra/spectral_axis.py
```diff
@@ -0,0 +1,98 @@
...
+            bin_edges = value
+            value = SpectralAxis._centers_from_edges(value)
+
+        obj = super().__new__(cls, value, unit=unit, observer=observer,
+                              target=target, radial_velocity=radial_velocity,
+                              redshift=redshift, doppler_rest=doppler_rest,
+                              doppler_convention=doppler_convention, **kwargs)
+
+        if bin_specification == "edges":
+            obj.bin_edges = bin_edges
```
nmearl, line 62: _Technically_, this should conflict with the `bin_edges` property -- I'm not sure how this manifests, but ✨ just to be sure ✨ we may want to store this as `obj._bin_edges`.

```suggestion
            obj._bin_edges = bin_edges
```

specutils/spectra/spectral_axis.py
rosteen: Two things: 
- "centers" is default, not "edges"
- I decided to implement this such that no matter what is input, the bin centers are always the thing that are stored in self.value, and self.bin_edges will have either the specified bin edges (if they were provided) or the calculated bin edges (if centers were provided). I figured this would mesh better with all the current code that expects bin centers from a `spec.spectral_axis` while still making the bin edges available when desired, rather than having to constantly check which is being returned in `spec.spectral_axis`. I'm very open to discussion on whether this is the right strategy.

specutils/spectra/spectral_axis.py
rosteen: Good call - I figured there would be a couple things to improve with my subclassing, been a while since I was in OOP land.

specutils/spectra/spectral_axis.py
rosteen: Also, I wasn't sure if I grasped 100% what `@lazyproperty` does, but my hope was that if obj.bin_edges was already set then it would never be evaluated (and thus, here, only be evaluated if centers were specified).

specutils/spectra/spectral_axis.py
rosteen: I wasn't sure if I grasped 100% what @lazyproperty does, but my hope was that if obj.bin_edges was already set then the `bin_edges` property would never be evaluated ( thus only evaluated if centers were specified). It seems to work as I expected in my testing.

nmearl: The `lazyproperty` in this case is just a lazy getter, meaning that the value of the property is computed only once, and later references simply use that same value, but otherwise it should be treated like a normal property getter. My concern in this case is that the property is _only_ a getter, and so setting it explicitly should technically not be allowed.

```python
In [7]: class Test():
   ...:     def __init__(self, *args, **kwargs):
   ...:         self._value = 0
   ...:
   ...:     @property
   ...:     def value(self):
   ...:         return self._value + 1
   ...:

In [8]:

In [8]: test = Test()

In [9]: test.value = 10
<ipython-input-9-7baea496662b> in <module>
----> 1 test.value = 10

AttributeError: can't set attribute
```

It's a strange consequence of the `lazyproperty` decorator that it allows overriding the calculated value:

```python
In [10]: class Test():
    ...:     def __init__(self, *args, **kwargs):
    ...:         self._value = 0
    ...:
    ...:     @lazyproperty
    ...:     def value(self):
    ...:         return self._value + 1
    ...:

In [11]: test = Test()

In [12]: test.value
Out[12]: 1

In [13]: test.value = 10

In [14]: test.value
Out[14]: 10
```

Which it technically should not, since properties and values on instances should behave differently (likewise, it's a standard to not name a property the same as an instance attribute).

I think to maintain the "technically correct" part of this would mean that the `obj.bin_edges` attribute should be named something different.

rosteen: Makes sense, thanks. I'll update that today.

rosteen:
This now closes #629, #495 and #646, and I'm calling it done (other than addressing further review comments).

rosteen:
@ibusko Since this is connected to your recent work, it would probably useful to have your eyes on it for a review if you don't mind.

ibusko:
@rosteen no problem. I added myself as a reviewer.

specutils/spectra/spectral_axis.py
```diff
@@ -0,0 +1,95 @@
...
+    Parameters
+    ----------
+    value : ndarray or `~astropy.units.Quantity` or `SpectralCoord`
+        Spectral axis data values.
+    unit : str or `~astropy.units.Unit`
+        Unit for the given data.
+    observer : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`, optional
+        The coordinate (position and velocity) of observer.
+    target : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`, optional
+        The coordinate (position and velocity) of observer.
```
ibusko, line 30: The coordinate (position and velocity) of target

ibusko, line 30: The coordinate (position and velocity) of target.

specutils/spectra/spectral_axis.py
```diff
@@ -0,0 +1,95 @@
...
+        return obj
+
+    def __quantity_subclass__(self, unit):
+        """
+        Overridden by subclasses to change what kind of view is
+        created based on the output unit of an operation.
+        """
+        return SpectralAxis, True
+
+    @staticmethod
```
ibusko, line 67: Just out of curiosity, is it possible that use cases exist where a 'bin center' is not the average of its edges? I am thinking of a (perhaps hypothetical) situation where a combination of a very non-linear dispersion curve with wide detector bins would justify such asymmetric bins. In that case, the code below this line would not work properly. I have in mind such things as the ACS grisms, where at the red end the dispersion becomes highly non-linear.

specutils/spectra/spectral_axis.py
rosteen: Good catch, fixed (also in spectra_coordinate.py)

specutils/spectra/spectral_axis.py
```diff
@@ -0,0 +1,95 @@
...
+        The coordinate (position and velocity) of observer.
+    radial_velocity : `~astropy.units.Quantity`, optional
+        The radial velocity of the target with respect to the observer.
+    redshift : float, optional
+        The redshift of the target with respect to the observer.
+    doppler_rest : `~astropy.units.Quantity`, optional
+        The rest value to use for velocity space transformations.
+    doppler_convention : str, optional
+        The convention to use when converting the spectral data to/from
+        velocity space.
```
nmearl:
```suggestion
```

Since these exist in the base class and are handled there, there is no need to copy them here in the subclass.

rosteen: Removed.

specutils/spectra/spectral_axis.py
nmearl: I think the above is a reasonable first assumption, and we ought to state this in the documentation. Future work may address customizable edge -> bin calculations, or there could be a diff check on spectral axis array and if they aren't consistent bin sizes, raise a warning to the user that they should pre-calculate the centers and feed those in instead.

specutils/spectra/spectral_axis.py
```diff
@@ -0,0 +1,78 @@
...
+        Must be "edges" or "centers". Determines whether specified axis values
+        are interpreted as bin edges or bin centers. Defaults to "centers".
+    """
+
+    def __new__(cls, value, *args, bin_specification="centers", **kwargs):
+
+        # Convert to bin centers if bin edges were given, since SpectralCoord
+        # only accepts centers
+        if bin_specification == "edges":
+            bin_edges = value
```
nmearl, line 33:
```suggestion
    diff_val = np.diff(value)
    
    if not np.allclose(diff_val, diff_val[0], atol=0.001):
        warnings.warn("Edge-based spectral axis does not seem to be linear. Conversions to bin centers may be incorrect.", AstropyUserWarning)
```

specutils/spectra/spectral_axis.py
```diff
@@ -0,0 +1,78 @@
...
+            value = SpectralAxis._centers_from_edges(value)
+
+        obj = super().__new__(cls, value, *args, **kwargs)
+
+        if bin_specification == "edges":
+            obj._bin_edges = bin_edges
```
nmearl, line 39:
```suggestion
            obj._bin_edges = values
```

specutils/spectra/spectral_axis.py
nmearl: @eteq Would this make sense as a temporary solution before we consider dealing with non-linear spectral axis conversions?

specutils/spectra/spectrum1d.py
```diff
@@ -114,48 +116,65 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
...
+                    bin_specification = "edges"
+                else:
+                    bin_specification = "centers"
+                self._spectral_axis = SpectralAxis(
                     spectral_axis, redshift=redshift,
                     radial_velocity=radial_velocity, doppler_rest=rest_value,
-                    doppler_convention=velocity_convention)
-            # If a SpectralCoord object is provided, we assume it doesn't need
+                    doppler_convention=velocity_convention,
+                    bin_specification = bin_specification)
```
nmearl, line 159:
```suggestion
                    bin_specification=bin_specification)
```

specutils/spectra/spectral_axis.py
eteq: Actually I think this warning is unnecessary.  non-linear spectral axes are quite common, but it's okay to assume a convention in this case and not bother with the warning.

specutils/spectra/spectral_axis.py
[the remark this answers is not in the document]
nmearl: Whoops! That suggestion wasn't supposed to be in the review -- thought I had removed it! Using `bin_edges` is completely fine in this context.

rosteen:
@jdavies-st I added the test you requested with an xfail, since (based on some initial poking at it) a SpectralAxis comes back from the round trip as a SpectralCoord. Minor note that there's not a SpectralCoord test in that file.

eblur:
This looks like a fine compromise for me in Issue #176 

I like the convenience property "bin_edges" as yes, I am quite lazy and prefer fast access to values that I use a lot.

That said, "bin_lo" and "bin_hi" is another convenience parameter for me that would be nice. (Since "low" and "high" values are not always consistent depending on which unit your working in, this could be "bin_left" and "bin_right" instead.

It's not a deal-breaker, though. I'm accustomed to writing my own X-ray spectrum subclass of Spectrum1D.

specutils/tests/test_spectral_axis.py
```diff
@@ -0,0 +1,121 @@
...
+def test_create_with_bin_edges():
+
+    wavelengths = np.linspace(500, 2500, 1001) * u.AA
+    spectral_axis = SpectralAxis(wavelengths, bin_specification="edges")
+
+    assert np.all(spectral_axis.bin_edges == wavelengths)
+    assert spectral_axis[0] == 501*u.AA
+
+    # Test irregular bin edges
+    wavelengths = np.array([500,510,550,560,590])*u.AA
```
eteq, line 46: (PEP8 nitpick)
```suggestion
    wavelengths = np.array([500, 510, 550, 560, 590])*u.AA
```

specutils/tests/test_spectral_axis.py
```diff
@@ -0,0 +1,121 @@
...
+    spectral_axis = SpectralAxis(wavelengths, bin_specification="edges")
+
+    assert np.all(spectral_axis.bin_edges == wavelengths)
+    assert np.all(spectral_axis == [505., 530., 555., 575.]*u.AA)
+
+
+# GENERAL TESTS
```
eteq, line 53: Am I right in guessing this is copied from `test_spectral_coord.py`?  If so, it might be more maintainable to use pytest's parametrized fixtures (https://docs.pytest.org/en/latest/fixture.html#parametrizing-fixtures) to have the same test function get run on both `SpectralCoord` and `SpectralAxis`?

eteq:
Thanks @eblur for the review!

> That said, "bin_lo" and "bin_hi" is another convenience parameter for me that would be nice. (Since "low" and "high" values are not always consistent depending on which unit your working in, this could be "bin_left" and "bin_right" instead.

Yeah I see this point - definitely something we want to enable. I think the complexity is that it then leads to at least the possibility of overlapping or non-continguous bins.  In either of those cases it's not clear to me that the way the wcs is implemented right now is adaptable to that.  But I think it makes sense to do this iteratively - so I'll make a follow-on issue about specifically this.

specutils/tests/test_spectral_axis.py
rosteen: Yes, I did base most of this on the `test_spectral_coord.py` tests. Since I'd have to do a little reading about parametrizing fixtures, perhaps it would be better to merge this as-is and file an issue for consolidating some tests?

PR #662, opened 2020-04-21
Fix spectral coord transformations

astrofrog:
While working on integrating ``SpectralCoord`` with FITS WCS, I carried out a series of in-depth accuracy tests and found that the values were very wrong when applying velocity transformations. I've annotated my changes below to explain the rationale for the changes. Rather than add an extensive accuracy test suite here, I'm focusing on adding this to the core, but in the mean time I thought I should fix these problems here too since they are serious.

### DIFF
--- a/specutils/spectra/spectral_coordinate.py
+++ b/specutils/spectra/spectral_coordinate.py
@@ -416,8 +416,8 @@ class SpectralCoord(u.Quantity):
         coordinate frame in that this calculates the radial velocity with
         respect to the *observer*, not the origin of the frame.
         """
-        return np.sum(self._calculate_radial_velocity(
-            self._observer, self._target), axis=0)
+        return self._calculate_radial_velocity(self._observer, self._target,
+                                               as_scalar=True)
 
     @property
     def redshift(self):
@@ -433,7 +433,7 @@ class SpectralCoord(u.Quantity):
         return self.radial_velocity.to('', equivalencies=RV_RS_EQUIV)
 
     @staticmethod
-    def _calculate_radial_velocity(observer, target):
+    def _calculate_radial_velocity(observer, target, as_scalar=False):
         """
         Compute the line-of-sight velocity from the observer to the target.
 
@@ -449,6 +449,7 @@ class SpectralCoord(u.Quantity):
         `~astropy.units.Quantity`
             The radial velocity of the target with respect to the observer.
         """
+
         # Convert observer and target to ICRS to avoid finite differencing
         #  calculations that lack numerical precision.
         observer_icrs = observer.transform_to(ICRS)
@@ -458,7 +459,12 @@ class SpectralCoord(u.Quantity):
 
         d_vel = target_icrs.velocity - observer_icrs.velocity
 
-        return d_vel.d_xyz * pos_hat.xyz
+        vel_mag = np.dot(d_vel.d_xyz, pos_hat.xyz)
+
+        if as_scalar:
+            return vel_mag
+        else:
+            return vel_mag * pos_hat.xyz
 
     @staticmethod
     def _norm_d_pos(observer, target):
@@ -521,10 +527,9 @@ class SpectralCoord(u.Quantity):
         init_obs_vel = self._calculate_radial_velocity(self.observer, target)
         fin_obs_vel = self._calculate_radial_velocity(observer, target)
 
-        # TODO: getting proper blueshifted results in the test involving
-        #  shifting an observer to the target velocity frame requires that we
-        #  negate the initial velocity between observer/target. Unsure why.
-        new_data = self._project_velocity_and_shift(-init_obs_vel, fin_obs_vel)
+        line_of_sight_unit_vec = self._norm_d_pos(observer, target).xyz
+
+        new_data = self._project_velocity_and_shift(init_obs_vel, fin_obs_vel, line_of_sight_unit_vec)
 
         new_coord = self._copy(value=new_data,
                                observer=observer,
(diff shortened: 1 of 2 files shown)


specutils/spectra/spectral_coordinate.py
```diff
@@ -416,8 +416,8 @@ def radial_velocity(self):
         coordinate frame in that this calculates the radial velocity with
         respect to the *observer*, not the origin of the frame.
         """
-        return np.sum(self._calculate_radial_velocity(
-            self._observer, self._target), axis=0)
+        return self._calculate_radial_velocity(self._observer, self._target,
+                                               as_scalar=True)
```
astrofrog, line 420: The length of the vector can't be found using ``sum``, and ``linalg.norm`` would return the absolute magnitude, so rather than duplicate code, I now use ``_calculate_radial_velocity`` with the option to return a signed scalar

specutils/spectra/spectral_coordinate.py
```diff
@@ -458,7 +459,12 @@ def _calculate_radial_velocity(observer, target):
 
         d_vel = target_icrs.velocity - observer_icrs.velocity
 
-        return d_vel.d_xyz * pos_hat.xyz
+        vel_mag = np.dot(d_vel.d_xyz, pos_hat.xyz)
+
+        if as_scalar:
+            return vel_mag
+        else:
+            return vel_mag * pos_hat.xyz
```
astrofrog, line 467: The previous implementation was incorrect - one wants to return a vector that has the total velocity magnitude times the unit vector between observer and target.

specutils/spectra/spectral_coordinate.py
```diff
@@ -521,18 +527,17 @@ def _change_observer_to(self, observer, target=None):
...
+        new_data = self._project_velocity_and_shift(init_obs_vel, fin_obs_vel, line_of_sight_unit_vec)
 
         new_coord = self._copy(value=new_data,
                                observer=observer,
                                target=target)
 
         return new_coord
 
-    def _project_velocity_and_shift(self, init_vel, fin_vel):
+    def _project_velocity_and_shift(self, init_vel, fin_vel, line_of_sight_unit_vec):
```
astrofrog, line 540: The issue with this method was that it actually had no concept of the direction between observer and target since both or either velocity vectors could be in the opposite direction. So we need to pass in a separate vector giving the actual observer to target direction with the correct orientation.

specutils/spectra/spectral_coordinate.py
```diff
@@ -548,12 +553,16 @@ def _project_velocity_and_shift(self, init_vel, fin_vel):
         new_data : `u.Quantity`
             Spectral axis data with velocity shift applied.
         """
-        # Project the velocity shift vector onto the line-on-sight vector
-        #  between the target and the new observation frame.
-        init_proj_vel = np.dot(init_vel, fin_vel / np.linalg.norm(fin_vel))
 
-        # Calculate the velocity shift between the two vectors
-        delta_vel = np.sum(fin_vel - init_proj_vel, axis=0)
```
astrofrog, line 556: np.sum does not return the vector magnitude, fixed now

specutils/spectra/spectral_coordinate.py
```diff
@@ -548,12 +553,16 @@ def _project_velocity_and_shift(self, init_vel, fin_vel):
...
+        # Project the velocity shift vector onto the line-of-sight vector
+        # between the target and the new observation frame.
+        init_proj_vel = np.dot(init_vel, line_of_sight_unit_vec) * line_of_sight_unit_vec
```
astrofrog, line 559: I now use the real unit vector from observer to target

specutils/spectra/spectral_coordinate.py
```diff
@@ -563,7 +572,7 @@ def _project_velocity_and_shift(self, init_vel, fin_vel):
         #  we may end up with a final velocity very close to, but not quite at,
         #  zero. In this case, set a tolerance for the final velocity; if it's
         #  below this tolerance, assume the delta velocity is essentially nan.
-        if np.isnan(delta_vel) or np.abs(np.sum(fin_vel, axis=0)) < 1e-7 * fin_vel.unit:
+        if np.isnan(delta_vel) or np.abs(np.linalg.norm(fin_vel)) < 1e-7 * fin_vel.unit:
```
astrofrog, line 575: As before, sum doesn't return the vector length

specutils/spectra/spectral_coordinate.py
```diff
@@ -676,7 +685,9 @@ def with_los_shift(self, target_shift=None, observer_shift=None):
         init_obs_vel = self._calculate_radial_velocity(observer_icrs, target_icrs)
         fin_obs_vel = self._calculate_radial_velocity(new_observer, new_target)
 
-        new_data = self._project_velocity_and_shift(init_obs_vel, fin_obs_vel)
+        line_of_sight_unit_vec = self._norm_d_pos(observer_icrs, target_icrs).xyz
```
astrofrog, line 688: I was a little puzzled here that I don't get the same result if I use ``new_observer`` and ``new_target`` - @eteq, is that because ``_norm_d_pos`` assumes the coordinates are in the same frame?

specutils/spectra/spectral_coordinate.py
dhomeier: Strictly I think the `np.abs` is redundant now, as it is acting on the 2-norm.

specutils/spectra/spectral_coordinate.py
eteq: Yes that's right.  And it probably shouldn't do that - I'd suggest adding this to the top of `_norm_d_pos`:
```
target_in_obs_frame = target.transform_to(observer)
```
and then use `target_in_obs_frame` everywhere that `target` is currently used.  That should be a no-op if they're already in the same frame, but will protect against mis-use like it is right now.
```

astrofrog:
@eteq - I'm adding detailed accuracy tests for SpectralCoord in the astropy core PR - I'm going to focus on that PR first, but after that let's see if it makes sense to copy them here?

astrofrog:
I'm going to close this, as there are now a number of other changes that should just be ported in one go from the core package.

PR #666, opened 2020-04-23
WIP: Display loaders in RTD using sphinx extensions

### LINKED ISSUE #394: Document the default loaders

javerbukh:
Using [Sphinx extensions](https://www.sphinx-doc.org/en/master/development/tutorials/helloworld.html), I am displaying the loader formats that are printed out with `Spectrum1D.read.list_formats(out=f)` in read the docs.

Resolves #394

### DIFF
--- a/docs/_ext/listformats.py
+++ b/docs/_ext/listformats.py
@@ -1,0 +1,30 @@
+from docutils import nodes
+from docutils.parsers.rst import Directive
+from specutils import Spectrum1D
+
+class ListFormats(Directive):
+
+    def run(self):
+        with open('read_formats.txt', 'w') as f:
+            Spectrum1D.read.list_formats(out=f)
+
+        file = open('read_formats.txt', 'r')
+
+        formats = file.readlines()
+        formats_as_str = ""
+
+        for line in formats:
+            formats_as_str += line
+
+        literal_block_node = nodes.literal_block(text=formats_as_str)
+        file.close()
+        return [literal_block_node]
+
+def setup(app):
+    app.add_directive('listformats', ListFormats)
+
+    return {
+        'version': '0.1',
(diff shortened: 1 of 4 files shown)


javerbukh:
Current way loaders are displayed as of the latest commit. Not the prettiest but it's what I could get done this sprint. There are more intricate ways to do this using different kinds of nodes and I will look into that more when I return to this ticket next sprint. 
Screen Shot 2020-04-23 at 1 52 06 PM

eteq:
@javerbukh - is there a reason why something like the *last* paragraph of https://github.com/astropy/specutils/issues/394#issuecomment-613064207 didn't work here?  

That is, instead of having an extension to output this and hard-coding the `txt` file, have the output of ``Spectrum1D.read.list_formats(out=f)`` directly included in the docstrings that are already coming out from ``.. automodapi::`` directives?

My concern as this stands is that it's not very maintainable and seems to be bypassing the built-in sphinx table formatting.

javerbukh:
@eteq My experience with Sphinx is just about this ticket, so I cannot give a great reason for extension over builtin_frames other than I could figure out how extensions work in the given amount of time, lol. I will say that the `.txt` file is not hardcoded, it is simply the only way to get the output of `list_formats` in a usable way. I can come back to this ticket the sprint after next if we cannot merge it as is for the time being.

javerbukh:
There is a better way to perform this task and the PR that is handling that is #675 . Closing this PR.

PR #667, opened 2020-04-25
SDSS loader updates and extensions

dhomeier:
1st commit has some clean-up of the "spSpec" and "spec" loaders, trying to read the flux units from file where possible, and switching to use just the 1st axis of the WCS for creating `spectral_axis`. The `_identify` functions are extended to allow automatic identification in HDULists and open file handles (the latter need to have been opened in `mode='rb'`, since FITS contents cannot be read from `_io.TextIOWrapper` objects).

2nd commit is adding @weaverba137's "spPlate" loader for `prospect` from the [PyAstro20 hackday](https://github.com/desihub/prospect/commit/ce435b7) to the SDSS loaders.

1. Since `InverseVariance` is now supported, `spec_loader` is now storing its `uncertainty` in the native form (same as `spPlate_loader` does). This avoids division-by-zero issues in creating the `Spectrum1D`, but perhaps needs to be checked if it breaks existing applications of the `SDSS-III/IV` type spectra.

2. The test for reading `spPlate` with the fluxes for all 640 fibres is reading something between 70 and 160 MB in `remote_data`, probably depending on OS caching (actual file size is 60 MB). I was hoping limiting the read to the first 20 or 30 spectra might reduce this a bit at least for the direct read, but since the loader is accessing the first 6 HDUs, reading the entire file is probably unavoidable. We might remove either the direct read or the `tempfile` read, if this is deemed too much network traffic - or is there an option for setting something like `HUGE_DATA` access in the tests?

### DIFF
--- a/specutils/io/default_loaders/sdss.py
+++ b/specutils/io/default_loaders/sdss.py
@@ -73,6 +73,30 @@ def spSpec_identify(origin, *args, **kwargs):
     return is_sdss
 
 
+def spPlate_identify(origin, *args, **kwargs):
+    """
+    Check whether given input is FITS with SDSS spPlate fibre spectral data.
+    This is used for Astropy I/O Registry.
+    """
+    if isinstance(args[2], fits.hdu.hdulist.HDUList):
+        hdulist = args[2]
+    elif isinstance(args[2], _io.BufferedReader):
+        hdulist = fits.open(args[2])
+    else:
+        hdulist = fits.open(args[0], **kwargs)
+
+    # Test telescope keyword and check if primary HDU contains data
+    # consistent with spSpec format
+    is_sdss = (hdulist[0].header['TELESCOP'] == 'SDSS 2.5-M' and
+               hdulist[0].header.get('FIBERID', 0) <= 0 and
+               isinstance(hdulist[0].data, np.ndarray) and
+               hdulist[0].data.shape[0] > 5)
+
+    if not isinstance(args[2], (fits.hdu.hdulist.HDUList, _io.BufferedReader)):
+        hdulist.close()
+    return is_sdss
+
+
 @data_loader(label="SDSS-III/IV spec", identifier=spec_identify, extensions=['fits'])
 def spec_loader(file_obj, **kwargs):
     """
@@ -171,4 +195,60 @@ def spSpec_loader(file_obj, **kwargs):
         hdulist.close()
 
     return Spectrum1D(flux=flux, spectral_axis=dispersion * dispersion_unit,
+                      uncertainty=uncertainty, meta=meta, mask=mask)
+
+
+@data_loader(label="SDSS spPlate", identifier=spPlate_identify, extensions=['fits'])
+def spPlate_loader(file_obj, limit=None, **kwargs):
+    """
+    Loader for SDSS spPlate files, reading flux spectra from all fibres into single array.
+
+    Parameters
+    ----------
+    file_obj: str, file-like, or HDUList
+           FITS file name, object (provided from name by Astropy I/O Registry),
+           or HDUList (as resulting from astropy.io.fits.open()).
+
+    limit : :class:`int`, optional
+        If set, only return the first `limit` spectra in `flux` array.
+
+    Returns
+    -------
+    Spectrum1D
+        The spectra represented by the wavelength solution from the header WCS
+        and the data array of the primary HDU (typically 640 along dimension 1).
+    """
+    if isinstance(file_obj, fits.hdu.hdulist.HDUList):
+        hdulist = file_obj
+    else:
+        hdulist = fits.open(file_obj, **kwargs)
+
+    header = hdulist[0].header
+    meta = {'header': header}
+    wcs = WCS(header).dropaxis(1)
+    if limit is None:
+        limit = header['NAXIS2']
+
+    bunit = header.get('BUNIT', '1e-17 erg / (Angstrom cm2 s)')
+    if 'Ang' in bunit and 'strom' not in bunit:
+        bunit = bunit.replace('Ang', 'Angstrom')
+    flux_unit = Unit(bunit)
+    flux = hdulist[0].data[0:limit, :] * flux_unit
+    uncertainty = InverseVariance(hdulist[1].data[0:limit, :] / flux_unit**2)
+
+    # dispersion along NAXIS1 from the WCS
+    wcs = WCS(header).dropaxis(1)
+    dispersion = wcs.pixel_to_world(np.arange(flux.shape[-1]))
+    # convert out of logspace (default for spSpec/spPlate spectra)?
+    if header.get('DC-Flag', 1) == 1:
+        dispersion = 10**dispersion
+    dispersion_unit = Unit('Angstrom')
+
+    mask = hdulist[2].data[0:limit, :] != 0
+    meta['plugmap'] = Table.read(hdulist[5])[0:limit]
+
+    if not isinstance(file_obj, fits.hdu.hdulist.HDUList):
+        hdulist.close()
+
+    return Spectrum1D(flux=flux, spectral_axis=dispersion*dispersion_unit,
                       uncertainty=uncertainty, meta=meta, mask=mask)
--- a/specutils/io/default_loaders/sdss.py
+++ b/specutils/io/default_loaders/sdss.py
@@ -5,12 +5,13 @@ Loader for SDSS individual spectrum files: spec_ files.
 """
 import os
 import re
+import _io
 
 from astropy.io import fits
 from astropy.table import Table
 from astropy.wcs import WCS
 from astropy.units import Unit, def_unit
-from astropy.nddata import StdDevUncertainty
+from astropy.nddata import StdDevUncertainty, InverseVariance
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


specutils/io/default_loaders/sdss.py
```diff
@@ -113,38 +159,96 @@ def spSpec_loader(file_obj, **kwargs):
...
+    if 'Ang' in bunit and 'strom' not in bunit:
+        bunit = bunit.replace('Ang', 'Angstrom')
+    flux_unit = Unit(bunit)
+    flux = hdulist[0].data[0:limit, :] * flux_unit
+    uncertainty = InverseVariance(hdulist[1].data[0:limit, :] / flux_unit**2)
+
+    # dispersion along NAXIS1 from the WCS
+    wcs = WCS(header).dropaxis(1)
+    dispersion = wcs.pixel_to_world(np.arange(flux.shape[-1]))
+    # convert out of logspace (default for spSpec/spPlate spectra)?
```
weaverba137, line 242: Yes log(lambda) is the default for spPlate, spSpec and spec files.  This is just a SDSS "thing".

specutils/io/default_loaders/sdss.py
```diff
@@ -113,38 +159,96 @@ def spSpec_loader(file_obj, **kwargs):
...
+
+    bunit = header.get('BUNIT', '1e-17 erg / (Angstrom cm2 s)')
+    # fix mutilated flux unit
+    bunit = bunit.replace('/cm/s/Ang', '/ (Angstrom cm2 s)')
+    if 'Ang' in bunit and 'strom' not in bunit:
+        bunit = bunit.replace('Ang', 'Angstrom')
+    flux_unit = Unit(bunit)
+    flux = hdulist[0].data[0, :] * flux_unit
+
+    uncertainty = StdDevUncertainty(hdulist[0].data[2, :] * flux_unit)
```
weaverba137, line 183: For the record, I double-checked the spSpec data model, and indeed, in those files (only) uncertainty is stored as standard deviation, not inverse variance.

specutils/io/default_loaders/sdss.py
```diff
@@ -113,38 +159,96 @@ def spSpec_loader(file_obj, **kwargs):
...
-    dispersion = 10**wcs.all_pix2world(np.vstack((np.arange(data.shape[0]),
-                                                  np.zeros((data.shape[0],)))).T,
-                                       0)[:, 0]
-    # dispersion = 10**hdulist[1].data['loglam']
+    return Spectrum1D(flux=flux, spectral_axis=dispersion * dispersion_unit,
+                      uncertainty=uncertainty, meta=meta, mask=mask)
+
+
+@data_loader(label="SDSS spPlate", identifier=spPlate_identify, extensions=['fits'])
+def spPlate_loader(file_obj, limit=None, **kwargs):
```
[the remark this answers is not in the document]
dhomeier: OK; I had thought if leaving that in it would actually make more sense to provide both `start` and `stop`, or a pair of `limits=[]`, so the user could extract a range of spectra. Only point would be some memory savings, I guess, but at 60 MB for the entire collection of spectra this should not really matter.

specutils/io/default_loaders/sdss.py
dhomeier: The only change here is really the check for `DC-Flag`, which I took from the documentation is the card indicating the logical (although I could not find ultimate confirmation). Anyway the default is left at "1" for "loglinear", but should the loader encounter a spectrum with `DC-Flag = 0`, it would assume linear dispersion.

specutils/io/default_loaders/sdss.py
dhomeier: Yes, I also verified for the reference spSpec in the tests that its `uncertainty` is indeed 1/sqrt of the corresponding spPlate row (although the test itself does not check this, as it would first have to mask the zeros in the two uncertainties).
The more critical part is the spec data, in which uncertainties had been stored as inverse variance, but were previously converted to standard deviations, because apparently at the time no model for `InverseVariance` existed. But since there is no direct conversion between the two available, it remains to be checked if this would break things downstream.

dhomeier:
Are there general rules what type a `Spectrum1D.uncertainty` should have? Of the other loaders, 6dGFS returns a `VarianceUncertainty`, ASDF seems to accommodate all 3 types. Manipulation tools like `resample1d` also have specific provisions for the 3 types, while `analysis.uncertainty` apparently assumes a `StdDevUncertainty`.

weaverba137:
@dhomeier , it may be that the specutils core team is waiting for you to fix the failed tests before reviewing it themselves.

dhomeier:
> @dhomeier , it may be that the specutils core team is waiting for you to fix the failed tests before reviewing it themselves.

Thanks, I assumed it was simply sitting in the pipeline since there are still some older PRs currently under review. The failed tests are in `jwst_s2d_multi_reader` which does not touch any files changed here; they also have not changed in master since, so I am not sure if a rebase would be of any use. Could be conflicts with other package versions as it only fails on this single run, or are these tests only carried out with the `spacetelescope/jwst` package installed?
Pinging @jdavies-st who should know more about the background.

jdavies-st:
Hmm, the builds on the `master` branch today are passing without problem.

but your build is pulling `jwst-0.15.1` and `asdf-2.6.0`, which could be a problem.

https://travis-ci.org/github/astropy/specutils/jobs/679542140#L895

Meanwhile the passing tests on master today are pulling `jwst-0.16.0` and `asdf-2.5.2`, which is expected to work.

https://travis-ci.org/github/astropy/specutils/jobs/686619861#L891

I've kicked the tires on the job to see if running it again gets it to work.

I think it would be good idea to make this Travis job an allowed failure.

dhomeier:
OK, I propose to ignore the failure here for now.

jdavies-st:
Well, kicking the tires made it pass.  As I expected.

So yeah, let's ignore. And I will file a PR to make this job in the Travis matrix an allowed failure.

jdavies-st:
And thanks for the heads-up.

specutils/io/default_loaders/sdss.py
```diff
@@ -26,29 +27,74 @@
...
-        # consistent with spSpec format
-        return (hdulist[0].header['TELESCOP'] == 'SDSS 2.5-M' and
-                isinstance(hdulist[0].data, np.ndarray) and
-                hdulist[0].data.shape[0] == 5)
+    if isinstance(args[2], fits.hdu.hdulist.HDUList):
+        hdulist = args[2]
+    elif isinstance(args[2], _io.BufferedReader):
+        hdulist = fits.open(args[2])
+    else:
+        hdulist = fits.open(args[0])
```
nmearl, line 62: What makes the location of the hdulist in `args` non-deterministic?

jdavies-st: `asdf 2.6.0` added a big validator efficiency on read.  `jwst.datamodels` specifically does not want to use that, and there was a period of about 2 weeks were our last release was not compatible with the latest `asdf 2.6.0`.  I.e. serialized files would be written out incorrectly.

specutils/io/default_loaders/sdss.py
```diff
@@ -26,29 +27,74 @@
...
+        hdulist = fits.open(args[0], **kwargs)
+
+    # Test if fits has extension of type BinTable and check for spec-specific keys
+    is_sdss = (hdulist[0].header['TELESCOP'] == 'SDSS 2.5-M' and
+               hdulist[0].header.get('FIBERID', 0) > 0 and
+               len(hdulist) > 1 and
+               isinstance(hdulist[1], fits.BinTableHDU) and
+               hdulist[1].header['TTYPE3'] == 'ivar')
+
+    if not isinstance(args[2], (fits.hdu.hdulist.HDUList, _io.BufferedReader)):
```
nmearl, line 47: @weaverba137 is there a reason the hdulist is kept open at all?

specutils/io/default_loaders/sdss.py
dhomeier: It's not really non-deterministic, but possibly a bit confusing from the calling sequence in `astropy.io.registry.identify_format` – according to the docs and as it appears also from the code, it should be calling the identifier with
`_identifiers[(data_format, data_class)](origin, path, fileobj, *args, **kwargs)`
thus the filename will be passed in `args[0]`; for `open(filename, mode='rb')` I found it to pass `args[:3]` as
```
None
<_io.BufferedReader name='spec-1323-52797-0012.fits'>
<_io.BufferedReader name='spec-1323-52797-0012.fits'>
```
whereas for the hdulist this becomes
```
None
None
[<astropy.io.fits.hdu.image.PrimaryHDU object at 0x1102962e0>, <astropy.io.fits.hdu.table.BinTableHDU object at 0x1103552e0>, <astropy.io.fits.hdu.table.BinTableHDU object at 0x110355520>, <astropy.io.fits.hdu.table.BinTableHDU object at 0x110355760>]
```
so apparently the FITS object is moved to `args[2]`, while a regular `fileobj` comes with copies in `args[1]` and `args[2]`. But perhaps it would be more consistent with the `registry` documentation to make the 3rd case
```
    elif isinstance(args[1], _io.BufferedReader):
        hdulist = fits.open(args[1])
```
(it works both ways).

specutils/io/default_loaders/sdss.py
dhomeier: Just as with the handling in #637 if an open file-like object is passed to the reader (same for the identifiers of course), it is left/returned in its original state.

nmearl: Ah, right. I suppose if it's closed in the identified check, it will not be available to the actually loader function.

specutils/io/default_loaders/sdss.py
nmearl: Strange. I'm not too worried about it, but it seems that the specutils documentation on custom loaders ought to be updated to reflect some of these nuances for new users.

nmearl:
Thanks for the contribution @dhomeier!

specutils/io/default_loaders/sdss.py
dhomeier: Yes, I think I would have had a hard time figuring any of this out without the earlier primer you gave me on using `identify`, and the examples in #637. https://docs.astropy.org/en/stable/io/registry.html#using-astropy-io-registry does not give a great lot of detail either, but I think summarising the main aspects relevant to `specutils` usage would make a good addition to the docs.
Also in principle most other loaders could probably also be enabled to auto-detect fileobj input.

PR #669, opened 2020-04-28
Uncertainty handling for line_flux

nmearl:
This PR includes uncertainty handling in the line flux analysis function. Based on conversations, this is may be a too generic solution to this particular problem.

In this case, we can't rely on the numpy functions to calculate e.g. sums due to needing the propagation machinery for uncertainties in the `NDData` classes.


specutils/analysis/flux.py
```diff
@@ -99,7 +99,16 @@ def _compute_line_flux(spectrum, regions=None):
...
+    if isinstance(calc_spectrum.uncertainty, VarianceUncertainty):
+        stddev_uncert = np.sqrt(calc_spectrum.uncertainty.array)
+    elif isinstance(calc_spectrum.uncertainty, InverseVariance):
+        stddev_uncert = np.sqrt(1/calc_spectrum.uncertainty.array)
+    else:
+        stddev_uncert = calc_spectrum.uncertainty.array
+
     line_flux = np.sum(calc_spectrum.flux[1:] * avg_dx)
+    line_flux.uncertainty = np.sqrt(
+        np.sum((stddev_uncert[1:] * avg_dx.value) ** 2)) * line_flux.unit
```
dhomeier, line 111:
```suggestion
        stddev_sq = calc_spectrum.uncertainty.array
    elif isinstance(calc_spectrum.uncertainty, InverseVariance):
        stddev_sq = 1/calc_spectrum.uncertainty.array
    else:
        stddev_sq = calc_spectrum.uncertainty.array**2

    line_flux = np.sum(calc_spectrum.flux[1:] * avg_dx)
    line_flux.uncertainty = np.sqrt(
        np.sum(stddev_sq[1:] * avg_dx.value**2) * line_flux.unit
```
You could save the `sqrt` in 2 of the cases.

What about zeros in an `InverseVariance` uncertainty? Might be unlikely within a line region, but perhaps an early check should be done before returning `inf` for the whole result.

dhomeier: This pertains to the existing flux integration as much as to the uncertainty:
with the new `SpectralAxis` it should be more accurate to use
` avg_dx = np.abs(np.diff(calc_spectrum.bin_edges))`
to integrate over the full `calc_spectrum.flux` resp. uncertainty (I am also a bit confused where the "average" is supposed to come in there – does it mean average over each bin width?).

nmearl: Good catch, thanks.

> What about zeros in an InverseVariance uncertainty?

Hmm, we could replace with a quick check for 0 values in the `InverseVariance` case and raise a warning, and then switch to using a `nansum`. Although, I think the user should be entirely aware of this fact and therefore specutils should not return reasonable data values at all in order to have the user deal with the zeros in their uncertainty.

nmearl: > I am also a bit confused where the "average" is supposed to come in there – does it mean average over each bin width?

`avg` may be a bit of misnomer -- it's essentially the dispersion bin sizes used to weight the flux contribution. You raise a good point though, this is really only necessary when e.g. the dispersion is in log space, in which case, it's more accurate to use the bin edges instead of the centers.

dhomeier: Had not even thought of the differences occurring in log space, but if the exact `bin_edges` are available, that is certainly a further improvement (`SpectralAxis._edges_from_centers` does not specifically seem to discriminate log vs. linear scales). But I was primarily concerned about the first element of flux/uncertainty being simply discarded!

eteq:
Ok I don't understand why none of the tests ran here.  But I can confirm I ran it locally and the relevant tests passed.

nmearl:
Runs fine for me too. Merging. Thanks @eteq!

PR #670, opened 2020-05-02
Ensure that existing manipulation and analysis functions use  `mask` attribute

ibusko:


specutils/analysis/uncertainty.py
```diff
@@ -83,8 +83,13 @@ def _snr_single_region(spectrum, region=None):
...
 
-    flux = calc_spectrum.flux
-    uncertainty = calc_spectrum.uncertainty.array * spectrum.uncertainty.unit
+    if hasattr(spectrum, 'mask') and spectrum.mask is not None:
+        flux = calc_spectrum.flux[~spectrum.mask]
+        uncertainty = calc_spectrum.uncertainty.array[~spectrum.mask] * \
+                      spectrum.uncertainty.unit
+    else:
+        flux = calc_spectrum.flux
+        uncertainty = calc_spectrum.uncertainty.array * spectrum.uncertainty.unit
```
nmearl, line 92:
```suggestion
        uncertainty = calc_spectrum.uncertainty.quantity
```

specutils/analysis/uncertainty.py
```diff
@@ -83,8 +83,13 @@ def _snr_single_region(spectrum, region=None):
     else:
         calc_spectrum = spectrum
 
-    flux = calc_spectrum.flux
-    uncertainty = calc_spectrum.uncertainty.array * spectrum.uncertainty.unit
+    if hasattr(spectrum, 'mask') and spectrum.mask is not None:
+        flux = calc_spectrum.flux[~spectrum.mask]
+        uncertainty = calc_spectrum.uncertainty.array[~spectrum.mask] * \
+                      spectrum.uncertainty.unit
```
nmearl, line 89:
```suggestion
        uncertainty = calc_spectrum.uncertainty.array[~spectrum.mask].quantity
```

specutils/analysis/flux.py
```diff
@@ -121,6 +129,15 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
...
     line_flux = _compute_line_flux(spectrum, regions)
 
+    # Account for the existence of a mask. A correction factor
+    # is applied to the continuum value to account for the
+    # data points that are not included in the line flux due to
+    # masking.
+    if hasattr(spectrum, 'mask') and spectrum.mask is not None:
+        valid = np.sum(np.where(calc_spectrum.mask, 0., 1.))
+        factor = valid / len(spectral_axis)
+        continuum *= factor
```
nmearl, line 139: I'm confused why this is necessary if we don't actually truncate the line flux calculation (that is, we instead replace with zeros).

specutils/analysis/uncertainty.py
ibusko: The correct construct is: 
```
uncertainty = calc_spectrum.uncertainty.quantity[~spectrum.mask]
```

specutils/analysis/flux.py
ibusko: This is not very clear to me either. My idea was that, if we remove data points from the line flux summation (by setting their fluxes to zero), the continuuum value should be corrected by a factor that reflects the flux that was removed. If we don't do that, we can get into situations where the EW of an emission line, say, can even become positive. Again, I am not sure how to apply this correction, if indeed a correction exists or is necessary.

nmearl: @eteq can you expand on the necessity of this?

specutils/analysis/flux.py
```diff
@@ -99,7 +99,15 @@ def _compute_line_flux(spectrum, regions=None):
     # Average dispersion in the line region
     avg_dx = (np.abs(np.diff(calc_spectrum.spectral_axis))).quantity
 
-    line_flux = np.sum(calc_spectrum.flux[1:] * avg_dx)
+    # Account for the existence of a mask.
+    if hasattr(spectrum, 'mask') and spectrum.mask is not None:
+        # Cannot use unmasked values because of average dispersion.
+        # Masked values must enter sum calculation valued as zeros.
+        flux_ = np.where(calc_spectrum.mask, 0, calc_spectrum.flux)[1:]
```
jdavies-st: So the `.mask` attribute is a boolean array where `True` means it is masked.  Because of this, it is probably a little clearer and more efficient (by a factor of 2) to do something like

```
flux_ = calc_spectrum.flux[~calc_spectrum.mask][1:]
```

And of course instead of returning an array of the same size with zeros in the masked areas, it is returning only the non-masked array values, which will also speed up the summation:

```
In [2]: a = np.arange(15).astype(np.float)

In [3]: mask = a > 9

In [4]: a[~mask]
Out[4]: array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])

In [5]: np.where(mask, 0, a)
Out[5]: array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9., 0., 0., 0., 0., 0.])

In [6]: %timeit a[~mask]
765 ns ± 8.23 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [7]: %timeit np.where(mask, 0, a)
1.6 µs ± 38.1 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

I see you've used the above method in other areas of the code.  👍

specutils/analysis/flux.py
```diff
@@ -121,6 +129,15 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None):
 
     line_flux = _compute_line_flux(spectrum, regions)
 
+    # Account for the existence of a mask. A correction factor
+    # is applied to the continuum value to account for the
+    # data points that are not included in the line flux due to
+    # masking.
+    if hasattr(spectrum, 'mask') and spectrum.mask is not None:
+        valid = np.sum(np.where(calc_spectrum.mask, 0., 1.))
+        factor = valid / len(spectral_axis)
```
jdavies-st, line 138: Since `.mask` is a boolean mask, something like

```
factor = (~calc_spectrum.mask).sum() / len(spectral_axis)
```

can be a little clearer and more efficient.

specutils/analysis/flux.py
nmearl: While this is true, it can have adverse affects on calculations that require the length of the array to be consistent (e.g. means):

```python
In [10]: flux[~mask]
Out[10]:
array([0.44012434, 0.18503118, 0.75635908, 0.80137293, 0.88490978,
       0.37460318])

In [11]: np.where(mask, 0, flux)
Out[11]:
array([0.44012434, 0.18503118, 0.        , 0.        , 0.75635908,
       0.80137293, 0.        , 0.88490978, 0.        , 0.37460318])

In [12]: np.mean(flux[~mask])
Out[12]: 0.5737334148059349

In [13]: np.mean(np.where(mask, 0, flux))
Out[13]: 0.3442400488835609
```

The route implemented now is certainly safer and more future-proof (since the `_compute_line_flux` is used as an intermediary in a lot of analysis functions), but this _could_ be circumvented by just being Very Careful if performance really is a huge bottleneck.

ibusko:
The two last commits above implement the #2 solution from @eteq comment.

specutils/analysis/flux.py
jdavies-st: In your example above though, only one mean is correct.  Masking doesn't mean treat the flux values as though they were 0, it means ignore them in the calculation.  By replacing the masked indices with a value of zero and then taking the mean, you're not ignoring them, but taking the mean of lots of zeros along with the good values, which gives the incorrect result, likely too low if the good values are positive.

Agree that one has to be careful.

Generally, each analysis function needs to take the input mask and do what is correct for that function I suspect.  And this will vary by function, as some might also depend on the binning or array length.  Because of this, I don't think it is a good idea to carry around various private class attributes that already have the mask applied in a particular way to the array.  Better to apply the mask, function-by-function as needed.

specutils/analysis/flux.py
```diff
@@ -7,7 +7,7 @@
 
 import numpy as np
 from .. import conf
-from ..manipulation import extract_region
+from ..manipulation import extract_region, LinearInterpolatedResampler
 from .utils import computation_wrapper
 import astropy.units as u
 from astropy.stats import sigma_clip
```
orifox, line 13:
```suggestion
from .utils import computation_wrapper
import astropy.units as u
from astropy.stats import sigma_clip
```

nmearl:
@ibusko Can you rebase this? Are you working on anything else in this PR?

ibusko:
I am not working on this PR. I can do the rebase, but on next week. I am attending this one-day class today, and off tomorrow.

ibusko:
@nmearl I spoke too soon. The conflicts are being caused by a change in master that totally overwhelmed the changes I had made on the branch. I am back at square one (almost) with this implementation. I moved the ticket back to "in progress" status.

ibusko:
@nmearl the conflicts are resolved now. I don't see any commits that should be squashed.

specutils/analysis/flux.py
```diff
@@ -71,6 +84,10 @@ def equivalent_width(spectrum, continuum=1, regions=None):
         will be assumed, otherwise units are required and must be the same as
         the ``spectrum.flux``.
 
+    mask_interpolation : `~specutils.manipulation.LinearInterpolatedResampler`
+        Interpolator class used to fill up the gaps in the spectrum's flux
+        array, when the spectrum mask is not None.
```
nmearl, line 89: Just some clarification. It was not immediately obvious why we'd be interpolating. But it's because there may be a excise operation that changes the shape of the flux/spectral axis without changing the shape of mask (and thus the mask would not be able to be applied in such cases).

```suggestion
    mask_interpolation : `~specutils.manipulation.LinearInterpolatedResampler`
        Interpolator class used to fill up the gaps in the spectrum's flux
        array after an excise operation to ensure the mask shape can always be applied when 
        the spectrum mask is not None.
```

nmearl:
This looks good. You have a few merge commits, however, so you may want to rebase.

Issue #671, opened 2020-05-03
specutils should not globally set enabled equivalencies


astrofrog:
Currently, importing specutils causes spectral equivalencies to be enabled by default for the rest of the Python session:

https://github.com/astropy/specutils/blob/ca093ea8757d17a728f1ee1adafc04bbeff083c4/specutils/spectra/spectrum1d.py#L18

This should be removed, and the equivalencies should only be enabled temporarily in the context where they are used (currently the astropy test suite breaks if specutils is installed because of this: https://github.com/astropy/astropy/issues/10283)

nmearl:
Fixes in #673.

PR #673, opened 2020-05-07
Restrict `u.spectral` equivalency to contexts where it is required

dhomeier:
Fix for #671

Fairly pedestrian implementation; @mhvk your proposed solution to use `setup_class`/`teardown_class` environments would require creating classes for the tests in question, right?

Leaving this for consideration now; perhaps some discussion is due if the tests now still test what they are supposed to test (i.e. would a user in a normal work environment expect those equivalencies to be in effect as well)?

### DIFF
--- a/specutils/spectra/spectral_region.py
+++ b/specutils/spectra/spectral_region.py
@@ -155,9 +155,10 @@ class SpectralRegion:
     def _valid(self):
 
         # Lower bound < Upper bound for all sub regions in length physical type
-        sub_regions = [(x[0].to('m'), x[1].to('m'))
-                       if x[0].unit.is_equivalent(u.m) else x
-                       for x in self._subregions]
+        with u.set_enabled_equivalencies(u.spectral()):
+            sub_regions = [(x[0].to('m'), x[1].to('m'))
+                           if x[0].unit.is_equivalent(u.m) else x
+                           for x in self._subregions]
 
         if any(x[0] >= x[1] for x in sub_regions):
             raise ValueError('Lower bound must be strictly less than the upper bound')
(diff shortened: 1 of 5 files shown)


dhomeier:
Concerning the failed doctests - need to think of a clean way to make this work in the docs. I am wondering if using equivalencies is really the best way to tackle the problem with lower/upper boundaries in the spectral ranges...

dhomeier:
Fixing the code rather than the tests hopefully will hold now. Concerning the failed `remote_data` tests, I got 404 errors from das.sdss.org yesterday roughly 2 out of 3 times.

The local astropy tests seem to succeed now at least without any failures potentially related to https://github.com/astropy/astropy/issues/10283.
@nmearl I've had a second look for any other equivalency uses that would depend on `u.spectral` set and might not be fully covered by the tests; `u.is_equivalent` checks for `spectral_axis` type quantities in `fitmodels`, `extract_spectral_region` and `spectrum_mixin` are already doing it the Right Way™*. Can you think of any uses of `Unit.to()` that might still depend on the equivalency?

*Rather, either one of the two right ways – had forgotten that `equivalency` can also be passed as an option; for the single operation uses this may be preferable to the `with` context.

mhvk:
@dhomeier - answering without looking: yes, `setup_class` and `teardown_class` require test classes, but there are also module-level equivalents.

dhomeier:
@mhvk thanks; since there are only three instances which are also easily managed with the kwarg, I think there is no need for a major restructuring of the tests at this point.

nmearl:
> I've had a second look for any other equivalency uses that would depend on u.spectral set and might not be fully covered by the tests; u.is_equivalent checks for spectral_axis type quantities in fitmodels, extract_spectral_region and spectrum_mixin are already doing it the Right Way™*. Can you think of any uses of Unit.to() that might still depend on the equivalency?

It seems you've covered all the bases here. The failures are due to inconsistencies with upstream currently (since we moved `SpectralCoord` into astropy proper).

I'm tempted to say that this is good as it stands, as it passes all my use cases currently.

nmearl:
Thanks @dhomeier!

PR #674, opened 2020-05-07
Backport SpectralCoord from astropy core and use SpectralCoord from astropy>=4.1

astrofrog:
This backports the ``SpectralCoord`` class now in astropy core to a ``specutils.extern`` folder so that specutils can continue to work with astropy<4.1. This backport is necessary because there are a number of  bugs (some causing issues of 100s of km/s in transformations) in the current implementation, and also to ensure consistency with the core package going forward.

Currently the tests do not pass, because the ``SpectralCoord`` class in astropy does not allow dimensionless quantities. I think that is the right behavior in that class but I see why one might want to allow it in ``SpectralAxis``, so I wonder whether the right fix is to add a private option to the ``SpectralQuantity`` class, called e.g. ``_allow_dimensionless``, default it to ``False`` there, but have ``SpectralAxis`` define it as ``True``?

This is an urgent issue to address, since if we want to do this we should get it in before the final 4.1 release of the core package.


specutils/analysis/flux.py
```diff
@@ -97,7 +97,7 @@ def _compute_line_flux(spectrum, regions=None):
         calc_spectrum = spectrum
 
     # Average dispersion in the line region
-    avg_dx = (np.abs(np.diff(calc_spectrum.spectral_axis))).quantity
+    avg_dx = (np.abs(np.diff(calc_spectrum.spectral_axis)))
```
astrofrog, line 100: Just FYI, this change is because most ufuncs applied to ``SpectralCoord`` now return ``Quantity`` except for a few explicitly specified cases that make physical sense.

specutils/spectra/spectrum1d.py
```diff
@@ -7,7 +7,7 @@
 from astropy.nddata import NDDataRef
 from astropy.utils.decorators import lazyproperty
 from .spectrum_mixin import OneDSpectrumMixin
-from .spectral_coordinate import SpectralCoord
+from ..extern.spectralcoord import SpectralCoord
```
nmearl, line 10:
```suggestion
```

This can be removed since we don't use `SpectralCoord` in `Spectrum1D` anymore (it's used through `SpectralAxis`).

eteq:
On the main (Astroy 4.1-critical) item of dimensionless units: I think instead of dimensionless, we really want `u.pixel` to explicitly state it's a spectral coord in pixel units.

astrofrog:
I've backported the fixes in https://github.com/astropy/astropy/pull/10339 and https://github.com/astropy/astropy/pull/10340 (but note that those PRs aren't merged yet)

specutils/spectra/spectrum1d.py
astrofrog: Fixed

astrofrog:
Actually as suggested by @mhvk we can just allow pixel units in SpectralAxis and not in SpectralQuantity/SpectralCoord (which makes sense since we don't allow pixel in say SkyCoord or Angle). That works nicely.

specutils/extern/spectralcoord/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+###############################################################################
+
+from distutils.version import LooseVersion
+
+from astropy import __version__ as astropy_version
+
+if LooseVersion(astropy_version) >= '4.1':
+    from astropy.coordinates import SpectralCoord
+else:
+    from .spectral_coordinate import SpectralCoord
```
eteq, line 47: I'm not sure if this is a mis-use of "deprecation warning", but I wonder if we want to add now something like "in the future you will need to use Astropy>=4.1 ? Because at some point we will want to stop maintaining this.

specutils/io/asdf/tags/spectra.py
```diff
@@ -9,7 +9,8 @@
                            tagged_tree_to_custom_tree)
 from astropy.io.misc.asdf.tags.unit.unit import UnitType
 
-from ....spectra import SpectralCoord, Spectrum1D, SpectrumList
+from ....extern.spectralcoord import SpectralCoord
```
eteq, line 12: Isn't there going to need to be a `SpectralCoordType` in astropy to get this to work, too? (and then that means it should be imported from there instead?)

astrofrog: Yes good point, will fix that

specutils/extern/spectralcoord/__init__.py
astrofrog: Won't we just bump the required version to astropy 4.1 at that point, and that's what users will get when they install specutils?

specutils/io/asdf/tags/spectra.py
astrofrog: So I think the correct approach is to just not define the Type class when astropy 4.1 is available since the ASDF tag and schemas are defined in astropy core for the class we then use in specutils. I'll fix that.

astrofrog:
@jdavies-st I'm seeing failures in the ASDF tests when running with astropy dev, and I was wondering whether you could investigate since you are much more familiar than me with ASDF? Basically with astropy dev (and in future 4.1) we'll want to be using the schema from astropy core and not defining it in specutils, which is what I tried to do here.

astrofrog:
Hmm on the other hand the CI is passing here, maybe we can merge this and then someone could investigate the ASDF failures with astropy 4.1 in a separate PR?

nmearl:
I imagine that once #562 is in, this failure will come up in the testing. We could just say "fixing the failing 4.1 builds should be handled there"?

specutils/extern/spectralcoord/__init__.py
eteq: Hmm.  I see your point but I think this is a special case.  That said I don't think we want to block this for that reason, so I guess I'm fine with just merging this without the warning and I can do a separate PR for just this topic.

eteq:
This touches a fair amount of stuff so to keep it from lingering on, @nmearl and @astrofrog and I agreed out-of-band it's better to address the asdf/astropy-dev problem as a separate issue.  So I'm going to ahead and approve and merge as it stands.

nmearl:
@eteq, @astrofrog Just a follow up, the issue seems to be solved by adding

```
      - $ref: "http://astropy.org/schemas/astropy/coordinates/spectralcoord-1.0.0"
```

to the `spectral_axis` portion of the `Spectrum1D` asdf schema. But this requires that 4.1 of astropy is released. A bit of catch-22.

astrofrog:
@nmearl - yeah, if we require astropy 4.1 we can get rid of all the schema stuff for spectralcoord anyway :-/

PR #675, opened 2020-05-07
Display list of loaders in specutils RTD

javerbukh:
Automatically display a list of loaders in RTD as provided by spectrum1d.read.list_formats().

Screen Shot 2020-05-07 at 12 04 34 PM

### DIFF
--- a/docs/spectrum1d.rst
+++ b/docs/spectrum1d.rst
@@ -72,6 +72,18 @@ astroquery, if the user has that package installed:
      >>> Spectrum1D.read(specs[0], format="SDSS-III/IV spec") # doctest: +SKIP
      <Spectrum1D(flux=<Quantity [30.596626,...]...>
 
+
+List of Loaders
+~~~~~~~~~~~~~~~
+
+The `~specutils.Spectrum1D` class has built-in support for various input and output formats.
+A full list of the supported formats is shown in the table below.
+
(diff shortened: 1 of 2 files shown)


specutils/spectra/list_of_loaders.py
```diff
@@ -0,0 +1,34 @@
...
+
+# Combine elements to create formatted table
+# which can be displayed using `automodapi`
+formatted_table = line_of_equals + "\n"
+for line in split_list:
+    formatted_table += line + "\n"
+formatted_table += line_of_equals + "\n"
+
+__doc__ = """
+List of loaders is as follows:
```
eteq, line 30: Maybe this line should be moved up into the rst file?  That is, the docstring here could be *just* the table, and any context like this test would be in the rst

javerbukh:
Currently, the local sphinx build looks like:
Screen Shot 2020-05-13 at 1 01 10 PM

However, when building this branch in RTD, it looks like:
Screen Shot 2020-05-13 at 1 06 10 PM

I am inclined to go with the local sphinx build being the correct version since there was a RTD error that happened to me earlier today and it seems like it is stuck building an older commit, where as the local sphinx version is the latest commit. In that case, this should be ready for a final review @eteq .

javerbukh:
@eteq @nmearl All tests are now passing and all requested changes have been addressed. This should be ready for a merge.

Issue #583, opened 2020-02-10
fit_continuum:  implement continuum fitting windows 


PatrickOgle:
The 'window'  and 'weights' parameter in specutils.fit_continuum should be implemented to allow users to indicate which spectral regions to be used in fitting, and what weights should be given to these regions.

PR #681, opened 2020-05-19
Implement the 'windows' parameter in fit_continuum

ibusko:
#583


specutils/tests/test_continuum.py
```diff
@@ -59,3 +62,91 @@ def test_continuum_calculation():
...
+
+    # Smooth in the same way fit_generic_continuum does.
+    spectrum_smoothed = median_smooth(spectrum, 3)
+
+    # Window selects the first half of the spectrum.
+    g1_fit = fit_continuum(spectrum_smoothed, window=(0.*u.um, 5.*u.um))
+
+    spectrum_normalized = spectrum / g1_fit(spectrum.spectral_axis)
+
+    y_continuum_fitted_expected = np.ones(shape=(spectrum_normalized.spectral_axis.shape))
```
nmearl, line 138: Could we instead check against the results of a full fit?

```suggestion
   g1_fit_orig = fit_continuum(spectrum_smoothed)
   y_continuum_full = spectrum / g1_fit_orig(spectrum.spectral_axis)
   y_continuum_fitted_expected = y_continuum_full[(spectrum.spectral_axis >= 0 * u.um) & (spectrum.spectral_axis <= 5 * u.m)]
```

specutils/tests/test_continuum.py
```diff
@@ -59,3 +62,91 @@ def test_continuum_calculation():
...
+    # Check that a full width window recovers the original, non-windowed fit.
+    g1_fit = fit_continuum(spectrum_smoothed, window=(0.*u.um, 10.*u.um))
+
+    spectrum_normalized = spectrum / g1_fit(spectrum.spectral_axis)
+
+    y_continuum_fitted_expected = np.array([1.15139925, 0.98509363, 0.73700614, 1.00911864, 0.913129,
+                                            0.93145533, 0.94904202, 1.04162879, 0.90851397, 0.9494352,
+                                            1.07812394, 1.06376489, 0.98705237, 0.94569623, 0.83502377,
+                                            0.91909416, 0.89662208, 1.01458511, 0.96124191, 0.94847744])
+
```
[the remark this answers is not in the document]
ibusko: Yes, this is better (I used the comparison with explicit values just because that's how the test over spectral regions was coded).

specutils/tests/test_continuum.py
ibusko: When fitting over the entire spectrum, the fit gets very bad because of the narrow line at 6.3 um. So it doesn't seem reasonable to use that bad fit as a comparison reference. Unless we want to decrease the tolerance to values in the range of 0.2 or something like that.

ibusko:
@PatrickOgle  I added a test that checks the fit on both sides of the Gaussian, using a constant continuum and no noise. It works as expected. I didn't try your scripts though, to see if I could spot the problem.

nmearl:
@PatrickOgle I notice in your example code that you're using two different units in your `SpectralRegion` creation: Angstroms in your lower bound, and nanometers in your upper bound. This would then always get the peak if we translate both to Angstrom.

PatrickOgle:
@nmearl I fixed that bug, and still get the misfitting
mis-fit_continuum

PatrickOgle:
Interestingly, I get a better fit if I only use region2:
fitwindows=[region2]
fit_region2_only

PatrickOgle:
@ibusko While it is understandable that using different fitting regions to either side gives different fits, I would expect that jointly fitting both regions would give a good fit.  Am I specifying fitwindows correction for the two-window case as?:

region1 = SpectralRegion(5125.*u.AA, 5175.*u.AA)
region2 = SpectralRegion(5225.*u.AA, 5275.*u.AA)
fitwindows=[region1,region2]
c_fit=fitting.fit_continuum(spec1d, model=Linear1D(0,0), exclude_regions=None, window=fitwindows, weights=None)

nmearl:
@ibusko I'm sensing something off in this code. If we look [here](https://github.com/astropy/specutils/pull/681/files#diff-e88bf14dcd8e8a539c0018d2cd451127R349), it looks like only a single `SpectralRegion` is getting fed to the `_fit_lines` function if `window` is specified as a list of `SpectralRegion`s. Can you verify?

ibusko:
Indeed, it appears that the underlying function `fit_lines` in `fitmodels.py` does not accept multiple windows.

ibusko:
I believe that's the reason I stopped short of trying a full continuum fit, when I was adding unit tests. The best I could do with this limitation is to check the red and blue sides separately. I apparently forgot that when creating the test with a tuple of windows.

ibusko:
Line 439 in `fitmodels.py` is where an exception gets raised when the window parameter is not a tuple of 2 quantities, but a tuple of two tuples (or two SpectralRegions).

PatrickOgle:
Can fitmodels.py be enhanced to handle a list of multiply Spectral Regions?

PatrickOgle:
It appears to work now with two windows.  Has that been implemented? 
fit_continuum_2windows

ibusko:
Function `fit_continuum` relies on function `fit_lines`, which in turn relies on private function `_fit_lines`. The top level `fit_lines` includes machinery to associate windows to spectral components in the model. That is, it requires that, in order to be able to pass a list of windows to it, the model also has to have a number of components identical with the number of windows in the list.

This works fine when fitting compound models, but may fail when fitting continua. Because, most often than not, the user wants to fit a single continuum model (say, a polynomial) throughout several windows that isolate regions devoid of spectral lines. In a sense , it's the exact opposite of what `fit_lines` is meant to do.

The actual handling of the windows (that is, picking up the correct indices in numpy arrays) is done by the underlying `_fit_lines` function. This suggests that fitting continua should perhaps call that function directly? 

In any case, this requires some more thought regarding architectural aspects of the software.

ibusko:
Closing and creating a new one to resolve conflicts (my fork got busted). Will preserve the pointer to this PR so the discussions remain accessible.

PR #698, opened 2020-07-02
Fit continuum over multiple windows.

ibusko:
This PR is in lieu of PR #681 , which got busted when my fork got busted. Discussions associated to this work can be found in #681 .

The implementation was actually simpler than it seemed at first. The code to fit a single model (the continuum) over multiple regions is already in place. To use it, it is just a matter of passing a single window to `fit_continuum`, using a single instance of `SpectralRegion`, but initialized with a tuple of regions. It is not enough to pass a sequence of tuples to the `window` parameter. An explicit instance of `SpectralRegion` does the trick.

### DIFF
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -679,13 +679,9 @@ fitted continuum, which returns a new object:
 
 
 When fitting over a specific wavelength region of a spectrum, one
-should use the `window` parameter to specify the region. This
-parameter accepts one single instance of a region. It is usually
-specified as a tuple of two wavelength values, but in cases where
-one wants to fit over multiple, disjoint regions, it is possible
-to pass to the parameter a single instance of a SpectralRegion.
-In this case, a Spectral Region can have multiple subregions, defined
-as a tuple of tuples in the constructor call:
+should use the `window` parameter to specify the region. Windows
+can be comprisewd of more than one wavelength interval; each interval
+is specified by a sequence:
 
 .. plot::
     :include-source:
@@ -697,7 +693,6 @@ as a tuple of tuples in the constructor call:
     import astropy.units as u
 
     from specutils.spectra.spectrum1d import Spectrum1D
-    from specutils.spectra import SpectralRegion
     from specutils.fitting.continuum import fit_continuum
 
     np.random.seed(0)
@@ -708,7 +703,7 @@ as a tuple of tuples in the constructor call:
 
     spectrum = Spectrum1D(flux=y * u.Jy, spectral_axis=x * u.um)
 
-    region = SpectralRegion(((4.*u.um, 5.*u.um), (8.*u.um, 10.*u.um)))
+    region = [(4.*u.um, 5.*u.um), (8.*u.um, 10.*u.um)]
 
     fitted_continuum = fit_continuum(spectrum, window=region)
 
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -689,7 +689,7 @@ is specified by a sequence:
     :context: close-figs
 
     import numpy as np
-
+import matplotlib.pyplot as plt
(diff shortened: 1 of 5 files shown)


nmearl:
Just to clarify, is the behavior different when doing

```python
fit_continuum(spectrum, window=[(0.*u.um, 5.*u.um),(8.*u.um, 10.*u.um)])
```

Vs.

```python
fit_continuum(spectrum, window=SpectralRegion((0.*u.um, 5.*u.um),(8.*u.um, 10.*u.um)))
```

ibusko:
Yes, it’s different. In the first case, it only picks up the first tuple.


On Jul 2, 2020, at 1:13 PM, Nicholas Earl <notifications@github.com<mailto:notifications@github.com>> wrote:



Just to clarify, is the behavior different when doing

fit_continuum(spectrum, window=[(0.*u.um, 5.*u.um),(8.*u.um, 10.*u.um)])

Vs.

fit_continuum(spectrum, window=SpectralRegion((0.*u.um, 5.*u.um),(8.*u.um, 10.*u.um)))

nmearl:
I think we want to have the first case be a shorter version of the second. That is, if provided `window=[(0.*u.um, 5.*u.um),(8.*u.um, 10.*u.um)]`, we should internally convert that to `window=SpectralRegion((0.*u.um, 5.*u.um),(8.*u.um, 10.*u.um))`.

ibusko:
Sounds OK by me. I will work on it.

On Jul 2, 2020, at 1:19 PM, Nicholas Earl <notifications@github.com<mailto:notifications@github.com>> wrote:



I think we want to have the first case be a shorter version of the second. That is, if provided window=[(0.*u.um, 5.*u.um),(8.*u.um, 10.*u.um)], we should internally convert that to window=SpectralRegion((0.*u.um, 5.*u.um),(8.*u.um, 10.*u.um)).

PatrickOgle:
I tested this using two different formats for the fit windows:
fitwindows1=[(5125.*u.AA, 5175.*u.AA),(5225.*u.AA, 5275.*u.AA)]
and
region1 = SpectralRegion(5125.*u.AA, 5175.*u.AA)
region2 = SpectralRegion(5225.*u.AA, 5275.*u.AA)
fitwindows3=[region1,region2]

This plot shows the results for one particular realization of the noise:
fit_continuum_2inputs_v1

The fit looks pretty good, and is the same regardless of the format of the windows (SpectralRegion or list of tuples).

However, I note that the fit is not always good, depending on the noise, which makes me suspicious that the fitting algorithm for fit_continuum is not robust.  This deserves a separate issue, and should not block this PR. Here is an example of a poor continuum fit, using the same windows but different random noise:
fit_continuum_2inputs_v2

docs/fitting.rst
```diff
@@ -678,6 +678,36 @@ fitted continuum, which returns a new object:
...
+    y += np.random.normal(0., 0.2, x.shape)
+    y += 3.2 * np.exp(-0.5 * (x - 5.6)**2 / 4.8**2)
+
+    spectrum = Spectrum1D(flux=y * u.Jy, spectral_axis=x * u.um)
+
+    region = [(4.*u.um, 5.*u.um), (8.*u.um, 10.*u.um)]
+
+    fitted_continuum = fit_continuum(spectrum, window=region)
+
+
```
dtn5ah, line 710:
```suggestion
y_fit = fitted_continuum(x*u.um)

plt.plot(x,y)
plt.plot(x,y_fit)
plt.title('Continuum Fitting')
plt.grid(True)
```

nmearl:
@PatrickOgle Can you open an issue describing your concerns with the robustness of the `fit_continuum` function? I'm going to go ahead and merge this.

PR #699, opened 2020-07-06
Propagate nans in flux to mask

### LINKED ISSUE #518: Decide if there are use cases for `flux_nanfilled` and if so implement it
This is a follow on to #509, as noted in https://github.com/astropy/specutils/pull/509#issuecomment-533177613.

Basically, that PR introduced a situation where it at first seemed like it would be useful to have something that produced a version of `spectrum.flux` with certain values masked.  While it turns out that specific need *may* be better suited to just carrying around the `mask` of  `Spectrum1D` and doing `spec.flux[~spec.mask]` as needed, the discussion revealed that there may be other cases where one may actually want a version more like a numpy masked array., which maintains the original shape (whereas `something[mask]` flattends to 1D). 

The two possible implementations that are already out there seem to be:
1. [Numpy MaskedArrays](https://docs.scipy.org/doc/numpy/reference/maskedarray.html) - for `Spectrum1D` that would basically just follow https://docs.astropy.org/en/stable/nddata/nddata.html#numpy-ma-maskedarray for the flux, spectral_axis, and uncertainties.
2. Use nans to indicate "masked" (this is how Pandas does it, for example).  I think @brechmos-stsci already implemented this in #509.
3. Do none of these, instead tell users to deal with the masking themselves.

In both cases this would result in properties `flux_masked`/`uncertainty_masked`/`spectral_axis_masked`, etc, which would be generated based on what's in the `mask` attribute.

These are both possible, but I tilt towards either 2 or 3.  I'm not a fan of masked arrays because they have historically (and I think currently?) not played very well with `Quantity` objects (of which `flux` and `spectral_axis` are), and have performance penalities.  This was part of the motivation for making `NDData` in fact, so it would be a bit odd to convert back to that instead of just carrying the spectrum around.  (That said, there are some other Python tools that do understand and respect masked arrays.)  The nan-masking, meanwhile, also has some support in the wider ecosystem (most notably, matplotlib does the expected thing of just hiding those points), but has the downside that NaN's are "infectious" - that is, many operations spread NaNs far and wide in downstream operations instead of "ignoring" them like masks. So that suggests to me we either do neither, or we use NaN masking but tell users "only use these as the *last* step if you're sure it's what you want, e.g. plotting something with matplotlib"

jdavies-st:
If the flux array in a `Spectrum1D` instance has NaN values, mask them in the `mask` attribute.

Tests are the most simple case, and I'm sure I'm missing something here.  Anyone point to me to other use cases so I can expand the tests?  @eteq 

~Resolves X #518?~

### DIFF
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -57,8 +57,10 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
         Contains uncertainty information along with propagation rules for
         spectrum arithmetic. Can take a unit, but if none is given, will use
         the unit defined in the flux.
-    mask : `~numpy.ndarray`-like and bool
-        Boolean array where values in the flux to be masked are set to `True`.
+    mask : `~numpy.ndarray`-like
+        Array where values in the flux to be masked are those that
+        ``astype(bool)`` converts to True. (For example, integer arrays are not
+        masked where they are 0, and masked for any other value.)
     meta : dict
         Arbitrary container for any user-specific information to be carried
         around with the spectrum container object.
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -18,9 +18,9 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
     """
     Spectrum container for 1D spectral data.
 
-    Note that "1d" in this case refers to the fact that there is only one
-    spectral axis.  `Spectrum1D` can contain "vector 1d spectra" by having the
-    ``flux`` have a shape with dimension great than 1 - the requirement then
+    Note that "1D" in this case refers to the fact that there is only one
+    spectral axis.  `Spectrum1D` can contain "vector 1D spectra" by having the
+    ``flux`` have a shape with dimension greater than 1.  The requirement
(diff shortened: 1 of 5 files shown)


specutils/spectra/spectrum1d.py
```diff
@@ -50,10 +50,15 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
...
     radial_velocity
         See `radial_velocity` for more information.
+    bin_specification : str
+        Either "edges" or "centers" to indicate whether the `spectral_axis`
+        values represent edges of the wavelength bin, or centers of the bin.
     uncertainty : `~astropy.nddata.NDUncertainty`
         Contains uncertainty information along with propagation rules for
         spectrum arithmetic. Can take a unit, but if none is given, will use
         the unit defined in the flux.
+    mask : `~numpy.ndarray`-like and bool
```
eteq, line 60: As I recall, it is *not* strictly necessary that this be a bool.  Rather it can be anything so long as "0"/"False" is "good".  That is, `mask` can either be a bool *or* a "DQ" array.

eteq (approved):

eteq:
I went ahead and implemented my suggestions so we can get this into an impending release.

PR #701, opened 2020-07-22
Update mask handling in line flux and equivalent width functions

rosteen:
Some further work addressing [issue 516](https://github.com/astropy/specutils/issues/516). The recent discussion between @nmearl and I provides some additional context for this PR. I believe the concern about variable-width wavelength bins with masked data removed is addressed by the fact that `Spectrum1D.bin_edges` returns appropriately spaced bin edges based on the remaining spectral_axis points if intermediate data points are removed, but it may turn out that I'm wrong and this solution isn't adequate.

### DIFF
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -117,7 +117,9 @@ def _compute_line_flux(spectrum, regions=None,
         mask = calc_spectrum.mask
         new_spec = Spectrum1D(flux=calc_spectrum.flux[~mask],
                               spectral_axis=calc_spectrum.spectral_axis[~mask])
-        return _compute_line_flux(new_spec)
+        interpolator = mask_interpolation(extrapolation_treatment='zero_fill')
+        sp = interpolator(new_spec, calc_spectrum.spectral_axis)
+        flux = sp.flux
     else:
         flux = calc_spectrum.flux
 
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -38,9 +38,11 @@ def line_flux(spectrum, regions=None,
         Region within the spectrum to calculate the gaussian sigma width. If
         regions is `None`, computation is performed over entire spectrum.
 
-    mask_interpolation : `~specutils.manipulation.LinearInterpolatedResampler`
+    mask_interpolation : ``None`` or `~specutils.manipulation.LinearInterpolatedResampler`
         Interpolator class used to fill up the gaps in the spectrum's flux
-        array, when the spectrum mask is not None.
+        array, when the spectrum mask is not None. If set to ``None``, the
+         masked spectral bins are excised from the data without interpolation
+         and the bin edges of the adjacent bins are extended to fill the gap.
 
     Returns
     -------
@@ -80,10 +82,12 @@ def equivalent_width(spectrum, continuum=1, regions=None,
         will be assumed, otherwise units are required and must be the same as
         the ``spectrum.flux``.
 
-    mask_interpolation : `~specutils.manipulation.LinearInterpolatedResampler`
+    mask_interpolation : ``None`` or `~specutils.manipulation.LinearInterpolatedResampler`
         Interpolator class used to fill up the gaps in the spectrum's flux
         array after an excise operation to ensure the mask shape can always be
-        applied when the spectrum mask is not None.
+        applied when the spectrum mask is not None. If set to ``None``, the
+        masked spectral bins are excised from the data without interpolation
+        and the bin edges of the adjacent bins are extended to fill the gap.
 
     Returns
     -------
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -143,7 +143,7 @@ def _compute_line_flux(spectrum, regions=None,
 
         if variance_q is not None:
             line_flux.uncertainty = np.sqrt(
-                np.sum(variance_q * avg_dx**2))
+                np.sum(variance_q * dx**2))
 
     # TODO: we may want to consider converting to erg / cm^2 / sec by default
     return line_flux
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -41,8 +41,8 @@ def line_flux(spectrum, regions=None,
     mask_interpolation : ``None`` or `~specutils.manipulation.LinearInterpolatedResampler`
         Interpolator class used to fill up the gaps in the spectrum's flux
         array, when the spectrum mask is not None. If set to ``None``, the
-         masked spectral bins are excised from the data without interpolation
-         and the bin edges of the adjacent bins are extended to fill the gap.
+        masked spectral bins are excised from the data without interpolation
+        and the bin edges of the adjacent bins are extended to fill the gap.
 
     Returns
     -------
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -9,6 +9,7 @@ import numpy as np
 from astropy.nddata import StdDevUncertainty, VarianceUncertainty, InverseVariance
 
 from .. import conf
+from ..spectra import Spectrum1D
 from ..manipulation import extract_region, LinearInterpolatedResampler
 from .utils import computation_wrapper
 import astropy.units as u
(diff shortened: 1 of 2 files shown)


rosteen:
I can't tag reviewers - @ibusko and @nmearl, this is probably of most interest to you two. Perhaps @eteq as well if you have time and attention to spare.

rosteen:
Ah, good point about removing the interpolator references, I had been erring on the side of leaving them in until this was working. 

The tests are failing, but it's because as far as I can tell the old interpolator method wasn't handling the mask - the interpolator was simply returning the original spectrum because it doesn't check for a mask on the spectrum and the requested spectral axis values were the same as the original so no interpolation was done. Thus the result was identical to that for the non-masked spectrum. In general one would expect the flux returned for the masked spectrum to differ from that of the non-masked case. So I also need to update the tests - I'll try to do that this afternoon.

specutils/analysis/flux.py
```diff
@@ -111,23 +112,17 @@ def _compute_line_flux(spectrum, regions=None,
     else:
         calc_spectrum = spectrum
```
nmearl, line 114:
```suggestion
    dx = np.abs(np.diff(calc_spectrum.spectral_axis.bin_edges))
    flux = calc_spectrum.flux
```

specutils/analysis/flux.py
```diff
@@ -111,23 +112,17 @@ def _compute_line_flux(spectrum, regions=None,
...
     if hasattr(calc_spectrum, 'mask') and calc_spectrum.mask is not None:
-        # For now, we interpolate over the masked values. A better solution
-        # would be to account for the masked values when computing the dispersion.
-        interpolator = mask_interpolation()
-        sp = interpolator(calc_spectrum, calc_spectrum.spectral_axis)
-        flux = sp.flux
+        mask = calc_spectrum.mask
+        new_spec = Spectrum1D(flux=calc_spectrum.flux[~mask],
+                              spectral_axis=calc_spectrum.spectral_axis[~mask])
+        return _compute_line_flux(new_spec)
```
nmearl, line 120:
```suggestion
        mask = calc_spectrum.mask
        dx = dx[~mask]
        flux =flux[~mask]
```

specutils/analysis/flux.py
```diff
@@ -111,23 +112,17 @@ def _compute_line_flux(spectrum, regions=None,
...
     else:
         flux = calc_spectrum.flux
```
nmearl, line 122:
```suggestion
```

specutils/analysis/flux.py
```diff
@@ -111,23 +112,17 @@ def _compute_line_flux(spectrum, regions=None,
...
 
-    line_flux = np.sum(flux * avg_dx)
+    dx = (np.abs(np.diff(calc_spectrum.spectral_axis.bin_edges)))
```
nmearl, line 124:
```suggestion
```

rosteen:
@nmearl At the risk of being snarky, you masked out all the low values, so if you take the average of what's left it's necessarily going to be higher. It doesn't matter if you interpolate between the large values, you're not going to get the small values back! 

Multiplying the bin size times the flux data points you have left after masking for larger bins (because the bin edges are now midway between the remaining values) is mathematically equivalent to doing a linear interpolation between them. Of course that doesn't account for the possibility of doing more complicated interpolations (or something like fitting a gaussian to whatever data points are available, be there a mask or not, and returning the line flux of _that_ as our answer).

I think the answer here is to take 
```
new_spec = Spectrum1D(flux=calc_spectrum.flux[~mask],
                              spectral_axis=calc_spectrum.spectral_axis[~mask])
```
and drop that into whatever the specified `mask_interpolation` function is, which I imagine is likely what the intent was in the first place. That would make it actually perform some interpolation, since the spectral_axis will have less values than the unmasked spectrum.

nmearl:
> At the risk of being snarky, you masked out all the low values, so if you take the average of what's left it's necessarily going to be higher.

@rosteen Not snarky at all! However (and I may being daft), I'm not quite following the point. We're taking the sum (not an average) of the product of the flux array and the bin widths -- that would suggest that _any_ amount masked out of the spectrum should produce less over all flux in the summation. 

Although, this does get to the root of the thing that's bothering me. The example was meant to show the effect of multiplying through the bin _sizes_, not their position. That is, regardless of their position (center or edge), their _differences_ would still reflect their width used to calculate the line flux. Thus my concern is the consistency of the physical interpretation of a truly "masked" spectrum where masked portions are _completely_ ignored vs. a spectrum in which masked bins are assumed to still contain flux (i.e. one in which the values are "interpolated" by variable bin widths).

> Multiplying the bin size times the flux data points you have left after masking for larger bins (because the bin edges are now midway between the remaining values) is mathematically equivalent to doing a linear interpolation between them.

Correct! And this I think exemplifies my presentiment above. It's the difference between

```python
[2, 3, --, --, 7, 10, 15, 10, 6, 3, --, --, --, 2, 3] * [1, 1, --, --, 1, 1, 1, 1, 1, 1, --, --, --, 1, 1]
```

and

```python
[2, 3, 7, 10, 15, 10, 6, 3, 2, 3] * [1, 3, 1, 1, 1, 1, 1, 1, 4, 1, 1]
```

and the answer may be "Nick, you're being ridiculously, it's definitely the second we're going for here". 😁

rosteen:
Whoops, pushed a change before I saw that comment (but I think you'll like the change)

I think I see more of what you're getting at now based on that comment, but I think I agree with "it's definitely the second we're going for here" because the mask isn't saying that there is _no flux_ in those bins, it's saying that we _haven't measured_ the flux in those bins, and thus the best way we have to estimate the (non-zero, presumably) flux in those bins is to interpolate based on the closest bins on either side. 

I do think it's a bit confusing/inconsistent that the masked values outside the outermost unmasked flux bins are considered "ignored completely, outside the area of interest" while the ones inside are interpolated, but I don't really see a better way to handle it (other than the fact, perhaps, that we can already mix use of a `region` and a `mask`)

eteq:
@rosteen @nmearl - to work this out I resorted to drawing diagrams on paper because I always need a picture when thinking about integrals, see the below... But I think I've identified a few different cases/interpretations for the mask and how they apply.  I think this PR currently represents the "linear" case, and I think that's fine/good.  But what I'm realizing is that we *also* need to consider the other two cases I'm showing here.  We don't necessarily need to implement it all in this PR, but worth thinking about.  

For a bit more context on these plots: in my head the line flux operation is the same as doing an integral of the flux over the spectrum. Since we are working with the "flux in boxes" interpretation for this operation, the diagram below shows how we interpret that - basically a train of square waves.  And that's what justifies the simple sum of flux * dx approach.  I've shown in the first case how I think that translates to linear interpolation (the pixel with the "M" is the masked one in this case), and I think it's safe to just carry that over.

However, there are two other cases:

"Other interpolation": if the user selects anything other than linear interpolation, it raises the question of whether what they really want is the integral over the *interpolated curve*, since it's not equivalent to the  . In my diagram I've shown a pathological interpolation just to illustrate the point here, that the area under the dashed curve is going to be quite different from the linear case *or* the "gap" case. I think in this case we may want to integrate the interpolated curve instead of doing the box approach. `scipy` has some functions to enable this, but I guess I don't know what the performance implications are there.

"gap": to me this is the mental image I have of the "one pixel is masked" case. You interpret the spectrum as though the pixel is not there *at all*, which leads to there being only 3 pixels, but two are wider now. I agree with @rosteen this is mathematically equivalent to the linear case as shown, but there are two caveats there: 1) the implementation is rather different so with more complex spectrum shapes the interpolation might become dicier, and 2) there is a *real* difference if you stop the integral somewhere in the middle of one of these bins, or, say, it's at the edge of the spectrum.

So my inclination is to say the approach in this PR is right for the linear interpolation case, but add the `None` option for `mask_interpolation` and treat it as the "gap" case. That could be done in this PR or a separate later one.  Then in definitely a separate later one we address the question of whether to integrate over a non-linear interpolator.

IMG_20200827_121855

rosteen:
@eteq I added gap case handling if `mask_interpolation=None` to `_compute_line_flux`, and updated the tests that were failing due to looking for incorrect values.

eteq:
Awesome! so just to be clear - this implements my "linear" and "gap" but not the more complex "other", right?  If so I'm happy with that scope for this (and will take another look to confirm/merge ASAP), and can make another issue for the "other"

ibusko:
The new implementation looks all right to me. The tests pass.

rosteen:
@eteq That is correct.

Thanks for the review @ibusko

eteq:
Was just about to push up a commit with that last fix so you didn't have to. But looks like you got there first.  Looks good @rosteen, merging!

PR #702, opened 2020-07-22
Adds new specutils data loader for SDSS MaNGA cube/rss data

havok2063:
This PR adds a new data loader for SDSS MaNGA cube and rss spectral data.  It also adds a convenience function, `io.registers.identify_spectrum_format` to identify a valid registered `specutils` file format given an input filename.  This function is essentially a wrapper for `astropy.io.registry.identify_format` setting origin to `read` and data_class to `Spectrum1D`.

### DIFF
--- a/docs/identify.rst
+++ b/docs/identify.rst
@@ -8,8 +8,11 @@ Identifying Spectrum1D Formats
 essentially acts as a wrapper on `~astropy.io.registry.identify_format`.
 
 This function is useful for identifying a spectrum file format without the need to
-read it in with Spectrum1d's `~specutils.Spectrum1D.read` method.  It returns the best guess as to a
-valid format from the list of ``Formats`` as given by `~astropy.io.registry.get_formats`.
+read it in with Spectrum1d's `~specutils.Spectrum1D.read` method.  It uses the
+same identification method as `read` however, so it simply provides a convenience
+of access outside of calling `read` without any change in underlying functionality.
+It returns the best guess as to a valid format from the list of ``Formats``
+as given by `~astropy.io.registry.get_formats`.
 
 General usage is as follows, passing in an absolute filename path:
 
--- a/docs/identify.rst
+++ b/docs/identify.rst
@@ -9,8 +9,8 @@ essentially acts as a wrapper on `~astropy.io.registry.identify_format`.
 
 This function is useful for identifying a spectrum file format without the need to
 read it in with Spectrum1d's `~specutils.Spectrum1D.read` method.  It uses the
-same identification method as `read` however, so it simply provides a convenience
-of access outside of calling `read` without any change in underlying functionality.
+same identification method as ``read`` however, so it simply provides a convenience
+of access outside of calling ``read`` without any change in underlying functionality.
 It returns the best guess as to a valid format from the list of ``Formats``
 as given by `~astropy.io.registry.get_formats`.
 
--- a/docs/identify.rst
+++ b/docs/identify.rst
@@ -7,21 +7,14 @@ Identifying Spectrum1D Formats
 `~specutils.Spectrum1D` file format from the list of registered formats, and
 essentially acts as a wrapper on `~astropy.io.registry.identify_format`.
 
-This function is useful for identifying a spectrum file format without the need to
-read it in with Spectrum1d's `~specutils.Spectrum1D.read` method.  It uses the
-same identification method as ``read`` however, so it simply provides a convenience
+This function is useful for identifying a spectrum file format without reading the
+whole file with the  `~specutils.Spectrum1D.read` method.  It uses the
+same identification method as ``read`` however, so it provides a convenience
 of access outside of calling ``read`` without any change in underlying functionality.
 It returns the best guess as to a valid format from the list of ``Formats``
 as given by `~astropy.io.registry.get_formats`.
 
-General usage is as follows, passing in an absolute filename path:
-
-.. code-block:: python
-
-    >>> from specutils.io.registers import identify_spectrum_format
-    >>> identify_spectrum_format("/path/to/file.fits")  # doctest: +SKIP
-
-An example is given to identify a SDSS MaNGA data cube file:
+For eample, to identify a SDSS MaNGA data cube file:
 
 .. code-block:: python
 
--- a/docs/identify.rst
+++ b/docs/identify.rst
@@ -1,0 +1,40 @@
+==============================
+Identifying Spectrum1D Formats
+==============================
+
+``specutils`` provides a convenience function,
+`~specutils.io.registers.identify_spectrum_format`, which attempts to guess the
+`~specutils.Spectrum1D` file format from the list of registered formats, and
+essentially acts as a wrapper on `~astropy.io.registry.identify_format`.
+
+This function is useful for identifying a spectrum file format without the need to
+read it in with Spectrum1d's `~specutils.Spectrum1D.read` method.  It returns the best guess as to a
+valid format from the list of ``Formats`` as given by `~astropy.io.registry.get_formats`.
+
+General usage is as follows, passing in an absolute filename path:
+
+.. code-block:: python
+
+    >>> from specutils.io.registers import identify_spectrum_format
+    >>> identify_spectrum_format("/path/to/file.fits")  # doctest: +SKIP
+
+An example is given to identify a SDSS MaNGA data cube file:
+
+.. code-block:: python
+
+    >>> from astropy.utils.data import download_file
+    >>> from specutils.io.registers import identify_spectrum_format
+    >>>
+    >>> url='https://dr15.sdss.org/sas/dr15/manga/spectro/redux/v2_4_3/8485/stack/manga-8485-1901-LOGCUBE.fits.gz'
+    >>> dd = download_file(url)
+    >>> identify_spectrum_format(dd)
+    'MaNGA cube'
+
+or a JWST extracted 1d spectral file:
+
+.. code-block:: python
+
+    >>> from specutils.io.registers import identify_spectrum_format
+    >>> path = '/data/jwst/jw00626-o030_s00000_nirspec_f170lp-g235m_x1d.fits'
+    >>> identify_spectrum_format(path)  # doctest: +SKIP
+    'JWST x1d'
--- a/docs/identify.rst
+++ b/docs/identify.rst
@@ -26,7 +26,7 @@ An example is given to identify a SDSS MaNGA data cube file:
     >>> from specutils.io.registers import identify_spectrum_format
     >>>
     >>> url='https://dr15.sdss.org/sas/dr15/manga/spectro/redux/v2_4_3/8485/stack/manga-8485-1901-LOGCUBE.fits.gz'
-    >>> dd = download_file(url)
+    >>> dd = download_file(url)  # doctest: +SKIP
     >>> identify_spectrum_format(dd)
     'MaNGA cube'
 
--- a/specutils/io/default_loaders/manga.py
+++ b/specutils/io/default_loaders/manga.py
@@ -1,14 +1,3 @@
-# !/usr/bin/env python
-# -*- coding: utf-8 -*-
-#
-# Filename: manga.py
-# Project: default_loaders
-# Author: Brian Cherinka
-# Created: Tuesday, 21st July 2020 2:40:56 pm
-# License: BSD 3-clause "New" or "Revised" License
-# Copyright (c) 2020 Brian Cherinka
-# Last Modified: Tuesday, 21st July 2020 2:40:56 pm
-# Modified By: Brian Cherinka
 
 
 from __future__ import print_function, division, absolute_import
@@ -114,4 +103,3 @@ def _load_manga_spectra(hdulist, per_unit=None, transpose=None):
 
     return Spectrum1D(flux=flux, meta={'header': hdr}, spectral_axis=wave,
                       uncertainty=ivar)
-
--- a/specutils/io/default_loaders/manga.py
+++ b/specutils/io/default_loaders/manga.py
@@ -1,21 +1,50 @@
 
-
-from astropy.io import fits
 import astropy.units as u
-from ...spectra import Spectrum1D
-from ..registers import data_loader
+from astropy.io import fits
 from astropy.nddata import InverseVariance
 
+import _io
+import contextlib
+from ...spectra import Spectrum1D
+from ..registers import data_loader
 
 __all__ = ["identify_manga_cube", "identify_manga_rss", "manga_cube_loader", "manga_rss_loader"]
 
 
-def _identify_sdss_fits(filename):
+@contextlib.contextmanager
+def _read_fileobj(*args, **kwargs):
+    """ Context manager for reading a filename or file object
+
+    Returns:
+        an Astropy HDUList
+    """
+    # access the fileobj or filename arg
+    # do this so identify functions are useable outside of Spectrum1d.read context
+    try:
+        fileobj = args[2]
+    except IndexError:
+        fileobj = args[0]
+
+    if isinstance(fileobj, fits.hdu.hdulist.HDUList):
+        hdulist = fileobj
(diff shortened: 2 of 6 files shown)


nmearl (dismissed):
Thanks, @havok2063, this looks great. Couple of small changes, but also one bigger change to suggest as well.

We've moved to allowing either string file paths or file object references in data loading. E.g. if you look at the [hst loaders](https://github.com/astropy/specutils/blob/master/specutils/io/default_loaders/hst_cos.py#L38), you'll see that they handle the case where a file object is passed to the reader. It'd be great if you could update these to handle that case as well.

havok2063:
Thanks for the comments @nmearl.  I've accepted your minor suggestions.  I've updated the manga loaders to use the `file_obj` nomenclature following the hst examples.  I originally followed the jwst example, which look like they haven't been updated either.

dhomeier:
Thanks for the update! Ideally accepting file handles or URLs would also work for the identifier, using information from the file content itself – that is essentially already implemented here, but will require slightly different treatment depending on the type of `file_obj`.
I suggest to have a look at e.g. `spec_identify` in the SDSS loader; you probably just need to add the check for `header['INSTRUME'] == 'MaNGA'` there.
Since the `identify_spectrum_format` function is not directly used in the loader, could you add some documentation when and how to use it?

The test failure seems due to an unrelated download failure in testing another loader, but it would be great if you could also add a test for the new format(s), assuming there are some example files on sdss.org (see e.g. `test_sdss_spec`).

havok2063:
@dhomeier Thanks for the comments.  I've updated the code so the identifier functions also accept a file object.  I've added some documentation on the `identify_spectrum_format` function.  I created a new page, `identify.html`, but am happy to incorporate it into a better spot.  I've also added two new tests `test_manga_cube` and `test_manga_rss`.  These simply pass in the public url to the data into `Spectrum1D.read`.  I tried following the old sdss test with `urllib` but ran into issues where Spectrum1D could not read the objects from the `urllib` response.   I was getting strange errors regarding `header does not have valid END card`.  I tried setting `ignore_missing_end=True` but then got an error about `empty or corrupt FITS file`.  When I download the data files locally or use the url stream directly, `Spectrum1D.` reads the file just fine.

docs/identify.rst
```diff
@@ -0,0 +1,40 @@
...
+
+An example is given to identify a SDSS MaNGA data cube file:
+
+.. code-block:: python
+
+    >>> from astropy.utils.data import download_file
+    >>> from specutils.io.registers import identify_spectrum_format
+    >>>
+    >>> url='https://dr15.sdss.org/sas/dr15/manga/spectro/redux/v2_4_3/8485/stack/manga-8485-1901-LOGCUBE.fits.gz'
+    >>> dd = download_file(url)
```
dhomeier, line 29:
```suggestion
    >>> dd = download_file(url)  # doctest: +SKIP
```
This made the doctest fail when not run with `remote_data=True` – don't know if one could add the `remote_data` decorator somehow for doctest, but I'd suggest to simply skip it.

docs/identify.rst
```diff
@@ -0,0 +1,40 @@
...
+==============================
+
+``specutils`` provides a convenience function,
+`~specutils.io.registers.identify_spectrum_format`, which attempts to guess the
+`~specutils.Spectrum1D` file format from the list of registered formats, and
+essentially acts as a wrapper on `~astropy.io.registry.identify_format`.
+
+This function is useful for identifying a spectrum file format without the need to
+read it in with Spectrum1d's `~specutils.Spectrum1D.read` method.  It returns the best guess as to a
+valid format from the list of ``Formats`` as given by `~astropy.io.registry.get_formats`.
```
dhomeier, line 12: Perhaps this could spell out still more clearly that it is using the same identification methods as `Spectrum1D.read` – I doubt many users will be familiar with the internals of `~astropy.io.registry`.

specutils/io/default_loaders/manga.py
```diff
@@ -0,0 +1,177 @@
...
+                and hdulist[1].header["NAXIS"] == 3)
+
+
+def identify_manga_rss(origin, *args, **kwargs):
+    """
+    Check whether the given file is a MaNGA RSS.
+    """
+    is_sdss = _identify_sdss_fits(*args, **kwargs)
+    with _read_fileobj(*args, **kwargs) as hdulist:
+        return (is_sdss and "FLUX" in hdulist and hdulist[1].header['INSTRUME'] == 'MaNGA'
```
dhomeier, line 70: Wondering if it's really worth delegating the `is_sdss` check to its own function, as the direct `hdulist[0].header["TELESCOP"] == "SDSS 2.5-M"` test would fit almost within the same space.
Since it is obviously working, it should be OK, but we might want to avoid extra file I/O operations from a performance point of view.

specutils/io/default_loaders/manga.py
```diff
@@ -0,0 +1,177 @@
...
+@data_loader("MaNGA cube", identifier=identify_manga_cube, dtype=Spectrum1D,
+             extensions=['fits'])
+def manga_cube_loader(file_obj, **kwargs):
+    """
+    Loader for MaNGA 3D rectified spectral data in FITS format.
+
+    Parameters
+    ----------
+    filename : str
+        The path to the FITS file
```
dhomeier, line 83:
```suggestion
    file_obj: str, file-like, or HDUList
          FITS file name, object (provided from name by Astropy I/O Registry),
          or HDUList (as resulting from astropy.io.fits.open()).
```

specutils/io/default_loaders/manga.py
```diff
@@ -0,0 +1,177 @@
...
+@data_loader("MaNGA rss", identifier=identify_manga_rss, dtype=Spectrum1D,
+             extensions=['fits'])
+def manga_rss_loader(file_obj, **kwargs):
+    """
+    Loader for MaNGA 2D row-stacked spectral data in FITS format.
+
+    Parameters
+    ----------
+    filename : str
+        The path to the FITS file
```
dhomeier, line 114:
```suggestion
    file_obj: str, file-like, or HDUList
          FITS file name, object (provided from name by Astropy I/O Registry),
          or HDUList (as resulting from astropy.io.fits.open()).
```

specutils/io/registers.py
```diff
@@ -167,3 +169,37 @@ def _load_user_io():
...
                     pass
+
+
+def identify_spectrum_format(filename):
+    """ Attempt to identify a spectrum file format
+
+    Given a filename, attempts to identify a valid file format
+    from the list of registered specutils loaders.  Essentially a wrapper for
+    `~astropy.io.registry.identify_format` setting **origin** to ``read`` and
+    **data_class_required** to `~specutils.Spectrum1D`.
```
dhomeier, line 180: I am curious if this can be generalised to other `spectra` objects, e.g. by adding a `dtype=Spectrum1D` kwarg.
There are one or two loaders for `SpectrumList` now, and #708 should add one for `SpectrumCollection`, so it would be great if the same function could be used for identification. Does not have to be solved in this PR, but then the documentation would also need to be updated a little bit.

dhomeier:
> incorporate it into a better spot. I've also added two new tests `test_manga_cube` and `test_manga_rss`. These simply pass in the public url to the data into `Spectrum1D.read`. I tried following the old sdss test with `urllib` but ran into issues where Spectrum1D could not read the objects from the `urllib` response. I was getting strange errors regarding `header does not `

Thanks a lot for the updates and sorry for pointing to a misleading example; I should have referred to the `test_*_stream` tests right away!

I especially like the implementation with the custom contextmanager – I really think this could be used by all loaders supporting all 3 `file_obj` types, so perhaps `_read_fileobj` could be moved out to `parsing_utils.py`. What do you think @nmearl ?

nmearl:
> I especially like the implementation with the custom contextmanager – I really think this could be used by all loaders supporting all 3 file_obj types, so perhaps _read_fileobj could be moved out to parsing_utils.py (maybe renamed to something like read_fileobj_or_hdulist). What do you think @nmearl ?

Yes, definitely. It's probably not necessary to do that here in this PR, but we should certainly open an issue to update the docs and the identifier/loading functions to use a generic context manager.

specutils/io/registers.py
havok2063: I made a small change to allow a user to specify a dtype that can either be `Spectrum1D`, `SpectrumList`, or `SpectrumCollection`, with the default being `Spectrum1D`.  It throws an assert otherwise.

specutils/io/default_loaders/manga.py
havok2063: I agree.  I've removed the redundant function and moved the the SDSS telescope check into the `cube` and `rss` identifiers

docs/identify.rst
havok2063: I've updated the docs here with another sentence or two.  Let me know what you think.

havok2063:
@dhomeier Thanks for the comments.  I've pushed the changes you've requested.

docs/identify.rst
```diff
@@ -0,0 +1,43 @@
...
+Identifying Spectrum1D Formats
+==============================
+
+``specutils`` provides a convenience function,
+`~specutils.io.registers.identify_spectrum_format`, which attempts to guess the
+`~specutils.Spectrum1D` file format from the list of registered formats, and
+essentially acts as a wrapper on `~astropy.io.registry.identify_format`.
+
+This function is useful for identifying a spectrum file format without the need to
+read it in with Spectrum1d's `~specutils.Spectrum1D.read` method.  It uses the
```
eteq, line 11:
```suggestion
This function is useful for identifying a spectrum file format without reading the
whole file with the  `~specutils.Spectrum1D.read` method.  It uses the
```

docs/identify.rst
```diff
@@ -0,0 +1,43 @@
...
+same identification method as ``read`` however, so it simply provides a convenience
```
eteq, line 12:
```suggestion
same identification method as ``read`` however, so it provides a convenience
```

docs/identify.rst
```diff
@@ -0,0 +1,43 @@
...
+
+General usage is as follows, passing in an absolute filename path:
+
+.. code-block:: python
+
+    >>> from specutils.io.registers import identify_spectrum_format
+    >>> identify_spectrum_format("/path/to/file.fits")  # doctest: +SKIP
+
+An example is given to identify a SDSS MaNGA data cube file:
+
```
eteq, line 25: I think the "general usage" isn't needed given that the MaNGA example is already quite clear
```suggestion
For eample, to identify a SDSS MaNGA data cube file:

```

specutils/io/registers.py
```diff
@@ -167,3 +169,45 @@ def _load_user_io():
...
+
+    Parameters
+    ----------
+    filename : str
+        The absolute filename to the object
```
eteq, line 185: it doesn't *have* to be absolute, right?  i.e. `../some-file.fits` should work even if it's not the best idea to do that?
```suggestion
        A path to a file to be identified
```

specutils/io/default_loaders/manga.py
```diff
@@ -0,0 +1,141 @@
...
+    Spectrum1D
+        The spectrum contained in the file.
+    """
+
+    if isinstance(file_obj, fits.hdu.hdulist.HDUList):
+        hdulist = file_obj
+    else:
+        hdulist = fits.open(file_obj, **kwargs)
+
+    spaxel = u.Unit('spaxel', represents=u.pixel, doc='0.5" spatial pixel', parse_strict='silent')
```
eteq, line 59: This is actually an interesting and good use of units.  I wonder if this should be moved to the public API so that users can do e.g. `specutils.<something>.spaxel`? 

(this could be done in a folllow-on PR if you don't want to deal with it right now though - I can make an issue to deal with later if that's your preference @havok2063)

specutils/io/default_loaders/manga.py
```diff
@@ -0,0 +1,141 @@
...
+    Spectrum1D
+        The spectrum contained in the file.
+    """
+
+    if isinstance(file_obj, fits.hdu.hdulist.HDUList):
+        hdulist = file_obj
+    else:
+        hdulist = fits.open(file_obj, **kwargs)
+
+    fiber = u.Unit('fiber', represents=u.pixel, doc='spectroscopic fiber', parse_strict='silent')
```
eteq, line 91: (ditto for the spaxel comment above)

eteq (approved):
This looks good as it stands, although I have several minor comments that I'm am suggesting to @havok2063 that he can take or leave depending on preference.

One higher-level thought (which doesn't need to hold up this PR, but might be easy to include) - should the rss format *also* have a `SpectrumCollection` option? I'm not familiar enough with it to know for sure, but it looks like it might be a case where it's more useful to think of it as a set of distinct fibers instead of spaxels?

nmearl:
> One higher-level thought (which doesn't need to hold up this PR, but might be easy to include) - should the rss format also have a SpectrumCollection option?

Just want to point out an issue here that `SpectrumCollection` isn't not a subclass of `NDData`, and thus doesn't leverage the astropy io machinery. We'd have to refactor `SpectrumCollection` a bit to allow e.g. `SpectrumCollection.read(...)`.

However, we _can_ still do `Spectrum1D.read(...)` and have it return a `SpectrumCollection` object, but it might be too inconsistent of a user workflow.

havok2063:
@eteq  Well admittedly I've always been a bit confused between some of the spectrum class distinctions outlined in the docs.  I was under the impression that `SpectrumCollection` was for a set of independent spectra that have different spectral axes, either a different length or different WCSes, and the FITS contains individual spectral axes for each spectrum.    If that's not the case, then yeah technically rss fibers could be a `SpectrumCollection`.  MaNGA rss files contain a set of fiber spectra (in a 2d-array) that all have the same spectral axis, i.e. same length and WCS.  There's one FITS extension containing the wavelength array for all spectral fibers.  The fibers aren't exactly independent of each other.  For example, if you plotted them on sky, you'd get a series of overlapping dithered fiber positions centered on a target.   It's the same situation as the Cubes, which I also read in as `Spectrum1D`.   I could be convinced of changing it, but less so if it requires some refactoring.  To me when I hear "SpectrumCollection" I think "a loose collection of random spectra I want to treat together".   I'd be happy to make this change in a future PR.

I've accepted your small changes.  They make sense to me.   Regarding the units, what do you mean by moving to the public API?  Does that mean moving these definitions into `astropy.units` or is this something that can be done within `specutils`?  If it's the latter, I can probably easily do that now.  If it's the former I might prefer creating a separate issue and punt this to a later PR since that means going through a whole new process.

dhomeier:
> Just want to point out here that `SpectrumCollection` isn't a subclass of `NDData`, and thus doesn't leverage the astropy io machinery. We'd have to refactor `SpectrumCollection` a bit to allow e.g. `SpectrumCollection.read(...)`.

That is just what I have done in #702 (by simply subclassing `SpectrumCollection` as`NDIOMixin`).

> @eteq Well admittedly I've always been a bit confused between some of the spectrum class distinctions outlined in the docs. I was under the impression that SpectrumCollection was for a set of independent spectra that have different spectral axes, either a different length or different WCSes, and the FITS contains individual spectral axes for each spectrum. If that's not the case, then yeah technically rss fibers could be a SpectrumCollection.

`SpectrumCollection` still is a relatively structured object, notably requiring all spectral axes to have the same length, but with generally different values. The loosest collection really is the `SpectrumList`.
But if your spectra have indeed all _identical_ `spectral_axis`, I agree that `Spectrum1D` would be the natural representation.

havok2063:
ahh ok, that makes sense.  I hadn't realized that subtle distinction between `SpectrumCollection` and SpectrumList`.  Thanks for clarifying.  Yeah the MaNGA data all have identical spectral axes in length and values.  It sounds like `Spectrum1D` is the right thing for them to be but happy to hear arguments otherwise.  If everyone agrees, then I think this is ready to be merged, and @eteq can create a new issue to flesh out what should be done with the units.

eteq:
Alright, yep I think all is good, and I'll make a separate issue for the units thing

PR #710, opened 2020-08-11
Fix factor of two error in SpectralRegion.from_center

rosteen:
The `SpectralRegion.from_center` method previously returned a region twice the size expected given the docs, which implied that the input was the full width of the region. I halved the input width in the calculations of the region upper and lower bounds so that the resulting width matches the input width, and made the language in the docs more explicit.


rosteen:
I can't tag reviewers here - heads up @eteq and @nmearl

PR #715, opened 2020-09-10
Fix the JWST x1d identifier

### LINKED ISSUE #714: specutils cannot identify some JWST x1d files
The `specutils` JWST identifer function for `x1d` files does not correctly identify all x1d files , e.g. `jw00643-o015_s00027_nirspec_f170lp-g235m_x1d.fits`, or `jw00736-o039_t001_miri_ch1-long_x1d.fits`.  These files were recently downloaded from MAST and checked.  Most likely something has changed in headers recently such that the existing identifier functions fails.  `s3d` and `s2d` files still identify correctly.

havok2063:
This PR fixes #714.  This removes the `SCI` constraint on the `identify_jwst_x1d_fits` function as this was interfering with the correct identification of x1d files downloaded from MAST.  Those files have a `SCI` extension preventing them from being correctly identified.  This fix should make this identifier either backwards or forwards compatible with new JWST x1d files, depending on if the MAST files are an older or newer version than when the identifier was originally written.  The null `EXTRACT1D` constraint on the `s3d` and `s2d` identifiers is enough to uniquely separate and identify files and avoid confusion between the 3 types.

### DIFF
--- a/specutils/io/default_loaders/jwst_reader.py
+++ b/specutils/io/default_loaders/jwst_reader.py
@@ -20,8 +20,7 @@ def identify_jwst_x1d_fits(origin, *args, **kwargs):
     """
     is_jwst = _identify_jwst_fits(args[0])
     with fits.open(args[0], memmap=False) as hdulist:
-        return (is_jwst and 'EXTRACT1D' in hdulist and ('EXTRACT1D', 2) not in hdulist
-            and "SCI" not in hdulist)
+        return (is_jwst and 'EXTRACT1D' in hdulist and ('EXTRACT1D', 2) not in hdulist)
 
 
 def identify_jwst_x1d_multi_fits(origin, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


jdavies-st (approved):
Looks good.  Thanks!

As an aside, when are SCI extensions written to `x1d` files?  There should be none, past or future, as the schema does not allow it.  Are these some very, very old files?

havok2063:
@jdavies-st I'm not sure when they were added, or even if it's supposed to be there.  I'm assuming the [JWST docs](https://jwst-pipeline.readthedocs.io/en/latest/jwst/data_products/science_products.html#x1d) is up-to-date.  I've only recently noticed that from the files I have access to on the isilon.  Combing through 1687 files (miri, nirspec), only 1 does not have the `SCI` extension.  The rest do and all have creation timestamps from 2020-7-28 to 2020-9-10.  It's very possible something is wrong with these files...  I'm attaching an example screenshot of the hdu.  

Screen Shot 2020-09-10 at 1 55 54 PM

jdavies-st:
Thanks @havok2063 .  I think this might be a bug in the CAL pipeline.  I'll investigate further.

Regardless, this PR looks good to go.  Thanks!

jdavies-st:
Yeah, that's a bug.

https://github.com/spacetelescope/jwst/issues/5344

Thanks for bringing it to my attention.

Issue #711, opened 2020-08-20
Spectrum1D.read cannot auto-identify format of gzipped files


havok2063:
The Spectrum1D.read method cannot auto-identify format of gzipped FITS files.  Spectrum1D can read in gzipped files ok when specifying the format keyword.  Without the format keyword, it throws an `OSError: Cannot read from/write to a closed file-like object (<gzip on 0x13234c6a0>).`  It probably has to do with getting the readable file object in `astropy.io.registry.read`.  Things fail when it gets inside the data loader method and tries to open the file a second time.  This issue might belong in the main `astropy` repo instead. 

To reproduce:
1. Download cube https://dr15.sdss.org/sas/dr15/manga/spectro/redux/v2_4_3/8485/stack/manga-8485-1901-LOGCUBE.fits.gz
2. Create an unzipped copy.  

```
from specutils import Spectrum1D

cube='/Users/bcherinka/Work/sdss/sas/dr15/manga/spectro/redux/v2_4_3/8485/stack/manga-8485-1901-LOGCUBE.fits.gz'
cube_unzip='/Users/bcherinka/Work/sdss/sas/dr15/manga/spectro/redux/v2_4_3/8485/stack/manga-8485-1901-LOGCUBE_unzip.fits'

# zipped fits fails 
s = Spectrum1D.read(cube)
...
OSError: Cannot read from/write to a closed file-like object (<gzip on 0x13234c130>).

# zipped fits reads correctly when specifying format
s = Spectrum1D.read(cube, format='MaNGA cube')
type(s)
specutils.spectra.spectrum1d.Spectrum1D

# unzipped fits correctly auto-identifies
s = Spectrum1D.read(cube_unzip)
type(s)
specutils.spectra.spectrum1d.Spectrum1D
```
To reproduce with existing loaders, you can download the file https://dr14.sdss.org/sas/dr14/eboss/spectro/redux/v5_10_0/spectra/lite/4055/spec-4055-55359-0596.fits, which is a `SDSS-III/IV spec` format and follow the same steps as above.  

Full traceback
```
s = Spectrum1d.read(cube)
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-36-3e77d30b6096> in <module>
----> 1 s=Spectrum1D.read(cube2)

~/anaconda3/envs/havokve/lib/python3.8/site-packages/astropy/nddata/mixins/ndio.py in __call__(self, *args, **kwargs)
     54
[...]

PR #717, opened 2020-09-17
Fixes for failures introduced in #702

eteq:
From merging #702 I missed that some of the errors were actually real and not red herrings. This fixes all but one of them, which it should xfail (which I *think* is an upstream issue)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -51,6 +51,8 @@ Bug Fixes
 
 - Fix handling of multi-dimensional mask slicing. [#704]
 
+- Fix identifier for JWST 1D loader. [#715]
+
 Documentation
 ^^^^^^^^^^^^^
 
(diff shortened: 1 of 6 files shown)


eteq:
(not that assuming this gets merged I should make a follow-on issue to investigate the one xfail's origin more definitively)

dhomeier:
Looks like the closed-file problem only popped up after adding the `read_fileobj_or_hdulist`, so I second the call for closer investigation.

eteq:
Aww geez, another one popped up in CI here.  I'm going to xfail it for the same reason on the theory that it's only for a few specific input file situations, but definitely needs investigation.

dhomeier:
> Looks like the closed-file problem only popped up after adding the `read_fileobj_or_hdulist` contextmanager, so I second the call for closer investigation.

Rather looks like another red herring; without the changes from #702 merged I am getting the same failures now as well. Seems to be exactly the problem reported in #711; will continue discussion there.

PR #719, opened 2020-09-30
Add common loaders from Data Central

aragilar:
For a bunch of AAT/other Australian surveys (both existing and to be released) we (Data Central) have been using these loaders to reduce duplicate code inside survey-specific loaders. Is this something specutils would be interested in having (I'd include the survey specific ones in a future PR, once we sort out where this code would sit inside specutils)?


aragilar:
@nmearl @eteq Any comments on this PR? I'd like to start creating PRs for the survey/instrument-specific loaders we've developed, but those PRs require the base loaders in the PR.

nmearl:
In a technical sense, this PR looks great. We would certainly want to move/rename the files to something more specific/descriptive (e.g. `dc_common`/`aat_common`, etc). @eteq thoughts?

aragilar:
I'm happy to rename the file, I figured if I named it `dc_common`, it would be assumed only Data Central created loaders could use it, whereas I though other groups may want to use those loaders as a basis for their surveys/data.

aragilar:
I've done the rename from `common` to `dc_common`.

aragilar:
I've fixed the tests, and added some additional changes we've made to our local copy.

nmearl:
Thanks @aragilar! We should be sure to follow the paradigm of the rest of the code base (e.g. function arguments shouldn't be one-per-line), but it's functionally fine, I think, for loader contributions.

aragilar:
Ah, sorry, I probably ran [black](https://github.com/psf/black) on it at some point, I'll try to remember to change that up before I push other loaders.

PR #722, opened 2020-09-30
Fix radial velocity/redshift setting

### LINKED ISSUE #721: Can't set radial velocity of Spectrum1D object
Trying to set the radial velocity of a Spectrum1D object raises an Attribute error.
This happens only in specutils version 1.1, but not in version 1.0.

Error Message:
```
Exception has occurred: AttributeError
SpectralAxis instance has no attribute 'with_radial_velocity'
```

Minimal Code:
```
import numpy as np
from specutils.spectra import Spectrum1D


wave = np.linspace(100, 200, 100) << u.AA
flux = np.ones(100) << u.one
spec = Spectrum1D(spectral_axis=wave, flux=flux, radial_velocity=0 * u.km / u.s)

spec.radial_velocity = 1 * u.km / u.s
```

nmearl:
Fixes #721. This PR addresses the issue in the transition to using `SpectralAxis` which broke the property setters on the `Spectrum1D` object.

### DIFF
--- a/specutils/spectra/spectral_axis.py
+++ b/specutils/spectra/spectral_axis.py
@@ -84,5 +84,5 @@ class SpectralAxis(SpectralCoord):
     def with_radial_velocity_shift(self, target_shift=None, observer_shift=None):
         if self.unit is u.pixel:
             raise u.UnitsError("Cannot transform spectral coordinates in pixel units")
-        super().with_radial_velocity_shift(target_shift=target_shift,
-                                           observer_shift=observer_shift)
+        return super().with_radial_velocity_shift(target_shift=target_shift,
+                                                  observer_shift=observer_shift)
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -376,7 +376,7 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
 
     @redshift.setter
     def redshift(self, val):
-        new_spec_coord = self.spectral_axis.with_redshift(val)
+        new_spec_coord = self.spectral_axis.with_radial_velocity_shift(val)
         self._spectral_axis = new_spec_coord
 
     @property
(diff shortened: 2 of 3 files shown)


javerbukh:
Received the following traceback when running the tests locally:
```
>       assert_quantity_allclose(spec.redshift, u.Quantity(0.4))

../../specutils/tests/test_spectral_axis.py:175: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

actual = <Quantity 0.45883691>, desired = <Quantity 0.4>, rtol = 1e-07, atol = None, kwargs = {}
np = <module 'numpy' from '/Users/javerbukh/Documents/specutils/.tox/test/lib/python3.8/site-packages/numpy/__init__.py'>
_unquantify_allclose_arguments = <function _unquantify_allclose_arguments at 0x11b8e9c10>

    def assert_quantity_allclose(actual, desired, rtol=1.e-7, atol=None,
                                 **kwargs):
        """
        Raise an assertion if two objects are not equal up to desired tolerance.
    
        This is a :class:`~astropy.units.Quantity`-aware version of
        :func:`numpy.testing.assert_allclose`.
        """
        import numpy as np
        from astropy.units.quantity import _unquantify_allclose_arguments
>       np.testing.assert_allclose(*_unquantify_allclose_arguments(
            actual, desired, rtol, atol), **kwargs)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 0.05883691
E       Max relative difference: 0.14709227
E        x: array(0.458837)
E        y: array(0.4)

../../.tox/test/lib/python3.8/site-packages/astropy/tests/helper.py:491: AssertionError
```

nmearl:
Thanks @javerbukh. Yea, we're aware. It's been a standing issue for @eteq and @astrofrog to look at.

eteq:
I think I identified the problem - it's that the way of translating the "new" and "old" redshifts isn't quite right.  There's a complex way to formulate it but I think the simpler approach I just pushed up in 1a56a8b should do the trick (I could have just suggested it in a review, but since I'd done it locally anyway to test it seemed easier to just push up a commit :shrug:).  Assuming this works I think it's set!

PR #724, opened 2020-10-14
Fix region extraction from Spectrum1D objects with multi-D flux

### LINKED ISSUE #690: Function `extract_region` fails with 3D data
When applying `extraction_region` to a `Spectrum1D` instance which was defined with a 3D flux array, the resulting `Spectrum1D` instance is broken.

For example, submiting a Spectrum1D with shape (100, 100, 2500) (2500 being the spatial axis size)  to `extract_region`, results in a Spectrum1D with shape (0, 100, 2500).

This suggests that `extract_region` wasn't coded to handle 3D data. The same problem may be affecting other parts of `specutils`.

rosteen:
I believe this fixes #690. I think I got the keywords that should be propagated through, but may have missed something.

### DIFF
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -51,7 +51,7 @@ def _to_edge_pixel(subregion, spectrum):
         if left_reg_in_spec_unit < spectral_axis[0]:
             left_index = 0
         elif left_reg_in_spec_unit > spectral_axis[-1]:
-            left_index = len(spectrum.spectral_axis)-1
+            left_index = len(spectrum.spectral_axis)
         else:
             try:
                 left_index = int(np.ceil(spectrum.wcs.world_to_pixel(
@@ -74,7 +74,7 @@ def _to_edge_pixel(subregion, spectrum):
                                                  u.spectral())
 
         if right_reg_in_spec_unit > spectral_axis[-1]:
-            right_index = len(spectrum.spectral_axis)-1
+            right_index = len(spectrum.spectral_axis)
         elif right_reg_in_spec_unit < spectral_axis[0]:
             right_index = 0
         else:
@@ -99,13 +99,17 @@ def extract_region(spectrum, region):
 
     Parameters
     ----------
-    spectrum: `~specutils.spectra.spectrum1d.Spectrum1D`
+    spectrum: `~specutils.Spectrum1D`
         The spectrum object from which the region will be extracted.
+
+    region: `~specutils.SpectralRegion`
+        The spectral region to extract from the original spectrum.
 
     Returns
     -------
-    spectrum: `~specutils.spectra.spectrum1d.Spectrum1D`
-        Excised spectrum.
+    spectrum: `~specutils.Spectrum1D` or list of `~specutils.Spectrum1D`
+        Excised spectrum, or list of spectra if the input region contained multiple 
+        subregions.
 
     Notes
     -----
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -108,7 +108,7 @@ def extract_region(spectrum, region):
     Returns
     -------
     spectrum: `~specutils.Spectrum1D` or list of `~specutils.Spectrum1D`
-        Excised spectrum, or list of spectra if the input region contained multiple 
+        Excised spectrum, or list of spectra if the input region contained multiple
         subregions.
 
     Notes
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -149,7 +149,18 @@ def extract_region(spectrum, region):
             if left_index > right_index:
                 left_index, right_index = right_index, left_index
 
-            extracted_spectrum.append(spectrum[left_index:right_index])
+            sub_spec = Spectrum1D(
+                        spectral_axis=spectrum.spectral_axis[left_index:right_index],
+                        flux=spectrum.flux[...,left_index:right_index],
+                        meta=spectrum.meta)
+
+            if spectrum.mask is not None:
+                sub_spec.mask=np.array(spectrum.mask)[...,left_index:right_index]
+
+            if spectrum.uncertainty is not None:
+                sub_spec.uncertainty=spectrum.uncertainty[...,left_index:right_index]
+
+            extracted_spectrum.append(sub_spec)
 
     # If there is only one subregion in the region then we will
     # just return a spectrum.
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -149,18 +149,7 @@ def extract_region(spectrum, region):
             if left_index > right_index:
                 left_index, right_index = right_index, left_index
 
-            sub_spec = Spectrum1D(
-                        spectral_axis=spectrum.spectral_axis[left_index:right_index],
-                        flux=spectrum.flux[...,left_index:right_index],
-                        meta=spectrum.meta)
-
-            if spectrum.mask is not None:
-                sub_spec.mask=np.array(spectrum.mask)[...,left_index:right_index]
-
-            if spectrum.uncertainty is not None:
-                sub_spec.uncertainty=spectrum.uncertainty[...,left_index:right_index]
-
-            extracted_spectrum.append(sub_spec)
+            extracted_spectrum.append(spectrum[..., left_index:right_index])
 
     # If there is only one subregion in the region then we will
     # just return a spectrum.
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -246,9 +246,26 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
         The first case is handled by the parent class, while the second is
         handled here.
         """
-        if len(self.flux.shape) > 1:
+
+        if self.flux.ndim > 1 or (type(item) == tuple and item[0] == Ellipsis):
+            if type(item) == tuple:
+                if len(item) == len(self.flux.shape) or item[0] == Ellipsis:
+                    spec_item = item[-1]
+                    if not isinstance(spec_item, slice):
+                        spec_item = slice(spec_item, spec_item+1, None)
+                        item = item[:-1] + (spec_item,)
+                else:
+                    # Slicing on less than the full number of axes means we want
+                    # to keep the whole spectral axis
+                    spec_item = slice(None, None, None)
+            else:
+                # Slicing with a single integer or slice uses the leading axis,
+                # so we keep the whole spectral axis, which is last
+                spec_item = slice(None, None, None)
+
             return self._copy(
                 flux=self.flux[item],
+                spectral_axis=self.spectral_axis[spec_item],
                 uncertainty=self.uncertainty[item]
                 if self.uncertainty is not None else None,
                 mask=self.mask[item] if self.mask is not None else None)
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -307,6 +307,17 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
 
         return self.__class__(**alt_kwargs)
 
+    @NDDataRef.mask.setter
+    def mask(self, value):
+        # Impose stricter checks than the base NDData mask setter
+        if value is not None:
+            value = np.array(value)
+            if not self.data.shape == value.shape:
+                raise ValueError(
+                    "Flux axis ({}) and mask ({}) shapes must be the "
+                    "same.".format(self.data.shape, value.shape))
+        self._mask = value
+
     @property
     def frequency(self):
         """
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,11 @@ New Features
 Bug Fixes
 ^^^^^^^^^
 
+- Fix ``extract_region`` behavior and slicing for ``Spectrum1D`` objects
+  that have multi-dimensional flux arrays. Extracting a region that extends
+  beyond the limits of the data no longer drops the last data point in the
+  returned spectrum. [#724]
+
 Documentation
 ^^^^^^^^^^^^^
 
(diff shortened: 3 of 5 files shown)


gkanarek:
Maybe include metadata & WCS?

nmearl:
Thanks, @rosteen. However, this should really be addressed in the [class indexer](https://github.com/astropy/specutils/blob/master/specutils/spectra/spectrum1d.py#L234) instead of a one-off fix here. Also, can you add a test to `test_region_extract.py` to handle this case?

rosteen:
@nmearl the thought had crossed my mind that fixing the indexer to accept e.g. spec[...,1:5] would fix this, but did the quick easy fix. I agree that putting in some more effort into the better lower-level fix instead is the better option, I'll hopefully have more time later this week to figure that out. I'll also add a test.

rosteen:
@nmearl I think this is ready for another look - the only tests failing locally are also failing in master, I think. Pretty sure I covered all the slicing cases, and I added a simple unit test for the multidimensional flux extraction case.

specutils/spectra/spectrum1d.py
```diff
@@ -246,9 +246,26 @@ def __getitem__(self, item):
         The first case is handled by the parent class, while the second is
         handled here.
         """
+
         if len(self.flux.shape) > 1:
```
dhomeier, line 250:
```suggestion
        if self.flux.ndim > 1 or (type(item) == tuple and item[0] == Ellipsis):
```
This would enable the `[..., lower, upper]` syntax for 1D and multi-D flux spectra alike.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -149,7 +149,11 @@ def extract_region(spectrum, region):
             if left_index > right_index:
                 left_index, right_index = right_index, left_index
 
-            extracted_spectrum.append(spectrum[left_index:right_index])
+            if len(spectrum.flux.shape)==1:
+                extracted_spectrum.append(spectrum[left_index:right_index])
+            else:
+                extracted_spectrum.append(spectrum[..., left_index:right_index])
+
```
dhomeier, line 156:
```suggestion
            extracted_spectrum.append(spectrum[..., left_index:right_index])
```
See suggestion for `Spectrum1D.__getitem__`, which would obviate the need for the test.
Note: double blank lines only between function definitions (E303)

dhomeier (changes requested):
Implementation looks good to me overall. Fundamentally I think it is unfortunate that `spectrum[lower:upper]` returns completely different things depending on whether flux is 1D or multi-D, but since this is established behaviour now, I guess that train has departed.
My suggested modification to `__getitem__` would amend this a little bit by allowing at least the `spectrum[..., lower:upper]` syntax to function the same independent of dimensionality. By comparison, slicing a 1d ndarray by `flux[..., m:n]` also is equivalent to `flux[m:n]`.

Two unrelated notes on `extract_spectral_region`:
https://github.com/astropy/specutils/blob/705875bc6ed7768d9b0dfd2dd9a14fefd15a4f03/specutils/manipulation/extract_spectral_region.py#L53-L54

incorrectly clips the spectrum one point short of its upper end if the`SpectralRegion` extends beyond its original length (`-1` should go).

And the docstring states `extract_region` invariably returns a `Spectrum1D`, when it is a list of spectra for regions with subregions.

specutils/manipulation/extract_spectral_region.py
rosteen: Good call, much cleaner

rosteen:
@dhomeier I'll have an update addressing your other comments tomorrow, thanks for pointing those things out.

rosteen:
I updated the docstring to reflect the possibility of list output, and fixed the upper bounds for outside-of-spectrum region limits. I also noticed that the change to always using [..., lower:upper] broke the case where the mask is provided as a simple list rather than a numpy array, so I added some code overriding the NDDataRef mask setter to force the mask to be an ndarray and check that its shape matches the shape of the flux array. 

This introduced a doctest failure that I'm currently stumped on how to fix, still thinking about that one. Suggestions welcome.

dhomeier (changes requested):
Just some extra suggestions for brevity – the correct usage is not entirely clear, since the class is exposed as any of
`~specutils.spectra.spectrum1d.Spectrum1D`, `~specutils.spectra.Spectrum1D`, `~specutils.Spectrum1D`. For convenience I would go for the latter, as that is probably how people are mostly using it (and 90 % of the existing references use the short version).

Tests for this function all seem to have passed, although the upper limit case you fixed in fact is missing tests. Could perhaps be added to `test_region_simple_check_ends` as
```
    region = SpectralRegion(0*u.um, 15*u.um)
    sub_spectrum = extract_region(spectrum, region)
    assert sub_spectrum.spectral_axis.value[0] == 1

    region = SpectralRegion(8*u.um, 30*u.um)
    sub_spectrum = extract_region(spectrum, region)
    assert sub_spectrum.spectral_axis.value[-1] == 25
```

specutils/manipulation/extract_spectral_region.py
```diff
@@ -101,10 +101,14 @@ def extract_region(spectrum, region):
     ----------
     spectrum: `~specutils.spectra.spectrum1d.Spectrum1D`
         The spectrum object from which the region will be extracted.
+    region: `~specutils.spectra.spectral_region.SpectralRegion`
```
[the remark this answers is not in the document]
nmearl: This does not matter as the `~` is truncates the linked reference so that it only displays "spectrum: Spectrum1D" in the documentation.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -101,10 +101,14 @@ def extract_region(spectrum, region):
...
+        The spectral region to extract from the original spectrum.
 
     Returns
     -------
-    spectrum: `~specutils.spectra.spectrum1d.Spectrum1D`
+    spectrum: `~specutils.spectra.spectrum1d.Spectrum1D` or list of
+              `~specutils.spectra.spectrum1d.Spectrum1D` if input region
+              contains multiple subregions
```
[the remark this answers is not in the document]
nmearl: See above.

specutils/manipulation/extract_spectral_region.py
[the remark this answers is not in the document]
dhomeier: `help(extract_region)` currently displays this
```
extract_region(spectrum, region)
    Extract a region from the input `~specutils.Spectrum1D`
    defined by the lower and upper bounds defined by the ``region``
    instance.  The extracted region will be returned as a new
    `~specutils.Spectrum1D`.
    
    Parameters
    ----------
    spectrum: `~specutils.spectra.spectrum1d.Spectrum1D`
...
```
so it's also somewhat inconsistent.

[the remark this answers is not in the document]
nmearl: Good point.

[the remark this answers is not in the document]
dhomeier: First forgot the header above, displaying the full path, which is a good thing. For the rest it might me down to taste...

dhomeier:
> This introduced a doctest failure that I'm currently stumped on how to fix, still thinking about that one. Suggestions welcome.

Missed that one. I think Sphinx simply wants a blank line between the parameters' blocks (before `returns`).

dhomeier:
> I updated the docstring to reflect the possibility of list output, and fixed the upper bounds for outside-of-spectrum region limits. I also noticed that the change to always using [..., lower:upper] broke the case where the mask is provided as a simple list rather than a numpy array, so I added some code overriding the NDDataRef mask setter to force the mask to be an ndarray and check that its shape matches the shape of the flux array.

I have to admit I did not think all possible inputs to `__getitem__` through; `Ellipsis` as first argument just seemed to cover any cases applicable to 1D flux spectra. I guess an extra check cannot harm.
On the error message, is specutils preferentially using f-strings by now?

rosteen:
@dhomeier the failing test I was referring to is actually in manipulation.rst, not the docstring in extract_spectral_region.py. Apparently the result array was previously being truncated, but now the whole thing is getting printed out:
```
269    >>> spec_w_unc.uncertainty # doctest: +ELLIPSIS

Expected:
    StdDevUncertainty([0.18823157, ..., 0.18823157])
Got:
    StdDevUncertainty([0.1886835, 0.1886835, 0.1886835, 0.1886835, 0.1886835,
                       0.1886835, 0.1886835, 0.1886835, 0.1886835, 0.1886835,
                       0.1886835, 0.1886835, 0.1886835, 0.1886835, 0.1886835, (etc)
```

dhomeier:
> Apparently the result array was previously being truncated, but now the whole thing is getting printed out:
> 
> ```
> 269    >>> spec_w_unc.uncertainty # doctest: +ELLIPSIS
> 
> Expected:
>     StdDevUncertainty([0.18823157, ..., 0.18823157])
> Got:
>     StdDevUncertainty([0.1886835, 0.1886835, 0.1886835, 0.1886835, 0.1886835,
> ```

I think that is supposed to be handled by the `doctest: +ELLIPSIS`, but the actual values have changed with the fix to the `left_index`/`right_index` calculation. Without tracking down the internals I'd say that is to be expected from now including the correct spectral range. I confirmed with various astropy and numpy versions that the old version returned `StdDevUncertainty([0.18823157, ...])`, and now it's `StdDevUncertainty([0.1886835, ...])`. The latter (new) value is also produced by the old version the changing the example to
`noise_region = specutils.SpectralRegion([(10, 7), (3, 1)] * u.GHz)`
i.e. forcing the subregions to stick exactly to the upper and lower bounds of the spectrum (10 GHz, 1 GHz) would have got the correct value; so I'd just change the expected value in the docs to 0.1886835.

rosteen:
Ha, I was so focused on thinking the ellipsis was the source of the problem that I didn't even notice the numbers themselves were slightly different. This is why it's so useful to have another set of eyes on things!

dhomeier:
Thanks, looks good, I am a bit puzzled by the docbuild error.
That test run in #562 raised the same warnings, but did not treat them as error, though calling the same version of `sphinx-build -W`. Maybe @eteq or @nmearl can comment if there is some sphinx configuration file still to be updated.
Anyway, while `fitting.rst` is using `noise_region_uncertainty`, I find no evidence for changed output before and after this PR.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -99,12 +99,15 @@ def extract_region(spectrum, region):
 
     Parameters
     ----------
-    spectrum: `~specutils.spectra.spectrum1d.Spectrum1D`
+    spectrum: `~specutils.Spectrum1D`
         The spectrum object from which the region will be extracted.
+    region: `~specutils.SpectralRegion`
```
dhomeier, line 104:
```suggestion
        The spectrum object from which the region will be extracted.

    region: `~specutils.SpectralRegion`
```

specutils/manipulation/extract_spectral_region.py
```diff
@@ -99,12 +99,16 @@ def extract_region(spectrum, region):
...
 
+    region: `~specutils.SpectralRegion`
+        The spectral region to extract from the original spectrum.
+
     Returns
     -------
-    spectrum: `~specutils.spectra.spectrum1d.Spectrum1D`
+    spectrum: `~specutils.Spectrum1D` or list of `~specutils.Spectrum1D`
+              if input region contains multiple subregions
         Excised spectrum.
```
rosteen, line 112:
```suggestion
    spectrum: `~specutils.Spectrum1D` or list of `~specutils.Spectrum1D`
        Excised spectrum, or list of spectra if the input region contained multiple subregions.
```

I think it was actually this line break causing problems, we'll see shortly...

dhomeier: Yes, I think so, looking at the correct line no. now.
Unfortunately the download error got in the way, but I tested the doc build locally and did not see the warning anymore after fixing the indentation.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -99,13 +99,16 @@ def extract_region(spectrum, region):
...
 
+    region: `~specutils.SpectralRegion`
+        The spectral region to extract from the original spectrum.
+
     Returns
     -------
-    spectrum: `~specutils.spectra.spectrum1d.Spectrum1D`
-        Excised spectrum.
+    spectrum: `~specutils.Spectrum1D` or list of `~specutils.Spectrum1D`
+        Excised spectrum, or list of spectra if the input region contained multiple subregions.
```
eteq, line 111:
```suggestion
    spectrum: `~specutils.Spectrum1D` or list of `~specutils.Spectrum1D`
        Excised spectrum, or list of spectra if the input region contained multiple 
        subregions.
```
Exceedingly minor pep8 item

eteq (changes requested):
This is almost ready, but I think it needs a changelog entry since it's technically an API change that the edge behavior is now different.  I think I'm OK with this change, but it needs to be recorded so it doesn't silently surprise users.

specutils/manipulation/extract_spectral_region.py
dhomeier: Having just looked at the details of pep8 in light of #621, they mandate to keep the 72-char limit for docstrings even where longer code lines are allowed, right? flake8 by default seems to assume the same for code and docs.

rosteen:
Added a changelog entry and rebased to master (including squashing a bunch of the tiny commits).

specutils/manipulation/extract_spectral_region.py
```diff
@@ -99,13 +99,17 @@ def extract_region(spectrum, region):
...
 
+    region: `~specutils.SpectralRegion`
+        The spectral region to extract from the original spectrum.
+
     Returns
     -------
-    spectrum: `~specutils.spectra.spectrum1d.Spectrum1D`
-        Excised spectrum.
+    spectrum: `~specutils.Spectrum1D` or list of `~specutils.Spectrum1D`
+        Excised spectrum, or list of spectra if the input region contained multiple
```
rosteen, line 111:
```suggestion
        Excised spectrum, or list of spectra if the input region contained multiple
```

eteq (approved):
LGTM now!  The build failure is again a red herring.

PR #726, opened 2020-10-20
Move data_prep.rst from jdaviz to specutils as Working with Spectral …

PatrickOgle:
Add section to ReadTheDocs main index  that discusses how to use Spectral Cube to read in a cube, then collapse
to create a Spectrum1D.

### DIFF
--- a/docs/spectral_cube.rst
+++ b/docs/spectral_cube.rst
@@ -1,0 +1,72 @@
+###########################
+Working with Spectral Cubes
+###########################
+
+Spectral cubes can be read in using the
+`SpectralCube <https://spectral-cube.readthedocs.io/en/latest/>` package then
+collapsed using various methods to yield a Spectrum1D object.  A specific
+example of this workflow is demonstrated here.
+
+Collapsing Cubes
+================
+Currently, the ``specutils`` library does not support loading three dimensional
+spectral cubes and manipulating their non-spectral axes, so parsing a spectral
+cube to a :class:`~specutils.Spectrum1D` object requires some programmatic
+manipulation using the ``spectral-cube`` library. Below is an example of
+loading a spectral cube object, collapsing the cube, and creating a
+:class:`~specutils.Spectrum1D` object.
+
+Loading a cube
+==============
+
+We'll use the example cube data found in the ``spectral-cube`` package, and
+load the data from the repository directly into a new :class:`~spectral_cube.SpectralCube`
+object::
+
+    >>> from astropy.utils.data import download_file
+    >>> from spectral_cube import SpectralCube
+    >>> sc = SpectralCube.read(download_file("https://github.com/radio-astro-tools/spectral-cube/raw/master/spectral_cube/tests/data/example_cube.fits"), format='fits')
+    >>> sc
+    SpectralCube with shape=(7, 4, 3) and unit=Jy / beam:
+    n_x:      3  type_x: RA---ARC  unit_x: deg    range:    52.231466 deg:   52.231544 deg
+    n_y:      4  type_y: DEC--ARC  unit_y: deg    range:    31.243639 deg:   31.243739 deg
+    n_s:      7  type_s: VRAD      unit_s: m / s  range:    14322.821 m / s:   14944.909 m / s
+
+We see that the example cube is small, with a shape of ``(7, 3, 4)``; two
+spatial axes ``RA`` and ``DEC``, and one spectral axis that's in velocity
+units.
+
+Extracting the spectral data
+============================
+
+Now that we have a data cube, our mission is to collapse this cube along the
+two spatial axes to get the representative spectral data. We can then parse the
+result into a :class:`~specutils.Spectrum1D` object. ``SpectralCube`` can take
+and apply arbitrary functions over axes of the data, in this case we can apply
+a simple median function over the spatial axes::
+
+    >>> import numpy as np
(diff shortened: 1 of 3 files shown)


eteq (approved):
Mostly looks good, but I added one suggestion to add some text giving a bit more context on how this is only one of the spectral cube formats, and we don't yet have a solution for the others.

docs/spectral_cube.rst
```diff
@@ -0,0 +1,72 @@
+###########################
+Working with Spectral Cubes
+###########################
+
+Spectral cubes can be read in using the
+`SpectralCube <https://spectral-cube.readthedocs.io/en/latest/>` package then
```
eteq, line 6: I think the CI failure is due to this:
```suggestion
`SpectralCube <https://spectral-cube.readthedocs.io/en/latest/>`_ package then
```

PatrickOgle: good catch!

javerbukh (approved):
Looks good. I would wait until tests are passing to merge, though. Any idea why they're failing?

docs/spectral_cube.rst
```diff
@@ -0,0 +1,80 @@
...
+###########################
+
+Spectral cubes can be read in using the
+`SpectralCube <https://spectral-cube.readthedocs.io/en/latest/>`_ package then
+collapsed using various methods to yield a Spectrum1D object.  A specific
+example of this workflow is demonstrated here.
+
+Note that the workflow described here is for spectral cubes that are rectified
+such that one of the axes is entirely spectral and all the spaxels have the same
+`spectral_axis` values - i.e., case 2 in
```
eteq, line 12: oops I think the test failure is my bad - this should be either
```suggestion
``spectral_axis`` values - i.e., case 2 in
```
or
```suggestion
`~specutils.Spectrum1D.spectral_axis` values - i.e., case 2 in
```

rosteen:
I'm pretty sure that (at least most of) the test failures are actually unrelated to this PR. Looks like test_loaders is hitting `ValueError: lookup_table should be an array with 1 dimensions.` running `gwcs_from_array`, I assume due to a change in an underlying package.

eteq:
Yeah, agreed @rosteen, there's several unrelated failures.  But since this is only the docs, I've been looking specificially at the third build in the matrix (the one that has `build_docs` as its command - as an aside, #727 is about making this more readable!), you'll see there's documentation-related failures.  It looks like this change fixed the error I said above, but now there's a new one at line 30. 

Once that's passing I think we can merge this without worrying about the other failures right now.

PatrickOgle:
@eteq Can this be fixed by an edit to the .rst file for this PR?

docs/spectral_cube.rst
```diff
@@ -0,0 +1,80 @@
...
+cube to a :class:`~specutils.Spectrum1D` object requires some programmatic
+manipulation using the ``spectral-cube`` library. Below is an example of
+loading a spectral cube object, collapsing the cube, and creating a
+:class:`~specutils.Spectrum1D` object.
+
+Loading a cube
+==============
+
+We'll use the example cube data found in the ``spectral-cube`` package, and
+load the data from the repository directly into a new :class:`~spectral_cube.SpectralCube`
```
PatrickOgle, line 31:
```suggestion
load the data from the repository directly into a new ``SpectralCube``
```

docs/spectral_cube.rst
```diff
@@ -0,0 +1,80 @@
...
+============================
+
+Now that we have a data cube, our mission is to collapse this cube along the
+two spatial axes to get the representative spectral data. We can then parse the
+result into a :class:`~specutils.Spectrum1D` object. ``SpectralCube`` can take
+and apply arbitrary functions over axes of the data, in this case we can apply
+a simple median function over the spatial axes::
+
+    >>> import numpy as np
+    >>> sc_spec = sc.apply_numpy_function(np.mean, axis=(1,2), projection=True)
```
nmearl, line 57:
```suggestion
    >>> sc_spec = sc.mean(axis=(1,2))
```

Just to offer an alternative that might make it conceptually easier for users, you can call the `mean` method directly which defaults to `projection=True` and avoid having to explain this edge case.

docs/spectral_cube.rst
```diff
@@ -0,0 +1,80 @@
...
+    >>> sc_spec.shape
+    (7,)
+
+The ``projection`` keyword returns a lower dimesional object that retains the WCS
+information that we'll need to use when constructing the `Spectrum1D` object.
```
nmearl, line 62:
```suggestion
```

javerbukh:
@PatrickOgle I think if you make the edits Nick suggested we can go ahead with merging this PR.

nmearl (approved):

docs/spectral_cube.rst
```diff
@@ -0,0 +1,80 @@
...
+Loading a cube
+==============
+
+We'll use the example cube data found in the ``spectral-cube`` package, and
+load the data from the repository directly into a new ``SpectralCube``
+object::
+
+    >>> from astropy.utils.data import download_file
+    >>> from spectral_cube import SpectralCube
+    >>> sc = SpectralCube.read(download_file("https://github.com/radio-astro-tools/spectral-cube/raw/master/spectral_cube/tests/data/example_cube.fits"), format='fits')
```
pllim, line 36: For future reference, any doctest that downloads stuff needs to be marked as remote data. Otherwise, you'll hit the server too much and cause unrelated failures.

xref #838

Issue #709, opened 2020-08-10
Smoothing spectra drops meta information about spectra


aragilar:
It appears https://github.com/astropy/specutils/blob/master/specutils/manipulation/smoothing.py#L97 does not pass the metadata on the spectra along, stripping it out. Was this intended, or should I make a PR fixing this.

nmearl:
Hi @aragilar, this was not intentional, so please feel free to provide a PR that passes on the meta information!

javerbukh:
Should this issue be closed? @nmearl @eteq

PR #732, opened 2020-11-12
Preserve stored meta on smooth

nmearl:
Resolves issue #709. Ensures that meta information on the spectrum object is passed through smoothing operations.


javerbukh (approved):
Looks good!

nmearl:
Thanks @javerbukh!

PR #736, opened 2020-11-26
Add github actions copied from astropy

### LINKED ISSUE #733: TST: Use GitHub Actions
Please see https://groups.google.com/g/astropy-dev/c/45frnTEAX-U

aragilar:
This I think should close #733. Feel free to edit this as needed (I'm not that familiar with github actions, I've mainly used other CI services, so feel free to completely rewrite/rebase this branch - I'm posting this so that specutils will have a CI running).


pllim:
Thank you! Since this is the initial implementation from a fork, it would only run on your fork but given how things are currently set up, it is not running. So you might have to push a temporary commit to let it run from your fork to prove that this works.

https://github.com/aragilar/specutils/actions

mwcraig:
THis looks pretty good. A few questions:

+ Does `specutils` need the 32-bit and parallel tests?
+ I don't think the label action will work unless you created a secret for this repository. 

OTher than that it looks good, but let's see what happens in the actions on your fork.

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,145 @@
+name: CI
+
+on:
+  push:
+    branches:
+    - master
```
mwcraig, line 6: Drop this for now to make the actions run on your fork. 

```suggestion
```

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,145 @@
...
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: (Allowed Failure) Python 3.7 with remote data and dev version of key dependencies
+            os: ubuntu-latest
+            python: 3.7
+            toxenv: py37-test-devdeps
```
eteq, line 95: This is what it is in the latest travis.yml - I'm not sure if there's a reason for that or not, but probably best to stay consistent.
```suggestion
            python: 3.8
            toxenv: py38-test-devdeps
```

.github/workflows/ci_cron_weekly.yml
```diff
@@ -0,0 +1,58 @@
...
+
+          # We check numpy-dev also in a job that only runs from cron, so that
+          # we can spot issues sooner. We do not use remote data here, since
+          # that gives too many false positives due to URL timeouts. We also
+          # install all dependencies via pip here so we pick up the latest
+          # releases.
+          - name: Python 3.7 with dev version of key dependencies
+            os: ubuntu-latest
+            python: 3.7
+            toxenv: py37-test-devdeps
```
eteq, line 26: (as above)
```suggestion
            python: 3.8
            toxenv: py38-test-devdeps
```

eteq (changes requested):
Thanks for this @aragilar ! A few things I noticed that need updating to be consistent with what `travis` was doing:


* We need a build that tests against the latest astropy LTS (not in astropy itself, of course, since it *is* the LTS).  This alread exists in `tox`, it just needs a matrix element.  `py36-test-astropylts-numpy116` is the toxenv you want - you can just add that to the matrix as another item
* There isn't a separate doc build right now, so  this should also get a `build_docs` in the matrix
* Since it's all subsumed, this PR should delete the `.travis.yml` too (yay!)
* I agree there's no need for the 32 bit/parallel runs, so can probably just delete that one.
* (one other thing that was easy to express as an inline suggestion)

eteq:
Oh and an *alternative* to my suggestion above is to get the readthedocs build set up like `astropy`, but I think it's probably better to do it in actions as a later step since it requires fiddling around with RTD settings (which I think requires admin rights right now).

pllim:
Re: RTD -- Yes, it is only one-click to enable PR build there but you do need to be admin there. Looking at https://readthedocs.org/projects/specutils/ , the following people can enable it:

* @keflavich 
* @nmearl 
* Whoever has the keys to `astropy` RTD account
* Whoever has the keys to `specutils` RTD account

aragilar:
I've added the inline suggestions, which looks like it's caused actions to run for the commit - https://github.com/aragilar/specutils/actions/runs/392730140. I'll have a go at the other changes, but feel free to push changes, rather than waiting for me to make them.

eteq:
> Re: RTD -- Yes, it is only one-click to enable PR build there but you do need to be admin there. 

I didn't realize it was this easy!  I just turned it on right now. So lets give that a try - I'm a bit worried about some of the quirks of the `specutils` docs getting in the way here, but we should just try it and see since the RTD approach is closer to what the user sees in the end anyway.

So @aragilar - never mind about my suggestion to add  `build_docs` to the matrix.

pllim:
> I'm a bit worried about some of the quirks

In your `.readthedocs.yml`, you can set it to allow warnings if you worry about warnings failing the RTD PR build. But then again, once you ignore warnings, they tend to multiply like rabbits and opens up a rabbit hole. So, ignore at your own risk.

Hope this helps!

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,123 @@
...
+
+          - name: Python 3.7 with all optional dependencies (MacOS X)
+            os: macos-latest
+            python: 3.7
+            toxenv: py37-test-alldeps
+            toxposargs: --durations=50
+
+          - name: Python 3.6 with astropy LTS
+            os: ubuntu-16.04
+            python: 3.7
```
mwcraig, line 63:
```suggestion
            python: 3.6
```

(I'm not a maintainer so I can't push to your branch)

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,123 @@
...
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Install language-pack-de and tzdata
+      if: startsWith(matrix.os, 'ubuntu')
+      run: sudo apt-get install language-pack-de tzdata
```
eteq, line 83:
```suggestion
      run: sudo apt-get install language-pack-de tzdata graphviz
```
I think this is what's necessary to get the doc build to run...

eteq:
Huzzah!  https://github.com/aragilar/specutils/runs/1482516182 looks like it all working except for the "allowed failure" (which I'm not sure we actually want anyway in the PRs vs the weeklies, but we can address that in a follow-on PR.

eteq (approved):

eteq:
I think this is probably ready now, but I'm just going to give a bit of time in case @nmearl wants to jump in (as the originator of the most recent iteration of the `.travis.yml`)

nmearl (approved):
This looks good to me. Although, it seems we might need to add another matrix entry in the future to include the external packages tox environment that includes the JWST pipeline to test the IO, but that might be fine as a follow-up.

pllim:
CI passed. Any reason why this is not yet merged? Thanks!

javerbukh:
Pinging @eteq and @nmearl to merge if everything looks good

eteq:
@pllim - the reason was that I was waiting for @nmearl... and missed that he approved it!  So merged just now.

Issue #735, opened 2020-11-25
'SpectralGWCS' object has no attribute 'to_fits'


bmiszalski:
Loading an SDSS spectrum with the 'SDSS-I/II spSpec' format loader and then writing to fits using format "wcs1d-fits" writer raises error: 

ValueError: Only Spectrum1D objects with valid WCS can be written as wcs1d: 'SpectralGWCS' object has no attribute 'to_fits'

Please see attached fits file and minumum example script.
We include in the script the output of running the script and environment info.
[SpectralGWCSBug.tar.gz](https://github.com/astropy/specutils/files/5594130/SpectralGWCSBug.tar.gz)

aragilar:
I think I tracked down the bug: the gwcs wcs is not API compatible with the astropy wcs, and wcs1d-fits assumes that `spec.wcs` is the astropy one. As the SDSS-I/II spSpec does not specify a wcs, it instead uses the gwcs wcs system, even though the astropy one would be sufficient. There are three possible fixes that I can see (ordered by how invasive they are):

 1. Set the wcs in the SDSS-I/II spSpec loader (thought this does not fix other cases where the gwcs wcs is defaulted to).
 2. Switch the default wcs in specutils to the astropy one (as based on reading the gwcs docs, you only need gwcs when you want a more general wcs system, which doesn't make sense as the default).
 3. Have gwcs ensure its API matches that of astropy (which may be more of a challenge in terms of maintenance, but to me seems to be required if we allow `spec.wcs` to be anything other than an astropy wcs object).

I can provide a PR with the first fix, but it seems like 2 or 3 are the real solution to fixing the bug.

PR #738, opened 2020-11-30
Use corrected WCS in SDSS loaders

aragilar:
See #735. This only solves the case for some of the SDSS loaders, the more general WCS issues raised by #735 still need to be worked out.

### DIFF
--- a/specutils/io/default_loaders/sdss.py
+++ b/specutils/io/default_loaders/sdss.py
@@ -3,7 +3,6 @@ Loader for SDSS individual spectrum files: spec_ files.
 
 .. _spec: https://data.sdss.org/datamodel/files/BOSS_SPECTRO_REDUX/RUN2D/spectra/PLATE4/spec.html
 """
-from math import log
 import os
 import re
 import _io
@@ -43,7 +42,7 @@ def _sdss_wcs_to_log_wcs(old_wcs):
     w0 = old_wcs.wcs.crval[0]
     w1 = old_wcs.wcs.cd[0,0]
     crval = 10 ** w0
-    cdelt = crval * w1 * log(10)
+    cdelt = crval * w1 * np.log(10)
     cunit = old_wcs.wcs.cunit[0] or Unit('Angstrom')
     ctype = "WAVE-LOG"
 
--- a/specutils/io/default_loaders/sdss.py
+++ b/specutils/io/default_loaders/sdss.py
@@ -3,6 +3,7 @@ Loader for SDSS individual spectrum files: spec_ files.
 
 .. _spec: https://data.sdss.org/datamodel/files/BOSS_SPECTRO_REDUX/RUN2D/spectra/PLATE4/spec.html
 """
+from math import log
 import os
 import re
 import _io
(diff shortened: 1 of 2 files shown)


specutils/io/default_loaders/sdss.py
```diff
@@ -3,6 +3,7 @@
 
 .. _spec: https://data.sdss.org/datamodel/files/BOSS_SPECTRO_REDUX/RUN2D/spectra/PLATE4/spec.html
 """
+from math import log
```
nmearl, line 6:
```suggestion
```

Just a nit-pick: I think we should try to stick to numpy functions since we use those for all our other math needs.

specutils/io/default_loaders/sdss.py
```diff
@@ -26,6 +27,35 @@
...
+
+    The FITS standard uses a natural log with a sightly different formulation,
+    see WCS Paper 3 (which discusses spectral WCS).
+
+    This function does the conversion from the SDSS WCS to FITS WCS.
+    """
+    w0 = old_wcs.wcs.crval[0]
+    w1 = old_wcs.wcs.cd[0,0]
+    crval = 10 ** w0
+    cdelt = crval * w1 * log(10)
```
nmearl, line 46:
```suggestion
    cdelt = crval * w1 * np.log(10)
```

nmearl (commented):
This looks good. I wasn't aware of this issue and am kind of surprised it hasn't come up before?

Can you add a small test with some WCS that is SDSS-esque and tests the conversion?

aragilar:
I've got a test, but I'm stuck on how to set wcs.wcs.cd (`wcs.wcs.cd[0,0] = 1.00000000000000E-04` doesn't work), any ideas?

ibusko:
@aragilar could you commit the test you have so we can try to make it work?

aragilar:
@ibusko I've pushed the test, feel free to push changes to this branch.

ibusko:
@aragilar looks like at that point in your test (3rd line), `sdss_wcs.wcs.cd` isn't initialized. There is a `sdss_wcs.wcs.pc` element initialized as 2x2 array though. Maybe you should stick the value in there instead?

aragilar:
`CD1_1` is where the wcs information is in the FITS file, so `sdss_wcs.wcs.cd[0,0] = 1.00000000000000E-04` needs to appear in the test somehow. Setting up `sdss_wcs.wcs.cd` is what is unclear to me.

ibusko:
Stepping thru the code with the debugger, I can see that the `WCS` constructor in the first line could be provided with a FITS header initialized with the CD keywords. The constructor has a keyword parameter named `header`, and that header is used to initialize a lot of stuff. It's passed to the constructor like in `WCS(naxis=2, header=fits_header)`. 

So, instead of setting CRVAL[0] and CD[0,0] after the WCS object is built, you would first create a FITS header instance, initialized with the values you need, and then pass it to the `WCS` constructor. I didn't try it myself, but I am pretty sure that it should work.

ibusko:
Now the test is failing because of a incompatibility in the parameters passed via the header. The WCS constructor call should have `naxis=1`. 

If you change that, the code passes thru the constructor but the test still fails at the assert statement. If you fix the values, I believe it will be good to go.

ibusko (approved):
The last commit fixed everything. The tests run to completion.

The CI build errors seem to be unrelated to this PR, since they show up on every other PR that relies on the same CI tool. I am approving this PR, pending an investigation of the causes of the failures.

aragilar:
I've rebased this to see if the unrelated failing tests pass now.

ibusko:
Tests look good. The one failure seems to be related to the SDSS server not being up, or something like that.

nmearl:
> Tests look good. The one failure seems to be related to the SDSS server not being up, or something like that.

It's actually due to an issue with the latest dev of numpy, however it's an allowed failured (github CI just doesn't allow you to show allowed failures as passing).

Thanks for the contribution, @aragilar!

PR #751, opened 2021-01-05
Fix linkcheck error and include py39

aragilar:
This fixes one of the linkcheck errors and adds py39 to tox. Depending on the openssl settings that the CI runs under, links to SDSS may not work as they need to update their webserver settings (see https://ssl-config.mozilla.org/)

PR #752, opened 2021-01-05
Stop fast-failing for now

aragilar:
It appears that the docs having issues with links causes all the tests to be skipped. Hopefully removing fast-failing fixes this.

#751 is where I'm trying to fix the link errors.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -11,7 +11,7 @@ jobs:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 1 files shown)


nmearl:
Re-running tests after #751.

aragilar:
Currently the docs are failing due to issues with the TLS connection to SDSS.

aragilar:
Can this be merged, as otherwise the SDSS TLS issue means that most tests are skipped?

eteq (approved):
I'm a bit confused why some of the more recent PRs don't seem to be fast-failing, but I agree it's better to be cautious about this.

Issue #744, opened 2020-12-09
Implement spectral_slab wrapper around region extraction


rosteen:
The actual functionality of `spectral-cube`'s `spectral_slab` method is handled by `manipulation.extract_region`, but in order to facilitate a smooth transition from `spectral-cube` we may want to provide an entry point to that functionality with the `spectral_slab` name.

rosteen:
Closed by #753 .

PR #753, opened 2021-01-05
Add spectral_slab function.

ibusko:
As per issue #744 

This PR also includes a bit of refactoring/cleanup of `test_region_extract.py`

### DIFF
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -163,7 +163,7 @@ def extract_region(spectrum, region):
     return extracted_spectrum
 
 
-def spectral_slab(spectrum, s1, s2):
+def spectral_slab(spectrum, lower, upper):
     """
     Extract a slab from the input `~specutils.Spectrum1D`
     defined by the lower and upper bounds defined by the ``region``
@@ -175,8 +175,8 @@ def spectral_slab(spectrum, s1, s2):
     spectrum: `~specutils.Spectrum1D`
         The spectrum object from which the region will be extracted.
 
-    s1, s2: spectral coordinates limiting the slab
-        Define the spectral coordinate bound of the region to extract
+    lower, upper: `~astropy.units.Quantity`
+        The lower and upper bounds of the region to extract
         from the original spectrum.
 
     Returns
@@ -191,6 +191,6 @@ def spectral_slab(spectrum, s1, s2):
     transition from spectral-cube.
 
     """
-    region = SpectralRegion(s1, s2)
+    region = SpectralRegion(lower, upper)
 
     return extract_region(spectrum, region)
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -3,7 +3,7 @@ from math import floor, ceil  # faster than int(np.floor/ceil(float))
 import numpy as np
 
 from astropy import units as u
-from .. import Spectrum1D
+from ..spectra import Spectrum1D, SpectralRegion
 
 __all__ = ['extract_region', 'spectral_slab']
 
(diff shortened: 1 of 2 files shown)


rosteen (changes requested):
The spectral-cube slab functions takes upper and lower spectral bounds as input rather than a region (see https://spectral-cube.readthedocs.io/en/latest/manipulating.html#extracting-a-spectral-slab), so to preserve back-compatibility I think this should take the bounds as input and construct the region internally.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -161,3 +161,36 @@ def extract_region(spectrum, region):
...
+    Returns
+    -------
+    spectrum: `~specutils.Spectrum1D` or list of `~specutils.Spectrum1D`
+        Excised spectrum, or list of spectra if the input region contained multiple
+        subregions.
+
+    Notes
+    -----
+    This is for now just a proxy for function `extract_region`, to ease the
+    transition from `spectral-cube`.
```
rosteen, line 191: It looks like this line is causing the tests to fail, with the quotes the docs are expecting `spectral-cube` to reference an object defined in this file (I think).
```suggestion
    transition from the spectral-cube package.
```

rosteen (approved):
Looks good to me. I noticed in comparing the output to that of the spectral-cube version that the upper bound is handled slightly differently. With an upper bound of 4000 A on my test cube, spectral-cube's version of spectral_slab generated output with a last spectral axis value of 4000.37, whereas this version has the last spectral axis value as 3999.45. I think this version makes more sense (and is consistent with other behavior in specutils) and I'm approving, but whoever writes the docs for #761 should be sure to note the difference.

javerbukh (approved):
This looks good, although I would suggest a final review from a scientist to make sure this is the expected workflow.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -161,3 +161,36 @@ def extract_region(spectrum, region):
...
+    `~specutils.Spectrum1D`.
+
+    Parameters
+    ----------
+    spectrum: `~specutils.Spectrum1D`
+        The spectrum object from which the region will be extracted.
+
+    s1, s2: spectral coordinates limiting the slab
+        Define the spectral coordinate bound of the region to extract
+        from the original spectrum.
```
nmearl, line 180:
```suggestion
    lower, upper: `~astropy.units.Quantity`
        The lower and upper bounds of the region to extract 
        from the original spectrum.
```

specutils/manipulation/extract_spectral_region.py
```diff
@@ -161,3 +161,36 @@ def extract_region(spectrum, region):
         extracted_spectrum = extracted_spectrum[0]
 
     return extracted_spectrum
+
+
+def spectral_slab(spectrum, s1, s2):
```
nmearl, line 166:
```suggestion
def spectral_slab(spectrum, lower, upper):
```

For consistency with other bounds-related functionality in specutils (e.g. `SpectralRegion`), we should reference the range using similar verbiage.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -161,3 +161,36 @@ def extract_region(spectrum, region):
...
+        Excised spectrum, or list of spectra if the input region contained multiple
+        subregions.
+
+    Notes
+    -----
+    This is for now just a proxy for function `extract_region`, to ease the
+    transition from spectral-cube.
+
+    """
+    region = SpectralRegion(s1, s2)
```
nmearl, line 194:
```suggestion
    region = SpectralRegion(lower, upper)
```

nmearl (changes requested):
Minor comments, but looks good. The SDSS servers are currently down, which is impacting the builds. I'll let this sit for a bit and see if they come back up, but otherwise it *looks* like local tests are passing for me.

PatrickOgle (approved):
Looks good.  I tested spectral_slab on a manga cube, and it returned the correct wavelength subset of the spec1d object. I didn't test edge case of wavelength bin equal to slab start/end wavelengths.

eteq:
LGTM too, so merging!  w.r.t. the SDSS servers, I think we need to update all the examples because the current URL is not very stable.  I thought I wrote an issue on this but can't find it right now - will try to make one later today.

PR #754, opened 2021-01-07
Subclass Spectrum1D from NDCube instead of NDDataRef

rosteen:
Creating this as a draft for now. I have all of the specutils tests passing, but still need to do a deeper dive to make sure that the features we were hoping to take advantage of through this change all work. Additionally, the NDCube developers are working towards a 2.0 release, so this will need to wait on that anyway (I'm working off the latest development version).

### DIFF
--- a/docs/nitpick-exceptions
+++ b/docs/nitpick-exceptions
@@ -32,6 +32,12 @@ py:obj ndcube.NDCube.world_axis_physical_types
 py:obj specutils.Spectrum1D.crop
 py:obj pixel_edges
 py:obj pixel_values
+py:obj ndcube.NDCube.wcs.world_axis_physical_types
+py:obj ndcube.NDCubeSequence
+py:obj sunpy.visualization.animator.ArrayAnimatorWCS
+py:obj ndcube.mixins.plotting.NDCubePlotMixin.plot
+py:obj matplotllib.pyplot.imshow
+py:obj matplotllib.pyplot.plot
 
 # SpectralAxis inherits methods which have warnings we can't fix here.
 # FIXME: https://github.com/astropy/sphinx-automodapi/issues/101
--- a/docs/nitpick-exceptions
+++ b/docs/nitpick-exceptions
@@ -18,6 +18,19 @@ py:obj SpectralCoord
 py:obj SpectralQuantity
 py:obj SkyCoord
 
+# Super class for Spectrum1D for which we don't control the API links.
+py:class ndcube.ndcube.NDCube
+py:class ndcube.ndcube.NDCubeBase
+py:class ndcube.ndcube.NDCubeABC
+py:class ndcube.mixins.plotting.NDCubePlotMixin
+py:class ndcube.mixins.ndslicing.NDCubeSlicingMixin
+py:obj ndcube.NDCube
+py:obj NDCube
+py:obj ExtraCoords
+py:obj GlobalCoords
+py:obj ndcube.NDCube.world_axis_physical_types
+py:obj specutils.Spectrum1D.crop
+
 # SpectralAxis inherits methods which have warnings we can't fix here.
 # FIXME: https://github.com/astropy/sphinx-automodapi/issues/101
 py:obj Quantity
--- a/docs/nitpick-exceptions
+++ b/docs/nitpick-exceptions
@@ -30,6 +30,8 @@ py:obj ExtraCoords
 py:obj GlobalCoords
 py:obj ndcube.NDCube.world_axis_physical_types
 py:obj specutils.Spectrum1D.crop
+py:obj pixel_edges
+py:obj pixel_values
 
 # SpectralAxis inherits methods which have warnings we can't fix here.
 # FIXME: https://github.com/astropy/sphinx-automodapi/issues/101
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -200,7 +200,6 @@ class Spectrum1D(OneDSpectrumMixin, NDCube):
                                  " last. Reshaping arrays to put spectral axis last.")
                     wcs = wcs.swapaxes(0, temp_axes[0])
                     if flux is not None:
-                        print(len(flux.shape), temp_axes[0])
                         flux = np.moveaxis(flux, len(flux.shape)-temp_axes[0]-1, -1)
                     if "mask" in kwargs:
                         if kwargs["mask"] is not None:
@@ -367,7 +366,6 @@ class Spectrum1D(OneDSpectrumMixin, NDCube):
                 else:
                     raise ValueError(f"Unclear how to slice with tuple {item}")
             else:
-                print(f"{type(item)}: {item}")
                 item = slice(item, item + 1, None)
         elif (isinstance(item.start, u.Quantity) or isinstance(item.stop, u.Quantity)):
             return self._spectral_slice(item)
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -182,32 +182,33 @@ class Spectrum1D(OneDSpectrumMixin, NDCube):
 
         # If a WCS is provided, check that the spectral axis is last and reorder
         # the arrays if not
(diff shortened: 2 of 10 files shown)


eteq:
For a bit more backstory for anyone seeing this: the reasoning behind this PR is that NDCube provides a bunch of functionality useful for data cubes that `Spectrum1D` can get essentially "for free" by subclassing from `NDCube` (and appropriate use of the `wcs`).  

Of course nothing is really free, so @rosteen and I had a chat about this out-of-band.  Here's a summary of I think what we settled on: 

* To actually usefully use the"cube" features, the `wcs` has to contain enough information for the APE14 API to understand the spectral and non-spectral axes. Many of the current loaders just produce `spectral_axes` instead of populating the `wcs`, so that should be fixed.  However, we're fairly sure that this doesn't *break* anything that currently works even if the `wcs` is not passed in that way, so that could be a follow-on PR (or PRs), doesn't have to be in this one.
* Relatedly, there's not any reason to think that this PR will break any machinery that currently works if you just pass in a `spectral_axis` and don't create a proper WCS for the other dimensions (and the tests reflect that too, even though they might not be that comprehensive for all the various types of input cubes one can imagine).
* It's probably a good thing to add to the `spectrum1d` initializer that if something is passed in that has `NDCube`-understandable wcs that does *not* follow the `specutils` convention that the last axis is spectral, to automatically put the spectral one to the end.  That's not critical though - more a "quality of life" improvement, so could be a follow-on PR. (Implementation note for that, though: it definitely should reshape uncertainty and mask, not just flux and the spectral axis/wcs!)
* Relatedly: there's a current "silent failure" where if you pass in something where the last axis is not spectral, this machinery will still proceed, and even assigns the last axis to be spectral, but incorrectly (i.e., it doesn't reshape the array) - that will need to be fixed, probably just instead by raising an exception if the specutils convention is known to not be followed for that input
* One more item that could be a follow-on issue or could be in this PR: Currently there's nothing forcing the world axis order (i.e., the output of `pixel_to_world`) to follow the "last dimension is spectral" convention.  It's not clear if this is desirable or not - one could argue that's more consistent with the specutils convention, but then again, the whole point of having a separate set of world and array axes is that they can be decoupled, so maybe we should just respect whatever the WCS says is the right answer.  (I can make a follow-on issue to discuss the "right" answer here if we're not sure, since we can start with just taking whatever the WCS is as the default)

rosteen:
Thanks for writing up the feedback @eteq. I just pushed a commit that addresses points 3 and 4 (inputs where the spectral axis isn't last, but we can figure out which axis to move based on the input WCS). I currently have it throwing a warning to the user to let them know this has happened.

The upshot of this is that we can update loaders (e.g. the MaNGA cube loader) that currently move axes around and construct a spectral axis to instead just input the raw flux and WCS, and the Spectrum1D machinery will handle the reordering now. The additional upshot is that all the WCS information will be kept, allowing things like cutting on spatial coordinates, unlike the case now where that information is lost on Spectrum1D creation.

I think it's reasonable to cover your other points in follow-up issues, so I'll move this to ready for review after I address the error I just introduced.

rosteen:
Ah, I just noticed that one of the test failures is that `ndcube` requires a more recent version of python (3.7+) than specutils currently requires, so the older astropy version tests are failing. If the aim is for `specutils` to continue supporting 3.6 for a while this might be a blocker.

rosteen:
Hm, I expected that updating the tox.ini file to remove the 3.6 envs would stop the CI 3.6 test from running (thus automatically fixing that failure), but it didn't.

nmearl:
You'll need to change/remove the 3.6 github actions workflow in `.github/worksflows/ci_workflow.yml`.

rosteen:
@nmearl Thanks for the tip! I hadn't familiarized myself with the github actions setup yet, looks like I know what I'm doing for an hour tomorrow.

rosteen:
I can't for the life of me get the current docs failures to happen on my local machine, and it's silly to be rerunning the tests up here for every single `nitpick-exceptions` addition. Any idea what the mismatch might be, @eteq @nmearl ?

docs/nitpick-exceptions
```diff
@@ -18,6 +18,21 @@ py:obj SpectralCoord
...
+py:class ndcube.mixins.ndslicing.NDCubeSlicingMixin
+py:obj ndcube.NDCube
+py:obj NDCube
+py:obj ExtraCoords
+py:obj GlobalCoords
+py:obj ndcube.NDCube.world_axis_physical_types
+py:obj specutils.Spectrum1D.crop
+py:obj pixel_edges
+py:obj pixel_values
+
```
nmearl, line 35:
```suggestion
py:obj ndcube.NDCube.wcs.world_axis_physical_types
py:obj ndcube.NDCubeSequence
py:obj sunpy.visualization.animator.ArrayAnimatorWCS
py:obj ndcube.mixins.plotting.NDCubePlotMixin.plot
py:obj matplotllib.pyplot.imshow
py:obj matplotllib.pyplot.plot

```

nmearl (changes requested):

nmearl:
NDcube seems have a lot of typos and/or broken api references. Just expanding the nitpicks seems to fix it until it can be addressed upstream.

To test locally, it should be enough to simply run `tox -e build_docs`.

rosteen:
The problem was that I was running that tox command locally, but despite my attempts to set up a clean environment in the ways indicated in the test log, I was still seeing `tox -e build_docs` pass happily on my machine when they were hitting errors up in cloud land. I didn't want to keep pushing up one addition to the nitpick file at a time - thanks for adding what I was missing.

specutils/spectra/spectrum1d.py
```diff
@@ -70,18 +72,31 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
...
             super().__init__(flux)
             return
 
+        # Handle initializing from NDCube objects
+        elif isinstance(flux, NDCube):
+            if flux.unit is None:
+                raise ValueError("Input NDCube missing unit parameter")
+
+            # Change the flux array from bare ndarray to a Quantity
+            q_flux = flux.data*u.Unit(flux.unit)
```
eteq, line 94: I suggest instead using this:
```suggestion
            q_flux = flux.data << u.Unit(flux.unit)
```
that's both the "recommended" way to do this now, and here it has the particular advantage of not copying the data (an importand consideration since cubes can be big). Alternatively, could do
```suggestion
            q_flux = u.Quantity(flux.data, unit=flux.unit, copy=copy)
```
and expose `copy` as an initializer keyword.

specutils/spectra/spectrum1d.py
```diff
@@ -39,7 +40,8 @@ class Spectrum1D(OneDSpectrumMixin, NDDataRef):
         dimension of flux, or one greater than the last dimension of flux
         if specifying bin edges.
     wcs : `~astropy.wcs.WCS` or `~gwcs.wcs.WCS`
-        WCS information object.
+        WCS information object that either has a spectral component or is
```
eteq, line 43: Just above here in `flux` I think you should specifically mention that `flux` can be an `NDCube` (and while you're at it, might as well mention it can be a `Spectrum1D`)

eteq (changes requested):
Looks great, @rosteen !  Just a couple inline items.  Additionally, one minor thing to be added: I think there should be a bit of a mention in the narrative docs that ndcube is involved - the two places I can think of are in the "overview of how specutils represents spectra" you can point out the in the "case 2" here, `NDCube` is a superclass.  Also worth at least mentioning briefly in the  "Working with Spectral Cubes" section.  We should fill that out a bit more fully, but that can be a follow-on PR.  For this one just a sentence along the lines of "To see more possible cube operations, see the NDCube docs" or similar.

eteq (approved):

eteq:
LGTM now, CI is passing (except for the allowed fail) and I think @nmearl's comments were all addressed, so I think it's ok to merge

nmearl:
Just to note for when we get around to addressing the issue that something in this PR is causing the allowed fail failure, since previous testing was passing.

Issue #746, opened 2020-12-09
Create method to extract "minimal bounding" region


rosteen:
`spectral-cube` has a method to extract the minimal valid sub-cube, see the docs [here](https://spectral-cube.readthedocs.io/en/latest/manipulating.html#extract-the-minimal-valid-subcube). If I recall correctly one can currently extract multiple regions from a `Spectrum1D` object with `extract_spectral_region`, but that cuts out any data in between the regions. It would be good to have a method to extract the minimum bounding cube (or lower dimensional equivalent) of any provided regions.

eteq:
I'm confused by this @rosteen - isn't the `spectral-cube` version of this about the *spatial* region?  I.e., drawing like a circular region in the *spatial* part of the cube and then extracting the smallest cube that contains that spatial region?

To be fair, this suggestion is still useful, but I'm not sure if it's the same as the `spectral-cube` method?

rosteen:
I think the `spectral-cube` version is agnostic about spectral vs spatial dimensions - it extracts the minimum non-masked cube by trimming in whatever dimensions are needed. I was far more ambiguous than I should have been in my description here, but the intent was to duplicate the `spectral-cube` functionality. @ibusko  created #774 to explicitly lay out the requirement to handle the spatial dimensions.

PR #755, opened 2021-01-11
Minimal bounding region in spectral extraction

ibusko:
As per #746 

This PR must be reviewed  and merged only **AFTER** PR #753 is merged.

This PR contains changes that generate conflicts with PR #753, thus it was developed on top of the 753 branch, and not on top of master. Thus the need for the ordered merging  (the relevant changes corresponding only to this PR are in the 2nd and later commits).

The CI doc build error seems to be coming from a different module (docs/fitting.rst).

### DIFF
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -7,7 +7,7 @@ import numpy as np
 from astropy import units as u
 from ..spectra import Spectrum1D, SpectralRegion
 
-__all__ = ['extract_region', 'extract_bounding_region', 'spectral_slab']
+__all__ = ['extract_region', 'extract_bounding_spectral_region', 'spectral_slab']
 
 
 def _to_edge_pixel(subregion, spectrum):
@@ -198,7 +198,7 @@ def spectral_slab(spectrum, lower, upper):
     return extract_region(spectrum, region)
 
 
-def extract_bounding_region(spectrum, region):
+def extract_bounding_spectral_region(spectrum, region):
     """
     Extract the entire bounding region that encompasses all sub-regions
     contained in a multi-sub-region instance of `~specutils.SpectralRegion`.
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -1,3 +1,5 @@
+import sys
+
 from math import floor, ceil  # faster than int(np.floor/ceil(float))
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


rosteen (approved):
Looks good to me (other than the minor conflicts to be resolved with the spectral_slab PR).

javerbukh (approved):
Looks good!

eteq:
#753 is in so I think this can be rebased, @ibusko ?

eteq:
(I rebased it myself, because I realized I can just merge it as it stands if I do that and the tests pass)

eteq:
Note: I'm realizing I'm not sure this is quite what #746 is about - see https://github.com/astropy/specutils/issues/746#issuecomment-770087237 there.  If @rosteen and/or @ibusko agree with my comment there, then I think we want to change the name of this PR's function to `extract_bounding_spectral_region`.  Alternatively we could leave it as-is and plan on adding on the "spatial" part of this function later, as it would then let you extract the minimal bounding region in *any* of the axes.  But if we go that route we should definitely make an issue recording it before merging this.

ibusko:
@eteq yes, the `minimal_subcube` function applies to both spatial and spectral regions. I haven't realized this before, I think I was driven by the description in #746 , which seems to apply to the spectral part only. Besides, the concept as applied to the spatial part only makes sense when masks are supported. In the spectral domain, we have `SpectralRegion` 's that may be thought as a kind of "mask". So, perhaps your first suggestion (change the function name) is the way to go. For now.

ibusko:
I renamed the function, and created issue #774 to record the need to implement the spatial handling.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -194,3 +196,64 @@ def spectral_slab(spectrum, lower, upper):
...
+
+    region: `~specutils.SpectralRegion`
+        The spectral region to extract from the original spectrum, comprised
+        of one or more sub-regions.
+
+    Returns
+    -------
+    spectrum: `~specutils.Spectrum1D`
+        Excised spectrum from the bounding region defined by the set of
+        sub-regions in the input `region` instance.
```
eteq, line 222: This should fix the docs failure in the PR
```suggestion
        sub-regions in the input ``region`` instance.
```

eteq:
In https://github.com/astropy/specutils/pull/755#pullrequestreview-580496961 I fixed what I think was the reason for the test failures.  Will merge this if the tests pass

eteq (approved):

PR #757, opened 2021-01-13
Add support for astropy registry priorities

### LINKED ISSUE #171: Data loaders need prioritization to resolve multiple matches
When trying to associate an identifier with my particular type of FITS file, the loader cannot decide between the generic FITS loader and my STIS-specific FITS loader.

```
IORegistryError: Format is ambiguous - options are: cubetest1, stis-format, tabular-fits
```
[Location of error in AstroPy source code](https://github.com/astropy/astropy/blob/9626265d77b8a21c113615c08bc6782deb52eaed/astropy/io/registry.py)

My identifier:
```python
# Define an optional identifier. If made specific enough, this circumvents the
# need to add `format="my-format"` in the `Spectrum1D.read` call.
def identify_stis_fits(origin, *args, **kwargs):
    from astropy.io import fits
    import os
    import six
    
    if isinstance(args[0], six.string_types) and (os.path.splitext(args[0].lower())[1] == '.fits'):
        with fits.open( args[0]) as f:
            return f[0].header.get('INSTRUME', '') == 'STIS'
    else:
        return False
```

Glue gets around this by allowing the specification of [loader priorities](http://docs.glueviz.org/en/stable/customizing_guide/customization.html#custom-data-loaders).

aragilar:
This moves from the old priority system, which does not work, to the proposed system in https://github.com/astropy/astropy/pull/11214. This would close #171 and #584.

This lacks direct tests, as monkeypatching the registry will probably hide problems, but I'm open to other ideas for explicitly testing for this.

### DIFF
--- a/specutils/io/registers.py
+++ b/specutils/io/registers.py
@@ -126,11 +126,9 @@ def data_loader(label, identifier=None, dtype=Spectrum1D, extensions=None,
 def custom_writer(label, dtype=Spectrum1D, priority=0):
     def decorator(func):
         if _astropy_has_priorities():
-            io_registry.register_writer(
-                label, Spectrum1D, func, priority=priority,
-            )
+            io_registry.register_writer(label, dtype, func, priority=priority)
         else:
-            io_registry.register_writer(label, Spectrum1D, func)
+            io_registry.register_writer(label, dtype, func)
 
         @wraps(func)
         def wrapper(*args, **kwargs):
--- a/specutils/io/registers.py
+++ b/specutils/io/registers.py
@@ -1,6 +1,7 @@
 """
 A module containing the mechanics of the specutils io registry.
 """
+import inspect
 import os
 import pathlib
 import sys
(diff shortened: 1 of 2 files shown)


specutils/io/registers.py
```diff
@@ -51,7 +62,10 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     def decorator(func):
-        io_registry.register_reader(label, dtype, func)
+        if _astropy_has_priorities():
```
nmearl, line 65: Is this check really necessary? If a user-defined reader/writer has no priority, seems we can just default to the lowest priority instead of resorting to reflection?

aragilar: Older versions of `io_registry.register_reader` won't accept the additional argument, so if we don't do this check, we'll get exceptions when trying to add a priority. Once all supported versions of astropy accept priorities, we can drop the check.

nmearl:
@aragilar it'd be great to also have a simple loader priority test.

aragilar:
I've added a test for the priorities, which should would for both astropy versions with and without priority support. Not sure if you want to wait until 4.3 comes out before merging this or not.

aragilar:
I've rebased this on top of main so that the fixes in https://github.com/astropy/specutils/pull/833 are included (so everything passes). Now that 4.3 is out, did we want to have another look at this?

nmearl (approved):
This looks great. Thanks @aragilar!

Issue #745, opened 2020-12-09
Implement moment method in spectrum data classes


rosteen:
One can calculate the n-th moment of a cube with a simple call of `cube.moment(order=n)` in `spectral-cube` (see https://spectral-cube.readthedocs.io/en/latest/moments.html). We should add a method like this to appropriate spectrum objects as part of the `spectral-cube` transition to `specutils`.

ibusko:
@rosteen do you envisage this as a method in the spectrum class (as is with spectral cube), or as a function similar to existing analysis and measurements functions in specutils? It seems to me, based on the current Spectrum1D API, that in specutils we moved away from having such kind of measurement functionality defined as  attributes of the spectrum object, instead having them being handled by functions. I ask because your wording above seems to imply otherwise. And there might be reasons for this that I am not aware of.

(this sort of touches on our discussion about the redshift slider)

rosteen:
I was actually about to pull in @eteq for his thoughts on exactly this question on the `spectral_slab` PR, since I realized that in spectral cube that's attached to the cube object itself, as opposed to specutils having separate analysis functions (as you said Ivo). Following the rest of specutils by having it separate does conflict with "maximum back compatibility" with spectral cube, but I think that I'd personally default to the specutils convention of having it in the analysis functions rather than attached to the Spectrum1D. Thoughts @eteq ?

nmearl:
I certainly think the precedent we've established for specutils should be maintained. That said, it might be worth using a proxy method with a deprecation decorator and plan to remove it in later releases:

```python
from ..manipulation import spectral_slab
from astropy.utils.decorators import deprecated


class Spectrum1D:
    ...
    @deprecated('0.2', alternative=':func:`~specutils.manipulation.spectral_slab.spectral_slab`', pending=False)
    def spectral_slab(self, *args, **kwargs):
        return spectral_slab(*args, **kwargs)
```

ibusko:
@nmearl in principle it should work, but it leads to a circular import. `spectral_slab` imports Spectrum1D. This might require some revamping of the file and import structure in `specutils`. Unless there is a solution I'm not seeing right now.

eteq:
Done in #758

PR #758, opened 2021-01-13
Implement moment function

ibusko:
As per #745 

This is for now being implemented as a function, not as a method in the `Spectrum1D` class. The rationale is that, in `specutils`, measurement and processing operations are implemented as functions in top-level packages such as `analysis`, `fitting`, and `manipulation`, and not as methods in `Spectrum1D` or `OneDSpectrumMixin`. I am not sure we should break this pattern now.

The `moment` method in `spectral_cube` accepts a parameter (how: cube | slice | ray | auto) that is meant to select specialized cube addressing patterns that optimize the calculation for certain cube and memory configurations. We leave this out of this PR for now (it's using the "cube" mode).

### DIFF
--- a/specutils/analysis/moment.py
+++ b/specutils/analysis/moment.py
@@ -4,12 +4,9 @@ spectral features.
 """
 
 import numpy as np
-from astropy.stats.funcs import gaussian_sigma_to_fwhm
-from astropy.modeling.models import Gaussian1D
 from ..manipulation import extract_region
 from . import centroid
 from .utils import computation_wrapper
-from scipy.signal import chirp, find_peaks, peak_widths
 
 
 __all__ = ['moment']
@@ -35,7 +32,7 @@ def moment(spectrum, regions=None, order=0):
     Returns
     -------
     moment: `float` or list (based on region input)
-        Moment of the spectrum
+        Moment of the spectrum. Returns None if (order < 0 or None)
 
     """
     return computation_wrapper(_compute_moment, spectrum, regions, order=order)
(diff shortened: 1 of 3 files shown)


rosteen (approved):
This looks good to me. The tests are passing for me locally, but it seems like there have been some connection errors in the tests across a few PRs in the Github actions tests. I still need to familiarize myself with the new machinery, not sure if it's just a  matter of kicking them off again or fixing something in the test infrastructure.

ibusko:
I’ve seen all these “canceled” test runs in several PRs, not just mine. It’s like something is timing out.

Sent from my iPhone

On Jan 15, 2021, at 11:48, rosteen <notifications@github.com> wrote:

﻿

@rosteen approved this pull request.

This looks good to me. The tests are passing for me locally, but it seems like there have been some connection errors in the tests across a few PRs in the Github actions tests. I still need to familiarize myself with the new machinery, not sure if it's just a matter of kicking them off again or fixing something.

nmearl:
Are there plans to include some documentation either with this, or in a follow up PR? If not, I'd also suggest opening an issue so we can be sure to address it eventually (preferably before the next release).

rosteen:
I'll open an issue to create a "so you want to do SpectralCube things" section to the docs addressing this and the other related PRs.

PatrickOgle (approved):
Tested moment order=0 and order=1 on manga cube and got sensible results.  Error regarding masks is fixed.  Ready to merge.

specutils/analysis/moment.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def moment(spectrum, regions=None, order=0, axis=-1):
+    """
+    Estimate the moment of the spectrum.
+
+
+    Parameters
+    ----------
+    spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
```
eteq, line 21: Will this also work on `SpectrumCollection`? *in principle* it should, and maybe we should add a test if it does (and if not make an issue to fix it later?)

eteq (changes requested):
One quick comment that should be an easy either "make a test" or "make an issue" solution, otherwise LGTM!

specutils/analysis/moment.py
ibusko: I verified now, it does not work with `SpectrumCollection`. I created issue #772 .

eteq: Ok, looks good - we can deal with it later then.  Will merge!

eteq (approved):

Issue #750, opened 2020-12-18
SpectralRegion() gives wrong error message when given tuples that are not Quantities


janerigby:
Running this code gives the wrong error message:
`
temp = [(2797.65, 2817.65), (8678.011033, 8698.011033), (9005.228311, 9025.228311)]
SpectralRegion((temp))
`
The error message it gives is "ValueError: SpectralRegion input must be a 2-tuple or a list of 2-tuples."

The error message was just plain wrong. The problem is NOT that "SpectralRegion input must be a 2-tuple or a list of 2-tuples", but that the input was a list of 2-tuples that weren't Quantities.

Issue #749, opened 2020-12-18
Output of fit_lines_derivative() is the wrong format to feed to the exclude_regions() paramter of fit_continuum_generic()


janerigby:
The specutils function fit_continuum_generic() will be fooled by strong emission lines.  A common use case is to mask those emission lines before fitting the continuum.  However, the output of find_lines_derivative() is the wrong format to be readable by the exclude_regions() parameter of fit_continuum_generic(), and there appears to be no conversion tool. 

A real-world example of this issue:  https://github.com/janerigby/NB/blob/master/spectutils_testdrive.ipynb

Here is the conversion function I made to get around it.  Spectools needs the functionality of converting its output linelists to SpectralRegions -- it's going to be a common concern.

```
def convert_lines_to_spectralregions(linelist, window=10., unit=u.AA) :
    temp =  [((x - window) * u.AA, (x + window)* u.AA)  for x in linelist['line_center'].value] # must be Quantities
    return (SpectralRegion(temp))
```

jdavies-st:
This issue was resolved by #759 and can be closed.  See

https://specutils.readthedocs.io/en/stable/spectral_regions.html#line-list-compatibility

and 

https://specutils.readthedocs.io/en/stable/api/specutils.SpectralRegion.html#specutils.SpectralRegion.from_line_list

Issue #742, opened 2020-12-08
fit_generic_continuum() documentation does not explain argument median_window


janerigby:
I'm trying to get fit_generic_continuum() to fit a continuum. The fit is terrible, and I'm wondering if the problem is the median_window argument. But, the documentation does not explain what that argument is, what its units are (Angstroms?  pixels), or what changing it is likely to do.

nmearl:
Thanks for the report. `median_smooth` is used in the median filter smoothing function which is performed on the spectrum prior to attempting to fit a continuum. Internally, it uses [scipy's `medfilt` function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.medfilt.html). The `median_window` is the same as the `kernal_size` used in the median filter function.

PR #759, opened 2021-01-17
Various improvements

nmearl:
- Addresses #750 by giving better error messages to the `SpectralRegion` objects.
- Addresses #749 by providing a new method `SpectralRegion.from_line_list` that allows passing in the result of the `fit_lines_*` functions to produce a `SpectralRegion` object from the line list table. Added example to docs.
- Address #742 by expanding continuum function docstrings and fixing some of the failing and/or incorrect plots generated in the documentation.
- Address issues with 0.16.0 gwcs release.
- Address issue with jwst loader test which previously used the parsing utilities to automatically find flux columns but now simply accesses them directly (marked test as xfail for now).
- Fix various issues with test infrastructure; tests should now pass consistently.


.github/workflows/label_pr.yml
```diff
@@ -1,13 +0,0 @@
...
-    types:
-    - opened
-
-jobs:
-  triage:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/labeler@main
-      with:
-        repo-token: "${{ secrets.GITHUB_TOKEN }}"
```
eteq, line 13: I'm not necessarily opposed to removing this, but is the problem just that the token isn't updated?  If so maybe we should fix that instead?  Or are you actively opposed to using the labeler at all?

docs/fitting.rst
```diff
@@ -635,47 +632,31 @@ convenience functions to perform exactly this task.  An example is shown below.
...
-
-    plt.plot(x, y)
-    plt.plot(x, y_continuum_fitted)
-    plt.title('Continuum Fitting')
-    plt.grid(True)
-
-The normalized spectrum is simply the old spectrum devided by the
-fitted continuum, which returns a new object:
-
-.. plot::
```
eteq, line 669: Why was this removed?  I think this is here specifically to demonstrate that you can do continuum normalization.

docs/spectral_regions.rst
```diff
@@ -209,6 +209,53 @@ An example of a multiple sub-region `~specutils.SpectralRegion`:
     >>> sub_spectra[1].spectral_axis
     <SpectralAxis [34., 35., 36., 37., 38., 39., 40.] nm>
 
+
+Line List Compatibility
```
eteq, line 213: Neat!  A good change, but maybe it should be called out in the changelog as a new feature since it's otherwise buried in this PR with several other things?

docs/spectral_regions.rst
```diff
@@ -209,6 +209,53 @@ An example of a multiple sub-region `~specutils.SpectralRegion`:
...
+-----------------------
+
+`~specutils.SpectralRegion` objects can also be created from the `~astropy.table.QTable` object returned from the line
+finding routines
```
eteq, line 217:
```suggestion
finding functions:
```
(quibble)

specutils/fitting/continuum.py
```diff
@@ -21,14 +21,18 @@ def fit_generic_continuum(spectrum, median_window=3, model=Chebyshev1D(3),
...
     spectrum : Spectrum1D
         The spectrum object overwhich the equivalent width will be calculated.
-
     model : list of `~astropy.modeling.Model`
         The list of models that contain the initial guess.
-
+    median_window : float
+        The width of the median smoothing kernel used to filter the data before
+        fitting the continuum. See `~scipy.signal.medfilt` for more
+        information.
```
eteq, line 29:
```suggestion
        fitting the continuum. See the ``kernel_size`` parameter of 
        `~scipy.signal.medfilt` for more information.
```

specutils/fitting/fitmodels.py
```diff
@@ -128,7 +128,9 @@ def find_lines_threshold(spectrum, noise_factor=1):
 
     # Threshold based on noise estimate and factor.
     uncertainty = spectrum.uncertainty
-    inds = np.where(np.abs(spectrum.flux) > (noise_factor*uncertainty.array) *
+    uncert_val = uncertainty.array if uncertainty is not None else 0
```
eteq, line 131:
```suggestion
    uncert_val = 0 if uncertainty is None else uncertainty.array 
```
(very minor style quibble)

eteq (changes requested):
Mostly looks good, just a few questions/comments.

I also have some thoughts about improving the docstring and parsing for `SpectralRegion`, but after working on it a bit I'm realizing it's probably best as a follow-on PR

.github/workflows/label_pr.yml
nmearl: We currently don't have any actions for the labeler right now -- no configuration file that specifies what changed to look for a what labels to apply, so I'm sure what the point of having the workflow is right now.

docs/fitting.rst
nmearl: Added back -- not sure how I missed it.

docs/spectral_regions.rst
nmearl: Yeah, there are several that I have listed that will go into the change log, but it's been happening usually as a PR before the release.

eteq: I didn't realize that's when you were populating it - we should definitely move away from that - it's dangerous to rely on memory, etc to update the changelog especially with more people potentially doing releases in the future.

But so as not to delay this PR any further I'll do this myself right now!

eteq:
Had to resolve a conflict, while adding a changelog entry, but otherwise looks good!

eteq (approved):

Issue #760, opened 2021-01-20
Conserve shape of array when fitting a function to a sub region of the spectrum


ibusko:
This came as a Developer Note in notebook https://spacetelescope.github.io/jdat_notebooks/jdat_notebooks/MRS_Mstar_analysis/JWST_Mstar_dataAnalysis_usecase.html  (cell 26)

The science case reported in the note is very specific: the user wants to approximate a region of the spectrum with a spline. So, even though the JIRA ticket (filed before the present github issue was filed) contains the term "fitting", I am assuming that the request does not apply to any kind of generic fitting, but instead applies to this narrower science case only. It seems to me that the requested operation would not make sense in cases where a physical spectrum model is being fitted to the data, since there seems to be no point in extending the fitted model with the original data, to the blue and/or red sides of the fitted model function.

eteq:
(note there's some discussion in #762 about the best solution to this problem)

PatrickOgle:
The title of this issue does not accurately describe the science case.  The corresponding PR (#762) should also be deprecated since it does something nonsensical from a science standpoint (resample a segment of the spectrum to a different wavelength grid while conserving the number of samples in that segment) .

I think what we want to do instead is this:
Create a spline-fitting function that replaces a segment of the input spectrum with the spline fit of that segment, maintaining the original wavelength sampling.

PR #762, opened 2021-01-21
Conserve shape of arrays when interpolating

ibusko:
See #760

### DIFF
--- a/specutils/manipulation/resample.py
+++ b/specutils/manipulation/resample.py
@@ -25,26 +25,73 @@ class ResamplerBase(ABC):
     extrapolation_treatment : str
         What to do when resampling off the edge of the spectrum.  Can be
         ``'nan_fill'`` to have points beyond the edges by set to NaN, or
-        ``'zero_fill'`` to be set to zero.
+        ``'zero_fill'`` to be set to zero
+
     """
     def __init__(self, extrapolation_treatment='nan_fill'):
         if extrapolation_treatment not in ('nan_fill', 'zero_fill'):
             raise ValueError('invalid extrapolation_treatment value: ' + str(extrapolation_treatment))
         self.extrapolation_treatment = extrapolation_treatment
 
-    def __call__(self, orig_spectrum, fin_spec_axis):
+    def __call__(self, orig_spectrum, fin_spec_axis, keep_shape=False):
         """
         Return the resulting `~specutils.Spectrum1D` of the resampling.
         """
-        return self.resample1d(orig_spectrum, fin_spec_axis)
+        return self.resample1d(orig_spectrum, fin_spec_axis, keep_shape=keep_shape)
 
     @abstractmethod
-    def resample1d(self, orig_spectrum, fin_spec_axis):
+    def resample1d(self, orig_spectrum, fin_spec_axis, keep_shape=False):
         """
         Workhorse method that will return the resampled Spectrum1D
         object.
         """
         return NotImplemented
+
+    def keep_shape(self, orig_spectrum, resample_spec_axis, resampled_flux_arr,
+                   resampled_unc):
+        """
+        Returns the resampled data as a patch on the original spectrum. This
+        is used when one needs the resampled data to be of same shape as the
+        input spectrum.
+        """
+        orig_spec_axis = orig_spectrum.spectral_axis.to(resample_spec_axis.unit)
+
+        indices_min = np.where(orig_spec_axis.value < resample_spec_axis[0].value)
+        indices_max = np.where(orig_spec_axis.value > resample_spec_axis[-1].value)
+
+        extended_flux = np.zeros(shape=orig_spectrum.flux.shape)
+
+        extended_flux[indices_min] = orig_spectrum.flux[indices_min]
+        extended_flux[indices_max] = orig_spectrum.flux[indices_max]
+
+        i1 = indices_min[0][-1] + 1
+        i2 = i1 + len(resampled_flux_arr)
+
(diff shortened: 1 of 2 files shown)


javerbukh (approved):
This looks good and passes all tests locally, although this is another situation where I think a science review is appropriate.

rosteen (commented):
Shouldn't this also update the spectral_axis values in the output spectrum? Currently, using the linear interpolation case from the tests (interpolating to spectral axis locations [3.5, 4.7, 6.8, 7.1] * u.AA), the interpolation changes the spectrum from:

`<Spectrum1D(flux=<Quantity [ 1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.] mJy>, spectral_axis=<SpectralAxis [ 1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.] Angstrom>, uncertainty=StdDevUncertainty([0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1. ]))>`

to 

`<Spectrum1D(flux=<Quantity [ 1. ,  2. ,  3. ,  3.5,  4.7,  6.8,  7.1,  8. ,  9. , 10. ] mJy>, spectral_axis=<SpectralAxis [ 1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.] Angstrom>, uncertainty=StdDevUncertainty([0.1 , 0.2 , 0.3 , 0.35, 0.47, 0.68, 0.71, 0.8 , 0.9 ,
                   1.  ]))>`

As you can see, the output spectrum now contains the flux values appropriate for the desired wavelength values, but the spectral axis still assigns those flux values to the original locations on the spectral axis.

ibusko:
@rosteen you're right. The spectral axis is not being updated. I will push a commit shortly to fix this.

rosteen (approved):
Looks good to me now after that fix, I don't see anything else that stands out.

PatrickOgle:
I don't completely understand the science use case, even after reading the jdat notebook.  Is the idea to:
1) replace a segment of a spectrum with a spline fit to that segment?
Or to
2) Do (1) and set the spectrum to zero outside of that segment.

The name of the PR "Conserve shape of arrays"  doesn't really get at the intent here.

ibusko:
I understood from the notebook case that the goal is to replace a segment of the spectrum with a fit, and keep the rest of the spectrum as is.

Sent from my iPhone

On Jan 29, 2021, at 11:08, Patrick Ogle <notifications@github.com> wrote:

﻿

I don't completely understand the science use case, even after reading the jdat notebook. Is the idea to:

  1.  replace a segment of a spectrum with a spline fit to that segment?
Or to
  2.  Do (1) and set the spectrum to zero outside of that segment.

The name of the PR "Conserve shape of arrays" doesn't really get at the intent here.

eteq:
I'm fine with this PR as it stands per se, *but* I'm also with @PatrickOgle that I don't understand which developer note this is addressing from #760. Can you clarify which note it is @ibusko ? (even just copy the text here is fine)

ibusko:
This is the developer note: 

"Developer note By fitting a spline to a sub region the spectral shapes are no longer the same. bbsub_spectra.flux.value - spline_spec.flux.value now brakes. Would need to trim the spectrum to the spline size to start looking closely for low contrast dust features and again measure their properties (see above). Some wrapper to stop repeating the same steps over and over would be nice"

eteq:
Hmm. I admit this is a bit ambiguous: I read this as asking for a wrapper that does the *other half* of this: i.e., a desire to trim the spectrum.

eteq:
Oh wait!  Right after sending that last comment I realized that you'd still need what's in this PR to do that.  Am I interpreting that right, @ibusko, that this PR is implementing "step 1", and then we might consider a wrapper that does this *in addition* to trimming to satisfy the developer note as posed?

(If so, then I think this is all set to go in if @PatrickOgle agrees, just doesn't completely address the developer note)

PatrickOgle:
It would make more sense to me to trim the input spectrum to the size of the spline-fitting region then subtract
the spline. However, I don't object to having this functionality as a 'keep shape' option to resample.py.

ibusko:
To conserve the shape of the spline (or other interpolator), we have to do nothing, because this is what the interpolators do that by default.

ibusko:
Yes, it sounds ambiguous. I had interpreted that the function should return only one spectrum instance, as the interpolators currently do. The reverse of what I implemented means that it has to return both the output of the interpolator, as a well as a trimmed version of the input spectrum. Is that what is meant by a 'wrapper'?

rosteen:
For the record, I interpreted the developer note the same way as Ivo: the user was annoyed that they couldn't simply do `bbsub_spectra.flux.value - spline_spec.flux.value` after creating spline_spec, since it was a different size than the original spectrum, and didn't want to have to go through the step of pulling a chunk out of the original spectrum to subtract the spline. However, I agree that it's ambiguous how they actually wanted the problem solved. 

It doesn't seem like it would be too hard (in a follow up) to allow subtracting a smaller spectrum from a larger one (leaving the mismatched regions unaffected) if we end up wanting to go that direction.

PatrickOgle:
I don't think this _keep_shape functionality should be in resample.py.  If we resample the spectrum but keep the same number of wavelength points, the wavelengths will not match up with the original spectrum.   

Also, I don't think this answers the science question of replacing a segment of the spectrum with a spline fit, and that should go somewhere else, not in resample.py. 

I suggest deprecating this PR.

PR #764, opened 2021-01-25
Add loader for OzDES

aragilar:
This adds a loader for [OzDES](http://www.mso.anu.edu.au/ozdes/). I've set the priority as 10, as it should be higher than the normal 1d fits loader.


dhomeier:
Are priorities working now?

nmearl:
Priorities aren't working yet as it won't be supported until astropy 4.3. I'd like to wait for that release to merge in #757 so the testing infrastructure will work appropriately. However, this should be fine to merge for now.

PR #765, opened 2021-01-25
Add loaders for the GAMA survey

aragilar:
Adds loaders for the GAMA survey. I haven't set priorities on these loaders, we probably need to discuss what they should be (as due to GAMA using data from other surveys, they likely need a higher priority than the original survey).


nmearl:
Thanks, @aragilar. As mentioned in the other loader PR, a follow up PR with adding priorities should be fine.

PR #766, opened 2021-01-25
Add GALAH loader

aragilar:
Adds a loader for the GALAH survey. Similar to #764, I've set the priority to 10 (this can be changed, but seems reasonable).


nmearl:
Thanks @aragilar!

PR #767, opened 2021-01-25
Add WiggleZ loader

aragilar:
Adds loader for WiggleZ. This shouldn't conflict with #765 as the identifiers should spot the difference, but a priority will need to be set such that the 1d-fits loader isn't used.


nmearl:
Thanks @aragilar! A follow-up PR with a priority implementation should be fine once that machinery is in place.

PR #770, opened 2021-01-27
Update Spectral Region docstrings and initializer

eteq:
This PR follows from #759 (and is currently based on @nmearl's) branch so I'm issuing it as a draft until that's merged and I can rebase it on master.

To see the diffs for just *this* PR and not #759, see https://github.com/eteq/specutils/compare/8b85d0ed783b2af726d1aff16fcf12143402741a..18c8c6958dad2893aa2338cc11c9b9fbfe1218c5

Mainly it cleans up the documentation and error messages to be (IMHO) much less confusing. #759 fixes the proximate issues #750 raises, but I think the underying issue of #750 is that the initializer is written in such a way that it was hard to generate case-relevant error messages, and the docstring was rather misleading (if not outright wrong).  This PR attempts to address that. (And in that direction, if @janerigby has some bandwidth to review this or at least say if this would also help with the ambiguity that #750 points out, even better!)

### DIFF
--- a/specutils/spectra/spectral_region.py
+++ b/specutils/spectra/spectral_region.py
@@ -7,12 +7,12 @@ import astropy.units as u
 
 class SpectralRegion:
     """
-    A `SpectralRegion` is a container class that enables some simplicty
-    in defining and passing a region (interval) for a spectrum.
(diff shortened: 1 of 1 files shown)


nmearl (approved):
Looks good!

Issue #773, opened 2021-01-29
Find an example spectrum that's a bit more reliable for the docs


eteq:
The SDSS spectrum at https://dr14.sdss.org/optical/spectrum/view/data/format=fits/spec=lite?plateid=1323&mjd=52797&fiberid=12 is used in several places in the specutils docs.  However, several times recently the server has been having issues and it messes up the doc builds when this happens.

I'd prefer if at all possible to avoid actually saving this somewhere in `specutils` or the `astropy` ecosystem, because we *want* to encourage users to use the original sources for archived data whenever possible. I asked @havok2063 about this and he suggested we use a URL that's on SAS (i.e. somewhere in e.g. https://data.sdss.org/sas/dr16/), as that may be a bit more reliable than the optical/spectrum may of accessing URLs.

To clarify: there's nothing special about that *particular* galaxy - I just went looking for an emission line galaxy because it's used to illustrate line-fitting.  But just about any starforming galaxy with a decent SDSS will do for this, and there are quite literally millions of them... I think part of the issue with this one is that it's from very early in the SDSS so it's not been processed by some of the later mechanisms - a galaxy from BOSS (a later stage of SDSS's spectroscopy) might be better because those seem to be a lot easier to find in SAS (the more reliable server).

cc @nmearl, who noticed it independently of me in #753

havok2063:
Here is your original spectrum https://dr16.sdss.org/optical/spectrum/view?plateid=1323&mjd=52797&fiberid=12 and its location on the SAS https://data.sdss.org/sas/dr16/sdss/spectro/redux/26/spectra/1323/.   I don't know if you need something newer vs something more stable.  One option is to keep the object you have now but simply change the url you use to grab the spectrum.  The SAS url can be more stable than the API url, oddly.  There's no real need to change what spectrum is used since SDSS will always keep the old observations around.

eteq:
Thanks @havok2063 ! I didn't realize this was still on the SAS - was thrown off by the "26" there!

PR #775, opened 2021-02-02
Fixes problem with non-loadable x1d files

ibusko:
Fixes the bug reported in jdaviz's issue #390. 

The bug was caused by the SRCTYPE keyword being moved around different extensions by different pipeline versions. Or, just not being there. 

Data created by pipeline 0.17.1 mostly works with the fix in here. That is, as long as the keyword is somewhere in the file, it works. Other pipeline versions do not create this keyword (they should, and they will from release B7.7.1, around beginning of March.

Link for 0.17.1 files https://stsci.box.com/s/5rxc5b41p04k8pux91iyn7aij5zhmm1l

### DIFF
--- a/specutils/io/default_loaders/jwst_reader.py
+++ b/specutils/io/default_loaders/jwst_reader.py
@@ -150,23 +150,12 @@ def _jwst_x1d_loader(file_obj, **kwargs):
             wavelength = Quantity(data["WAVELENGTH"])
 
             # Determine if FLUX or SURF_BRIGHT column should be returned
-            # based on whether it is point or extended source
+            # based on whether it is point or extended source.
+            #
+            # SRCTYPE used to be in primary header, but was moved around. As
+            # per most recent pipeline definition, it will be either in the
+            # EXTRACT1D or the SCI extension.
             srctype = hdu.header.get("srctype")
-
-            # SRCTYPE is in primary header because there is only one spectrum
-            if srctype is None:
-                # srctype = hdulist["PRIMARY"].header.get("srctype")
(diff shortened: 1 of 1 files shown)


ibusko:
I am adding Brian as an unofficial reviewer.

havok2063:
@ibusko I've tested that this works with 0.17.1 nirspec x1d files that have the correct SRCTYPE keyword in the EXTRACT1D header.  I've also tested the existing 0.17.1 miri mrs x1d files fail due to the missing keyword, but work when I manually add it back in.  I unofficially approve this.  

You may want to add some tests in `tests/test_jwst_reader.py` with a fake HDU to ensure it correctly works and fails with the respective inputs or lack thereof.

ibusko:
> @ibusko I've tested that this works with 0.17.1 nirspec x1d files that have the correct SRCTYPE keyword in the EXTRACT1D header. I've also tested the existing 0.17.1 miri mrs x1d files fail due to the missing keyword, but work when I manually add it back in. I unofficially approve this.
> 
> You may want to add some tests in `tests/test_jwst_reader.py` with a fake HDU to ensure it correctly works and fails with the respective inputs or lack thereof.

There is one already: lines 148-158 in `tests/test_jwst_reader.py`

havok2063:
Ahh true, I didn't see that.

javerbukh (approved):
Works with 0.17.1 data. Nice fix!

eteq (approved):
LGTM! Merging.

eteq:
Oh, and the test failure looks to be #773 - we really should fix that since it's possible it's masking other things now... :shrug:

PR #778, opened 2021-02-05
Update doc examples of sdss spectrum

### LINKED ISSUE #773: Find an example spectrum that's a bit more reliable for the docs

eteq:
Closes #773 following @havok2063's suggestion of just switching to the SAS URLs.

Note, though, that while replacing this I also stumbled into a couple other SDSS spectra that might be concerning for the reasons #773 wanted to replace the 1d spectra - specifically https://github.com/astropy/specutils/blob/50752a0d58e3bd62153f21c92269e20b8a2ec637/specutils/tests/test_loaders.py#L174 and a few lines below it.  I can't seem to find plateid 4055 in the same parts of SAS, though. @havok2063, do you have any insight here?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -51,7 +51,7 @@ may have downloaded from some archive, or reduced from your own observations.
 
     Now we load the dataset from it's canonical source:
 
-    >>> f = fits.open('https://dr14.sdss.org/optical/spectrum/view/data/format=fits/spec=lite?plateid=1323&mjd=52797&fiberid=12')  # doctest: +IGNORE_OUTPUT +REMOTE_DATA
+    >>> f = fits.open('https://data.sdss.org/sas/dr16/sdss/spectro/redux/26/spectra/1323/spec-1323-52797-0012.fits')  # doctest: +IGNORE_OUTPUT +REMOTE_DATA
     >>> # The spectrum is in the second HDU of this file.
     >>> specdata = f[1].data # doctest: +REMOTE_DATA
     >>> f.close() # doctest: +REMOTE_DATA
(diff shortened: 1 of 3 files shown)


havok2063:
@eteq Most likely because that plate is an eboss plate, while the original spectra you were using is an old sdss or boss plate.  We separate eboss observations from the older ones.  Here is the directory with the spectrum used in the test, https://data.sdss.org/sas/dr16/eboss/spectro/redux/v5_13_0/spectra/lite/4055/.

nmearl (approved):

Issue #772, opened 2021-01-29
moment function doesn't work with SpectrumCollection


ibusko:
Results in an error like this:

`ValueError: operands could not be broadcast together with shapes (2,10000) (4,10000)`

dhomeier:
https://github.com/astropy/specutils/blob/625e05499386e4c48467a4a141d6763e9147fcbc/specutils/analysis/moment.py#L68-L78
is not doing the right thing for `SpectrumCollection` since `spectral_axis` already has the same shape as `flux`.
Also the final subtraction does not work for non-scalar `m1` (both `Spectrum1D` with multi-D flux and `SpectrumCollection`). Should be something like this (untested and probably not working for >2D):
```python
    if len(flux.shape) > len(spectral_axis.shape):
        dispersion = np.tile(spectral_axis, [flux.shape[0], 1])

    if order == 1:
        return np.sum(flux * dispersion, axis=axis) / np.sum(flux, axis=axis)

    if order > 1:
        m0 = np.sum(flux, axis=axis)
        m1 = np.sum(flux * dispersion, axis=axis) / np.sum(flux, axis=axis)
        if len(dispersion.shape) > 1:
            m1 = np.tile(m1, [dispersion.shape[-1], 1]).T

        return np.sum(flux * (dispersion - m1)**order, axis=axis) / m0
```

pllim:
Fixed in #781

PR #781, opened 2021-02-09
Fixes bug in moment calculation

ibusko:
as per bug report #772 

The first implementation didn't handle well `SpectrumCollection`. It also had problems with higher order moments (n > 1) with cubes in the spatial direction (axis=0 or 1). This PR fixes that. 

A few additional unit test cases exercise these (previously failing) modes.


rosteen (approved):
Looks good to me.

nmearl (approved):

PR #782, opened 2021-02-11
Replace spectrum segment by smooth spline interpolation

ibusko:
As per issue #760 

See relevant discussion in PR #762 

Example of replacement while preserving the input (default):

`model_replace(input_spectrum, spline_knots)`

image

Example of replacement while zeroing the non-replaced regions:

`model_replace(input_spectrum, spline_knots, extrapolation_treatment='zero_fill')`

image

Example of uncertainty array after replacement:

image

Note that the above uncertainty array results regardless of the extrapolation treatment that was selected. Is this what we want?


PatrickOgle (approved):
Looks good. I think this will satisfy the notebook-driven use case of spline-fitting regions of a spectrum.  I am agnostic on whether zero-fill or input data fill should be default.

specutils/manipulation/replace_segment.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    Generates a new spectrum with a region replaced by a smooth spline.
+
+    Parameters
+    ----------
+
+    spectrum : `~specutils.Spectrum1D`
+        The spectrum to be modified.
+
+    spline_knots : array of `~astropy.units.Quantity`
```
eteq, line 18:
```suggestion
    spline_knots : 1d `~astropy.units.Quantity`
```
If I'm understanding right?  "array of quantity" sounds like an array of quantity objects rather than a quantity with array-like shape

specutils/manipulation/replace_segment.py
```diff
@@ -0,0 +1,81 @@
...
+    -------
+    spectrum : `~specutils.Spectrum1D`
+        The spectrum with the region replaced by spline values, and data
+        values or zeros outside the spline region. The spectral axis will
+        have the same units as the spline knots.
+
+    Notes
+    -----
+    The uncertainty attribute, if present in the input spectrum, is not
+    carried over into the output.
```
eteq, line 38: I think it may make sense to just interpolate the uncertainty *exactly* the wsame way as the flux, right? It might be this should be an option (``interpolate_uncertainty=True``) since sometimes that's not the right thing, but as a "sometimes reasonable default" I think it might be a good idea

eteq (changes requested):
I very much like this in concept (particularly with the first of the two in the comment above as the default behavior).  I have a few inline comments, and one "big picture one":

Specifically, I think we don't want this to be spline-only.  In fact for many science cases what you really want is to do a model fit using the model fit tool, and then interpolate *using that model*. I think that better reflects the use case expressed in #760.

There's an intertwined problem, that we don't yet have a spline "model", mainly because of the limitation that unlike most of the other models, splines don't generally make any sense as a model *without* having been fit. That's a problem that needs fixing but I think we don't want to bake a "work-around" to that missing feature into `specutils` unless we have to.

So my concrete suggestion would be to change the `spline_knots` argument to be something that can either be:
1. An `astropy.modeling` model object, which is just assumed to have already been fit and is just used as `flux_spline_2` is right now
2. An `astropy.modeling` *class* (not an object), which is then fit to the spectrum before (eventually this will be the way to fit the spline)
3. A quantity/array.  This will then be interpreted *exactly* as the current `spline_knots` does now.  Once spline is possible via 2 above, we might or might not eventually deprecate this option.

If we go with that suggestion the function should also be renamed as `model_replace` or `fit_replace`.

specutils/manipulation/replace_segment.py
ibusko: Fixed.

specutils/manipulation/replace_segment.py
ibusko: I added uncertainty handling to the code.

ibusko:
@eteq in the cases where an astropy model is used instead of a spline, do you have a suggestion about how to specify the spectral axis range of that model? With the spline knots array, the range is implicit in the extreme values of the array. But a model, either a class or a fitted instance, doesn't carry spectral axis information with it. 

I am trying basically to design a call sequence for the function that makes more sense, or is more convenient, to the end user, without being overwhelming. The immediate thought would be to add yet another argument to the function, in the form of a `SpectralRegion` or something equivalent. Or perhaps (I like this better)  keep the `spline_knots` array as a mandatory argument, and have the fitting model defined by a keyword argument that defaults to None. The fitting range would then be always defined, and the user would have control on what model to fit (None will default to the existing spline fit). 

I'm sure you have a better sense than I have on what end users, or the "typical" user, would like to see there.

eteq:
@ibusko - sorry it took me so long to get back to this!

For the astropy model item, good point, how about a hybrid of your two suggestions:
* one keyword called `model` or perhaps `model_or_knots`, which has the options I listed in https://github.com/astropy/specutils/pull/782#pullrequestreview-598731958 , *as well as* an option of an integer
* Add a `replace_region` mandatory argument that is a SpectralRegion.  In the integer case above (which should probably be the default with some not-unreasonable number like 10?), the knots are just taken to span the `SpectralRegion` with the given number of knots. Then we still get the simple case you're suggesting in "i like this better" while keeping the flexibility?

(everything else looks good, so I think we're almost there!)

specutils/manipulation/model_replace.py
```diff
@@ -0,0 +1,183 @@
...
+        parameter `model` below should define spline knots explicitly in the
+        form of an `~astropy.units.Quantity` list or array
+
+    model :
+        An `~astropy.modeling` model object, which is assumed to have
+        already been fit to the spectrum;
+        Or
+        a list or array of spectral axis values to be used as spline knots.
+        They all should share the same units, which can be different
+        from the units of the input spectrum spectral axis;
```
eteq, line 34:
```suggestion
        from the units of the input spectrum spectral axis, but most
        be of compatible physical type;
```

eteq:
This is looking great now aside from one very minor thing above.  The one missing piece is my suggestion of passing in a model class, but as I think about it (and some out-of-band conversation with @ibusko) I'm not sure that's really a good idea anyway. So I think this is good to merge as-is (I'll just accept my suggestion here since it's so minor).

eteq (approved):

PR #786, opened 2021-02-26
Remove frozen SpectralCoord now that we require astropy 4.1

rosteen:
It looks like the last discussion of this was in #679 . Since then, the requirements have been updated to include `astropy>=4.1`, but the use of our frozen version of `SpectralCoord` was never removed in favor of the live astropy version. I'm not aware of any outstanding problems with the live version that preclude using it, so I updated the imports and such.


nmearl (approved):
Good catch. Thanks!

PR #789, opened 2021-03-08
Ignore priorities when getting loader extensions

nmearl:
This should resolve a build failure using astropy 4.2+ in which priorities are included in the returned tuples of the ``registry._readers`` dictionary.

### DIFF
--- a/specutils/io/registers.py
+++ b/specutils/io/registers.py
@@ -137,7 +137,16 @@ def get_loaders_by_extension(extension):
     loaders : list
         A list of loader names that are associated with the extension.
     """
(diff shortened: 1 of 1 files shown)


eteq (approved):
This definitely fixes it and is ok in that respect so I'm ok with this, but just to clarify, @nmearl, don't we want the priorities to be respected *if present*?

nmearl:
@eteq The `get_loaders_by_extension` function really only exists so users can get a list of support formats for a particular extension. In that case, it only needs to return the `format` names for each loader given an extension. I suppose we could make it return e.g. a table with a `name` and `priority` column, and whether it has reading/writing capabilities, but we should open another issue about that and address in another PR.

PR #798, opened 2021-03-12
Fix 2SLAQ_QSO loader error

eteq:
This should fix the last remaining build failure.  I confess I don't really understand these Data Central loaders (@aragilar might?), but if I'm understanding right it looks like this is a case where the `GAMA_2SLAQ_QSO_CONFIG` was updated in the tests but not the loader itself (specifically, the details of how to interpret the wcs).  So this just copy-and-pastes `GAMA_2SLAQ_QSO_CONFIG` from the tests to the loaderm which seems to do the job.


nmearl (approved):
Love seeing those ✅✅✅!

PR #805, opened 2021-04-09
MNT: Remove unused astropy config code

pllim:
This is probably leftover from package template. This package does not use the `astropy` configuration system at all.

Also see astropy/astropy#11497


nmearl (approved):
Thanks!

PR #806, opened 2021-04-12
Changed positional to keyword arguments when calling fit_continuum

tomasstolker:
The `weights` parameter was not correctly passed by `fit_generic_continuum` to `fit_continuum` so the `NotImplementedError` did not occur. The issue was solved by changing some of the positional arguments to keyword arguments.


nmearl (approved):
Great catch, thanks @tomasstolker!

PR #810, opened 2021-04-19
Update logging to use a non-root logger

### LINKED ISSUE #728: logging directly to root conflicts with other package loggers
Logging directly to the root , i.e.`logging.info` messes up the logging of packages that have `specutils` as  a dependency which use a named logger with its own stream handler.   Using `logging.info, etc` calls `logging.basicConfig`which creates a new StreamHandler at the top-level root logger, injecting itself as a parent to all other loggers.  This causes log messages to be displayed twice, once by the defined logger and again by the root.  There are certainly downstream (user-based) solutions to this, but it might be better to have `specutils` use a defined logger.

rosteen:
Fixes #728. This also removes some code from the __init__ file that was messing with the default logging config, which I think we would rather leave in the control of the user as discussed in #728.

I'm still deciding the best way to set the logger in each file; I'll move this from draft to ready once I've finalized it. In the meantime, comments are welcome.

### DIFF
--- a/specutils/spectra/spectrum_mixin.py
+++ b/specutils/spectra/spectrum_mixin.py
@@ -1,4 +1,5 @@
 from copy import deepcopy
+import logging
 
 import numpy as np
 import astropy.units.equivalencies as eq
@@ -8,7 +9,6 @@ from astropy.utils.decorators import lazyproperty, deprecated
 from astropy.wcs.wcsapi import HighLevelWCSWrapper
 
 from specutils.utils.wcs_utils import gwcs_from_array
-from ..utils.logger import log
 
 DOPPLER_CONVENTIONS = {}
 DOPPLER_CONVENTIONS['radio'] = u.doppler_radio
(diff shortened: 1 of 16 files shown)


specutils/spectra/spectrum_mixin.py
```diff
@@ -9,6 +8,7 @@
 from astropy.wcs.wcsapi import HighLevelWCSWrapper
 
 from specutils.utils.wcs_utils import gwcs_from_array
+from ..utils.logger import log
```
saimn, line 11: Better to use `logging.getLogger(__name__)` everywhere, this way it's possible to show the module when the message was emitted in the messages.

rosteen: Solid point, I erred on the side of being concise but I can deal with a few more lines for good reason.

rosteen:
Looks like the (non-allowed) failure was due to a file retrieval timeout last time it ran.

pllim:
That failing example with timeout should have been marked with `# doctest: +REMOTE_DATA`

pllim:
In a separate PR, should deal with the "allowed failures" or they gonna come bite you when `astropy` 4.3 is released.

pllim:
I think the fixes should be okay. Might want to add user-facing doc to tell them how to control the logging verbosity.

havok2063 (approved):
This looks good to me.  I've tested that it no longer inserts as the root logger and messages don't print twice.

pllim (approved):
LGTM and works for Brian C. Thanks!

PR #811, opened 2021-04-20
Update manga cube loader, fix bugs in Spectrum1D init

rosteen:
This updates the manga cube loader to take advantage of #754 by initializing the `Spectrum1D` with the WCS rather than a spectral axis. I also had to fix a few bugs that this surfaced in the `Spectrum1D` initialization code, which now properly reshapes an input mask as well as handling reshaping `uncertainty` inputs that are `astropy.nddata.NDUncertainty` instances.

### DIFF
--- a/specutils/io/default_loaders/manga.py
+++ b/specutils/io/default_loaders/manga.py
@@ -1,7 +1,7 @@
-
 import astropy.units as u
 from astropy.io import fits
 from astropy.nddata import InverseVariance
+from astropy.wcs import WCS
 
 from ...spectra import Spectrum1D
 from ..registers import data_loader
@@ -53,7 +53,7 @@ def manga_cube_loader(file_obj, **kwargs):
 
     spaxel = u.Unit('spaxel', represents=u.pixel, doc='0.5" spatial pixel', parse_strict='silent')
     with read_fileobj_or_hdulist(file_obj, **kwargs) as hdulist:
-        spectrum = _load_manga_spectra(hdulist, per_unit=spaxel, transpose=True)
+        spectrum = _load_manga_spectra(hdulist, per_unit=spaxel)
 
     return spectrum
 
@@ -83,13 +83,13 @@ def manga_rss_loader(file_obj, **kwargs):
     return spectrum
 
 
-def _load_manga_spectra(hdulist, per_unit=None, transpose=None):
+def _load_manga_spectra(hdulist, per_unit=None):
     """ Return a MaNGA Spectrum1D object
 
-    Returns a Spectrum1D object for a MaNGA data files.  Set
-    `transpose` kwarg to True for MaNGA cubes, as they are flipped relative
-    to what Spectrum1D expects.  Use the `per_unit` kwarg to indicate the
-    "spaxel" or "fiber" unit for cubes and rss files, respectively.
+    Returns a Spectrum1D object for a MaNGA data files. Use the `per_unit`
+    kwarg to indicate the "spaxel" or "fiber" unit for cubes and rss files,
+    respectively. Note that the spectral axis will automatically be moved to
+    be last during Spectrum1D initialization.
 
     Parameters
     ----------
@@ -97,8 +97,6 @@ def _load_manga_spectra(hdulist, per_unit=None, transpose=None):
         A MaNGA read astropy fits HDUList
     per_unit : astropy.units.Unit
         An astropy unit to divide the default flux unit by
-    transpose : bool
-        If True, transpose the data arrays
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


nmearl (approved):
Nice catch @rosteen! Thanks.

specutils/io/default_loaders/manga.py
```diff
@@ -53,7 +53,7 @@ def manga_cube_loader(file_obj, **kwargs):
 
     spaxel = u.Unit('spaxel', represents=u.pixel, doc='0.5" spatial pixel', parse_strict='silent')
     with read_fileobj_or_hdulist(file_obj, **kwargs) as hdulist:
-        spectrum = _load_manga_spectra(hdulist, per_unit=spaxel, transpose=True)
+        spectrum = _load_manga_spectra(hdulist, per_unit=spaxel)
```
havok2063, line 56: You mention the mask and uncertainty are now reshaped.  Is that why the `transpose` was removed?  The SDSS MaNGA files, including flux, is originally stored transposed to what `Spectrum1D` expected?  Does `Spectrum1D` fix this and transposes the entire cube internally?  I want to make sure we don't end up with transposed fluxes but correctly oriented masks and uncertainties.

havok2063 (commented):
Just took a look and added a few comments/questions, but in principle this looks good to me.

specutils/io/default_loaders/manga.py
```diff
@@ -110,18 +108,11 @@ def _load_manga_spectra(hdulist, per_unit=None, transpose=None):
         unit = unit / per_unit
 
     hdr = hdulist['PRIMARY'].header
-    wave = hdulist['WAVE'].data * u.angstrom
-
-    if transpose:
-        flux = hdulist['FLUX'].data.T * unit
-        ivar = InverseVariance(hdulist["IVAR"].data.T)
-        # SDSS masks are arrays of bit values storing multiple boolean conditions.
```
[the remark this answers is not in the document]
rosteen: Sure, I'll add it back in. Didn't mean to drop a useful comment!

specutils/io/default_loaders/manga.py
nmearl: According to the the tests, the shape is what we would expect. It may be worth updating the [unit tests](https://github.com/astropy/specutils/blob/main/specutils/tests/test_loaders.py#L153) to also check the shape of the uncertainties and mask.

rosteen: Sorry, was going to reply to this earlier and got distracted. The transpose was needed because `Spectrum1D` requires the spectral axis to be the last axis of the flux array, and it was removed because now `Spectrum1D` always performs this check and reshapes the arrays (flux, uncertainty and mask) if needed in its initialization code. It also warns the user if it did this reshaping to give them a heads up that the `Spectrum1D` is shaped differently than the input data.

There is a subtle difference in how it's handled now, which is that rather than transposing the input arrays, `Spectrum1D` now uses `np.moveaxis` internally (because the WCS needs to be updated to match the arrays and the astropy WCS API only has an equivalent of `moveaxis`, with no transpose method). So previously if your input array axes were say `[a, b, c]` (with 'a' being the spectral axis) the resulting `Spectrum1D` would have axes `[c, b, a]`. Now the resulting `Spectrum1D` will have axes `[b, c, a]`.

havok2063: Thanks for the explanation.  This is helpful.  We may want to update the user-facing documentation with a section explaining what Spectrum1D now does to input data.  It might be confusing to users who load their data in and find it's been reoriented.  With maybe a toy example.

javerbukh (commented):
In the CubevizExample notebook in jdaviz:
```
with fits.open(fn) as hdulist:
    for hdu in hdulist:
        data_label = f"{fn}[{hdu.name}]"

        if hdu.data is None or not hdu.is_image or "IMG" in hdu.name or "PSF" in hdu.name:
            continue
        try:
            sc = Spectrum1D.read(hdu, format="MaNGA cube")
            viz.app.data_collection[hdu.name] = sc
        except Exception as e:
            print(e)
            continue
```
Prints "File-like object does not have a 'write' method, required for mode 'ostream'." Any ideas?

rosteen:
Ah, the MaNGA cube reader is meant to be used on `fn` there, not an individual hdu. You also shouldn't need to provide a format string in that case, the `Spectrum1D.read` machinery should automatically detect that it's a MaNGA cube.

javerbukh (approved):
That works, thank you!

eteq (approved):

dhomeier:
@rosteen when using this loader I am getting the `spectral_axis` in meter(!) instead of Å as in the previous and v1.3 release version. I guess since it is a Quantity it will work all the same, but should perhaps be documented in the changelog.

havok2063:
meters is the incorrect unit for the spectral axis for MaNGA data.  Angstroms is the correct unit.  So it should ultimately be corrected, but if the bug will linger for a set of minor/patch releases, then adding a line item in the documentation is a good idea, for those versions.

dhomeier:
For 1.3 the switch to `NDCube` has been reverted and, I think, been postponed to a more stable interface. So if I am not mistaken in all the released versions the spectral axis from `hdulist['WAVE']` in Å is still used, and hopefully this can be addressed in good time.

Issue #809, opened 2021-04-19
Bin edges from unequally-spaced centers


zmoon:
For unequal spacing of centers, the algorithm
https://github.com/astropy/specutils/blob/57dd2b28e5122e00e28333043192d30ec9f78a9d/specutils/spectra/spectral_axis.py#L45
produces bins where some bin center locations don't match the input.

This seems potentially problematic, or at least misleading. Maybe a warning should be raised?

#### Examples

Including https://github.com/astropy/specutils/issues/176#issue-318981747

```python
def edges_from_centers(centers):
    a = np.insert(centers, 0, 2*centers[0] - centers[1])
    b = np.append(centers, 2*centers[-1] - centers[-2])
    edges = (a + b) / 2
    
    return edges

cases = [
    np.r_[1.0, 2.0, 3.0, 4.0],  # ok (equally spaced)
    np.r_[1.0, 2.0, 2.3, 3.0],
    np.r_[1, 2, 4, 6, 8, 9, 10],
]
    
for case in cases:
    edges = edges_from_centers(case)
    print("  input centers:", " "*3 + "".join(f"{x:<6}" for x in case))
    print(" computed edges:", "".join(f"{x:<6}" for x in edges))
    print("implied centers:", " "*3 + "".join(f"{x:<6}" for x in (edges[:-1] + edges[1:])/2))
    print()
```
Output:
```
  input centers:    1.0   2.0   3.0   4.0   
 computed edges: 0.5   1.5   2.5   3.5   4.5   
implied centers:    1.0   2.0   3.0   4.0   

  input centers:    1.0   2.0   2.3   3.0   
 computed edges: 0.5   1.5   2.15  2.65  3.35  
implied centers:    1.0   1.825 2.4   3.0   

  input centers:    1     2     4     6     8     9     10    
 computed edges: 0.5   1.5   3.0   5.0   7.0   8.5   9.5   10.5  
implied centers:    1.0   2.25  4.0   6.0   7.75  9.0   10.0  
```

rosteen:
Hey @zmoon, just wanted to let you know I saw this but didn't have much time to think about it this week. I remember some discussion about this from a while back as well, I'll look back at those issues/PRs and give it some thought next week. Agreed that it might be useful to have a warning if there are use cases where it might come up.

rosteen:
[...]

PR #816, opened 2021-04-28
Raise an error if bin_edges is called for unevenly spaced centers

rosteen:
Addresses #809. I could be convinced that this should just be a warning instead of an error, but I think since the calculated bin edges don't reproduce the input centers we should error and not return them at all. See the linked issue for a little more discussion of why I didn't just implement a better edge calculation algorithm.


pllim:
CI is failing and needs to be fixed.

dhomeier:
Probably needs to be rebased as some tests were generally failing at the time, but this test looks very unfinished!

PR #818, opened 2021-05-03
Adds support for JWST c1d spectroscopic files

havok2063:
This PR adds support for JWST combined 1d spectroscopic files (c1d).  These files are structurally the same as x1d files, so I've consolidated the identifiers and loaders.  I've also added a case when the `SRCTYPE` header keyword is present but has `value='UNKNOWN'`, where I issue a warning and default to using the `FLUX` and `ERROR` columns.  The original `SRCTYPE` error is issued when the `SRCTYPE` header value cannot be found.   

The logging syntax mirrors that of https://github.com/astropy/specutils/pull/810, but this PR is not dependent on that PR.

### DIFF
--- a/specutils/io/default_loaders/jwst_reader.py
+++ b/specutils/io/default_loaders/jwst_reader.py
@@ -241,16 +241,17 @@ def _jwst_spec1d_loader(file_obj, extname='EXTRACT1D', **kwargs):
             if "srctype" in hdu.header:
                 srctype = hdu.header.get("srctype", None)
 
+            # checking if SRCTPYE is missing or UNKNOWN
+            if not srctype or srctype == 'UNKNOWN':
+                log.warning('SRCTYPE is missing or UNKNOWN.  Defaulting to srctype="POINT".')
+                srctype = 'POINT'
+
             if srctype == "POINT":
                 flux = Quantity(data["FLUX"])
                 uncertainty = StdDevUncertainty(data["ERROR"])
             elif srctype == "EXTENDED":
                 flux = Quantity(data["SURF_BRIGHT"])
                 uncertainty = StdDevUncertainty(hdu.data["SB_ERROR"])
-            elif srctype == 'UNKNOWN':
-                flux = Quantity(data["FLUX"])
-                uncertainty = StdDevUncertainty(data["ERROR"])
-                log.warning('SRCTYPE is UNKNOWN.  Defaulting to loading FLUX and ERROR data.')
             else:
                 raise RuntimeError(f"Keyword SRCTYPE is {srctype}.  It should "
                                    "be 'POINT' or 'EXTENDED'. Can't decide between `flux` and "
--- a/specutils/io/default_loaders/jwst_reader.py
+++ b/specutils/io/default_loaders/jwst_reader.py
@@ -3,6 +3,7 @@ from astropy.units import Quantity
 from astropy.table import Table
 from astropy.io import fits
 from astropy.nddata import StdDevUncertainty
+import logging
 import numpy as np
 import asdf
 from gwcs.wcstools import grid_from_bounding_box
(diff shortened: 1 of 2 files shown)


specutils/io/default_loaders/jwst_reader.py
```diff
@@ -155,18 +233,24 @@ def _jwst_x1d_loader(file_obj, **kwargs):
...
                 flux = Quantity(data["SURF_BRIGHT"])
                 uncertainty = StdDevUncertainty(hdu.data["SB_ERROR"])
-
+            elif srctype == 'UNKNOWN':
+                flux = Quantity(data["FLUX"])
+                uncertainty = StdDevUncertainty(data["ERROR"])
+                log.warning('SRCTYPE is UNKNOWN.  Defaulting to loading FLUX and ERROR data.')
             else:
                 raise RuntimeError(f"Keyword SRCTYPE is {srctype}.  It should "
                                    "be 'POINT' or 'EXTENDED'. Can't decide between `flux` and "
```
ibusko, line 256: There are cases where *no* `SRCTYPE` keyword exists at all in the header (there is a pemding Jira bug fix ticket). Shouldn't the `flux` variable be defined in the same way for the cases `SRCTYPE=UNKNOWN`, and `SRCTYPE=None`?

havok2063: that's a good question, which I did consider.  Talking to some pipeline folks, it seems that the `SRCTYPE` keyword should never be missing, but we've certainly found bugs in files where that was the case.  So I was originally thinking to leave the `raise` error to catch files where the `SRCTYPE` keyword is missing.   But I think it depends a bit on what we want Jdaviz or specutils to do when a user tries to load these files.  Should Jdaviz/specutils throw an error or still load them using some default logic?  I don't have a strong preference either way.   

If we switched to loading the file with some defaults, as we do with UNKNOWN, I could issue a new warning in the log.  In those cases, is it possible to catch these warnings and issue a warning snackbar message in Jdaviz ?

ibusko: I don't know if it is possible to catch warnings issued via the `logging` package. A possible workaround would be to have the code raise a custom exception, and have `jdaviz` code to catch that specific type of exception.

ibusko: In our tagup today, it was clarified that a decision was made a while ago (in the context of the NIRISS parser, PR#394) that the `jdaviz` tools should load the files anyway, even when no `SRCTYPE` keyword is present. In that case, a warning message of some sort (snackbar perhaps) could inform the user that the file is not compliant but is being loaded anyway using a default setting. The `MOSViz NIRISS parser uses a default of `EXTENDED`.

havok2063: Ok, sounds good.  I also clarified that idea on the tag up today.  I think we can default to loading something and issue a warning.  I'll update the code and ping when you it's ready.

havok2063: @ibusko @rosteen I've updated the code and pushed the changes.  The loader now sets the default `SRCTPYE`  for missing or UNKNOWN values to `POINT`.

ibusko: I wonder if this will generate user confusion with users, since the MOSViz NIRISS parser uses a default of `EXTENDED`. https://github.com/spacetelescope/jdaviz/blob/main/jdaviz/configs/mosviz/plugins/parsers.py

rosteen: Does quietly setting the `SRCTYPE` to "EXTENDED"  in Mosviz affect anything such that the user would even notice it was done? I don't think we need to worry about that here, since this explicitly warns the user about what it's doing (whereas in Mosviz the parser _knows_ what the sourcetype for NIRISS data should be and just silently adds it if needed without the user needing to worry about it). Seems like the warning that something might be wrong with the way the file was constructed is more important than the choice of what to use for the missing keyword.

ibusko (approved):
Works as expected.

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -155,18 +233,25 @@ def _jwst_x1d_loader(file_obj, **kwargs):
...
+            # or missing.  If that's the case, default to using POINT as the SRCTYPE.
+            # Error out only when SRCTYPE is a bad value.
             srctype = None
             if "srctype" in hdu.header:
-                srctype = hdu.header.get("srctype")
+                srctype = hdu.header.get("srctype", None)
+
+            # checking if SRCTPYE is missing or UNKNOWN
+            if not srctype or srctype == 'UNKNOWN':
+                log.warning('SRCTYPE is missing or UNKNOWN.  Defaulting to srctype="POINT".')
```
eteq, line 246: I find logging messages like this are much easier to understand if they say where they come from (of course one can search the line number, but most people don't...)
```suggestion
                log.warning('SRCTYPE is missing or UNKNOWN in JWST x1d loader. 
                             Defaulting to srctype="POINT".')
```

eteq (approved):
LGTM too! This has been sitting for a while, so I'm just going to accept the minor suggestion I made so I can merge this.  @havok2063, if you disagreed, though, feel free to issue a follow-on PR reverting that commit and I'm happy to discuss it there.

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -155,18 +233,26 @@ def _jwst_x1d_loader(file_obj, **kwargs):
...
+            # Error out only when SRCTYPE is a bad value.
             srctype = None
             if "srctype" in hdu.header:
-                srctype = hdu.header.get("srctype")
+                srctype = hdu.header.get("srctype", None)
+
+            # checking if SRCTPYE is missing or UNKNOWN
+            if not srctype or srctype == 'UNKNOWN':
+                log.warning('SRCTYPE is missing or UNKNOWN in JWST x1d loader. 
+                             Defaulting to srctype="POINT".')
```
eteq, line 247: whoops...
```suggestion
                log.warning('SRCTYPE is missing or UNKNOWN in JWST x1d loader. '
                            'Defaulting to srctype="POINT".')
```

havok2063:
@eteq those changes are fine by me!

PR #819, opened 2021-05-18
docs: fix inaccuracy about custom loading

PaulPrice:
Some of the examples lower down show loaders returning a
SpectrumList, so it doesn't just have to be a Spectrum1D.


nmearl:
Thanks for the update!

nmearl (approved):

PR #822, opened 2021-06-02
Use np.swapaxes to match wcs operation correctly

rosteen:
After a vacation break, I came back and looked at this code with fresh eyes and immediately said "how the heck did I convince myself I needed to use `np.moveaxis` instead of `np.swapaxes`?" 

@eteq @nmearl can I get a sanity check here - I believe the way I have it coded now the RA/Dec axes in a 3D flux cube end up switched compared to the WCS, which this PR should fix. I think I had somehow erroneously gotten into my head at some point that the WCS operation was a move, not a swap.

### DIFF
--- a/specutils/tests/test_spectrum1d.py
+++ b/specutils/tests/test_spectrum1d.py
@@ -5,6 +5,7 @@ import numpy as np
 import pytest
 from astropy.nddata import StdDevUncertainty
 from astropy.coordinates import SpectralCoord
+from astropy.wcs import WCS
 
 from .conftest import remote_access
 from ..spectra import Spectrum1D
@@ -106,6 +107,20 @@ def test_create_with_spectral_coord():
     assert isinstance(spec.spectral_axis, SpectralCoord)
     assert spec.spectral_axis.size == 50
 
+
+def test_create_from_cube():
+
+    flux = np.arange(24).reshape([2,3,4])*u.Jy
+    wcs_dict = {"CTYPE1": "RA---TAN", "CTYPE2": "DEC--TAN", "CTYPE3": "WAVE-LOG",
+                "CRVAL1": 205, "CRVAL2": 27, "CRVAL3": 3.622e-7,
+                "CRDELT1": -0.0001, "CRDELT2": 0.0001, "CRDELT3": 8e-11,
+                "CRPIX1": 0, "CRPIX2": 0, "CRPIX3": 0}
+    w = WCS(wcs_dict)
+
+    spec = Spectrum1D(flux=flux, wcs=w)
+
+    assert spec.flux.shape == (4,3,2)
+    assert spec.flux[3,2,1] == 23*u.Jy
 
 def test_spectral_axis_conversions():
     # By default the spectral axis units should be set to angstroms
--- a/specutils/tests/test_spectrum1d.py
+++ b/specutils/tests/test_spectrum1d.py
@@ -121,6 +121,7 @@ def test_create_from_cube():
 
     assert spec.flux.shape == (4,3,2)
     assert spec.flux[3,2,1] == 23*u.Jy
+    assert np.all(spec.spectral_axis.value == np.exp(np.array([1,2])*w.wcs.cdelt[-1]/w.wcs.crval[-1])*w.wcs.crval[-1])
 
 def test_spectral_axis_conversions():
     # By default the spectral axis units should be set to angstroms
--- a/specutils/tests/test_spectrum1d.py
+++ b/specutils/tests/test_spectrum1d.py
@@ -113,7 +113,7 @@ def test_create_from_cube():
     flux = np.arange(24).reshape([2,3,4])*u.Jy
     wcs_dict = {"CTYPE1": "RA---TAN", "CTYPE2": "DEC--TAN", "CTYPE3": "WAVE-LOG",
                 "CRVAL1": 205, "CRVAL2": 27, "CRVAL3": 3.622e-7,
-                "CRDELT1": -0.0001, "CRDELT2": 0.0001, "CRDELT3": 8e-11,
+                "CDELT1": -0.0001, "CDELT2": 0.0001, "CDELT3": 8e-11,
                 "CRPIX1": 0, "CRPIX2": 0, "CRPIX3": 0}
     w = WCS(wcs_dict)
 
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -206,23 +206,23 @@ class Spectrum1D(OneDSpectrumMixin, NDCube):
                              " last. Reshaping arrays to put spectral axis last.")
                     wcs = wcs.swapaxes(0, temp_axes[0])
                     if flux is not None:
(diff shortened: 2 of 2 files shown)


eteq:
Oh that's a tricky one.  I'm not sure I even know which is the right answer here.  I guess here's what I see as the science-centric perspective thinking about a spectral-spatial-spatial to have a concrete way to think about it.  Here's what I see as the result of `spectrum[..., spec_idx]` in the two cases hgere:

1. `moveaxis`: this is the right thing  to reproduce the behavior of `original_cube[spec_idx, :, :]` and have that be "the spatial plane", which I think is what the science user generally expects
2. `swapaxis`: this gives `original_cube[spec_idx, :, :].T`, which is not as intuitive. And it's much worse with like a spatial-spectral-spatial-time or similar >3D dataset

So based on that it feels to me like `moveaxis` is the right answer.  But I think you might be right that this is *not* consistent with WCS? But I could be wrong about that.  I am constantly confused about transposes in WCS land. My approahc is usually transpose-until-its-right :shrug: ...

eteq:
Put another way: I wonder if the fix is really to either re-jigger the WCS (dangerous...), or alternatively, present a "transposed" view of the data to the WCS?

(Note this whole discussion might be easier if we have a 4D cube, as that will help make sure we don't do something that's purely a transposed-too-many-times issue vs a more complicated reordering. I dunno if we have an example of that, but I could probably cook up a fake one if we want it.)

pllim:
@eteq , but the CI passed, so whatever this change is, it didn't trip the existing functionality that is being tested, correct?

Regardless, I think this PR needs a new test (or a few) to test the original problem that this PR is trying to fix.

pllim:
p.s. The "allowed failure" failures are unrelated as I see the same failures in `main`.

rosteen:
@pllim Agreed that some new test cases are needed here. I can probably work on that tomorrow/early next week.

eteq:
Yeah, I'm 98% sure we don't have any tests that involve the more complex sort of cubes I was talking about here.  So it doesn't surprise me that the tests aren't failing.  But agreed that means more tests are desirable.  I'm just still not 100% sure what the "right" answer is for those tests!

(That said... maybe this is really test-driven devlopment!  I.e., the tests might be a better way to understand the right answer.)

eteq:
Given how tricky this is I'm going to ask for a few more perspectives: @Cadair @PatrickOgle @keflavich.

The TL;DR version is: given that specutils *requires* the spectral axis be the last axis, what's the more logical way to re-order N-D cubes: "swapping" the last and the spectral axis, or "moving" the spectral axis to the end (which preserves the order of all the other axes). 

The second question is: is whatever the answer is above ok even if the WCS expects it to be transposed or something?

(And @Cadair, you don't need to say "I told you so" - I'm already thinking it :wink:)

Cadair:
> (And @Cadair, you don't need to say "I told you so" - I'm already thinking it wink)

I will however say that it only requires it to be the last axis because you made it. You can unmake it.

I think if you are going to do this you want to only ever present a view of the array in a different order (although doing this with dask arrays makes me extremely nervous) and secondly to reorder the WCS https://github.com/sunpy/ndcube/blob/main/ndcube/wcs/wrappers/reordered_wcs.py#L9 might me useful.

keflavich:
This is awkward.  `spectral-cube` requires the spectral axis to be the _first_ axis, axis zero, since coordinates in python are usually backwards, i.e. the x, y and z dimensions are sliced as `[z,y,x]`.

However, I think the order shouldn't matter - we should just pass the right order to the wcs.

I can't say either swapping or moving has any intrinsic advantage.  An image, in python, should be laid out as `[y,x]`, though, so if you're dealing with data that are in `[z,y,x]` order, I'd say use `moveaxis`.  If the data are `[z,x,y]`, they're already weird, so it doesn't matter which you use.

I also agree that these choices should be driven by explicit use cases rather than general hypothesizing.

havok2063:
I can be convinced otherwise but I think it makes sense to preserve the order, and stay as close to `[z,y,x]` as possible, and use `moveaxis`.  I think it's important to preserve the slicing order (for non-spectral axes) between when users slice regular python data arrays and slice via `specutils` objects so as not to confuse users.  If you're reordering the spatial axes it will confuse users unless you put multiple warnings and prompts in the code and the docs.  Users will spend a lot of time loading data both ways just to validate pixels are in the correct spot and they aren't going crazy.  Internally I think it's fine to do whatever.

PatrickOgle:
I would like Spectrum1D to preserve whatever order the axes are in when you extract the data array using astropy.io.fits.
That way, the Spectrum1D and data arrays would be indexed the same way.

rosteen:
I just create #852 for any further discussion of axis ordering, so that we can merge and close this PR. The most important thing here is that, regardless of whether we decide to change the order later, as the code currently stands the WCS does not correctly match the data in the `flux` array. This PR fixes the `__init__` method so the WCS corresponds to the correct data in the `flux` array (by doing a `swap` on both, rather than a `swap` on one and a `move` on the other) and should be merged.

eteq (changes requested):
This looks right to me, *but* I think it still needs a test.  That is, something that fails in the current `main` but passes with this PR.

Ideally this would just be a cube created as part of the test a la how some of the loaders just load a "fits file" defined in the test itself, no need for an actual real data file since it's just the dimensionality/etc that matters.

specutils/tests/test_spectrum1d.py
```diff
@@ -107,6 +108,20 @@ def test_create_with_spectral_coord():
     assert spec.spectral_axis.size == 50
 
 
+def test_create_from_cube():
+
+    flux = np.arange(24).reshape([2,3,4])*u.Jy
+    wcs_dict = {"CTYPE1": "RA---TAN", "CTYPE2": "DEC--TAN", "CTYPE3": "WAVE-LOG",
+                "CRVAL1": 205, "CRVAL2": 27, "CRVAL3": 3.622e-7,
+                "CRDELT1": -0.0001, "CRDELT2": 0.0001, "CRDELT3": 8e-11,
```
eteq, line 116: typo (necessary for the suggestion below to work)
```suggestion
                "CDELT1": -0.0001, "CDELT2": 0.0001, "CDELT3": 8e-11,
```

specutils/tests/test_spectrum1d.py
```diff
@@ -107,6 +108,20 @@ def test_create_with_spectral_coord():
...
+                "CRPIX1": 0, "CRPIX2": 0, "CRPIX3": 0}
+    w = WCS(wcs_dict)
+
+    spec = Spectrum1D(flux=flux, wcs=w)
+
+    assert spec.flux.shape == (4,3,2)
+    assert spec.flux[3,2,1] == 23*u.Jy
```
eteq, line 123: this checks that the spectral axis is actually right
```suggestion
    assert spec.flux[3,2,1] == 23*u.Jy
    assert np.all(spec.spectral_axis.value == np.exp(np.array([1,2])*w.wcs.cdelt[-1]/w.wcs.crval[-1])*w.wcs.crval[-1])
```

eteq (approved):

PR #823, opened 2021-06-04
Stop infinitely recursing in some model fitting cases

rosteen:
I was running into a nasty infinite recursion while trying to fit `Polynomial1D` or `Linear1D` models, possibly due to some quirk of the way `fit_lines` was getting called in a wrapper class in a multiprocessing pool. Regardless of the exact details, this change prevents the infinite recursion. However, I'm somewhat surprised that it didn't break anything else and I'm not 100% clear about the initial motivation for overriding the attribute getters and setters, so I'm very open to suggestions on better ways to fix this.

### DIFF
--- a/specutils/tests/test_utils.py
+++ b/specutils/tests/test_utils.py
@@ -24,7 +24,7 @@ def test_quantity_model():
 
     assert uc(10*u.nm).to(u.m) == 0*u.m
 
-def test_pickle_quantity_model():
+def test_pickle_quantity_model(tmp_path):
     """
     Check that a QuantityModel can roundtrip through pickling, as it
     would if fit in a multiprocessing pool.
@@ -33,10 +33,12 @@ def test_pickle_quantity_model():
     c = modeling.models.Chebyshev1D(3)
     uc = QuantityModel(c, u.AA, u.km)
 
-    with open("qmodel.pkl", "wb") as f:
+    pkl_file = tmp_path / "qmodel.pkl"
+
+    with open(pkl_file, "wb") as f:
         pickle.dump(uc, f)
 
-    with open("qmodel.pkl", "rb") as f:
+    with open(pkl_file, "rb") as f:
         new_model = pickle.load(f)
 
     assert new_model.input_units == uc.input_units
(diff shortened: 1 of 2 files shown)


nmearl:
Interesting. Can you add a test that exemplifies the problem you faced?

pllim (commented):
I'll have to sit this one out as I don't know what is going on here. Looks like @eteq reviewed Craig's original implementation in #342 , while unit support in models (first introduced in astropy/astropy#4855) was still kinda new.

The fact that CI didn't fail means the case for `nm = 'unitless_model'` is not tested. This should be remedied regardless.

rosteen:
I was able to figure out what was triggering the recursion, so I'll be able to write a test for it next week. 

The problem occurs when unpickling a pickled `QuantityModel` object, as might happen for example after a multiprocessing thread sends its result back to the main processing pool. The unpickling process tries to access the QuantityModel's `__getstate__`, which in turn tries to use `__getattr__` to retrieve the `unitless_model` and throws it into the recursion. The thing I was worried about was that excluding `unitless_model` from this custom `__getattr__` method would prevent it from being retrieved on the other end of a pickling at all, but in my testing it seems like it round trips properly.

rosteen:
Added a test that would fail on main due to a `RecursionError` to check that `QuantityModel` successfully round-trips through a `pickle.dump`/`pickle.load`.

pllim:
Did you verify that it actually errors out as expected without your patch? Maybe a draft PR separately with just the test as a proof?

rosteen:
I did, but I can make a draft PR to show that as well.

pllim:
If you have URL to that log, it works too. 😸

rosteen:
I ran it locally, how about a screenshot?

Screen Shot 2021-06-07 at 10 00 00 AM

pllim:
Sometimes CI disagrees with local run, but if the other maintainers think this is sufficient, fine by me.

rosteen:
@pllim see #824 for more proof! =D

nmearl (approved):
Nice sleuthing, @rosteen! I do think we eventually need to revisit our workarounds in light of astropy 4.0 and the modeling changes to see if these proxy models are strictly necessary, but this is great. Thanks!

PR #824, opened 2021-06-07
For dummy PR showing test failure

rosteen:
Proof that this test fails on main for #823 .


pllim:
```
E   RecursionError: maximum recursion depth exceeded
!!! Recursion detected (same locals & position)
```

Consider it proven. Thanks for humoring me!

PR #825, opened 2021-06-10
Update changelog with changes since 1.2 release

rosteen:
I'd like to get a release out soon with the recursion bug fix from #823. However, NDCube still hasn't done their 2.0 release, so we might want to revert the change to subclassing from NDCube instead of releasing with a pin to the dev version, in which case this would 1.2.1 rather than 1.3.


rosteen:
I removed #754 and #811 (the two PRs related to the NDCube integration) and added #782 , which got merged yesterday.

Issue #832, opened 2021-06-23
wcs.world_axis_physical_types check breaks apStar_loader


dhomeier:
Since merging #754 its tests fail as
```python
________________________________________________________________________________ test_apStar_loader ________________________________________________________________________________

tmpdir = local('/private/var/folders/bb/f_z2vsjd21d_p0jpqxzhhwhh00027w/T/pytest-of-derek/pytest-52/test_apStar_loader0')

    @pytest.mark.remote_data
    def test_apStar_loader(tmpdir):
        apstar_url = ("https://data.sdss.org/sas/dr16/apogee/spectro/redux/r12/"
                      "stars/apo25m/N7789/apStar-r12-2M00005414+5522241.fits")
        with set_temp_cache(path=str(tmpdir)):
            filename = download_file(apstar_url, cache=True)
>           spectrum = apStar_loader(filename)

lib/python3.8/site-packages/specutils/io/default_loaders/tests/test_apogee.py:17: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
lib/python3.8/site-packages/specutils/io/registers.py:111: in wrapper
    return func(*args, **kwargs)
lib/python3.8/site-packages/specutils/io/default_loaders/apogee.py:146: in apStar_loader
    return Spectrum1D(data=data * unit,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'Spectrum1D' object has no attribute '_data'") raised in repr()] Spectrum1D object at 0x12cb64fd0>
flux = <Quantity [0., 0., 0., ..., 0., 0., 0.] 1e-17 erg / (Angstrom cm2 s)>
[...]

PR #833, opened 2021-06-23
Handle new `PhysicalType` class in `spectrum_from_column_mapping`

### LINKED ISSUE #831: Setting custom units in `spectrum_from_column_mapping` fails with astropy 4.3
These tests on v1.3 are failing with astropy 4.3rc1:
```python
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_writer[wavelength] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_writer[frequency] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_writer[energy] - KeyError: 'energy'
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_writer[wavenumber] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[wavelength-1] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[wavelength-2] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[wavelength-3] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[frequency-1] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[frequency-2] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[frequency-3] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[energy-1] - KeyError: 'energy'
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[energy-2] - KeyError: 'energy'
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[energy-3] - KeyError: 'energy'
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[wavenumber-1] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[wavenumber-2] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
FAILED lib/python3.9/site-packages/specutils/tests/test_loaders.py::test_tabular_fits_multid[wavenumber-3] - assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
```
the majority are failures on the `flux` unit like
```python
        spectrum.write(tmpfile, format='tabular-fits', overwrite=True)
    
        cmap = {spectral_axis: ('spectral_axis', wlu[spectral_axis]),
                'flux': ('flux', 'erg / (s cm**2 AA)'),
                'uncertainty': ('uncertainty', None)}
    
        # Read it back again and check against the original
        spec = Spectrum1D.read(tmpfile, format='tabular-fits', column_mapping=cmap)
>       assert spec.flux.unit == u.Unit('erg / (s cm**2 AA)')
E       assert Unit("W / (Angstrom m2)") == Unit("erg / (Angstrom cm2 s)")
E        +  where Unit("W / (Angstrom m2)") = <Quantity [ 5.15047686e-10,  3.85273149e-09,  5.70890511e-10,\n            1.13556564e-09,  9.54001763e-10,  6.51391251...     -3.15269245e-10,  7.58969220e-10, -7.72825215e-10,\n           -2.36818607e-10, -4.85363548e-10] W / (Angstrom m2)>.unit
E        +    where <Quantity [ 5.15047686e-10,  3.85273149e-09,  5.70890511e-10,\n            1.13556564e-09,  9.54001763e-10,  6.51391251...     -3.15269245e-10,  7.58969220e-10, -7.72825215e-10,\n           -2.36818607e-10, -4.85363548e-10] W / (Angstrom m2)> = <Spectrum1D(flux=<Quantity [ 5.15047686e-10,  3.85273149e-09,  5.70890511e-10,\n            1.13556564e-09,  9.54001763...                  1.77558228e-06, 2.75493960e-06, 2.77997341e-06,\n                   1.53889118e-06, 2.20309679e-06]))>.flux
E        +  and   Unit("erg / (Angstrom cm2 s)") = <class 'astropy.units.core.Unit'>('erg / (s cm**2 AA)')
E        +    where <class 'astropy.units.core.Unit'> = u.Unit
```
the reason being `spectrum_from_column_mapping` is failing the check
https://github.com/astropy/specutils/blob/3a86739d53663b528ee751eb55d01de3c252022d/specutils/io/parsing_utils.py#L110
as `unit.physical_type` in astropy 4.3 has become an object of type `PhysicalType` instead of a plain string; this can be simply fixed by checking for `'spectral flux' in str(cm_unit.physical_type)` instead.

Somewhat related, `spectral_axis` type 'energy' is not recognised since energy units like `u.eV` have now become
`PhysicalType({'energy', 'torque', 'work'})`
and are accordingly written to FITS as
`TTYPE1  = 'energy/torque/work'`

I have to think about how to proceed best with this case. The `spectral_axis` column would need to be referenced as `table['energy/torque/work']`, but I'd prefer to have the writer not use that ugly `TTYPE` in the first place.
Setting custom units on 'energy' seems broken anyway, as the test simply uses `u.eV` before _and_ after.

dhomeier:
To fix #831 and #824.

This converts the new ~astropy.units.physical.PhysicalType class from 4.3 (https://github.com/astropy/astropy/pull/11204) to string, if necessary, and also uses the appropriate type only from types like `PhysicalType({'energy', 'torque', 'work'})`. Updated tests to test setting of custom `spectral_axis` units and types as well.

### DIFF
--- a/specutils/io/parsing_utils.py
+++ b/specutils/io/parsing_utils.py
@@ -104,8 +104,8 @@ def spectrum_from_column_mapping(table, column_mapping, wcs=None):
 
             if not isinstance(cm_unit, u.Unit):
                 cm_unit = u.Unit(cm_unit)
-            cm_type =  str(cm_unit.physical_type)
-            if 'length in cm_type' or 'frequency' in cm_type or 'energy' in cm_type:
+            cm_type = str(cm_unit.physical_type)
+            if 'length' in cm_type or 'frequency' in cm_type or 'energy' in cm_type:
                 # Spectral axis column information
                 kwarg_val = kwarg_val.to(cm_unit, equivalencies=u.spectral())
             elif 'spectral flux' in cm_type:
--- a/specutils/io/parsing_utils.py
+++ b/specutils/io/parsing_utils.py
@@ -104,11 +104,10 @@ def spectrum_from_column_mapping(table, column_mapping, wcs=None):
 
             if not isinstance(cm_unit, u.Unit):
                 cm_unit = u.Unit(cm_unit)
-            cm_type = str(cm_unit.physical_type)
-            if 'length' in cm_type or 'frequency' in cm_type or 'energy' in cm_type:
+            if cm_unit.physical_type in ('length', 'frequency', 'energy'):
                 # Spectral axis column information
                 kwarg_val = kwarg_val.to(cm_unit, equivalencies=u.spectral())
-            elif 'spectral flux' in cm_type:
+            elif 'spectral flux' in str(cm_unit.physical_type):
                 # Flux/error column information
                 kwarg_val = kwarg_val.to(cm_unit, equivalencies=u.spectral_density(1 * u.AA))
         elif tab_unit:
--- a/specutils/io/parsing_utils.py
+++ b/specutils/io/parsing_utils.py
@@ -99,8 +99,8 @@ def spectrum_from_column_mapping(table, column_mapping, wcs=None):
             kwarg_val = u.Quantity(table[col_name], tab_unit)
 
             # Attempt to convert the table unit to the user-defined unit.
-            log.debug(f"Attempting auto-convert of table unit {tab_unit} to "
-                      f"user-provided unit {cm_unit}.")
+            log.debug("Attempting auto-convert of table unit '%s' to "
+                      "user-provided unit '%s'.", tab_unit, cm_unit)
 
             if not isinstance(cm_unit, u.Unit):
                 cm_unit = u.Unit(cm_unit)
--- a/specutils/io/parsing_utils.py
+++ b/specutils/io/parsing_utils.py
@@ -99,18 +99,18 @@ def spectrum_from_column_mapping(table, column_mapping, wcs=None):
             kwarg_val = u.Quantity(table[col_name], tab_unit)
 
             # Attempt to convert the table unit to the user-defined unit.
-            log.debug("Attempting auto-convert of table unit '%s' to "
-                          "user-provided unit '%s'.", tab_unit, cm_unit)
+            log.debug(f"Attempting auto-convert of table unit {tab_unit} to "
+                      f"user-provided unit {cm_unit}.")
 
             if not isinstance(cm_unit, u.Unit):
                 cm_unit = u.Unit(cm_unit)
-            if cm_unit.physical_type in ('length', 'frequency'):
+            cm_type =  str(cm_unit.physical_type)
+            if 'length in cm_type' or 'frequency' in cm_type or 'energy' in cm_type:
                 # Spectral axis column information
                 kwarg_val = kwarg_val.to(cm_unit, equivalencies=u.spectral())
-            elif 'spectral flux' in cm_unit.physical_type:
+            elif 'spectral flux' in cm_type:
                 # Flux/error column information
-                kwarg_val = kwarg_val.to(
-                    cm_unit, equivalencies=u.spectral_density(1 * u.AA))
+                kwarg_val = kwarg_val.to(cm_unit, equivalencies=u.spectral_density(1 * u.AA))
         elif tab_unit:
             # The user has provided no unit in the column mapping, so we
             # use the unit as defined in the table object.
--- a/specutils/tests/test_loaders.py
+++ b/specutils/tests/test_loaders.py
@@ -522,14 +522,14 @@ def test_tabular_fits_writer(tmpdir, spectral_axis):
         spectrum.write(tmpfile, format='tabular-fits')
     spectrum.write(tmpfile, format='tabular-fits', overwrite=True)
 
-    cmap = {spectral_axis: ('spectral_axis', wlu[spectral_axis]),
+    cmap = {spectral_axis: ('spectral_axis', 'micron'),
             'flux': ('flux', 'erg / (s cm**2 AA)'),
             'uncertainty': ('uncertainty', None)}
 
     # Read it back again and check against the original
     spec = Spectrum1D.read(tmpfile, format='tabular-fits', column_mapping=cmap)
     assert spec.flux.unit == u.Unit('erg / (s cm**2 AA)')
-    assert spec.spectral_axis.unit == spectrum.spectral_axis.unit
+    assert spec.spectral_axis.unit == u.um
     assert quantity_allclose(spec.spectral_axis, spectrum.spectral_axis)
     assert quantity_allclose(spec.flux, spectrum.flux)
     assert quantity_allclose(spec.uncertainty.quantity,
@@ -566,13 +566,13 @@ def test_tabular_fits_multid(tmpdir, ndim, spectral_axis):
                              spectrum.uncertainty.quantity)
 
     # Test again, using `column_mapping` to convert to different flux unit
-    cmap = {spectral_axis: ('spectral_axis', wlu[spectral_axis]),
+    cmap = {spectral_axis: ('spectral_axis', 'THz'),
             'flux': ('flux', 'erg / (s cm**2 AA)'),
             'uncertainty': ('uncertainty', None)}
 
     spec = Spectrum1D.read(tmpfile, format='tabular-fits', column_mapping=cmap)
     assert spec.flux.unit == u.Unit('erg / (s cm**2 AA)')
-    assert spec.spectral_axis.unit == spectrum.spectral_axis.unit
+    assert spec.spectral_axis.unit == u.THz
     assert quantity_allclose(spec.spectral_axis, spectrum.spectral_axis)
     assert quantity_allclose(spec.flux, spectrum.flux)
     assert quantity_allclose(spec.uncertainty.quantity,
--- a/specutils/tests/test_loaders.py
+++ b/specutils/tests/test_loaders.py
@@ -522,6 +522,7 @@ def test_tabular_fits_writer(tmpdir, spectral_axis):
         spectrum.write(tmpfile, format='tabular-fits')
     spectrum.write(tmpfile, format='tabular-fits', overwrite=True)
 
+    # Map to alternative set of units
     cmap = {spectral_axis: ('spectral_axis', 'micron'),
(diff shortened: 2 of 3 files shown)


specutils/io/parsing_utils.py
```diff
@@ -99,18 +99,18 @@ def spectrum_from_column_mapping(table, column_mapping, wcs=None):
...
-            log.debug("Attempting auto-convert of table unit '%s' to "
-                          "user-provided unit '%s'.", tab_unit, cm_unit)
+            log.debug(f"Attempting auto-convert of table unit {tab_unit} to "
+                      f"user-provided unit {cm_unit}.")
 
             if not isinstance(cm_unit, u.Unit):
                 cm_unit = u.Unit(cm_unit)
-            if cm_unit.physical_type in ('length', 'frequency'):
+            cm_type =  str(cm_unit.physical_type)
+            if 'length in cm_type' or 'frequency' in cm_type or 'energy' in cm_type:
```
rosteen, line 108:
```suggestion
            if 'length' in cm_type or 'frequency' in cm_type or 'energy' in cm_type:
```
Looks like one of the quotes is misplaced here.

dhomeier: Thanks, yes; I should stay in Emacs+Flycheck for this stuff.

dhomeier:
I've simplified/reverted some of the comparisons since https://github.com/astropy/astropy/pull/11625#issuecomment-866972382 clarified that `PhysicalType` has better support for string operations than it appeared.

pllim:
The Windows failure is unrelated. And the remaining "allowed" failures _appear_ unrelated.

The Windows failure is a symptom of a larger problem of not marking remote data access as `# doctest: +REMOTE_DATA`. As a results, all the jobs hammer the same service around the same time.

```
035     >>> file = download_file(filename, cache=True)
UNEXPECTED EXCEPTION: URLError(gaierror(11001, 'getaddrinfo failed'))
```

dhomeier:
> The Windows failure is a symptom of a larger problem of not marking remote data access as `# doctest: +REMOTE_DATA`. As a results, all the jobs hammer the same service around the same time.

And that download has a lot of redirections I noticed when fetching the file manually; don't know if that causes additional tensions.

The other failures are due to #832, not directly related to astropy.

pllim (approved):

pllim:
I think the changes are reasonable, so I approved but I'll let the package maintainer merge.

specutils/io/parsing_utils.py
```diff
@@ -99,18 +99,17 @@ def spectrum_from_column_mapping(table, column_mapping, wcs=None):
             kwarg_val = u.Quantity(table[col_name], tab_unit)
 
             # Attempt to convert the table unit to the user-defined unit.
-            log.debug("Attempting auto-convert of table unit '%s' to "
-                          "user-provided unit '%s'.", tab_unit, cm_unit)
+            log.debug(f"Attempting auto-convert of table unit {tab_unit} to "
+                      f"user-provided unit {cm_unit}.")
```
nmearl, line 103: The logging library in python is optimized to use `%s` formatting ([ref](https://docs.python.org/3/howto/logging.html#optimization), [ref](https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles)), which is why you'll find it throughout the codebase. I realize it's not strictly enforced, so this is probably a fine change (easier to parse > optimization), but just for future reference.

specutils/tests/test_loaders.py
```diff
@@ -522,14 +522,14 @@ def test_tabular_fits_writer(tmpdir, spectral_axis):
         spectrum.write(tmpfile, format='tabular-fits')
     spectrum.write(tmpfile, format='tabular-fits', overwrite=True)
 
-    cmap = {spectral_axis: ('spectral_axis', wlu[spectral_axis]),
+    cmap = {spectral_axis: ('spectral_axis', 'micron'),
```
nmearl, line 525: Why was this changed? It seems to defeat the purpose of the parameterized test.

nmearl (changes requested):

specutils/io/parsing_utils.py
pllim: Interesting... but does this line run so much that it warrants optimization?

```python
In [14]: %timeit log.debug(f"'{x.unit}'")
3.59 µs ± 176 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [15]: %timeit log.debug("'%s'" % x.unit)
1.49 µs ± 5.85 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

dhomeier: It's actually `log.debug("'%s'", x.unit)` which is probably using `.format()` internally:
```python
In [8]: %timeit log.debug("'%s'", x.unit)
526 ns ± 8.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [9]: %timeit log.debug("'%s'".format(x.unit))
611 ns ± 11.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
Still, the speedup in this context is very modest compared to the time the actual code takes
```python
In [10]: %timeit x.to(u.nm)
10.6 µs ± 199 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [11]: %timeit x.to(u.Hz, equivalencies=u.spectral())
64.5 µs ± 2.22 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
I find the original version easy enough to parse; just needs to be known since f-strings are otherwise strongly endorsed now in astropy core.

nmearl: @dhomeier is exactly right, the logging package is built to be passed arguments which are evaluated as-needed. In the case of f-strings, interpolation happens before logging visibility criteria is evaluated, which can lead to cases where string errors are raised, nothing gets logged, and the program exits because of the automatic evaluation of `__str__`. In `%s`-style, interpolation happens within the logging package and only as needed. It will therefore be caught as logging-related issue, raises an error, and the code will continue executing.

Additionally, log aggregation (i.e. the sentry package), works because the `%s` string template is the same between similar logging statements. Obviously, we don't make prolific use of of this, nor do we experience string-related log-avoiding bugs, but there is a reason that python did not switch the package to using `{}` or f-style string interpolation with python 3.

pllim: FWIW, just because `astropy` uses it doesn't mean `specutils` has to. I don't think `astropy` uses much `logging` internally.

Though you probably want to document this somewhere in your dev docs, if not already.

specutils/tests/test_loaders.py
dhomeier: `wlu[spectral_axis]` is simply the original unit `disp` was created with, so for `spec.spectral_axis` the original test did not cover custom unit conversion by the mapping (and therefore did not find the bug of not identifying `energy` as a `physical_type` correctly).
I guess this might be moved to a separate check, but I don't see any problems with testing `spectral_axis` and `flux` units in the same run.

specutils/io/parsing_utils.py
dhomeier: No, I also took this solely as a recommendation, but still think it a good idea to synchronise the styles of coordinated packages unless there are more important considerations against it like here.

dhomeier:
Switched back to the original log() format and added some comments to explain what's done in the `custom_mapping` tests.

specutils/io/parsing_utils.py
dhomeier: > Though you probably want to document this somewhere in your dev docs, if not already.

I don't think the dev docs here include anything on code style; might be best to just link to https://docs.astropy.org/en/latest/development/codeguide.html
But I believe the performance considerations could be of value there, too, even if logging is currently used very little (found some direct usage of `logging` in the FITS scripts but hardly any of its custom `AstropyLogger`) – maybe just adding a note to the recommendation on f-strings.

pllim: I wouldn't bring `AstropyLogger` into this... 😆 There was a separate discussion, which resulted in doc update clarifying that the logger is for `astropy` internal use only.

If you want to put the doc upstream in `astropy` dev docs, I'll have to loop in @embray here since he's working on dev docs revamp over there. They would have a better idea whether this particular doc should go.

dhomeier: I had no intention to switch to `AstropyLogger` here, was just looking where `logging` is used inside astropy, if at all.
If it isn't, there's obviously not much need to mention its own formatting interface. But it probably could not hurt either...

nmearl (approved):

nmearl:
Thanks @dhomeier! Great PR.

Issue #779, opened 2021-02-06
LSRD.transform_to(ICRS) deprecated


dhomeier:
Running tests locally with astropy 4.2 gives me this error on collection:
```python
________________________________________________ ERROR collecting lib/python3.9/site-packages/specutils/tests/test_spectral_axis.py ________________________________________________
lib/python3.9/site-packages/specutils/tests/test_spectral_axis.py:71: in <module>
    LSRD.transform_to(ICRS),  # different frame
/opt/lib/python3.9/site-packages/astropy/coordinates/baseframe.py:1207: in transform_to
    warnings.warn("Transforming a frame instance to a frame class (as opposed to another "
E   astropy.utils.exceptions.AstropyDeprecationWarning: Transforming a frame instance to a frame class (as opposed to another frame instance) will not be supported in the future.  Either explicitly instantiate the target frame, or first convert the source frame instance to a `astropy.coordinates.SkyCoord` and use its `transform_to()` method.
```
I could not see how the error is getting ignored in the CI *dev job, but got rid of the warning by changing to either
`SkyCoord(LSRD).transform_to(ICRS)` or `LSRD.transform_to(ICRS())` (analogously to `LSRD_DIR_STATIONARY.transform_to(ICRS())`).

gully:
Similar warning message here in the basic usage:

```python
import specutils
import numpy as np
import astropy.units as u

spec = specutils.Spectrum1D(flux=np.random.normal(1.0, 0.01, size=500)*u.erg/u.s, 
                     spectral_axis=np.linspace(1, 1.5, num=500)*u.micron)
```

yields:

```
WARNING: AstropyDeprecationWarning: Transforming a frame instance to a frame class (as opposed to another frame instance) will not be supported in the future.  Either explicitly instantiate the target frame, or first convert the source frame instance to a `astropy.coordinates.SkyCoord` and use its `transform_to()` method. [astropy.coordinates.baseframe]
WARNING:astropy:AstropyDeprecationWarning: Transforming a frame instance to a frame class (as opposed to another frame instance) will not be supported in the future.  Either explicitly instantiate the target frame, or first convert the source frame instance to a `astropy.coordinates.SkyCoord` and use its `transform_to()` method.
```

dhomeier:
@gully which specutils and astropy versions did you see this with?
I only ever saw the DeprecationWarning in the tests, not with the above example; neither in astropy 4.3rc1 nor 5.0.dev.
[...]

PR #836, opened 2021-06-24
Fix test_spectral_axis by passing frame instances

dhomeier:
Fixes deprecation warning #779 in setup of `test_spectral_axis.py`.
Tested against astropy 4.1-4.3rc1.


eteq (approved):

eteq:
Yep definitely right.  Thanks @dhomeier !

Issue #837, opened 2021-07-01
Implement "read_many" or similar for SpectrumList


eteq:
Right now `SpectrumList` uses the same I/O scheme as all the otherloaders, where the implicit assumption is generally a *single* file is loaded into a single object.  However, this is not the case for some applications of `SpectrumList`.  A specific concrete examplesi the JWST MIRI spectrograph in MRS mode, which has as pipeline outputs 12 different 1d spectra that all have different shapes but are also all extracted from the same raw data exposure (effectively for this conversation, it's 12 separate detectors that cover different wavelength ranges).  So it seems logical to stick them into a single `SpectrumList`.

So my proposal is adding a method to `SpectrumList` along the lines of `read_many`, which would work pretty much the same as `read` but explicitly is expecting a list of file-like objects, a list of file names, or if it's a single string it's interpreted as either a directory name or a glob pattern.  Each then get individually loaded with `Spectrum1D` (which I guess `read_many` would pass ``kwargs``into), and then stuffed into the SpectrumList.

What I do *not* know is whether that runs afoul of some of the unified I/O machinery.  Someone will have to try it and see I think.  It may not even be critical to use the unified I/O machinery *at all* if we are just defering all the file-specific loading to `Spectrum1D`, but it's worth at least checking.

(note: this might eventually also be desirable for `SpectrumCollection` since one can image a pile of files that all have the same shape and wanting to put them into one `SpectrumCollection`.  But I think that's probably best implemented instead with some way to easily "collapse" a `SpectrumList` into a `SpectumCollection`, since there's not much of a performance advantage to loading them from `SpectrumCollection` if each file has to be parsed individually anyway.  At any rate, definitely a follow-on since the `SpectumList` use case is much more concrete.)

aragilar:
Does this assume that there's a loader for `Spectrum1D` (rather than only `SpectrumList`, which may be the case where there is multiple spectra in a single file)? What would be the distinction between `read_many` and joining a `SpectrumList` for each file?

kelle:
I think this is high priority.

Issue #835, opened 2021-06-23
TST: Mark remote data tests and doctests as remote data


pllim:
https://github.com/astropy/specutils/pull/833#issuecomment-867105354

This is important to avoid all the jobs hammering service providers. Only one of the jobs should do it for any given push events.

* For test functions, use `@pytest.mark.remote_data` decorators.
* For doctest, use `# doctest: +REMOTE_DATA` at the end of each affected line. Support for code blocks will be available once astropy/pytest-doctestplus#137 is merged and released.

dhomeier:
That would essentially involve adding `@pytest.mark.remote_data` to every `@remote_access` test?

Important because a number of the tests use fairly heavy downloads with plates or spectral cubes _O_(100 MB).

pllim:
What is `@remote_access`? I never seen that decorator before.

dhomeier:
It's pretty much all over the place in `test_loaders.py`. 😄
Looking at its definition
https://github.com/astropy/specutils/blob/3a86739d53663b528ee751eb55d01de3c252022d/specutils/tests/conftest.py#L16-L19
it looks actually like this should already skip them for `remote_data=False`; but what happens if the flag is not set at all (or rather, are there runs the do not set `remote_data` in any way)?

pllim:
Errr... Why duplicate code? Why not just get rid of `remote_access` and use `remote_data`?

For doctest, you cannot use `@remote_data`, you need `pytest-remotedata` _and_ `pytest-doctestplus` to get `# doctest: +REMOTE_DATA` feature.

pllim:
No, the Zenodo stuff you have to figure out, though in principle, I disagree on hiding it within a fixture like this, but I am not a maintainer here.

dhomeier:
When first working on this it might be a bit puzzling where the data come from, but otherwise I don't have any particular issues with that.
But looking e.g. at the skips in `test_loaders` it seems evident that they are in fact skipped in all but the `CI / (Allowed Failure)` run, including those marked only`@remote_access`.

pllim:
Okay, if you want to keep `remote_access`, that is fine, but someone has to go through all the doctest and mark remote data approrpiately.

pllim:
I see REMOTE_DATA marked in docs now. Though I didn't verify if you caught it all. Good enough.

PR #838, opened 2021-07-08
Load a set of _x1d files in SpectrumList.

ibusko:
Relates to discussion in #837  

I implemented the basic "read from list of files/objects", and the "read from directory" functionalities. Still missing is the "read from glob pattern" mode.


pllim (commented):
You might need to rebase and squash. I see some unrelated commits at the beginning and a lot of little commits in between.

Disclaimer: I am not familiar with specutils loader internals.

specutils/tests/test_loaders.py
```diff
@@ -1273,3 +1273,64 @@ def test_galah_guess(remote_data_path):
...
+# to this data set. We use just a few files for the current tests. Maybe
+# more files will be needed when more functionality is added later on.
+filename_storage = []
+@remote_access([
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch1-long_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch1-medium_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch1-short_x1d.fits'},
+])
+def test_loaddata_miri_mrs(remote_data_path):
+    filename_storage.append(remote_data_path)
```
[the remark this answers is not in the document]
ibusko: Indeed. But the `@remote_access` decorator doesn't seem to work with setup functions. Apparently it must be applied to `test_` functions.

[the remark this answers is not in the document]
ibusko: The "merge from astropy/main" commits were generated by GitHub itself. Maybe this is why I can't see then in my local copy? 
image

specutils/tests/test_loaders.py
```diff
@@ -1273,3 +1273,64 @@ def test_galah_guess(remote_data_path):
...
+
+    # format is explicitly set
+    specs = SpectrumList.read(filename_storage, format="JWST x1d MIRI MRS")
+
+    assert len(specs) == 3
+    for spec in specs:
+        assert isinstance(spec, Spectrum1D)
+        assert spec.spectral_axis.unit == u.Unit("um")
+
+    # auto-detect format
```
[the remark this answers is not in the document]
ibusko: Good to keep that in mind. I will leave it as is for now though, since we don't know yet how this is going to evolve. Maybe more tests will be required anyway.

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -119,6 +146,52 @@ def jwst_x1d_multi_loader(file_obj, **kwargs):
...
+            file_list = glob.glob(os.path.join(input, "*_x1d.fits"), recursive=True)
+        else:
+            file_list = glob.glob(input, recursive=True)
+
+    else:
+        file_list = input
+
+    spectra = []
+    for file_obj in file_list:
+        sp =  _jwst_x1d_loader(file_obj, **kwargs)
```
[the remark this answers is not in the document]
ibusko: The specutils loaders raise `FileNotFound` when encounter that situation. I think we should preserve the same behavior here.

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -119,6 +146,52 @@ def jwst_x1d_multi_loader(file_obj, **kwargs):
...
+          or a directory path with a glob pattern already set.
+
+    Returns
+    -------
+    SpectrumList
+        A list of the spectra that are contained in all the files.
+    """
+    # If input is a list, go read each file. If directory, glob-expand
+    # list of file names.
+    if not isinstance(input, list):
```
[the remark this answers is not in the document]
ibusko: I tried that. The code may fail then because it may try to use `os.path.isdir` with a list or tuple.

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -12,7 +15,31 @@
...
+    # file path string).
+    if isinstance(input, str):
+        if os.path.isdir(input):
+            return True
+
+        if len(glob.glob(input)) > 0:
+            return True
+
+    # or it can be either a list of file names, or a list of file objects
+    elif len(input) > 1:
```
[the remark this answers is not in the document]
ibusko: Good catch. It should be indeed `> 0`.

specutils/io/default_loaders/jwst_reader.py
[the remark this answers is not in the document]
dhomeier: The situation is slightly different here as you want to accept wildcards (or even more so, with a directory as input, when this would fail once there is any file present that is not a valid FITS spectrum, right?)
Maybe needs some more discussion how strict this should be (or add an option to fail or pass on unreadable files.

specutils/tests/test_loaders.py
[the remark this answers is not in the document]
dhomeier: For a given file there would normally only be the two cases of specifying `format` explicitly or not, seems unlikely more options could be coming in there.

specutils/tests/test_loaders.py
```diff
@@ -1273,3 +1273,64 @@ def test_galah_guess(remote_data_path):
...
+
+    assert len(specs) == 3
+    for spec in specs:
+        assert isinstance(spec, Spectrum1D)
+        assert spec.spectral_axis.unit == u.Unit("um")
+
+
+# loading from a directory via glob
+@pytest.mark.remote_data
+def test_spectrum_list_directory_miri_mrs():
```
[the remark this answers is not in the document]
ibusko: Done

specutils/io/default_loaders/jwst_reader.py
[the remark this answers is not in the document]
ibusko: @dhomeier that's why it uses a glob pattern when accessing a directory with nothing else specified.

specutils/tests/test_loaders.py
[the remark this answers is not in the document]
dhomeier: Now it apparently contains changes from #833 and possibly other unrelated PRs; looks a bit like you might have rebased this branch onto `astropy/main` instead of `git rebase `astropy/main read-many`?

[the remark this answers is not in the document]
ibusko: Possibly. I'll check on that.

specutils/io/default_loaders/jwst_reader.py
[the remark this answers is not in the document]
dhomeier: Ah, I see, so it only iterates through anything ending on `_x1d.fits`. Maybe that pattern could be customised (independently from offering the full path name pattern as `input`), but I don't know how strictly those naming conventions are enforced with JWST files.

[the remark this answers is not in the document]
ibusko: @dhomeier yes, I think there is room in here to discuss other possible choices and combinations for the input parameters that specify the files to be read. For one, the `_x1d.fits` pattern is hard-coded, but should be a keyword parameter instead (defaulting to `_x1d.fits`). Passing a path name pattern is part of the requirements, but doesn't work without more extensive work with the existing subclassing/delegation architecture of `SpectrumList`.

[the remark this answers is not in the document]
pllim: Loading one file, throwing error is fine. But what if you have 100 files, do you really want it to error out on the last invalid file after loading the other 99? Or do you want it to continue on anyway with a warning?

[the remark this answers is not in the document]
ibusko: I'll let this decision to the higher-ups. I don't know much about specutils' requirements regarding these kinds of situations.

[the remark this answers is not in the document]
dhomeier: I think this will be the first example where a single call would access many files.
Yes, best let the maintainers chime in; I'd repeat my suggestion to add a kwarg option, perhaps defaulting to `warn` with choices of "raise" and "silent".

javerbukh:
I am able to load `x1d` files into a SpectrumList object, so this looks good to me! They were not MIRI MRS files but instead nirspec level 3 data, not sure if that affects things. Also noted that `jdaviz` does not yet support loading SpectrumList objects, I'm sure that's already a ticket though. Nice work!

javerbukh (approved):
Worked for me.

dhomeier:
> Also noted that `jdaviz` does not yet support loading SpectrumList objects, I'm sure that's already a ticket though. Nice work!

https://github.com/spacetelescope/jdaviz/issues/707

rosteen:
@ibusko I'm working on reviewing this, did some rebasing locally to clean up the commit history and I think I have it looking reasonable now. Do you mind if I squash a bunch of the smaller commits to clean up the history a little further?

specutils/io/default_loaders/jwst_reader.py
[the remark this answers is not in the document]
rosteen: I'm in favor of a kwarg option as you recommended @dhomeier. I think I'm going to commit @pllim's suggestion as a starting point and then iterate off of that with some additional suggestions.

ibusko:
@rosteen please go ahead and clean it up. No worries.

rosteen:
@pllim @dhomeier I've cleaned up the commit history and added options (`silent` and `warn`) for handling missing files in the list, would you mind taking another look at this to see if it looks good to you?

pllim:
Failure is unrelated but it does block the full CI from running here, so probably needs a fix either has part of this PR or a new PR that this PR depends on. This example needs to be marked as remote data.

```
_________________________ [doctest] spectral_cube.rst __________________________
026 
027 We'll use a ``MaNGA cube`` for our example, and load the data from the
028 repository directly into a new ``Spectrum1D`` object:
029 
030 .. code-block:: python
031 
032     >>> from astropy.utils.data import download_file
033     >>> from specutils.spectra import Spectrum1D
034     >>> filename = "https://stsci.box.com/shared/static/28a88k1qfipo4yxc4p4d40v4axtlal8y.fits"
035     >>> file = download_file(filename, cache=True)
```

rosteen:
I added some remote data markers to the offending docs file.

docs/spectral_cube.rst
```diff
@@ -85,7 +85,7 @@ spectral regions from the cube.
 
     >>> import astropy.units as u
     >>> from specutils.manipulation import spectral_slab
-    >>> ss = spectral_slab(sc, 5000.*u.AA, 5003.*u.AA)
+    >>> ss = spectral_slab(sc, 5000.*u.AA, 5003.*u.AA) # doctest: +REMOTE_DATA
     >>> ss.shape  #doctest:+SKIP
```
pllim, line 89: I don't know why even set up the test with remote data if you then skip the important stuff from doctest... But that is a discussion for another day.

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -192,6 +219,67 @@ def jwst_x1d_multi_loader(file_obj, **kwargs):
     return _jwst_spec1d_loader(file_obj, extname='EXTRACT1D', **kwargs)
 
 
+@data_loader("JWST x1d MIRI MRS", identifier=identify_jwst_miri_mrs,
+             dtype=SpectrumList, extensions=['*'])
+def jwst_x1d_miri_mrs_loader(input, missing="raise", **kwargs):
+    """
+    Loader for JWST x1d MIRI MRS spectral data in FITS format
```
pllim, line 226:
```suggestion
    Loader for JWST x1d MIRI MRS spectral data in FITS format.
```

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -192,6 +219,67 @@ def jwst_x1d_multi_loader(file_obj, **kwargs):
...
+
+    A single data set consists of a bunch of _x1d files corresponding to
+    a variety of wavelength bands. This reader reads one by one and packs
```
pllim, line 229:
```suggestion
    a variety of wavelength bands. This reader reads them one by one and packs
```

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -192,6 +219,67 @@ def jwst_x1d_multi_loader(file_obj, **kwargs):
...
+    the result into a SpectrumList instance.
+
+    Parameters
+    ----------
+    input : list with str or file-like
```
pllim, line 234:
```suggestion
    input : list of str or file-like
```

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -192,6 +219,67 @@ def jwst_x1d_multi_loader(file_obj, **kwargs):
...
+        List of FITS file names, or objects (provided from name by
+        Astropy I/O Registry). Alternatively, a directory path on
+        which glob.glob runs with pattern an implicit pattern "_x1d.fits",
+        or a directory path with a glob pattern already set.
+    missing : str
```
pllim, line 239:
```suggestion
    missing : {'warn', 'silent'}
```

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -192,6 +219,67 @@ def jwst_x1d_multi_loader(file_obj, **kwargs):
...
+            file_list = glob.glob(input, recursive=True)
+    else:
+        file_list = input
+
+    spectra = []
+    for file_obj in file_list:
+        try:
+            sp =  _jwst_spec1d_loader(file_obj, **kwargs)
+        except FileNotFoundError as e:
+            if missing.lower() == "warn":
```
pllim, line 265: I don't see why we need to allow for mixed-case values. This is not FITS header.

```suggestion
            if missing == "warn":
```

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -192,6 +219,67 @@ def jwst_x1d_multi_loader(file_obj, **kwargs):
...
+                log.warning(f'Failed to load {file_obj}: {repr(e)}')
+                continue
+            elif missing.lower() == "silent":
```
pllim, line 268:
```suggestion
            elif missing == "silent":
```

specutils/io/default_loaders/jwst_reader.py
```diff
@@ -192,6 +219,67 @@ def jwst_x1d_multi_loader(file_obj, **kwargs):
...
+                continue
+            else:
+                raise FileNotFoundError(f"Failed to load {file_obj}: {repr(e)}. "
+                                        "To suppress this error, set argument missing='warn'")
+
```
pllim, line 273:
```suggestion
```

specutils/tests/test_loaders.py
```diff
@@ -1273,3 +1273,85 @@ def test_galah_guess(remote_data_path):
         assert isinstance(spectra[1].uncertainty, StdDevUncertainty)
         assert spectra[1].meta.get("label") is not None
         assert spectra[1].meta.get("header") is not None
+
+
+# We cannot use remote_access directly in the MIRI MRS tests, because
```
pllim, line 1278: I am confused. Comment says you cannot use `remote_access` directly, but the code uses it directly. What is going on here?

specutils/tests/test_loaders.py
```diff
@@ -1273,3 +1273,85 @@ def test_galah_guess(remote_data_path):
...
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch2-short_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch3-long_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch3-medium_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch3-short_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch4-long_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch4-medium_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch4-short_x1d.fits'},
+])
+def test_loaddata_miri_mrs(remote_data_path):
+    filename_list.append(remote_data_path)
```
pllim, line 1302: This is appending to a global var without declaring the var as global. Is this really working as intended?

specutils/tests/test_loaders.py
```diff
@@ -1273,3 +1273,85 @@ def test_galah_guess(remote_data_path):
...
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch2-medium_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch2-short_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch3-long_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch3-medium_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch3-short_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch4-long_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch4-medium_x1d.fits'},
+    {'id': '5082863', 'filename': 'combine_dithers_all_exposures_ch4-short_x1d.fits'},
+])
+def test_loaddata_miri_mrs(remote_data_path):
```
pllim, line 1301: What is this test supposed to do? I thought this was a fixture but looks like it is its own test function now.

If this is meant to set up the global var for other tests, why not use `setup_module`?

specutils/tests/test_loaders.py
```diff
@@ -1273,3 +1273,85 @@ def test_galah_guess(remote_data_path):
...
+    with pytest.raises(FileNotFoundError):
+        specs = SpectrumList.read(filename_list, format="JWST x1d MIRI MRS")
+
+    # Skip missing file silently
+    specs = SpectrumList.read(filename_list, format="JWST x1d MIRI MRS", missing='silent')
+
+    assert len(specs) == 12
+    for spec in specs:
+        assert isinstance(spec, Spectrum1D)
+        assert spec.spectral_axis.unit == u.Unit("um")
```
pllim, line 1319: No need for overhead to pass string into constructor.
```suggestion
        assert spec.spectral_axis.unit == u.micron
```

specutils/tests/test_loaders.py
```diff
@@ -1273,3 +1273,85 @@ def test_galah_guess(remote_data_path):
...
+
+    # Auto-detect format
+    specs = SpectrumList.read(filename_list[1:])
+
+    assert len(specs) == 12
+    for spec in specs:
+        assert isinstance(spec, Spectrum1D)
+        assert spec.spectral_axis.unit == u.Unit("um")
```
pllim, line 1337:
```suggestion
        assert spec.spectral_axis.unit == u.micron
```

specutils/tests/test_loaders.py
```diff
@@ -1273,3 +1273,85 @@ def test_galah_guess(remote_data_path):
...
+    tmp_dir = tmpdir.strpath
+    for file_path in filename_list[1:]:
+        shutil.copy(file_path, tmp_dir)
+
+    specs = SpectrumList.read(tmp_dir)
+
+    assert len(specs) == 12
+    for spec in specs:
+        assert isinstance(spec, Spectrum1D)
+        assert spec.spectral_axis.unit == u.Unit("um")
```
pllim, line 1357:
```suggestion
        assert spec.spectral_axis.unit == u.micron
```

pllim (commented):
Code seems okay but I am confused by the test setup.

docs/spectral_cube.rst
```diff
@@ -85,7 +85,7 @@ spectral regions from the cube.
 
     >>> import astropy.units as u
     >>> from specutils.manipulation import spectral_slab
-    >>> ss = spectral_slab(sc, 5000.*u.AA, 5003.*u.AA)
+    >>> ss = spectral_slab(sc, 5000.*u.AA, 5003.*u.AA) # doctest: +REMOTE_DATA
```
pllim, line 88: xref #835

dhomeier: That's in fact an interesting point – since loading `sc` has already been decorated `+REMOTE_DATA` above, what would a non-remote test actually do here if the spectrum is not available?

pllim: Any code that uses the remote data product is basically remote data and needs to be marked as such to be tested properly. If you don't add it here, you'll get exception.

specutils/io/default_loaders/jwst_reader.py
rosteen: I just figured it's the most user-friendly way to handle string matching *shrug*

specutils/tests/test_loaders.py
rosteen: @ibusko wrote these tests (I just added to them), perhaps he can shed light on why this was necessary. The answer might be that he wasn't familiar (I'm not either) with `setup_module`, in which case we can change it...

rosteen: Reading back further in the history, I imagine it's this:

> Indeed. But the `@remote_access` decorator doesn't seem to work with setup functions. Apparently it can be applied to `test_` functions only.

pllim: Can we guarantee that this test function will always run first? What if you enable parallel mode using `pytest-xdist`? 💭

pllim: After some offline discussions, @rosteen has convinced me that the tests are running as expected.

pllim (approved):

docs/spectral_cube.rst
rosteen: @dhomeier It would error because of `sc` being undefined (as I saw earlier...)

PR #839, opened 2021-07-08
Fix loader failures in remote data test

rosteen:
The `Spectrum1D.__init__` code now accounts for the possibility of `None` values in WCS physical axis types. Also, the apStar loader no longer passes the original WCS into the `Spectrum1D` it creates (that WCS was previously overwritten by the manually specified spectral axis, it doesn't make sense to keep since it doesn't match the reshaped data the loader creates for the `Spectrum1D`).


nmearl (approved):
Nice and straightforward -- thanks!

Issue #841, opened 2021-07-14
Equivalent width measurements are different from IRAF


ganesh1076:
Whenever we used specutils to measured EW its measurement is not matching with the measurement from IRAF and the measured value of EW at 5890 A from IRAF is 3.1 A and from specutils is 1.5363474 A.
What could be wrong? Are we missing something in our code?
I am attaching the normalized spectrum for the 5890 \AA and the code below.

import pandas as pd
from astropy import units as u
from specutils.spectra import Spectrum1D, SpectralRegion
from specutils.analysis import equivalent_width
import matplotlib.pyplot as plt

data=pd.read_csv('/home/ganesh/line_5890.txt',header=None)

y=u.Quantity(data[1],u.dimensionless_unscaled)
x=u.Quantity(data[0],u.angstrom)
spectrum = Spectrum1D(spectral_axis=x,flux=y)
Haregion=SpectralRegion(5860*u.angstrom, 5920*u.angstrom)
equivalent_width(spectrum,regions=Haregion)
[line_5890.txt](https://github.com/astropy/specutils/files/6818459/line_5890.txt)

dhomeier:
No error on your side really.
The function is using a simple multiplication of `continuum` with the length of the spectral axis over the boundary point (_centres_), but the actual spectral flux is calculated as in `line_flux`, which is integrating using over `spectral_axis.bin_edges`:
https://github.com/astropy/specutils/blob/1f6eded24d74ca40ec3f370b0b56573f5e314345/specutils/analysis/flux.py#L179-L182
making the continuum spectrum effectively one wavelength step shorter than the flux spectrum. I guess this was missed when `bin_edges` were introduced to represent the actual wavelength range covered by the spectrum. Changing this to
```python
    dx = (np.abs(spectral_axis.bin_edges[-1] - spectral_axis.bin_edges[0]))
```
produces EW = `<Quantity 2.79417127 Angstrom>` for your data; 3.002 Å for a narrower region [5870, 5910] Å.

@nmearl in 843ef92 `_compute_line_flux` was changed to use `bin_edges`, but `_compute_equivalent_width` was not – does this sound correct?

ganesh1076:
Thank you for the clarification, Derek!

dhomeier:
Thanks for bringing this to our attention @ganesh1076!
[...]

PR #843, opened 2021-07-25
Use spectral_axis.bin_edges to integrate continuum flux in equivalent_width

dhomeier:
#841 shows a significant underestimate of the EW from `_compute_equivalent_width` for relatively coarsely sampled spectra, which appears to be due to inconsistent integration of the spectral/line flux (using the spectral axis range over `bin_edges` in `_compute_line_flux`) and continuum (using `spectral_axis` points/bin centres).
Modifying the continuum integration to use `bin_edges` as well, decreased tolerances of the tests, which lets them fail with the old version, and added one test using the `edges` definition with coarser sampling (which would result in errors of several Å using the old calculation).

### DIFF
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -176,7 +176,7 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None,
 
     spectral_axis = calc_spectrum.spectral_axis
 
-    dx = (np.abs(spectral_axis[-1] - spectral_axis[0]))
+    dx = (np.abs(spectral_axis.bin_edges[-1] - spectral_axis.bin_edges[0]))
 
     line_flux = _compute_line_flux(spectrum, regions,
                                    mask_interpolation=mask_interpolation)
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -173,22 +173,21 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None,
             flux=spectrum.flux[~mask],
             spectral_axis=spectrum.spectral_axis[~mask])
 
-    if continuum == 1:
-        continuum = 1*calc_spectrum.flux.unit
+    # Calculate the continuum flux value
+    continuum = u.Quantity(continuum, unit=spectrum.flux.unit)
 
-    spectral_axis = calc_spectrum.spectral_axis
+    # If continuum is provided as a single scalar/quantity value, create an
+    # array the same size as the processed spectrum. Otherwise, assume that the
+    # continuum was provided as an array and use as-is.
+    if continuum.size == 1:
+        continuum = continuum * np.ones(spectrum.flux.size)
 
-    if regions is not None:
-        cont_spec = Spectrum1D(flux=continuum*np.ones(spectral_axis.shape),
-                               spectral_axis=calc_spectrum.spectral_axis)
-        cont_flux = _compute_line_flux(cont_spec, mask_interpolation=mask_interpolation)
-    else:
-        cont_flux = continuum * (np.abs(spectral_axis.bin_edges[-1] - spectral_axis.bin_edges[0]))
-
-    if mask is not None:
-        line_flux = _compute_line_flux(spectrum, regions, mask_interpolation=mask_interpolation)
-    else:
-        line_flux = _compute_line_flux(calc_spectrum, mask_interpolation=mask_interpolation)
+    cont_spec = Spectrum1D(flux=continuum, 
+                           spectral_axis=spectrum.spectral_axis)
+    cont_flux = _compute_line_flux(cont_spec,
+                                   mask_interpolation=mask_interpolation)
+    line_flux = _compute_line_flux(spectrum,
+                                   mask_interpolation=mask_interpolation)
 
     # Calculate equivalent width
     ew = (cont_flux - line_flux) / continuum
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -121,6 +121,7 @@ def _compute_line_flux(spectrum, regions=None,
         mask = calc_spectrum.mask
         new_spec = Spectrum1D(flux=calc_spectrum.flux[~mask],
                               spectral_axis=calc_spectrum.spectral_axis[~mask])
+
         if mask_interpolation is None:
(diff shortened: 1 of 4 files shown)


dhomeier:
I found it's not so straightforward to determine the correct `dx` with `regions` set, so with that option I have now switched to constructing a continuum spectrum to pass to `_compute_line_flux` in the same manner as the actual spectrum.
I was concerned about potential performance impacts as extracting `regions` is quite costly, but it seems the main cost came from doing the extraction twice (first in `extract_region` and then again inside `_compute_line_flux`).
At least for the case where no mask interpolation is applied, directly using `calc_spectrum` from the first extraction should produce the same results. While the old version took almost exactly twice as long for `equivalent_width` with `regions` as `line_flux`, the two functions are now clocking in nearly identical.

dhomeier:
The docstring has a confusing reference to a `gaussian sigma`, and I am not sure if I understand the first sentence in the `mask_interpolation` doc. I think the latter could also better clarify the meaning of the `None` option – the current description sounds like it _might_ exclude the masked regions entirely just as if they had been cut out from `regions`, but then
> the bin edges of the adjacent bins are extended to fill the gap

sounds more like the edge values (or their mean) are assumed as constant flux within the masked region; in that case it's not clear what is the difference to `LinearInterpolatedResampler`. And numerically the two seem to produce identical results, for the spectrum in `test_equivalent_width_masked`:
```python
>>> equivalent_width(spectrum_masked, mask_interpolation=LinearInterpolatedResampler)
<Quantity -719.8996202179319 um>
equivalent_width(spectrum_masked, mask_interpolation=None)
<Quantity -719.8996202179319 um>
```

nmearl:
Hmm. Good points. It seems as though the docstring may have been copied from another location [specifically the (`gaussian_sigma_width`](https://github.com/astropy/specutils/blob/main/specutils/analysis/width.py#L31) function). I see this same text in a few locations that don't make sense.

It's also not clear why the equivalent width calculation takes a `mask_interpolation` argument at all, as opposed to say a boolean `interpolate_mask` argument, since a user cannot actually change the behavior of the interpolation currently. The `line_flux` function _seems_ to be written as though one could pass a _different_ interpolator, but the `extrapolation_treatment` argument being passed in explicitly is unique to the linear interpolator...

If you have the time, I'd say go ahead and update the docstring to reflect that the regions specify where in the spectrum to calculate the equivalent width.

It also doesn't make sense that the `equivalent_width` function accepts a _list_ of `SpectralRegions`, as it's just passed to the `extract_region` function, producing a disjointed spectrum where there is more than one region defined. In that case, an equivalent width value doesn't mean much.

This might be worth a follow up issue for discussion.

specutils/analysis/flux.py
```diff
@@ -169,20 +169,29 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None,
...
+        cont_flux = _compute_line_flux(cont_spec, mask_interpolation=mask_interpolation)
+    else:
+        cont_flux = continuum * (np.abs(spectral_axis.bin_edges[-1] - spectral_axis.bin_edges[0]))
 
-    line_flux = _compute_line_flux(spectrum, regions,
-                                   mask_interpolation=mask_interpolation)
+    if mask is not None:
+        line_flux = _compute_line_flux(spectrum, regions, mask_interpolation=mask_interpolation)
+    else:
+        line_flux = _compute_line_flux(calc_spectrum, mask_interpolation=mask_interpolation)
```
nmearl, line 191: This part confuses me at the moment. `calc_spectrum` should have already dealt with regions and masks at this point, so I'm not sure why we're reverting back to using `spectrum` when it seems the line flux ought to be calculated from `calc_spectrum`. It looks as though previously, the mask wasn't being considered?

dhomeier: Frankly I did not understand how the mask was taken care of in the current version either, and have not tracked that down. In my first version I also used `calc_spectrum` throughout; this broke the test for `equivalent_width(spectrum_masked, mask_interpolation=FluxConservingResampler)`
as it returned the same EW as the `LinearInterpolatedResampler` one. However with other example data like the SDSS-1323 spectrum from the main documentation I could not find any differences between the two methods either, and in light of your comments maybe the results should indeed be the same?

dhomeier:
> It's also not clear why the equivalent width calculation takes a `mask_interpolation` argument at all, as opposed to say a boolean `interpolate_mask` argument, since a user cannot actually change the behavior of the interpolation currently. The `line_flux` function _seems_ to be written as though one could pass a _different_ interpolator, but the `extrapolation_treatment` argument being passed in explicitly is unique to the linear interpolator...

See the comment on `calc_flux`, maybe the argument value really has no effect.

> If you have the time, I'd say go ahead and update the docstring to reflect that the regions specify where in the spectrum to calculate the equivalent width.
> 
> It also doesn't make sense that the `equivalent_width` function accepts a _list_ of `SpectralRegions`, as it's just passed to the `extract_region` function, producing a disjointed spectrum where there is more than one region defined. In that case, an equivalent width value doesn't mean much.

There were two use cases I could think of –
1. effecting my original interpretation of the `mask_interpolation=None` setting, i.e. ignoring any line flux (or absorption) inside the masked region (in that case the range between two regions each covering part of the line). But obviously that would not deliver a meaningful measure of line strength in any sense.
2. specifying a set of regions to measure the co-added EW of a whole line series or even a molecular band. Cannot think of any likely application of this, but at least it is not entirely nonsensical.

nmearl:
@dhomeier your PR works great locally, and makes sense for the issue you are addressing despite the weird concerns that have popped up. I'm going to go ahead and merge this, but would you be able to open a new issue with a few bullet points highlighting some of these concerns so we can follow up with a broader discussion?

specutils/analysis/flux.py
nmearl: Hmm. The case of using `FluxConservingResampler` _should_ be different, albeit very close to the case of the `LinearInterpolatedResampler`, so I would not expect them to be the same. But it's not clear why using `calc_spectrum` vs `spectrum` broke the tests.

nmearl:
@dhomeier I've pushed some commits to your branch. These address a few of our issues with parsing the behavior of the function by simplifying a lot of the more obtuse code; it should be easier to follow and result in the same values you identified.

PR #850, opened 2021-08-17
Subclass Spectrum1D from `NDArithmeticMixin`

dhomeier:
This implements a required modification to restore arithmetic functionality that has been removed from `NDCube` in https://github.com/sunpy/ndcube/pull/457 .
Subclassing `Spectrum1D` from `~astropy.nddata.NDArithmeticMixin` here; might be up for discussion if this should be done directly in `OneDSpectrumMixin` instead.

Note that `ndcube` has removed the arithmetic operations since they are not (yet) WCS-aware; I think `specutils` at this point does not offer consistent handling of addition on different spectral axes etc. either.
E.g. one can `+` two spectra with completely different `spectral_axis` as long as their shapes are matching without having as much as a warning raised.

To elaborate a but further; recently there has been some discussion on the mailing list on recommendations for correctly co-adding spectra on different wavelength grids along the lines implemented in https://github.com/hamogu/spectrum/blob/master/spectrum/coadd.py
Obviously this has never been further integrated in Astropy, and @hamogu has not been working on it for years, but there is certainly demand for some more automatic functionality beyond what is shown in https://specutils.readthedocs.io/en/stable/manipulation.html#resampling.

### DIFF
--- a/docs/nitpick-exceptions
+++ b/docs/nitpick-exceptions
@@ -26,6 +26,7 @@ py:class ndcube.mixins.plotting.NDCubePlotMixin
 py:class ndcube.mixins.ndslicing.NDCubeSlicingMixin
 py:obj ndcube.NDCube
 py:obj NDCube
+py:obj NDCube.plotter
 py:obj ExtraCoords
 py:obj GlobalCoords
 py:obj ndcube.NDCube.world_axis_physical_types
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,7 +21,7 @@ install_requires =
     gwcs!=0.16.0
     scipy
     asdf>=2.5
-    ndcube@git+https://github.com/sunpy/ndcube.git@947799ea213728d6215cb791508ef186ca53b24f
+    ndcube@git+https://github.com/sunpy/ndcube.git@v2.0.0rc1
 
 [options.extras_require]
 test =
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -4,7 +4,7 @@ import logging
 import numpy as np
 from astropy import units as u
 from astropy.utils.decorators import lazyproperty
-from astropy.nddata import NDUncertainty
+from astropy.nddata import NDUncertainty, NDArithmeticMixin
 
 from .spectral_axis import SpectralAxis
 from .spectrum_mixin import OneDSpectrumMixin
@@ -18,7 +18,7 @@ __all__ = ['Spectrum1D']
 log = logging.getLogger(__name__)
 
 
-class Spectrum1D(OneDSpectrumMixin, NDCube):
+class Spectrum1D(OneDSpectrumMixin, NDCube, NDArithmeticMixin):
     """
     Spectrum container for 1D spectral data.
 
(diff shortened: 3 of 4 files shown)


Cadair:
Given the work on integrating reproject support into NDCube this summer, the eventual plan is to allow add etc to work using reproject to align the WCSes if they are not already aligned. If that works for specutils we would love to have contributions upstream to support this. :)

(p.s. if this upstream change really breaks stuff for you let us know)

dhomeier:
Most Linux tests are failing with
> http://azure.archive.ubuntu.com/ubuntu/pool/main/l/language-pack-de-base/language-pack-de-base_20.04+20210121_all.deb not found
which has been replaced with language-pack-de-base_20.04+20210802_all.deb, but I could not find any setup for this in our own configuration.

dhomeier:
> which has been replaced with language-pack-de-base_20.04+20210802_all.deb, but I could not find any setup for this in our own configuration.

Hidden in .github/workflows; probably need to pull in [#astropy/12068](https://github.com/astropy/astropy/pull/12068).

rosteen:
Thanks @dhomeier. I think that I'm in favor of subclassing `NDArithmeticMixin` for now in the interest of not breaking anything that currently works, while ongoing discussions of better WCS-aligned arithmetic happen. I do think that it would be better to do the subclassing in `OneDSpectrumMixin` (which is where we also subclass `NDIOMixin`).

I just opened a PR to (hopefully) address those CI failures in the same way as the astropy PR you linked.

nmearl:
Why is this moved to the mixin? Ideally, the mixin is supposed to support 1D-specific functionality, but the arithmetic operates over multi-dimensional data.

rosteen:
Well, partly because no one objected when I said I thought it belonged there 😆 . My logic was that the NDIOMixin was there, so perhaps other mixins should go there as well? If that isn't the underlying logic of the mixin class then I'm happy to move it back.

nmearl:
😁 Originally, the `NDArithmeticMixin` and `NDIOMixin` were part of the `NDDataRef` super class that `Spectrum1D` inherited from. It must have been moved to the `OneDSpectrumMixin` class when `NDCube` was adopted. However, it should exist on the `Spectrum1D`, since that would be constant regardless of a specific dimensional mixin (i.e. imagine if we had `TwoDSpectrumMixin`, `NDArithmeticMixin` and `NDIOMixin` should still exist on the core `Spectrum1D` class).

rosteen:
Gotcha. I'll go ahead and move them both over to `Spectrum1D` then. I hope you'll forgive me for not realizing/remembering that `Spectrum1D` means "any dimensionality" but `OneDSpectrumMixin` means "no but really, only 1D here." 😵 

I'll also add the nitpick exception needed to make the docs happy after moving up to 2.0.0rc1.

rosteen:
Oh, how fun:

```
E   TypeError: Cannot create a consistent method resolution
E   order (MRO) for bases NDArithmeticMixin, NDIOMixin, NDCube
```

rosteen (approved):

dhomeier:
Thanks all for clarifying the mixins structure and merging!

PR #851, opened 2021-08-19
Run update before installing language pack in CI

rosteen:
Fix CI failures (see e.g. #850), same as in https://github.com/astropy/astropy/pull/12068.


eteq (approved):
I must say I don't understand why we need the German language pack... but hey don't look to closely if it works.

PR #856, opened 2021-09-01
MIRI x1d files now load in Specviz

ibusko:
Fixes a bug that affects the loading of MIRI x1d files, but very likely other instruments' x1d files as well.

The unit test function exercises the code against x1d files from 3 pipeline versions: 1.1, 1.2.3, and 1.3

### DIFF
--- a/specutils/tests/test_loaders.py
+++ b/specutils/tests/test_loaders.py
@@ -1367,7 +1367,7 @@ def test_jwst_miri_x1d(remote_data_path):
 
     data = Spectrum1D.read(remote_data_path)
 
-    assert type(data) is Spectrum1D
+    assert isinstance(data, Spectrum1D)
     assert data.shape in [(745,), (388,), (1091,)]
     assert data.unit == u.Jy
     assert data.spectral_axis.unit == u.um
--- a/specutils/tests/test_loaders.py
+++ b/specutils/tests/test_loaders.py
@@ -1357,17 +1357,87 @@ def test_spectrum_list_directory_miri_mrs(tmpdir):
         assert spec.spectral_axis.unit == u.micron
 
 
-# loading a Spectrum1D from a MIRI x1d file
+# x1d and c1d assorted files
+
 @remote_access([
-    {'id': "5363560", 'filename':"jw00623-c1012_t002_miri_p750l_x1d.fits"},     # pipeline 1.2.3
-    {'id': "5363560", 'filename':"jw00212-o035_t007_miri_ch2-medium_x1d.fits"}, # pipeline 1.1
-    {'id': "5363560", 'filename':"jw00623-o057_t008_miri_ch1-long_x1d.fits"},   # pipeline 1.3.1
+    {'id': "5394931", 'filename':"jw00623-c1012_t002_miri_p750l_x1d.fits"},                  # pipeline 1.2.3
+    {'id': "5394931", 'filename':"jw00787-o014_s00002_niriss_f150w-gr150c-gr150r_c1d.fits"}, # pipeline 1.2.3
+    {'id': "5394931", 'filename':"jw00623-o057_t008_miri_ch1-long_x1d.fits"},                # pipeline 1.3.1
+    {'id': "5394931", 'filename':"jw00626-o064_t007_nirspec_g235h-f170lp_x1d.fits"},         # pipeline 1.3.1
 ])
-def test_jwst_miri_x1d(remote_data_path):
+def test_jwst_x1d_c1d(remote_data_path):
 
     data = Spectrum1D.read(remote_data_path)
 
     assert isinstance(data, Spectrum1D)
-    assert data.shape in [(745,), (388,), (1091,)]
+    assert data.shape in [(388,), (5,), (1091,), (3843,)]
     assert data.unit == u.Jy
     assert data.spectral_axis.unit == u.um
+
+
+# utility functions to be used with list comprehension in SpectrumList checking
+def assert_multi_isinstance(a, b):
+    assert isinstance(a, b)
+
+def assert_multi_equals(a, b):
+    assert a == b
+
+
+@remote_access([
+    {'id': "5394931", 'filename':"jw00624-o027_s00001_nircam_f356w-grismr_x1d.fits"}, # pipeline 1.2.3
+])
+def test_jwst_nircam_x1d_multi_v1_2_3(remote_data_path):
+
+    data = SpectrumList.read(remote_data_path)
+
+    assert isinstance(data, SpectrumList)
+    assert len(data) == 3
+    [assert_multi_isinstance(d, Spectrum1D) for d in data]
+    [assert_multi_equals(d.shape, r) for d,r in zip(data, [(459,), (336,), (962,)])]
+    [assert_multi_equals(d.unit, u.Jy) for d in data]
+    [assert_multi_equals(d.spectral_axis.unit, u.um) for d in data]
+
+
+@remote_access([
+    {'id': "5394931", 'filename':"jw00660-o016_s00002_nircam_f444w-grismr_x1d.fits"}, # pipeline 1.3.1
+])
+def test_jwst_nircam_x1d_multi_v1_3_1(remote_data_path):
+
+    data = SpectrumList.read(remote_data_path)
+
+    assert isinstance(data, SpectrumList)
+    assert len(data) == 4
+    [assert_multi_isinstance(d, Spectrum1D) for d in data]
+    [assert_multi_equals(d.shape, r) for d,r in zip(data, [(1166,), (786,), (1157,), (795,)])]
+    [assert_multi_equals(d.unit, u.Jy) for d in data]
+    [assert_multi_equals(d.spectral_axis.unit, u.um) for d in data]
+
+
+@remote_access([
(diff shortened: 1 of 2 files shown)


specutils/tests/test_loaders.py
```diff
@@ -1355,3 +1355,19 @@ def test_spectrum_list_directory_miri_mrs(tmpdir):
...
+@remote_access([
+    {'id': "5363560", 'filename':"jw00623-c1012_t002_miri_p750l_x1d.fits"},     # pipeline 1.2.3
+    {'id': "5363560", 'filename':"jw00212-o035_t007_miri_ch2-medium_x1d.fits"}, # pipeline 1.1
+    {'id': "5363560", 'filename':"jw00623-o057_t008_miri_ch1-long_x1d.fits"},   # pipeline 1.3.1
+])
+def test_jwst_miri_x1d(remote_data_path):
+
+    data = Spectrum1D.read(remote_data_path)
+
+    assert type(data) is Spectrum1D
```
pllim, line 1370: Is there a reason to not use `isinstance`?

```suggestion
    assert isinstance(data, Spectrum1D)
```

pllim:
There are a lot of merge commits here, is it possible to rebase?

specutils/tests/test_loaders.py
ibusko: Not really. I copied the code pattern from a number of similar tests in `test_jwst_reader.py.` Then I moved the test to `test_loaders.py` so it could benefit from the remote loading fixture that is already in place there. `test_jwst_reader.py` uses only internally-generated test data. I will change that to `isinstance`

ibusko:
The merge commits come from me updating the branch in my fork. I suppose they are already in `main`. Shouldn't they disappear when the branch gets merged back in `main`? I ask because my rebasing tool shows just the two commits I actually commited.

pllim:
> Shouldn't they disappear

I am not sure. I guess it depends on which merge button you choose to use. Over at `astropy`, we just ask people to rebase so the merge commits don't show up to confuse people.

pllim:
I guess I'll leave the question of rebase to `specutils` maintainer.

ibusko:
@havok2063 do I need a specific branch of `specutils` to read `c1d` files? I tried with one from version 1.2.3 and got that "Format could not be identified" error.

ibusko:
@havok going back to the `x1d` files: they have "ERROR" in 1.1, and "FLUX_ERROR" in 1.2.3 and 1.3.1. Should those differences be tested against as well?

havok2063:
@busko yeah the FITS changes occurred in version 1.2.3 of the pipeline.    I guess the reason for having that test would mainly be to check that `specutils` can still load data from 1.1.   Since we're still pre-launch, I don't know if we need to care too much about old pipeline files, since no one will ever see or use that data, especially after launch.   But since it's still lingering in the Portal for now I could see including that test.  I don't have a strong opinion either way on it.

pllim:
> if we need to care too much about old pipeline files

I don't think we should. How about we check for CAL_VER and throw exception if it is too old? Then it is clear to the user that they need to rerun with new pipeline.

havok2063:
Rerun the pipeline or download the latest data.   I think before launch you could probably pin a minimum version of JWST data that Jdaviz will support.   The problem with doing this **right** now is that it could cut off people who are still looking at older rehearsal data with Jdaviz through the Portal.    But I agree that you don't want Jdaviz to support defunct versions of the pipeline, especially of pre-launch data.

pllim:
> older rehearsal data

Well, they can use older version of specutils/jdaviz if they are stuck with old stuff. 🤷

havok2063:
@plim On their local systems of course, but not if they are viewing Jdaviz via the web in MAST.

havok2063:
@ibusko I'm ok removing the 1.1 file test.   I think minimally the tests should test your solution works for `x1d` and `c1d` files, and for multiple `x1d` instrument files, to test they can all successfully convert into `Spectrum1D`.   I suppose if the tests pass the column name check is implicit.    Ideally the tests would also check that `x1d multi` and `c1d multi` files can be successfully converted in `SpectrumList` objects as well.

ibusko:
@havok2063 do you have examples of `x1d multi` and `c1d multi` files? If they are in your box folder, how can I identify them?

havok2063:
@ibusko they are in the box folder.  I just open up the files and see how many `EXTRACT1D` or `COMBINE1D` extensions are in them.  I'm not sure if there is a header keyword that indicates that.  The multi-extension files are mostly `niriss` or `nircam` files, usually `EXP_TYPE=`,  `NIS_WFSS` , `NIS_SOSS`, or `NRC_WFSS`.

rosteen:
`jw00787-o014_s00033_niriss_f150w-gr150c-gr150r_x1d.fits` and `jw00787-o014_s00004_niriss_f150w-gr150c-gr150r_x1d.fits` are xid_multi files from that box folder. They both load fine with `SpectrumList.read()` but currently give an IORegistryError if you try to load them with `Spectrum1D.read`

ibusko:
@rosteen thanks for the pointers. The error I was getting is that same error you see when using `Spectrum1D.read()` on the `c1d` and multi files. It seems rigth, since we have `SpectrumList` to handle these formats. I was using  `Spectrum1D.read()`. But do you think the `SpectrumList.read()` functionality should be included in `Spectrum1D.read()`? I really don't recall conversations that we might have had already on the subject; perhaps we had then. Does it ring any bell?

rosteen:
I personally don't think we need to kick over automatically to `SpectrumList.read`, I like the custom error message that tells the user to use `SpectrumList.read` instead in these cases. But I was confused as to why that error message (from `jwst_x1d_single_loader` ) wasn't getting triggered and I instead got the format identification failure message instead. And maybe there is reason on the MAST side to automatically fall back on `SpectrumList.read` somehow...if we did do that, I would at least want a warning message displayed that the user is getting back something unexpected.

havok2063:
In our notebook for MAST, we use https://github.com/spacetelescope/jdaviz/blob/4d38952bdba2a250de7a9c18584f1d87ff6c4728/jdaviz/core/data_formats.py#L53 to identify the file format and relevant helper configuration.  It currently correctly identifies whether a file format is `JWST x1d` or `JWST x1d multi`.  The problem has been that Specviz only expects a single `Spectrum1D` object, using `Spectrum1D.read()`, there was no helper config that can load `SpectrumList` objects, and Mosviz wasn't in a place we could use it.  I know there's been some work folding in support for `SpectrumList` objects, using `SpectrumList.read()`, but I'm not sure how far along that is, and if it's in the new Mosviz?  

Ideally, `Jdaviz` should automatically determine whether a file should be loaded as a `Spectrum1D` or `SpectrumList`, using the loaders.  In the interim, we can add some logic in our notebook that looks something like 
```
# identify file format (x1d or x1d multi)
if "x1d":
   s = Specviz()
   s.load_spectrum(file)
elif "x1d multi":
  m = Mosviz()
  m.load_data(file)
```
Once there's valid support for multi extension science (x1d, c1d, and s2d), then we'd update the following dictionary with the relevant format to helper mapping.  https://github.com/spacetelescope/jdaviz/blob/4d38952bdba2a250de7a9c18584f1d87ff6c4728/jdaviz/core/data_formats.py#L13

ibusko:
I added unit test cases to test a few file formats:

- MIRI x1d - pipeline 1.2.3, 1.3.1
- NIRCAM x1d multi - pipeline 1.2.3, 1.3.1
- NIRCAM c1d - pipeline 1.2.3
- NIRISS c1d - pipeline 1.2.3 

Please let me know in case there are any combinations missing.

havok2063:
That looks pretty good.  Maybe add a single test for a NIRSPEC x1d file?  e.g. `1.3.1/jw00626-o064_t007_nirspec_g235h-f170lp_x1d.fits` in Box.  And change the NIRCAM to a `c1d multi`, e.g. `1.2.3/jw00624-o027_s00001_nircam_f356w-grismr_c1d.fits'` in Box.  Then I'd say that covers it with a case for all 4 instruments and at least one `x1d`, `c1d`, `x1d multi` and `c1d multi`.

ibusko:
@havok2063  Both `c1d` files in my list above are multi-extension. Sorry, I forgot to say that in the table. I added the NIRSPEC `x1d` file to the tests.

havok2063:
@ibusko ok cool.  If you want a single `c1d` file, you could switch the NIRISS file to `1.2.3/jw00787-o014_s00002_niriss_f150w-gr150c-gr150r_c1d.fits`.

ibusko:
I will add the NIRISS single extension file, and not remove the c1d already there.

rosteen (approved):
Looks good to me, thanks. I'm happy about the additional test coverage.

I'll see if I can clean up the merge commits before I merge.

rosteen:
I'm unclear about why the one CI test keeps cancelling, but I think this is ready to merge.

pllim:
Was someone manually hitting the cancel button? Let me try to restart.

rosteen:
I don't think so, I already restarted once.

pllim:
`os: ubuntu-16.04` might be too old. If it still self-cancel, try update it to at least Ubuntu 18.x 🤞

rosteen:
Looks like it ran this time 😕 🤷

rosteen:
Failure is an unrelated URL timeout, which (I'm pretty sure) I saw pass earlier. Merging.

PR #860, opened 2021-09-08
Set loader priorities so surveys override generic

### LINKED ISSUE #859: Correct priorities for loaders
I was going through some auto-detection tests and found that sometimes the wrong loader was used. It appears that as the default priority is 0, and the 1d-wcs loader has a priority of 5, that the 1d-wcs loader was used when a survey specific loader should have been used (the value of 5 was set in https://github.com/astropy/specutils/pull/632). I see two solutions:
1. We bump all survey specific loaders to 10 (or higher than 6)
2. We drop 1d-wcs to below 0 (e.g. -10 to give some freedom to play with ordering)

The latter solution seems simpler to me (as we only need change the 1d-wcs, and possibly the tabular-fits loaders, which has a priority of 6 currently), but if there's a reason for their higher priority, then the former solution would be better.

aragilar:
This modifies loader priorities using the following rules:
 * Leave the generic fits loader priorities alone, otherwise set generic loaders to -10
 * For survey-specific fits loaders, set to 10
 * Don't change non-fits survey loaders

Fixes #859.

### DIFF
--- a/specutils/io/default_loaders/aaomega_2df.py
+++ b/specutils/io/default_loaders/aaomega_2df.py
@@ -43,7 +43,7 @@ def identify_aaomega(origin, *args, **kwargs):
 
 @data_loader(
     label=AAOMEGA_LOADER, extensions=FITS_FILE_EXTS, dtype=SpectrumList,
-    identifier=identify_aaomega
+    identifier=identify_aaomega, priority=10,
 )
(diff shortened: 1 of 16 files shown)


nmearl (approved):

nmearl:
Thanks @aragilar!

PR #861, opened 2021-09-08
Allow force to be passed down to astropy registry

aragilar:
Astropy's registry register functions allow overriding existing readers, writers and identifiers. This allows specutils loaders and writers to use that override functionality.


nmearl:
Neat addition, thanks!

PR #865, opened 2021-09-13
Add uncertainty to JWST s3d loader

ibusko:
The loader was ignoring the ERR image extension in JWST s3d data cubes. 

This PR adds the uncertainty to the product. It is assuming that the ERR extensions are always present.


ibusko:
This is superseded by #866.

Closing.

Issue #653, opened 2020-04-10
Specutils template matching needs to handle resampling without returning nan


eblur:
I am working on a tutorial for fitting an SDSS quasar spectrum with a template and solving for redshift:
https://drive.google.com/file/d/1cOPKSH3ED_0fokcf07yPj2t2mqYwtnrP/view?usp=sharing

([Astropy Tutorials Issue #441](https://github.com/astropy/astropy-tutorials/issues/441))

Due to the nature of the template, the model has to be resampled onto the observed spectral axis in order to calculate chi^2 (see the first three lines of the `_chi_square_for_templates`). If the template axis doesn't span the observed axis over the full range of redshift values provided, then the resampling function returns 'nan', which breaks everything.

The function needs a mask to ignore regions that can't be resampled, or at least return a value other 'nan' for regions that are out-of-bounds of the template.

nmearl:
@eblur This case _should_ already be supported with the `extrapolation_treatment` argument in the resamplers. By default it is `nan_fill`, but you can use `zero_fill` instead. There's [an example](https://specutils.readthedocs.io/en/stable/manipulation.html#splicing-combining-multiple-spectra) in the docs as well.

eblur:
The `template_comparison` library doesn't provide any arguments or keywords for passing that information to the resampler.  Thus, in order to apply this library to a very simple example (a spectrum downloaded directly to the SDSS archive), the user would have to re-write much of this library.

So, I see two possible ways forward:
(1) Make zero_fill extrapolation the default for the samplers returned in the `_resampler` function, or
(2) Provide a `resample_fill` (or similarly named) keyword in the `template_match` and `template_redshift` functions that allow the user to refine the template sampling method.

I suggest (2) because I can think of some examples where (1) could cause problems.

Also `template_redshift` should have `resample_method` keyword, like the `template_match` function.

dhomeier:
Speaking of which, is there an imperative reason `resample` does not also offer some basic extrapolation option like `const_fill` (i.e. instead of setting the `off_edges` to zero, to the last valid flux value off the region)? Of course the simple solution would be quite sensitive to noise, so a better implementation should average over some region close to the edge, and at that point the options (and possibly also the temptation to abuse by extrapolating too far) may quickly get out of control.
Still it seems for points say within one original step width this could often be a better approximation than zero.

camipacifici:
[...]

Issue #652, opened 2020-04-10
Template matching tools need to support InverseVariance case for calculating chi^2


eblur:
I noticed that the `specutils.analysis.template_comparison._chi_square_for_templates` function treats the uncertainty array as the pure uncertainty value, when this is not necessarily the case with inverse variance values.

I became aware of this while working on a tutorial for fitting SDSS spectra:
https://drive.google.com/file/d/1cOPKSH3ED_0fokcf07yPj2t2mqYwtnrP/view?usp=sharing

(See also [Astropy Tutorials Issue #441](https://github.com/astropy/astropy-tutorials/issues/441))

It seems necessary to type check the `observed_spectrum.uncertainty` before deciding how to handle it.

PR #872, opened 2021-09-23
Fix bug with template_comparison method

javerbukh:
Fix bug with incorrect chi2 value being returned as the smallest chi2.

A few pieces that need to be addressed before this is taken out of draft form:

- [ ]  Using a range of redshift values vs one redshift value returns the correct redshift but different minimum chi2 values, is this expected or does this mean the method is still working incorrectly? I am using this notebook for testing: https://github.com/spacetelescope/jwebbinar_prep/blob/main/spec_analysis/jw8_redshift_fitting.ipynb .
- [ ] `template_match` and `template_redshift` return the same values but in a different order. What is the preferred return order for the return values: `redshift`, `minimum chi2`, `list of chi2 values`, `smallest chi2 index in list`, and `normalized_template_spectrum`?
- [ ] If I am returning so many values, does it make sense to pack them into a dictionary instead of a tuple? This is probably a question for @pllim 

@camipacifici can you please answer these questions for me and make sure this PR fixes the original bug? Thank you!


specutils/analysis/template_comparison.py
```diff
@@ -153,20 +151,27 @@ def template_match(observed_spectrum, spectral_templates,
...
-        normalized_spectral_template, chi2 = _chi_square_for_templates(
-            observed_spectrum, spectral_templates, resample_method)
+        else:
+            normalized_spectral_template, chi2 = _chi_square_for_templates(
+                observed_spectrum, spectral_templates, resample_method)
+
+        chi2_list.append(chi2)
 
-        return normalized_spectral_template, chi2, 0, chi2_list
+        return normalized_spectral_template, chi2, 0, chi2_list, final_redshift
```
pllim, line 174: Whether you attached new value to the end like this or use `dict`, that is `specutils` maintainer's call. It will be a breaking change either way.

pllim (commented):
I'd rather you use `# noqa` to get rid of `flake8` line too long warning than to use `\` continuation.

camipacifici:
> Fix bug with incorrect chi2 value being returned as the smallest chi2.
> 
> A few pieces that need to be addressed before this is taken out of draft form:
> 
> * [ ]   Using a range of redshift values vs one redshift value returns the correct redshift but different minimum chi2 values, is this expected or does this mean the method is still working incorrectly? I am using this notebook for testing: https://github.com/spacetelescope/jwebbinar_prep/blob/main/spec_analysis/jw8_redshift_fitting.ipynb .
> * [ ]  `template_match` and `template_redshift` return the same values but in a different order. What is the preferred return order for the return values: `redshift`, `minimum chi2`, `list of chi2 values`, `smallest chi2 index in list`, and `normalized_template_spectrum`?
> * [ ]  If I am returning so many values, does it make sense to pack them into a dictionary instead of a tuple? This is probably a question for @pllim
> 
> @camipacifici can you please answer these questions for me and make sure this PR fixes the original bug? Thank you!

1. the chi^2 should be exactly the same when the same redshift value is used. When the range of redshift is used, there might be a better match than the single redshift. I will check this.
2. I would like as order: normalized spectrum (at the correct redshift), best redshift, index of smallest chi^2 (1 value for single redshift, 2 values for range of redshifts), minimum chi^2, list of chi^2 values.
3. I leave this to you.

Testing now.

camipacifici:
The function works correctly now!

javerbukh:
@camipacifici Reordered the return values and updated tests. I did not implement the `chi2_min` value as a coordinate feature for a range of redshifts so I will work on that next.

rosteen (approved):
I think this looks good, I don't mind the additional information being returned - seems like it could be useful. 

Side note, I had some half finished work addressing #652 and #653, this is good motivation for finishing that and getting everything needed for improving template_comparison into the next release.

rosteen:
Unclear to me why the one CI job got stuck at "Starting job...", I cancelled it after 30 minutes.

pllim:
I fixed that in d13ae65da587f4acddc2fe9cd3d88445c0b9549e on `main`. Please rebase.

eteq:
Looks like this is all set, so merging now!

PR #873, opened 2021-09-28
MNT: Replace log.warn with log.warning

pllim:
`log.warn()` is deprecated.


nmearl (approved):
Good catch.

nmearl:
P.S. branch naming on point. 👌

PR #879, opened 2021-10-06
extract_region: return_single_spectrum

### LINKED ISSUE #875: Add a single_spectrum keyword to extract_region
Right now if you use `extract_region` on a multi-section spectral region, it splits the spectrum into separate distinct `Spectrum1D` objects for each sub-region. This is problematic for some workflows, for example when trying to do line fitting or similar operations on discontinuous spectral features (@camipacifici may have a more concrete example).

So I propose we add a keyword to `extract_region` that allows it to return a *single* spectrum for the multi sub-region case.  This could be just `return_single_spectrum` which would default to False but when True when trigger the discussion above.  But that's kind of a long keyword name.  Any other ideas?

kecnry:
if `return_single_spectrum` (defaults to `False`) is `True`, `extract_region` returns a single Spectrum1D object with concatenated internal arrays, with no filled-values for gaps, but removing duplicate entries in the case where the sub-regions overlap.  

An alternate approach would be to use [re-sampling](https://specutils.readthedocs.io/en/stable/manipulation.html#splicing-combining-multiple-spectra), but this seems like overkill for this case where we know that the individual Spectrum1D objects came from the same original spectrum.

This logic could eventually be split out and generalized to support splicing/concatenating any multiple Spectrum1D objects (and with options to add a fill-value in the gaps, etc), but that would require consistency checks that can be avoided here.

Closes #875

### DIFF
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -96,7 +96,7 @@ def _to_edge_pixel(subregion, spectrum):
     return left_index, right_index
 
 
-def extract_region(spectrum, region):
+def extract_region(spectrum, region, return_single_spectrum=False):
     """
     Extract a region from the input `~specutils.Spectrum1D`
     defined by the lower and upper bounds defined by the ``region``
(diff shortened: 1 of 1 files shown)


keflavich:
> I'm wondering if it makes sense to just drop the WCS here rather than copying it, since the WCS as originally defined no longer gives you the right coordinates in this concatenated spectrum. Any thoughts on that @keflavich ? 

Can we change to lookuptable WCS?  that would be preferable

Issue #869, opened 2021-09-16
Update WCS in spectral_slab output


PatrickOgle:
While spectral slab does output the correct sub-wavelength range as a new Spectrum1D, it does not appear to update the WCS to match.  This means going through a lot of trouble to update the wcs manually, like this:

import astropy.wcs as fitswcs
from specutils.manipulation import spectral_slab

my_wcs = fitswcs.WCS(header_cube)
spec1d = Spectrum1D(flux = cube*u.Jy, wcs=my_wcs)
trim=[1.57E-6*u.m, 1.62E-6*u.m]
spec1d_t = spectral_slab(spec1d, trim[0], trim[1])
cube_trimmed = spec1d_t.flux.value

#Adjust wcs to trimmed cube because spectral_slab doesn't
new_wcs = spec1d_t.wcs
new_wcs.wcs.crval[0] = trim[0].value  #Change wave crval to match trim region
print(new_wcs)

spec1d_trimmed = Spectrum1D(flux=cube_trimmed * u.Jy, wcs=new_wcs)

PR #882, opened 2021-10-13
First Pass WCS extract region bound fix

duytnguyendtn:
Addresses #869
@PatrickOgle identified a bug in `extract_regions` where the WCS wavelength reference value (CRVAL0) wasn't being updated to the new bounds of the extracted region. Implemented Patrick's solution in this PR. Still need to figure out what the appropriate solution would be for GWCS though...


pllim:
As discussed, if we want to revisit, the solution should adhere to APE 14 high-level API if possible. That way, it would naturally work for both FITS WCS and GWCS.

PR #886, opened 2021-10-18
BUG: Fix SpectralRegion crash for pixels

### LINKED ISSUE #846: SpectralRegion interprets pixel regions as a wavenumber
Tested on Windows 10, Python 3.8.5, Specutils 1.3.

The following code snippet fails:
```
# Create a spectral region from 0-10pix, centered on 5
from specutils import SpectralRegion
x = SpectralRegion.from_center(center=5*u.pix, width=10*u.pix)
```
and results in the following ValueError:
```
    192 
    193         if any(x[0] >= x[1] for x in sub_regions):
--> 194             raise ValueError('Lower bound must be strictly less than the upper bound')
    195 
    196         return True

ValueError: Lower bound must be strictly less than the upper bound
```

It appears SpectralRegion is treating pixel quantities as frequency/wavenumber (i.e. reversed order) since it's not a "length" quantity:
```
>>> from astropy import units as u
>>> u.pix.physical_type
'unknown'
```
https://github.com/astropy/specutils/blob/dca5e7ceccef7dd57efef425435aa0b51d6bb795/specutils/spectra/spectral_region.py#L81-L82

Considering pixel-specified SpectralRegions are valid:
```
>>> from specutils import SpectralRegion
>>> from astropy import units as u
>>> SpectralRegion(1*u.pix, 2*u.pix)
Spectral Region, 1 sub-regions:
  (1.0 pix, 2.0 pix) 
```
I believe this is a proper bug

pllim:
Fix #846

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ Bug Fixes
 
 - Speed up JWST s3d loader and reduce memory usage. [#874]
 
+- ``SpectralRegion`` can now handle pixels. [#886]
+
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


pllim:
`test_tabular_fits_writer` failures are unrelated.

PR #891, opened 2021-10-22
use astropy modeling units in fit_lines

kecnry:
This pull request makes use of the new support for units within astropy modeling (introduced in astropy 4.0, specutils already requires 4.1+) and in doing so removes the need for significant code that handled stripping and re-adding units internally.

Edit: apparently not all models in astropy support being fitted with units, so perhaps we can't entirely get rid of the internal stripping/re-adding for all cases.


specutils/fitting/fitmodels.py
```diff
@@ -418,8 +418,8 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(),
     window_indices = None
     if window is not None and isinstance(window, (float, int)):
         center = model.mean
-        window_indices = np.nonzero((spectrum.spectral_axis >= center-window) &
-                                    (spectrum.spectral_axis < center+window))
+        window_indices = np.nonzero((dispersion >= center-window) &
+                                    (dispersion < center+window))
```
rosteen, line 422: Good catch!

rosteen (approved):
Looks good, thanks!

Issue #889, opened 2021-10-20
Spectrum1D.with_spectral_unit does not change spectral unit and overwrites WCS


kecnry:
`Spectrum1D.with_spectral_unit` seems create a copy of the `Spectrum1D` object with meta-data storing the original unit and with a newly created WCS from the original `spectral_axis` (so would overwrite any original WCS if it had been provided), but does not actually change units.

```
import specutils
import astropy.units as u

sp = specutils.Spectrum1D(spectral_axis=[0,1,2,3]*u.cm, flux=[0.1, 0.2, 0.3, 0.4]*u.erg/u.AA/u.cm**2/u.s)

sp.spectral_axis.unit
# Angstroms

sp.wcs.unit
# Angstroms

sp_wsu = sp.with_spectral_unit(u.um)

sp_wsu.spectral_axis.unit
# Angstroms

sp_wsu.wcs.unit
# Angstroms
```

rosteen:
Closed by #1119

Issue #676, opened 2020-05-07
What is the intended and actual result of `spectrum.with_spectral_unit`?


dhomeier:
When reviewing the handling of units inside of `Spectrum1D` for #673 (to ensure things like `spectral_axis.to('GHz')` work as expected), I noticed the `with_spectral_unit` method, which also has a test in `test_spectral_axis_conversions` verifying that it functions, but not really what its function should be.
From the docstring one might expect it to return a `Spectrum1D` copy with `spectral_axis` converted to the specified unit, but from the code it becomes quickly clear that to only modifies `spectrum.wcs`. Or rather tries to:
```
>>> spectrum = Spectrum1D(flux=[26.0, 52.4] * apu.Jy, spectral_axis=[400, 500] * u.AA)
>>> newspec = spectrum.with_spectral_unit('micron')
>>> newspec.wcs
<WCS(output_frame=SpectralFrame, input_frame=CoordinateFrame, forward_transform=Model: SpectralTabular1D
N_inputs: 1
N_outputs: 1
Parameters: 
  points: (array([0, 1]),)
  lookup_table: [400. 500.] Angstrom
  method: linear
  fill_value: nan
  bounds_error: True)>
```
for a lookup-table WCS. A "true" (`CRVAL`/`CD1_1`-constructed) WCS raises
```
>>> newspec = wcs1dspec.with_spectral_unit('micron')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/derek/lib/python3.8/site-packages/specutils/spectra/spectrum_mixin.py", line 213, in with_spectral_unit
    new_wcs, new_meta = self._new_spectral_wcs(
  File "/Users/derek/lib/python3.8/site-packages/specutils/spectra/spectrum_mixin.py", line 282, in _new_spectral_wcs
    meta['original_unit'] = self.wcs.unit[0]
AttributeError: 'WCS' object has no attribute 'unit'
```
But the first use case does not seem to have any effect other than adding a
```
>>> newspec.meta
OrderedDict([('original_unit', Unit("Angstrom"))])
```
to the copied spectrum, since `_new_spectral_wcs` internally is creating the new `wcs` as
`gwcs_from_array(self.spectral_axis)`
which discards any modifications to `wcs.unit` made up to that point.

Any insights on the use case for this?

nmearl:
[...]

PR #892, opened 2021-10-22
Spectrum1D.with_spectral_units: fallback on spectral_axis unit if not available in WCS

kecnry:
for cases where the attached WCS does not have a `unit` attribute, this now allows a fallback on `spectral_axis.unit`.  This prevents the unit attribute error in #676, but does not address the actual lack of unit conversion brought up in #676 and #889.


rosteen (approved):
Looks pretty non-controversial to me, thanks.

PR #899, opened 2021-11-17
Allow for different uncertainty types in template_comparison

### LINKED ISSUE #652: Template matching tools need to support InverseVariance case for calculating chi^2

rosteen:
Closes #652 by allowing for `VarianceUncertainty` and `InverseVariance` in `template_comparison` calculations.

### DIFF
--- a/specutils/analysis/template_comparison.py
+++ b/specutils/analysis/template_comparison.py
@@ -1,4 +1,5 @@
 import numpy as np
+from astropy.nddata import StdDevUncertainty, VarianceUncertainty, InverseVariance
 
 from ..manipulation import (FluxConservingResampler,
                             LinearInterpolatedResampler,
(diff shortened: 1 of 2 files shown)


rosteen:
It's not clear to me why a doc file I didn't change is suddenly causing the `build_docs` test to fail.

keflavich (approved):

pllim:
For the record: matplotlib/matplotlib#21668

PR #900, opened 2021-11-18
[TST]: Fix build_docs test by registering options in conf file

rosteen:
Apparently matplotlib 3.5 broke the recognition of doctest option flags from pytest-doctestplus within `plot` directives in the docs, so we have to register these in `conf.py` to get the docs build to pass. Thanks to @pllim for pointing me to the solution here. See https://github.com/matplotlib/matplotlib/issues/21668.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -29,6 +29,7 @@ import os
 import sys
 import datetime
 from importlib import import_module
+import doctest
 
 try:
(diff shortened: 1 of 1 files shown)


nmearl:
Nice fix!

nmearl (approved):

PR #901, opened 2021-11-23
Update changelog and GWCS version pin for 1.5.0 release

rosteen:
Simple enough. We need to do a release with astropy 5.0 compatibility fixes.


rosteen:
The dev test failures are just remote data retrieval failures. Merging.

PR #904, opened 2021-12-01
Add collapse methods to Spectrum1D

rosteen:
Adds `mean`, `median`, `max`, `min`, and `sum` methods to `Spectrum1D` to collapse the flux array. These currently respect the mask on the input `Spectrum1D`, but don't do anything with any attached uncertainties. 

I was a bit undecided on whether to return `Spectrum1D` where it makes sense (i.e. when the collapse is done over a non-spectral axis, such that the spectral axis is unchanged) versus just returning a `Quantity` array with the collapsed flux in all instances. Opinions welcome.

Opening as a draft for now until I add test coverage.

### DIFF
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -10,7 +10,6 @@ from .spectral_axis import SpectralAxis
 from .spectrum_mixin import OneDSpectrumMixin
 from .spectral_region import SpectralRegion
 from ..utils.wcs_utils import gwcs_from_array
-from ..manipulation import extract_region
 from astropy.coordinates import SpectralCoord
 from ndcube import NDCube
 
@@ -443,7 +442,7 @@ class Spectrum1D(OneDSpectrumMixin, NDCube, NDIOMixin, NDArithmeticMixin):
         reg = SpectralRegion(start, stop)
         return extract_region(self, reg)
 
-    def collapse(self, method, axis=None, physical_type=None, spectral_region=None):
+    def collapse(self, method, axis=None, physical_type=None):
         """
         Collapse the flux array given a method. Will collapse either to a single
         value (default), over a specified numerical axis or axes if specified, or
@@ -466,9 +465,6 @@ class Spectrum1D(OneDSpectrumMixin, NDCube, NDIOMixin, NDArithmeticMixin):
             The axis or axes over which to collapse the flux array.
         physical_type : str, optional
             Either 'spectral' or 'spatial'.
-        spectral_region : :class:`~specutils.SpectralRegion`
-            Limits the operation to only collapse the flux within the specified
-            spectral region.
 
         Returns
         -------
@@ -490,13 +486,7 @@ class Spectrum1D(OneDSpectrumMixin, NDCube, NDIOMixin, NDArithmeticMixin):
             if physical_type is not None:
                 raise ValueError("physical_type must be 'spatial' or 'spectral'")
 
-        if spectral_region is not None:
-            spec_to_collapse = extract_region(self, spectral_region,
-                                              return_single_spectrum=True)
-        else:
-            spec_to_collapse = self
-
-        collapsed_flux = collapse_func[method](spec_to_collapse.flux, axis=axis)
+        collapsed_flux = collapse_func[method](self.flux, axis=axis)
 
         # Return a Spectrum1D if we collapsed over the spectral axis, a Quantity if not
         if axis in (-1, None, len(self.flux.shape)-1):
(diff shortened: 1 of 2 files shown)


specutils/spectra/spectrum1d.py
```diff
@@ -442,6 +442,81 @@ def _spectral_slice(self, item):
...
+
+        if physical_type is not None and axis is not None:
+            raise ValueError("Cannot set both axis and physical_type inputs")
+
+        if physical_type == 'spectral':
+            axis = -1
+        elif physical_type == 'spatial':
+            # generate tuple if needed for multiple spatial axes
+            axis = tuple([x for x in range(len(self.flux.shape) - 1)])
+        else:
```
nmearl, line 485: This doesn't seem to handle the case where `physical_type` is `None` and `axis` is not set? Perhaps by default, we assume that the default collapse axis is spectral.

specutils/spectra/spectrum1d.py
```diff
@@ -442,6 +442,81 @@ def _spectral_slice(self, item):
...
+            if physical_type is not None:
+                raise ValueError("physical_type must be 'spatial' or 'spectral'")
+
+        # Set masked locations to NaN for the calculation, since the `where` argument
+        # does not seem to work consistently in the numpy functions.
+        flux_to_collapse = self.flux.copy()
+        flux_to_collapse[np.where(self.mask != 0)] = np.nan
```
nmearl, line 492: Isn't the mask already a boolean array? Does simply using `flux_to_collapse[~self.mask] = np.nan` not work?

specutils/spectra/spectrum1d.py
```diff
@@ -442,6 +442,81 @@ def _spectral_slice(self, item):
...
+
+        Returns
+        -------
+        :class:`~specutils.Spectrum1D` or :class:`~astropy.units.Quantity`
+
+        """
+        collapse_func = {"mean": np.nanmean, "max": np.nanmax, "min": np.nanmin,
+                         "median": np.nanmedian, "sum": np.nansum}
+
+        if physical_type is not None and axis is not None:
```
nmearl, line 477: I'm not sure I see why you wouldn't be able to set each parameter? That is, if I want to collapse over the spectral data, but my spectral data is two-dimensional, why couldn't I use `physical_type = 'spectral'` and `axis = 1`?

nmearl:
Nice! Just a couple of comments. Also, would it be more useful if the `method` argument could take a function instead of a string, e.g. `method=np.mean`? This would allow users to provide their own collapse functions while preserving the easy `spec.mean/spec.median/etc` shortcuts.

nmearl:
I do also want to say that using the term `physical_type` is a bit overloaded in this case, since Astropy defines it a bit differently. However, I can't really think of a better term to use... 😅 Perhaps `physical_axis`?

specutils/spectra/spectrum1d.py
rosteen: I assumed that the default case with neither set is to collapse the entire thing to a single value.

specutils/spectra/spectrum1d.py
rosteen: I wanted to avoid conflicts in the laziest way possible, rather than having to check that the axis is correct given the string input if both are provided. If this is too lazy I can be more robust, but my thought was that if you're inputting the `axis` values you probably already know what axis is what, whereas a user who is a little less familiar with the data structure might find the string input easier to just get out something that makes physical sense.

rosteen:
> Also, would it be more useful if the `method` argument could take a function instead of a string, e.g. `method=np.mean`?

I like this idea, I'll make that change.


> I do also want to say that using the term `physical_type` is a bit overloaded in this case, since Astropy defines it a bit differently. However, I can't really think of a better term to use... 😅 Perhaps `physical_axis`?

Oh yeah, good catch - I totally forgot about that despite having dealt with it recently. `physical_axis` might work, I'll think about it. Thanks for the review!

rosteen:
@nmearl Thanks again for the review. I decided to absorb the `physical_type` parameter into the `axis` parameter instead of having it separate, to keep things simpler. I also opened up the possibility of the user providing a callable function for the `collapse` method, but left providing a string name as the default so users don't necessarily have to remember to use e.g. `np.nanmean` instead of `np.mean` to take an average.

specutils/spectra/spectrum1d.py
```diff
@@ -442,6 +442,84 @@ def _spectral_slice(self, item):
...
+            else:
+                raise ValueError("String axis input must be 'spatial' or 'spectral'")
+
+        # Set masked locations to NaN for the calculation, since the `where` argument
+        # does not seem to work consistently in the numpy functions.
+        flux_to_collapse = self.flux.copy()
+        flux_to_collapse[np.where(self.mask != 0)] = np.nan
+
+        # Leave open the possibility of the user providing their own method
+        if callable(method):
```
nmearl, line 494: Should this be absorbed as an `elif` in the above `if` statement? Seems that its current `else` relies on it being a string?

rosteen: The previous statement is checking `axis`, this is checking `method`.

nmearl (approved):

PR #906, opened 2021-12-09
Add collapse docs, fix bug in collapse of unmasked spectra

rosteen:
I should have put the docs in the original PR, but I thought I might be lazier than this and put it off longer. Here they are though, with the added bonus of fixing a bug in the collapse code for unmasked spectra (apparently `np.where(self.mask != 0)` actually evaluates to [0] if `mask` is `None` 😱 ).


nmearl (approved):
Looks good!

Issue #796, opened 2021-03-12
SpectralRegion requires frequencies to be in descending order


PatrickOgle:
As mentioned in a previous github issue:
[https://github.com/astropy/specutils/issues/615](#615)

The order of arguments in Spectral region shouldn't matter.  However, it still does and gives an
error for this example:

noise_region = SpectralRegion(531.*u.GHz, 559.*u.GHz)

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-81-8c6a1914651a> in <module>
      3 
      4 #noise_region = SpectralRegion(559.*u.GHz, 531.*u.GHz)
----> 5 noise_region = SpectralRegion(531.*u.GHz, 559.*u.GHz)
      6 #noise_region2 = SpectralRegion(531.*u.um, 559.*u.um)
      7 

~/miniconda3/envs/science/lib/python3.8/site-packages/specutils/spectra/spectral_region.py in __init__(self, *args)
     76         #  Check validity of the input sub regions.
     77         #
---> 78         if not self._valid():
     79             raise ValueError('SpectralRegion 2-tuple lower extent must be less than upper extent.')
     80 

~/miniconda3/envs/science/lib/python3.8/site-packages/specutils/spectra/spectral_region.py in _valid(self)
    165 
    166         if any(x[0] >= x[1] for x in sub_regions):
--> 167             raise ValueError('Lower bound must be strictly less than the upper bound')
    168 
    169         return True

ValueError: Lower bound must be strictly less than the upper bound

This is very confusing as data in frequency units are typically in ascending order.
Recommend adding code so that the order of the arguments in SpectralRegion doesn't matter.

EfremBraun:
Agreed. I'm using energy units, and it's extremely odd having to reverse all my spectral data before forming the Spectrum1D object.

Issue #615, opened 2020-03-20
Relax the requirement that SpectralRegion be in a increasing order


eteq:
To see the issue at hand, the following snippet is illustrative:
```
>>> spectral_axis = np.linspace(10, 1, 200) * u.GHz
>>> flux = np.random.randn(len(spectral_axis)) * u.Jy
>>> spec = specutils.Spectrum1D(spectral_axis=spectral_axis, flux=flux)
>>> manipulation.extract_region(spec, specutils.SpectralRegion([(3, 0)] * u.GHz))
<Spectrum1D(flux=...>
```
That works fine and extracts a spectrum around the 0-3 GHz range.  *However* the following gives an error:
```
>>> manipulation.extract_region(spec, specutils.SpectralRegion([(3, 0)] * u.GHz))
...
ValueError: Lower bound must be strictly less than the upper bound
```
That error is confusing because there's really no reason why the order should matter at all in spectral region - it's a one-dimensional segment so the user shouldn't have to care about getting the order right.  That's not just academic, either: frequency, wavelength, and energy can swap a seemingly "natural" order for the axis, so the user might use various different conventions for the ordering.  So they should be free to do that and have the `extract_region` machinery not freak out.

(Note that this has additional consequences than just `extract_region` itself because `extract_region` is used quite widely in other places where an analysis or manipulation step is applied to a region)

nmearl:
I don't understand the error. In the snippet you show, you're calling `manipulation.extract_region(spec, specutils.SpectralRegion([(3, 0)] * u.GHz))` the same way each time?

abeelen:
I am seeing a similar issue here : 
I am working with `specutils.Spectrum1D` where the `spectrum_axis` is defined in frequency, and thus, I wand to define `SpectralRegion` as  `specutils.SpectralRegion(1*u.GHz, 2*u.GHz)` which leads to the same

abeelen:
This also imply a strange behavior on the `from_center` method, `specutils.SpectralRegion.from_center(1*u.GHz, 0.1*u.GHz)` leads to the same error

abeelen:
Related to that, it seems that `specutils.extract_region` expect a wavelength ordered spectral axis

nmearl:
@abeelen Thanks for the report! In general, frequency space should be descending (to match ascending wavelength space), but the issue with `from_center` is definitely a bug.

nmearl:
[...]

PR #911, opened 2021-12-17
Allow descending spectral axes and more flexible SpectralRegion bounds

rosteen:
Addresses #796  and #615 and additionally no longer requires that spectral axes be ascending in wavelength space by fixing the previous failure in `world_to_pixel` in the case of descending spectral axes. This also removes the need for some internal unit conversions to Angstrom, which is probably a minor performance improvement.

I'm still working on some failing unit tests, so I'm opening as draft for now.

### DIFF
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -13,20 +13,16 @@ __all__ = ['extract_region', 'extract_bounding_spectral_region', 'spectral_slab'
 def _edge_value_to_pixel(edge_value, spectrum, order, side):
     spectral_axis = spectrum.spectral_axis
     if order == "ascending":
-        if side == "right":
-            if edge_value > spectral_axis[-1]:
-                return len(spectral_axis)
-        elif side == "left":
-            if edge_value < spectral_axis[0]:
-                return 0
+        if edge_value > spectral_axis[-1]:
+            return len(spectral_axis)
+        if edge_value < spectral_axis[0]:
+            return 0
 
     elif order == "descending":
-        if side == "right":
-            if edge_value < spectral_axis[-1]:
-                return len(spectral_axis)
-        elif side == "left":
-            if edge_value > spectral_axis[0]:
-                return 0
+        if edge_value < spectral_axis[-1]:
+            return len(spectral_axis)
+        if edge_value > spectral_axis[0]:
+            return 0
 
     try:
         if hasattr(spectrum.wcs, "spectral"):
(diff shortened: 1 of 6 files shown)


pllim:
Looks like CI passed. Is this ready for review?

rosteen:
Yes, I moved it from draft to ready for review this morning.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -29,69 +60,36 @@ def _to_edge_pixel(subregion, spectrum):
 
     """
     # TODO: spectral regions cannot handle strictly ascending spectral axis
-    #  values. Instead, convert to length space if axis given in a desceninding
+    #  values. Instead, convert to length space if axis given in a descending
     #  unit space (e.g. frequency). We should find a more elegant solution.
     spectral_axis = spectrum.spectral_axis
+    if spectral_axis[-1] > spectral_axis[0]:
```
eteq, line 66: I think this still leaves the limitation that the spectrum has to be *in order* even though the order now can be either way.  It's a rare, but not unheard of, case that a spectrum has overlapping or otherwise non-consecutive axes. In that case this check won't really reflect that. I wonder if we should note somewhere, maybe as a warning in the docs or similar, that spectral axes need to be monotonic?

An alternative would be to do something more complex like:
1. use `argsort` to re-sort the spectral axis
2. Do the operations we currently do on the assumed-increasing axis
3. "unsort" to get the relevant parts of the original spectrum and extract just that

But I'm not sure this is worth it if you want to get this out ASAP, so we could defer it to the future if need be.

eteq (changes requested):
I love a PR that has more deletions than additions while still adding a feature!

While this broadly looks good to me, see one of my comments for a possible more general approach (which you can either try, @rosteen, or if you think it's too much for now we can leave it as a future to-do because I know this PR is already solving by far the most important use case).

The other thing I think this needs, though, is at least one unit test that explicitly checks that the answer is right for a "backwards" spectral axis.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -29,69 +60,36 @@ def _to_edge_pixel(subregion, spectrum):
 
     """
     # TODO: spectral regions cannot handle strictly ascending spectral axis
-    #  values. Instead, convert to length space if axis given in a desceninding
+    #  values. Instead, convert to length space if axis given in a descending
```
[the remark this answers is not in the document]
rosteen: Yes, good call.

specutils/manipulation/extract_spectral_region.py
rosteen: Hm, I think there are higher priority things that I would want to address before spending time on that case. I'm leaning toward just making an issue for follow-up, but I'll take a least a few minutes and see if something simple would work. If I don't change it to handle non-monotonic axes I'll add a note to the docs.

eteq (approved):
LGTM now!  I agree the failure seems to be a red herring.

PR #912, opened 2021-12-20
BUG: Transpose m1 before transposing again

pllim:
This should fix the original error reported in spacetelescope/jdaviz#947 ? Without this patch, you see the following error when trying to calculate moment 2 in Cubeviz plugin:

`ValueError: operands could not be broadcast together with shapes (33,39,2059) (39,33,2059)`

TODO:

- [x] Is this the correct fix, @rosteen et al.?
- [x] Add change log.
- [x] Add test.

### DIFF
--- a/specutils/tests/test_analysis.py
+++ b/specutils/tests/test_analysis.py
@@ -17,7 +17,7 @@ from ..analysis import (line_flux, equivalent_width, snr, centroid,
                         snr_derived, fwzi, is_continuum_below_threshold)
 from ..fitting import find_lines_threshold
 from ..manipulation import snr_threshold, FluxConservingResampler
-from ..tests.spectral_examples import simulated_spectra
+from ..tests.spectral_examples import simulated_spectra  # noqa
 
 
 def test_line_flux():
@@ -55,6 +55,7 @@ def test_line_flux_uncertainty():
     assert quantity_allclose(lf.uncertainty, u.Quantity(0.01544415, u.Jy * u.AA))
 
 
+@pytest.mark.filterwarnings('ignore:invalid value encountered in true_divide:RuntimeWarning')
 def test_line_flux_masked():
 
     np.random.seed(42)
@@ -65,7 +66,7 @@ def test_line_flux_masked():
 
     g = models.Gaussian1D(amplitude=2000*u.mJy, mean=0.56*u.um, stddev=0.01*u.um)
     flux = g(wavelengths) + 1000 * u.mJy
-    noise = 400 * np.random.random(flux.shape)* u.mJy
+    noise = 400 * np.random.random(flux.shape) * u.mJy
     flux += noise
 
     spectrum = Spectrum1D(spectral_axis=wavelengths, flux=flux)
@@ -87,7 +88,7 @@ def test_line_flux_masked():
     assert quantity_allclose(result.value, 720.61116, atol=0.001)
 
 
-def test_line_flux_uncertainty():
+def test_line_flux_uncertainty_2():
     np.random.seed(42)
 
     spec = Spectrum1D(spectral_axis=np.arange(10) * u.AA,
@@ -122,7 +123,7 @@ def test_equivalent_width():
     assert quantity_allclose(result, expected, atol=0.0025*u.GHz)
 
 
-def test_equivalent_width_masked ():
+def test_equivalent_width_masked():
 
     np.random.seed(42)
 
@@ -132,7 +133,7 @@ def test_equivalent_width_masked ():
 
     g = models.Gaussian1D(amplitude=2000*u.mJy, mean=0.56*u.um, stddev=0.01*u.um)
     flux = g(wavelengths) + 1000 * u.mJy
-    noise = 400 * np.random.random(flux.shape)* u.mJy
+    noise = 400 * np.random.random(flux.shape) * u.mJy
     flux += noise
 
     spectrum = Spectrum1D(spectral_axis=wavelengths, flux=flux)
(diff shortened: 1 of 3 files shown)


eteq (commented):
Curse you transposes!

More seriously, this looks in concept right, but it *might* need to be `(m1.transpose(-1, m1.size[-1] )` or similar (to *not* transpose the spectral axis. but I think the following test will determine for sure:
1. create a `Spectrum1D` with 4 dimensions - say, (2,3,4,100), with the length-100 being the spectral dimension
2. compute the moment
3. Ensure that the output has shape (2,3,4)

If that passes, then I think we're all good here, and if not, just transpose until it works :wink:

pllim:
Instead of writing a new test, I try to adapt an existing one. Without this patch in `moment.py`, running the new test would error out as follows:

```
        # higher order
>       moment_2 = moment(spectrum, order=2)

specutils/tests/test_analysis.py:1029:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
specutils/analysis/moment.py:42: in moment
    return computation_wrapper(_compute_moment, spectrum, regions,
specutils/analysis/utils.py:19: in computation_wrapper
    return func(spectrum, **kwargs)
specutils/analysis/moment.py:83: in _compute_moment
    return np.sum(flux * (dispersion - m1) ** order, axis=axis) / m0
.../astropy/coordinates/spectral_quantity.py:88: in __array_ufunc__
    result = super().__array_ufunc__(function, method, *inputs, **kwargs)
...
        # Call our superclass's __array_ufunc__
>       result = super().__array_ufunc__(function, method, *arrays, **kwargs)
E       ValueError: operands could not be broadcast together with shapes (9,10,10000) (10,9,10000)

.../astropy/units/quantity.py:614: ValueError
```

specutils/tests/test_analysis.py
```diff
@@ -87,7 +88,7 @@ def test_line_flux_masked():
     assert quantity_allclose(result.value, 720.61116, atol=0.001)
 
 
-def test_line_flux_uncertainty():
+def test_line_flux_uncertainty_2():
```
pllim, line 91: Bonus: Fixed duplicate function name.

p.s. As you can see, I also couldn't help myself with some PEP 8...

specutils/tests/test_analysis.py
```diff
@@ -996,16 +997,16 @@ def test_moment_cube():
     noise = np.random.normal(0., 0.01, frequencies.shape) * u.Jy
     flux = g(frequencies) + noise
 
-    # use identical arrays in each spaxel. The purpose here is not to
+    # do not use identical arrays in each spaxel. The purpose here is not to
     # check accuracy (already tested elsewhere), but dimensionality.
 
-    flux_multid = np.broadcast_to(flux, [10,10,flux.shape[0]]) * u.Jy
+    flux_multid = np.broadcast_to(flux, [9, 10, flux.shape[0]]) * u.Jy
```
pllim, line 1003: This is the modified test case.

specutils/tests/test_analysis.py
```diff
@@ -1020,27 +1021,17 @@ def test_moment_cube():
...
     # higher order
     moment_2 = moment(spectrum, order=2)
 
-    assert moment_2.shape == (10,10)
+    assert moment_2.shape == (9, 10)
     assert moment_2.unit.is_equivalent(u.GHz**2)
     assert quantity_allclose(moment_2, 816.648*u.GHz**2, atol=0.01*u.GHz**2)
 
-    # spatial higher order (what's the meaning of this?)
-    moment_2 = moment(spectrum, order=2, axis=1)
```
pllim, line 1035: This no longer works with modified test case, some failure broadcasting shapes. And since the author does not know what is the meaning of this usage, might as well remove it?

pllim:
@eteq , are my test stuff correct? Please re-review.

pllim:
Remote data job failure is unrelated.

specutils/tests/test_analysis.py
eteq: I don't understand this particular line.  Why not just start with a shape of ``(9,10)`` instead of doing this broadcasting?

specutils/tests/test_analysis.py
eteq: This test case definitely needs to work!  I think the author is saying it's not clear this has *physical* meaning, but I do have one: a collection of spectra from one instrument (e.g. a fiber spectrograph or some IFUs that are "slicers") have this same shape (nfiber, nspatial, nspectral), and sometimes one wants to know, say, where a trace is in the spatial axes.  So that's the order=1, axis=1 case.

More to the point, if this isn't working that means something's not quite broadcasting right in the implementation... So I think this case needs to be addressed (I can't vouch for the *values*, although I assume whoever wrote this checked them.  But I think the shape check should still be there, just that it should be (9, 10000) in the changed shape

specutils/tests/test_analysis.py
pllim: You are using numpy random generator with seed above. I am afraid I will change more than I intended to.

specutils/tests/test_analysis.py
pllim: I don't think this test case of `order=2, axis=1` ever worked when your spatial-spatial slice is not a square, regardless of my fix or not. In fact, the logic of this test case does not even go through my fix because of the `axis` check.

eteq: Ooh, gotcha.  yes, then this can merge then now that you point out it's definitely unrelated to this fix.  But in definitely *should* work for non-square axes!  So I'll make an issue to track that down.

eteq (approved):

pllim:
@rosteen , I might have messed up. This test case I added here didn't take account of the many times specutils swapped axes when we load JWST data into Cubeviz. Now when I use this patch and calculate moment 2 for a JWST cube, the result is rotated. I should have check. I am sorry!

If we can sort this out quickly tomorrow before release, then great. If not, you might need to revert this PR before release.

PR #913, opened 2021-12-29
MNT: Stop using catch_warnings and other updates

pllim:
MNT: Stop using catch_warnings; also see astropy/astropy#12633

MNT: Misc PEP 8 fixes

DOC: Update RTD config and delete unneeded files


pllim:
Eh, why is the RTD PR builder not set up here? Someone with RTD admin access need to do this: https://docs.readthedocs.io/en/stable/pull-requests.html

pllim:
This should be uncontroversial. I did similar patches to many other projects already. Since no one wants to review, I am just gonna YOLO.

PR #918, opened 2022-02-08
Fix slicing to actually drop and recreate WCS as intended

rosteen:
From the comments in the code, it looks like the intent here was always to create a new lookup table SpectralGWCS when slicing, but the `copy` method defaults to `self.wcs` and thus sliced spectra were ending up with the original spectrum's WCS. This doesn't solve the fact that spatial WCS components may end up incorrect if the spatial dimensions are sliced, but it at least fixes the spectral part. We may be able to fall back more on the NDCube machinery for spatial slicing than we are now, I'll open a follow-up issue for that. I'll also add a test for this case shortly.

### DIFF
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -369,11 +369,12 @@ class Spectrum1D(OneDSpectrumMixin, NDCube, NDIOMixin, NDArithmeticMixin):
                 spectral_axis=self.spectral_axis[spec_item],
                 uncertainty=self.uncertainty[item]
                 if self.uncertainty is not None else None,
-                mask=self.mask[item] if self.mask is not None else None)
+                mask=self.mask[item] if self.mask is not None else None,
+                wcs=None)
 
         if not isinstance(item, slice):
             if isinstance(item, u.Quantity):
-                raise ValueError("Indexing on a single spectral axis values is not"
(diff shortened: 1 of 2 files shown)


rosteen:
See https://github.com/astropy/specutils/issues/919 for followup to do with spatial WCS axes.

keflavich:
lgtm but I'm not giving formal approval b/c of my persistent unfamiliarity w/the code base.  Agree, though, copies should have copies of WCS!

eteq (approved):
I think part of the thinking here is that the original `wcs` has some context information that should get copied over. That said, I totally agree this is an improvement, because the current behavior is way more confusing.

I wonder, though, if the previous wcs should remain, say, in `meta` with a name like `'previous_wcs'` or something? I could be argued either way, though, as I can see good or bad sides.  Happy to merge this once you've decided either way on this, @rosteen .

rosteen:
@eteq I decided to take your suggestion and retain it in `meta` as `original_wcs`, mainly for the case where someone might override their original spectrum with the slice and thus lose access to the original WCS.

eteq:
Sounds good. I have a minor niggle that this should be documented more clearly somewhere, but... since this wasn't documented *before* I'm thinking to merge this and make that better in #919.

PR #922, opened 2022-02-10
Remove logging, handle test warnings, etc

### LINKED ISSUE #315: Change the data used for tests to not create a bunch of warnings
Currently the specutils tests yield a bunch of warnings of the sort like this:
```
/.../astropy/units/format/generic.py:474: UnitsWarning: 'erg/cm2/s/A' contains multiple slashes, which is discouraged by the FITS standard
    core.UnitsWarning)
```
this is not desirable because it hides "real" errors/warnings and makes it harder to parse what's going on with the tests.

I think the real fix is not to update the test data to be something that actually doesn't yield warnings... but if that's not possible we could just supress the warnings in those tests using either the standard python warning machinery or some of the pytest warning-catching tools.

cc @drdavella

### LINKED ISSUE #397: Fix all warnings in the test suite and turn warnings to errors
Currently the test suite yields quite a lot of warnings. See e.g., https://travis-ci.org/astropy/specutils/jobs/457642734 .  Most if not all of these are avoidable and are indeed signs that we shouldn't be doing something we're doing.  We should fix whatever is causing those warnings at a minimum.  

As an additional bit, maybe we should turn on the pytest warnings-to-errors feature?  That would keep these from creeping in in the future...

### LINKED ISSUE #920: Cannot silence warning downstream because you use logger (replace logger with warnings?)
https://github.com/astropy/specutils/blob/12ff269c7a2a2a9b57ed860327a0f01220a8157d/specutils/spectra/spectrum1d.py#L18

The warning is driving me crazy and I cannot use `warnings` to suppress it.

I can suppress it like this but it seems very heavy-handed because that affects everything for the whole session.

```python
import specutils
specutils.spectra.spectrum1d.log.setLevel(40)
```

pllim:
This PR accomplishes the following:

* Fix #920 
* Fix #397, close #315
* Move pytest fixture from `tests/spectral_examples.py` to `conftest.py`
* Remove unused imports and check for that in `tox.ini`
* Fix broken `__all__` and check for that in `tox.ini`
* Move `flake8` setting from `tox.ini` to `setup.cfg` for consistent behavior between IDE and CI. Change list from select to ignore. Fix some.
* Enable `xfail_strict=true` in testing.
* Other minor clean-ups

xref spacetelescope/jdaviz#1084


specutils/__init__.py
```diff
@@ -10,25 +10,11 @@
...
-import sys
-
-__minimum_python_version__ = "3.5"
-
-
-class UnsupportedPythonError(Exception):
-    pass
-
-
-if sys.version_info < tuple((int(val) for val in __minimum_python_version__.split('.'))):
```
pllim, line 24: You don't need this. This is now defined in `setup.cfg`.

specutils/io/default_loaders/subaru_pfs_spec.py
```diff
@@ -16,7 +15,7 @@
 from ..registers import data_loader
 from ..parsing_utils import _fits_identify_by_name, read_fileobj_or_hdulist
 
-__all__ = ['spec_identify', 'spec_loader']
+__all__ = ['identify_pfs_spec', 'pfs_spec_loader']
```
pllim, line 18: Not sure how this ever worked. This is a real bug.

specutils/tests/test_analysis.py
```diff
@@ -87,19 +86,6 @@ def test_line_flux_masked():
     assert quantity_allclose(result.value, 720.61116, atol=0.001)
 
 
-def test_line_flux_uncertainty():
```
pllim, line 90: This is a duplicate test.

rosteen (approved):
Thanks, I was getting tired of `logging` in different packages interfering with each other in weird ways, and I'm fine with a `verbose` flag for printing extra information. I appreciate the extra cleanup as well.

PR #929, opened 2022-02-18
Reverts PR 912 due to rotated result in JWST data

pllim:
Reverts actual fix portion of #912 (BUG: Transpose m1 before transposing again)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,8 +7,6 @@ New Features
 Bug Fixes
 ^^^^^^^^^
 
-- Fixed cube transpose bug in moment analysis for non-square spatial arrays. [#912]
-
 - Fixed ``spectral_slab`` crashing when ``spectral_axis`` has unit of pixels and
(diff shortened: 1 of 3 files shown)


pllim:
Hold on, don't merge yet. I am getting more confused by the minute.

pllim:
p.s. Neither fixed the rotated JWST data problem!

eteq (approved):
LGTM. 

For broader context, after some out-of-band discussion @pllim @rosteen and I realized not only are we not confident in the right answer here, we're not even sure it makes sense to *have* the axis keyword. But it would be good to get a release of `specutils` out soon.  So in lieu of making a fix that we then have to revert, we're leaving this as-is for now, to be more appropriately fixed in the next specutils version.

rosteen:
FWIW I'm pretty sure that the "with 912" answer is correct, and that the JWST data rotation lies somewhere in the translation between `specutils` and the `jdaviz` backend, but I'm happy to spend some more time thinking about this after the next bugfix release.

PR #934, opened 2022-02-23
Fix formatting in analysis docs.

pjs902:
An extra ":" on line 99 was messing up the formatting of the next code block.


eteq:
Just to provide clarity here: in sphinx docs, either `::` *or* ``.. codeblock: python`` can be used to indicate "treat this paragraph as a code block".  I'm guessing someone accidentally did both at the same time here.

But regardless, thanks for the fix @pjs902 !

PR #937, opened 2022-03-22
Some updates to specutils.analysis functions

chloe-mt-cheng:
- Updated `_normalize_for_template_matching()`: changed `np.sum` to `np.nansum` on lines 60 and 61.  Due to resampling of the spectra (resampling methods fill arrays with NaNs to have points beyond the edges set to NaN), all redshifts compared in `template_redshift()` were producing a `chi2` of NaN (even for a test case using the correct redshift).
- Updated `_chi_square_for_templates()`: changed `np.sum` to `np.nansum` on line 137 for the same reason as above.
- Updated `template_redshift()`: Added `final_spectrum = redshifted_spectrum` on line 292 and added `final_spectrum` to return instead of `redshifted spectrum`.  This is important specifically for if you feed a list of redshifts - the function was not returning the spectrum redshifted by the best-fitting redshift (i.e. the one corresponding to `chi2_min`), but rather was returning the spectrum redshifted by the last redshift in the list (not necessarily the best-fitting one).

### DIFF
--- a/specutils/analysis/template_comparison.py
+++ b/specutils/analysis/template_comparison.py
@@ -31,7 +31,10 @@ def _normalize_for_template_matching(observed_spectrum, template_spectrum, stdde
     if stddev is None:
         stddev = observed_spectrum.uncertainty.represent_as(StdDevUncertainty).quantity
     num = np.nansum((observed_spectrum.flux*template_spectrum.flux) / (stddev**2))
-    denom = np.nansum((template_spectrum.flux / stddev)**2)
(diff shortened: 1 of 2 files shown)


rosteen:
@chloe-mt-cheng Sorry that I dropped the ball on this for so long! I updated the tests after confirming that the differences make sense, and made one more change to ensure that the same number of bins get summed for the numerator and denominator in the nansums.

rosteen (approved):

PR #938, opened 2022-03-23
centroid: implement propagated uncertainties

kecnry:
* follows same pattern as in `line_flux` of setting an uncertainty attribute on the returned quantity if the input `Spectrum1D` has assigned uncertainties.
* implements a private `uncertainty._convert_uncertainty` function to convert the `Spectrum1D.uncertainty` to any assumed type.  This is now used in `line_flux` as well.
* hardcodes dispersion uncertainty to zero for now, but keeps the terms in the equations in case we want to support passing/storing dispersion uncertainties in the future.  Alternatively, the equations would simplify somewhat significantly if we drop support for uncertainties on dispersion entirely.

### DIFF
--- a/specutils/analysis/location.py
+++ b/specutils/analysis/location.py
@@ -4,9 +4,11 @@ spectral features.
 """
 
 import numpy as np
+from astropy.nddata import StdDevUncertainty
 
 from ..spectra import SpectralRegion
 from ..manipulation import extract_region
+from .uncertainty import _convert_uncertainty
 
 
 __all__ = ['centroid']
@@ -19,7 +21,9 @@ def centroid(spectrum, region):
     Parameters
     ----------
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
-        The spectrum object overwhich the centroid will be calculated.
+        The spectrum object overwhich the centroid will be calculated.  If the uncertainty
+        is populated, the returned quantity will include an uncertainty attribute with
+        the propagated uncertainty (in stddev).
 
     region: `~specutils.utils.SpectralRegion` or list of `~specutils.utils.SpectralRegion`
         Region within the spectrum to calculate the centroid.
@@ -53,8 +57,9 @@ def centroid(spectrum, region):
 
 def _centroid_single_region(spectrum, region=None):
     """
-    Calculate the centroid of the spectrum based on the flux and uncertainty
-    in the spectrum.
+    Calculate the centroid of the spectrum based on the flux in the spectrum.
+    The returned quantity object will have an uncertainty attribute which
+    will be populated if ``spectrum`` has uncertainties assigned.
 
     Parameters
     ----------
@@ -80,16 +85,49 @@ def _centroid_single_region(spectrum, region=None):
     else:
         calc_spectrum = spectrum
 
+    if spectrum.uncertainty is not None:
+        flux_uncert = _convert_uncertainty(calc_spectrum.uncertainty, StdDevUncertainty)
+    else:
+        # dummy value for uncertainties to avoid extra if-statements when applying mask
+        flux_uncert = np.zeros_like(spectrum.flux)
+
     if hasattr(spectrum, 'mask') and spectrum.mask is not None:
         flux = calc_spectrum.flux[~calc_spectrum.mask]
         dispersion = calc_spectrum.spectral_axis[~calc_spectrum.mask].quantity
+        flux_uncert = flux_uncert[~calc_spectrum.mask]
     else:
         flux = calc_spectrum.flux
         dispersion = calc_spectrum.spectral_axis.quantity
 
-    if len(flux.shape) > 1:
+    if len(flux.shape) == 2:
         dispersion = np.tile(dispersion, [flux.shape[0], 1])
+    elif len(flux.shape) > 2:
+        raise ValueError("spectrum must be 1D or 2D")
+
+    # centroid is the flux-weighted mean of the dispersions, the uncertainties
+    # need to be scaled to the numerator/denominator, so we'll compute those in advance.
 
     # the axis=-1 will enable this to run on single-dispersion, single-flux
     # and single-dispersion, multiple-flux
-    return np.sum(flux * dispersion, axis=-1) / np.sum(flux, axis=-1)
+    numerator = np.sum(flux*dispersion, axis=-1)
+    denom = np.sum(flux, axis=-1)
+    centroid = numerator/denom
+
+    if spectrum.uncertainty is not None:
+        disp_uncert = 0.0 * dispersion.unit
+
+        # uncertainty for each flux*dispersion term, frationally added in quadrature
+        num_term_uncerts = flux*dispersion * np.sqrt((flux_uncert/flux)**2 +
+                                                     (disp_uncert/dispersion)**2)
+        # uncertainty for the numerator sum, added in quadrature
+        num_uncert = np.sqrt(np.sum(num_term_uncerts**2, axis=-1))
+        # uncertainty for the denom sum, added in quadrature
+        denom_uncert = np.sqrt(np.sum(flux_uncert**2, axis=-1))
+
+        # centroid uncertainty, fractional added in quadrature of numerator and denom
+        centroid.uncertainty = numerator/denom * np.sqrt((num_uncert/numerator)**2 +
+                                                         (denom_uncert/denom)**2)
+    else:
+        centroid.uncertainty = None
+
+    return centroid
--- a/specutils/analysis/location.py
+++ b/specutils/analysis/location.py
@@ -21,9 +21,9 @@ def centroid(spectrum, region):
     Parameters
     ----------
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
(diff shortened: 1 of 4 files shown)


kecnry:
CI failure is unrelated and already a known issue in astropy.

specutils/tests/test_analysis.py
```diff
@@ -467,6 +467,7 @@ def test_centroid(simulated_spectra):
 
     assert isinstance(spec_centroid, u.Quantity)
     assert np.allclose(spec_centroid.value, spec_centroid_expected.value)
+    assert hasattr(spec_centroid, 'uncertainty')
```
rosteen, line 471:
```suggestion
quantity_allclose(spec_centroid.uncertainty, 4.4585e-06*u.um, rtol = 5e-5)
```

rosteen (commented):
Looks good, thanks for the addition. I made one suggestion to check the actual value of uncertainty as well as the existence of the attribute in the test. If you want to accept that I'll go ahead and approve/merge.

specutils/tests/test_analysis.py
rosteen: Hold on, that actual fails...what did I mess up... 😆

kecnry:
I'll also just add the same comment I did in #939 - `line_flux` uses (and seems to return) variance, but `centroid` here returns standard deviation.  We should probably be consistent (and document) and/or wrap the uncertainty in the `StdDevUncertainty` class (but then that feels even worse to be hacked on top of the quantity object)... but which do we choose?

I think at some point attaching the uncertainty to the quantity object should be revisited, as it also is easily dropped when converting units, but that's well outside the scope of these two PRs.

specutils/tests/test_analysis.py
kecnry: Maybe add a comment that the actual value has not been properly vetted yet (the math hasn't been reviewed, afaik).  But putting something in to ensure it doesn't accidentally change is probably still a good idea.

specutils/analysis/uncertainty.py
```diff
@@ -186,3 +187,26 @@ def _snr_derived(spectrum, region=None):
         return signal / noise
     else:
         return 0.0
+
+
+def _convert_uncertainty(uncertainty, to_class):
```
eteq, line 192: This should really be in `astropy.nddata`.  And hopefully will be in 5.1 with stropy/astropy#12057 !  We should change this to use that if the version is > 5.0.x, and fall back on this implementation (or a simple copy-paste of 12057 into here?)  if not.  That way we won't be forced to require Astropy 5.1 which isn't even out yet.

specutils/analysis/location.py
```diff
@@ -80,16 +85,49 @@ def _centroid_single_region(spectrum, region=None):
...
+        flux_uncert = flux_uncert[~calc_spectrum.mask]
     else:
         flux = calc_spectrum.flux
         dispersion = calc_spectrum.spectral_axis.quantity
 
-    if len(flux.shape) > 1:
+    if len(flux.shape) == 2:
         dispersion = np.tile(dispersion, [flux.shape[0], 1])
+    elif len(flux.shape) > 2:
+        raise ValueError("spectrum must be 1D or 2D")
```
eteq, line 105: I'm a bit confused on this.  Shouldn't it be OK for it to be *any* dimension, and the centroiding is just applied only along the spectral dimension?

specutils/analysis/location.py
```diff
@@ -19,7 +21,9 @@ def centroid(spectrum, region):
     Parameters
     ----------
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
-        The spectrum object overwhich the centroid will be calculated.
+        The spectrum object overwhich the centroid will be calculated.  If the uncertainty
```
eteq, line 24:
```suggestion
        The spectrum object over which the centroid will be calculated.  If the uncertainty
```
(minor pre-existing typo)

specutils/analysis/location.py
```diff
@@ -19,7 +21,9 @@ def centroid(spectrum, region):
...
+        is populated, the returned quantity will include an uncertainty attribute with
+        the propagated uncertainty (in stddev).
```
eteq, line 26:
```suggestion
        the propagated uncertainty (as Standard Deviation-style uncertainties).
```

specutils/analysis/location.py
```diff
@@ -53,8 +57,9 @@ def centroid(spectrum, region):
 
 def _centroid_single_region(spectrum, region=None):
     """
-    Calculate the centroid of the spectrum based on the flux and uncertainty
-    in the spectrum.
+    Calculate the centroid of the spectrum based on the flux in the spectrum.
+    The returned quantity object will have an uncertainty attribute which
```
eteq, line 61:
```suggestion
    The returned quantity object will have a ``.uncertainty`` attribute which
```

specutils/analysis/location.py
```diff
@@ -53,8 +57,9 @@ def centroid(spectrum, region):
...
+    will be populated if ``spectrum`` has uncertainties assigned.
```
eteq, line 62:
```suggestion
    will be populated if ``spectrum`` has uncertainties assigned, or ``None`` if not.
```

eteq (changes requested):
Mostly some minor suggestions, overall this looks generally good.

There is one comment about the "must be 1d or 2d" that is concerning, but that might be a pre-existing limitation that shouldn't be solved in this PR (I don't know what the current behavior is in >2D!).

On the broader topic of how to encode the uncertainties: I think the ship has sailed somewhat in that line_flux is following the `.uncertainty` -> stddev-style uncertainties. I think that's a fine starting point, as it is infinitely better than no uncertainties and a "better" way does not yet exist.

Going forward I think we may want to encode these instead either as 
1. StdDevUncertainty objects, with a deprecation period on the "old" way of just passing stddev style uncertainties without a wrapper object, or
2. use `astropy.uncertainties`.  Unfrotunately that does not yet have a clean way to handle stddev/var/ivar-style uncertainties, though, so that's probably a project for the future.

Regardless, though, I think it would be good to add an `uncertainty_type` attribute which is set to `std`.  That's how uncertainty objects encode it, and is sort of a soft duck-typing that might give us a fighting chance of following option 1 without a ton of pain.

specutils/analysis/location.py
kecnry: probably, but then we need an additional argument to control which index is the spectral dimension.  That is probably out of scope here... but you could also argue that the new `ValueError` is out of scope, so I'm happy to remove that and/or open a new ticket to support any number of dimensions.

specutils/analysis/location.py
```diff
@@ -80,16 +85,49 @@ def _centroid_single_region(spectrum, region=None):
...
+        num_term_uncerts = flux*dispersion * np.sqrt((flux_uncert/flux)**2 +
+                                                     (disp_uncert/dispersion)**2)
+        # uncertainty for the numerator sum, added in quadrature
+        num_uncert = np.sqrt(np.sum(num_term_uncerts**2, axis=-1))
+        # uncertainty for the denom sum, added in quadrature
+        denom_uncert = np.sqrt(np.sum(flux_uncert**2, axis=-1))
+
+        # centroid uncertainty, fractional added in quadrature of numerator and denom
+        centroid.uncertainty = numerator/denom * np.sqrt((num_uncert/numerator)**2 +
+                                                         (denom_uncert/denom)**2)
```
eteq, line 129: This should give the same answer, but eliminates two `np.sqrt` operations (which in my experience are usually the permformance limitations)
```suggestion
        num_uncertsq = np.sum(num_term_uncerts**2, axis=-1)
        # uncertainty for the denom sum, added in quadrature
        denom_uncertsq = np.sum(flux_uncert**2, axis=-1)

        # centroid uncertainty, fractional added in quadrature of numerator and denom
        centroid.uncertainty = numerator/denom * np.sqrt(num_uncertsq * numerator**-2 +
                                                         (denom_uncertsq * denom)**-2)
```

eteq (approved):
I independently derived this just to check the propogation - if you want to check my math here it is:


I made the slightly different (simpler) assumption of no uncertainty on the dispersion, but it comes to the same result, so huzzah! All is good here.  I did leave one inline suggestion for a minor performance optimization, though, which you can take or leave as you wish.

Two other large-scale thoughts:
1. this derivation is only correct if the pixel uncertainties are uncorrelated.  That's fine, because the correlated case is way more complicated anyway, and I don't think we need to say we'll support that right now.  But we should just note in the docs somewhere that the uncertainty propogation assumes all pixel uncertainties are un-correlated.
2. I agree we can leave the `uncertainty_type` change to a follow-on PR.

So I'm approving with the proviso that @kecnry looks at my suggestion and accepts it or not as you like, and at least consider putting something in the docstrings about the uncorrelated assuption.

rosteen:
Dev failure is also present on main (need to take a look at that...) so I'm merging.

specutils/analysis/uncertainty.py
rosteen: @eteq I'm revisiting this now - are there good reasons to keep `_convert_uncertainty` rather than just bumping the `astropy` requirement to 5.1 in the next `specutils` release and using the `represent_as` machinery now available on the uncertainty objects?

PR #939, opened 2022-03-31
equivalent_width: propagate line_flux uncertainty

kecnry:
Propagates the already assigned uncertainties from `line_flux` through the equations used to compute `equivalent_width`.  We likely need to think about the type of uncertainty being returned though, ~as `line_flux` uses (and seems to return) variance, but centroid was implemented in #938 to use standard deviation~ (edit: see https://github.com/astropy/specutils/pull/939#discussion_r856450839).

This follows the same pattern as elsewhere to assign the uncertainty as an attribute on the returned quantity object, but it should be noted that these will be dropped when manipulating the quantity externally (unit conversion, for example), so may not be an ideal pattern.

### DIFF
--- a/specutils/analysis/flux.py
+++ b/specutils/analysis/flux.py
@@ -193,7 +193,17 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None,
     dx = np.abs(np.diff(spectrum.spectral_axis.bin_edges))
     ew = np.sum((1 - line_flux / cont_flux) * dx)
 
-    return ew.to(spectrum.spectral_axis.unit)
+    # NOTE: uncertainty gets lost during unit translation, so we have to convert first
+    ew = ew.to(spectrum.spectral_axis.unit)
+    if line_flux.uncertainty is not None:
+        # continuum has no uncertainty, here we assume no uncertainty on the bin edges (dx) as well
+        # we need to add the line_flux uncertainties in quadrature for the sum above (which was
+        # summed over the length of dx)
+        ew.uncertainty = np.sqrt((line_flux.uncertainty.value**2)*len(dx)) * dx.unit
+    else:
+        ew.uncertainty = None
+
+    return ew
 
 
 def is_continuum_below_threshold(spectrum, threshold=0.01):
--- a/specutils/tests/test_analysis.py
+++ b/specutils/tests/test_analysis.py
@@ -101,6 +101,7 @@ def test_equivalent_width():
     result = equivalent_width(spectrum)
 
     assert result.unit.is_equivalent(spectrum.wcs.unit, equivalencies=u.spectral())
+    assert hasattr(result, 'uncertainty')
 
     # Since this is an emission line, we expect the equivalent width value to
     # be negative


specutils/analysis/flux.py
```diff
@@ -193,7 +193,17 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None,
     dx = np.abs(np.diff(spectrum.spectral_axis.bin_edges))
     ew = np.sum((1 - line_flux / cont_flux) * dx)
 
-    return ew.to(spectrum.spectral_axis.unit)
+    # NOTE: uncertainty gets lost during unit translation, so we have to convert first
```
eteq, line 196: Is there an issue for this?  I didn't realize this, definitely needs to be fixed if we can! (not part of this PR though)

specutils/analysis/flux.py
```diff
@@ -193,7 +193,17 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None,
...
+    ew = ew.to(spectrum.spectral_axis.unit)
+    if line_flux.uncertainty is not None:
+        # continuum has no uncertainty, here we assume no uncertainty on the bin edges (dx) as well
+        # we need to add the line_flux uncertainties in quadrature for the sum above (which was
+        # summed over the length of dx)
+        ew.uncertainty = np.sqrt((line_flux.uncertainty.value**2)*len(dx)) * dx.unit
```
eteq, line 202: `uncertainty` is not guaranteed to be "stddev" style uncertainties - there's a whole heap of complexity there but for now I would say the thing to do is something like this (which basically follows a similar implementation in `_compute_line_flux` above):
```suggestion
        if isinstance(line_flux.uncertainty, StdDevUncertainty):
            variance_q = line_flux.uncertainty.quantity ** 2
        elif isinstance(line_flux.uncertainty, VarianceUncertainty):
            variance_q = line_flux.uncertainty.quantity
        elif isinstance(line_flux.uncertainty, InverseVariance):
            variance_q = 1/line_flux.uncertainty.quantity
        else:
            warnings.warn(f"Uncertainty type "
                          f"'{line_flux.uncertainty.uncertainty_type}' "
                          f"was not recognized by equivalent_width. Proceeding "
                          f"without uncertainty in result.",
                          AstropyUserWarning)
            variance_q = None
            
        if variance_q is None:
            ew.uncertainty = None
        else:
            ew.uncertainty = np.sqrt(variance_q.value*len(dx)) * dx.unit
            
```

specutils/analysis/flux.py
```diff
@@ -193,7 +193,17 @@ def _compute_equivalent_width(spectrum, continuum=1, regions=None,
     dx = np.abs(np.diff(spectrum.spectral_axis.bin_edges))
     ew = np.sum((1 - line_flux / cont_flux) * dx)
 
-    return ew.to(spectrum.spectral_axis.unit)
+    # NOTE: uncertainty gets lost during unit translation, so we have to convert first
+    ew = ew.to(spectrum.spectral_axis.unit)
+    if line_flux.uncertainty is not None:
+        # continuum has no uncertainty, here we assume no uncertainty on the bin edges (dx) as well
```
eteq, line 199: We should probably *eventually* provide a way to give uncertainty through the continuum. Problem is, I'm not sure what the best way is. I would be OK with any of:
1. Allow the `continuum` keyword to be an `astropy.uncertainty`, although that doesn't really play smoothly yet with nddata uncertainties
2. Add a `continuum_uncertainty` keyword (or `continuum` as a 2-tuple?) that uses the NDData uncertainty objects to encode the continuum uncertainty
3. punt the problem for the future and record it as an issue.

Regardless of which is chosen, though, the docstring should be amended to explicitly say that continuum uncertainty is assumed to be 0.

eteq (changes requested):
Mostly there!  But this needs to address the case of non-std uncertainties (and get a test case or two exercising that).  See my suggestion which I think you can just take as-is to support this.

eteq:
Oh, and I agree that it's a bit of an issue that currently it's just a raw stddev uncertainty. I'll leave a commend in #938 about this but I think that can be easily pasted over here to be done consistently!

specutils/analysis/flux.py
kecnry: I don't think there's an issue filed.  @pllim made the good point that dropping custom attributes added to a quantity might actually be desired behavior as keeping them might suggest they were properly handled during the conversion (for example, if you set a float to `quantity.uncertainty` and then do math on that object, would you expect the float to remain fixed or react somehow to the math?).  But I do think this limitation is a reason that we should expose the uncertainties in a more "officially supported" way.

specutils/analysis/flux.py
kecnry: Currently `line_flux` just attaches a quantity to the uncertainty attribute as well, so I don't think we we should include the check here unless/until we change that behavior across the board.  

It does appear to be in standard deviation, though, despite my original comment... it just uses variance internally for the computation. See [lines 156-158](https://github.com/astropy/specutils/blob/d1cc4c45f6369c62d76d680f27bb72a47df19235/specutils/analysis/flux.py#L156-L158).

eteq: Oh gotcha, probably as part of the follow-on PR that adds uncertainty type it would be good to just check the type is stddev in case that changes in the future or something, but not a high priority now that you've clarified that.

eteq:
@kecnry - while checking the uncertainty on this PR I think I uncovered an error in the EW calculation itself (which of course also affects the uncertainty itself).  I think I trace the problem to here: https://github.com/astropy/specutils/commit/c5aaf73661e0ab603b5377bf9206c11befbec82e  which re-wrote this incorrectly thinking that `line_flux` was the *per-pixel* flux (i.e. "f-lambda" for wavelength), rather than the already-computed line flux. 

I can try to explain this more clearly, but it might be easier to just do it in code.  @kecnry would you prefer I:
1. Make a PR to your branch
2. Push some commits straight to this branch, or
3. Merge this (which I think has correctly propogated errors, but for an incorrect EW formula) as-is, and then make an issue/PR fixing the underlying issue?

kecnry:
Any of the above work for me - I don't really have a preference.  If you don't have permissions to commit directly to my branch, I'll be on the lookout for a PR and merge it in.

eteq:
Alright, I'm going to go ahead with option 3 then since that at least gets something in here.  It's also *possible* that there's some clever cancellation that happens in the math here that I'm missing, in which case option 3 actually does yield the right thing, so here goes nothing!

eteq (approved):

PR #942, opened 2022-04-06
MNT: Remove empty file and outdated docs nitpick

### LINKED ISSUE #940: Empty file: specutils/spectra/spectral_coordinate.py
Why is this file even there? https://github.com/astropy/specutils/blob/main/specutils/spectra/spectral_coordinate.py

pllim:
I don't know the history but pretty sure these are not needed anymore?

Fix #940


pllim:
Silence means no objection, so merging.

PR #944, opened 2022-04-06
New Spectrum1D.with_spectral_unit() like spectral-cube

### LINKED ISSUE #941: Port over CUNIT handling from spectral-cube
Due to astropy/astropy#3658 , you get the following behavior in `specutils`, which can be surprising to users expecting the `CUNITn` to be respected.

```python
from astropy.wcs import WCS

w = WCS({'WCSAXES': 3,
         'CRPIX1': 21.299551407997, 'CRPIX2': 31.024649715622, 'CRPIX3': 1.0,
         'CDELT1': -5.5555555555556e-05, 'CDELT2': 5.5555555555556e-05, 'CDELT3': 1.25,
         'CUNIT1': 'deg',  'CUNIT2': 'deg', 'CUNIT3': 'Angstrom',
         'CTYPE1': 'RA---TAN', 'CTYPE2': 'DEC--TAN', 'CTYPE3': 'AWAV',
         'CRVAL1': 54.1925, 'CRVAL2': -35.99917, 'CRVAL3': 4749.931640624999,
         'LONPOLE': 180.0, 'LATPOLE': -35.99917,
         'CSYER1': 1.51033062781e-05, 'CSYER2': 5.76768648658e-06,
         'MJDREF': 0.0, 'RADESYS': 'ICRS'})
print(w.wcs.cunit)
```

Get this:

```python
['deg', 'deg', 'm']
```

Want this:

```python
['deg', 'deg', 'Angstrom']
```

As a result, when this WCS is passed into `Spectrum1D`, you get this:

```python
import numpy as np

spec = Spectrum1D(flux=np.ones((10, 10, 10)) * u.nJy, wcs=w)
print(spec.spectral_axis.unit)  # gets 'm' but want 'Angstrom'
```

In `spectral-cube`, [there are code to work around this](https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/spectral_axis.py) but why was it not ported over here? If it was, I can't find it so would appreciate someone pointing me to it.

cc @keflavich

pllim:
My attempt to fix #941 but it doesn't work like I thought it would. Did I miss something?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@
 New Features
 ^^^^^^^^^^^^
 
+- New ``Spectrum1D.with_spectral_unit()`` to convert spectral axis unit
+  in a similar way as ``spectral-cube`` package. [#944]
+
 Bug Fixes
 ^^^^^^^^^
 
--- a/specutils/spectra/spectral_axis.py
+++ b/specutils/spectra/spectral_axis.py
@@ -1,4 +1,7 @@
+import warnings
+
 import astropy.units as u
+from astropy import constants
 from astropy.utils.decorators import lazyproperty
 from astropy.coordinates import SpectralCoord
 import numpy as np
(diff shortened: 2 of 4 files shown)


keflavich:
@eteq noted that we need this machinery to play well with spectralcoord https://docs.astropy.org/en/stable/api/astropy.coordinates.SpectralCoord.html

pllim:
Does someone more familiar with all this stuff want to take this PR over? I am not sure how to proceed.

keflavich:
sorry my comment was terse - we were talking about how to integrate specutils + spectral-cube, and this was one of the points we hit.  I think we just need more discussion about this to understand where the interaction layers are b/w spectralcoord and this WCS machinery.  My intent was to open that discussion here.  Maybe that means one of us takes over the PR or maybe we just need a new one on top.

keflavich:
Documentation for how to wrap FITS-WCS into gWCS: https://gwcs.readthedocs.io/en/latest/#a-step-by-step-example-of-constructing-an-imaging-gwcs-object

eteq:
FYI, after some out-of-band discussion we are thinking an approach at least vaguely like https://gwcs.readthedocs.io/en/latest/#a-step-by-step-example-of-constructing-an-imaging-gwcs-object would be a good way to make it so that we can turn what start out as FITS-WCS files into a gwcs-style object that can be more easily adapted in some of the specutils machinery.

pllim:
I don't think this is getting anywhere and I forgot what I was trying to do.

PR #946, opened 2022-04-12
Add helper functions to control spectral shifting

### LINKED ISSUE #820: redshift applied in the wrong direction in Spectrum1D object
Hi, I just tried filling in the redshift value for a Spectrum1D object after the redshift information became available, and the spectral axis was immediately multiplied by (1+z). I would normally expect that the wavelength will be reported in observed frame by default and not changed when the user adds the measured redshift to the object but that the redshift allows the user to convert from observed frame to rest-frame. As a user, I would also expect that if there is a "to_rest" operation, it will apply to the spectral axis and to the flux and uncertainty as well, taking account the units to treat the f_lambda and f_nu units properly (the latter is already in issue #741 https://github.com/astropy/specutils/issues/741). I would also hope to be able to know whether the wavelength/fluxes are in rest or observed frame (maybe a keyword with rest=True or False ?).

Here is a simple code that I used in a Notebook based on a demo ran by Erik Tollerud:
```
from astropy.utils.data import download_file
%matplotlib inline
import matplotlib.pyplot as plt

spec_fn = download_file('https://stsci.box.com/shared/static/b22b1fzhimtdqfp8597m4bg67kovvauu.fits', cache=True)

jwst_spec = Spectrum1D.read(spec_fn)
plt.plot(jwst_spec.spectral_axis, jwst_spec.flux)

jwst_spec.redshift = 1.5
plt.plot(jwst_spec.spectral_axis, jwst_spec.flux, color='r')

aragilar:
The setters for Spectrum1d.redshift and Spectrum1d.radial_velocity shift
the spectral axis, which from #820 is not always what users expect. This
adds three helpers, one which does the current shifting (and is
explicitly documented to do that), and two which simply change the
values of redshift and radial_velocity.

Closes #820.

### DIFF
--- a/docs/spectrum1d.rst
+++ b/docs/spectrum1d.rst
@@ -133,6 +133,49 @@ however, specutils will recognize ``flux`` values input as ``NaN`` and set the
 mask to ``True`` for those values unless explicitly overridden.
 
 
+Including Redshift or Radial Velocity
+-------------------------------------
+
+The :class:`~specutils.Spectrum1D` class supports setting a redshift or radial
+velocity upon initialization of the object, as well as updating these values.
+The default value for redshift and radial velocity is zero - to create a
+:class:`~specutils.Spectrum1D` with a non-zero value, simply set the appropriate
+attribute on object creation:
+
+.. code-block:: python
+
+    >>> spec1 = Spectrum1D(spectral_axis=np.arange(5000, 5010)*u.AA, flux=np.random.sample(10)*u.Jy, redshift = 0.15)
+    >>> spec2 = Spectrum1D(spectral_axis=np.arange(5000, 5010)*u.AA, flux=np.random.sample(10)*u.Jy, radial_velocity = 1000*u.Unit("km/s"))
+
+By default, updating either the ``redshift`` or ``radial_velocity`` attributes
+of an existing :class:`~specutils.Spectrum1D` directly uses the
+:meth:`specutils.Spectrum1D.shift_spectrum_to` method, which also updates the
+values of the ``spectral_axis`` to match the new frame. To leave the
+``spectral_axis`` values unchanged while updating the ``redshift`` or
+``radial_velocity`` value, use the :meth:`specutils.Spectrum1D.set_redshift_to`
+or :meth:`specutils.Spectrum1D.set_radial_velocity_to` method as appropriate.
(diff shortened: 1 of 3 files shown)


rosteen (commented):
I think this looks good, thanks for the update (and preserving the existing functionality). I'm working on adding a section about this to the narrative docs before I approve and merge, mind if I push that to this branch so you can be a second pair of eyes on it?

aragilar:
Sure, go ahead.

I think deprecating and (eventually) removing directly writing to redshift would be a good idea, but I figure it's not an urgent fix, so a generic warning would suffice for now, but if we want to switch to a deprecation warning immediately, I'm happy if someone pushes that change though.

rosteen:
I agree that the warning is sufficient for now. I pushed the added doc section, if you wouldn't mind looking it over once and giving me a thumbs up that would be great. Either way I'll merge by the end of the day.

rosteen (approved):

aragilar:
Looks good to me!

rosteen:
Remote data failure is unrelated, merging.

stephjuneau:
Hi @aragilar , thanks for adding the functionality to update the redshift value without shifting the spectrum. This will be particularly useful for large spectroscopic surveys like SDSS and DESI for which the redshift is known but the spectra are still published in observed frame. Regarding the "default" behavior of shifting the spectrum, does the code correctly apply the redshift to all of the wavelength, flux and uncertainty? They all change with redshift (and differently for f_lambda versus f_nu units for the flux). Thanks again! Looking forward to trying it soon.

aragilar:
@stephjuneau I haven't changed the shifting behaviour, only moved the code that did the shifting from the setter to a separate function. I don't know what specutils' policy around backwards compatibility is, but maybe the corrections to flux/uncertainty can be added to the shifting code as a flag (which is off by default for now, with a change in the future for 2.0?).

pllim:
> specutils' policy around backwards compatibility

I don't think there is a formal policy but given this is an Astropy Coordinated Package, it should not stray too far from the core policy; i.e., breaking changes should only go in major releases, while non-breaking API changes may go to minor releases, and bugfix release should be strictly bugfix. That said, `astropy` does not use semantic versioning, so what constitutes a "major release" for specutils is up to the package release manager here, but once set should stay consistent going forward, if possible. Hope this helps!

PR #954, opened 2022-05-24
DOC: Do not use code that emits warning in example

pllim:
This is a follow-up of #946 . If you do not want people to use it, do not put it in example. Here, I replaced the example with the new "proper" way to set the redshift in a backward compatible way.


pllim:

rosteen (approved):

PR #962, opened 2022-07-01
Fitting improvements

rosteen:
Working toward a couple improvements to fitting, mainly doing parameter estimates based on the region being fit rather than the whole spectrum, and returning information needed to estimate fit parameter uncertainties from the underlying scipy fitter results

### DIFF
--- a/specutils/analysis/location.py
+++ b/specutils/analysis/location.py
@@ -14,7 +14,7 @@ from .uncertainty import _convert_uncertainty
 __all__ = ['centroid']
 
 
-def centroid(spectrum, region):
+def centroid(spectrum, regions=None):
     """
     Calculate the centroid of a region, or regions, of the spectrum.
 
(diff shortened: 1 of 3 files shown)


rosteen:
Alright, this probably merits another test or two, but moving this to ready for review in case anyone has feedback. Mainly this changes the default fitter to include uncertainties from the underlying astropy fitter in the output. It also gives the user the option to get the `fit_info` back from the underlying fitter in case they want to do something with e.g. the covariance matrix themselves.

keflavich (approved):

specutils/analysis/location.py
```diff
@@ -14,7 +14,7 @@
 __all__ = ['centroid']
 
 
-def centroid(spectrum, region):
+def centroid(spectrum, regions=None):
```
pllim, line 17: This broke Jdaviz. Is this extra "s"  a typo? There is no deprecation period.

Example log: https://github.com/spacetelescope/jdaviz/runs/7290806527?check_suite_focus=true

rosteen: Gah, good call. See #963

PR #963, opened 2022-07-11
Properly deprecate region keyword in centroid

rosteen:
#962 was a little overzealous, and I forgot to provide a deprecation period for the old input parameter name.

### DIFF
--- a/specutils/analysis/location.py
+++ b/specutils/analysis/location.py
@@ -14,7 +14,7 @@ from .uncertainty import _convert_uncertainty
 __all__ = ['centroid']
 
 
-def centroid(spectrum, regions=None):
+def centroid(spectrum, regions=None, region=None):
     """
     Calculate the centroid of a region, or regions, of the spectrum.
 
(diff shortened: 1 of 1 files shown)


specutils/analysis/location.py
```diff
@@ -14,7 +14,7 @@
 __all__ = ['centroid']
 
 
-def centroid(spectrum, region):
+def centroid(spectrum, regions=None, region=None):
```
pllim, line 17: This would be more backwards compatible.
```suggestion
def centroid(spectrum, region=None, regions=None):
```

rosteen: Good call

rosteen:
After thinking about it for another minute, I might roll back your suggestion actually. If the user is just putting in the region argument positionally rather than explicitly giving the keyword, that will work fine in the new code since it will just go into `regions` as expected. If I put `region` before `regions` then if the region is provided just based on position, it will unnecessarily raise the deprecation warning.

pllim:
Good point...

rosteen:
Apparently I never committed adding the imports, sheesh. Should be good now.

pllim (approved):
I think this works. Though if you wanna be fancy, you can use https://docs.astropy.org/en/stable/api/astropy.utils.decorators.deprecated_renamed_argument.html but you don't have to.

This fixes the traceback but it won't fix Jdaviz CI. Downstream, there needs to be a follow-up PR after this is merged to handle this API change appropriately.

rosteen:
I think I'll leave it as-is. Thanks for the review!

rosteen:
I don't know how duplicate commits for the last two merged PRs ended up in here, I'll deal with that tomorrow. Don't want to merge until it's cleaned up.

pllim:
If you click "resolve conflicts" on GitHub UI, weird things like that happens. Hopefully the history will clear itself up when you manually rebase.

rosteen:
There we go, local main was messed up, had to reset.

PR #964, opened 2022-07-12
added functions to Spectrum1D that allow for multiplication, addition…  

ChrisSWDavis:
… and subtraction from the left as well as from the right.

This makes the manipulation of spectra through various operations slightly more flexible and easier for users, so they don't have to add/subtract/multiply floats and ints etc from the right every time.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 New Features
 ^^^^^^^^^^^^
 
+- Added handling for arithmetic with constants from the left to Spectrum1D. [#964]
+
 Bug Fixes
 ^^^^^^^^^
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,10 +4,10 @@
 New Features
 ^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


rosteen:
Could you give an example of something that uses `__rsub__`, `__rmul__` or `__radd__`? I admit to not really understanding what capability this is adding.

rosteen:
Alright, I answered my own question - I didn't even realize we were missing this capability, thanks for the contribution! I'll add a test or two and then get this in.

ChrisSWDavis:
Hi, apologies for not responding sooner, I got quite busy over the last couple of days. Yeah the changes let you do things like

`float_number + Spectrum_1D_instance `

instead of having to do

`Spectrum_1D_instance + float_number`

every time (where Spectrum_1D_instance is of type Spectrum1D, and float_number is a float).

rosteen:
Sorry for the delay, testing this lead me down a rabbit hole of something that I think will require an upstream fix in `astropy`. I added a changelog entry and tests, I think this is good to go. Thanks again for the contribution.

rosteen (approved):

PR #961, opened 2022-06-28
uncertainty_type exposed for analysis functions that implement uncertainty

kecnry:
analysis.line_flux, equivalent_width, and centroid currently return a quantity object with the `.uncertainty` attribute set to a float (see #938, #939).  This PR also sets `.uncertainty_type = 'stddev'` as an intermediate way to expose that these are standard-deviation uncertainties until a more permanent framework for exposing uncertainties attached to the quantity objects is implemented.

Closes #958

PR #968, opened 2022-07-27
guassian_fwhm and gaussian_sigma_width: propagated uncertainties

kecnry:
Similar to #938 and #939 and making use of the new `uncertainty_type` introduced in #961, this propagates uncertainties through the `gaussian_fwhm` and `gaussian_sigma_width` functions.

Note that the values in the test coverage have not been validated (they are just set to the current returned values).

### DIFF
--- a/specutils/analysis/width.py
+++ b/specutils/analysis/width.py
@@ -4,9 +4,11 @@ spectral features.
 """
 
 import numpy as np
+from astropy.nddata import StdDevUncertainty
 from astropy.stats.funcs import gaussian_sigma_to_fwhm
 from ..manipulation import extract_region
 from . import centroid
+from .uncertainty import _convert_uncertainty
 from .utils import computation_wrapper
 from scipy.signal import find_peaks, peak_widths
 
@@ -186,8 +188,15 @@ def _compute_gaussian_fwhm(spectrum, regions=None):
     This is a helper function for the above `gaussian_fwhm()` method.
     """
 
-    fwhm = _compute_gaussian_sigma_width(spectrum, regions) * gaussian_sigma_to_fwhm
+    sigma = _compute_gaussian_sigma_width(spectrum, regions)
 
+    fwhm = sigma * gaussian_sigma_to_fwhm
+    if sigma.uncertainty is not None:
+        fwhm.uncertainty = sigma.uncertainty * gaussian_sigma_to_fwhm
+    else:
+        fwhm.uncertainty = None
+
+    fwhm.uncertainty_type = 'std'
     return fwhm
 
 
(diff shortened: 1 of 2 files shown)


kecnry:
hmmm... CI is failing because the uncertainties attached to the centroid quantity are getting dropped when wrapped in the `computation_wrapper` when testing support for multiple spectra in the same call.  I'm going to mark this as draft until we come up with a solution for that.

specutils/analysis/width.py
```diff
@@ -215,8 +231,31 @@ def _compute_gaussian_sigma_width(spectrum, regions=None):
...
+        num_term_uncerts = dx * dx * flux * np.sqrt(2*(dx_uncert/dx)**2 + (flux_uncert/flux)**2)
+        # uncertainty (squared) for the numerator, added in quadrature
+        num_uncertsq = np.sum(num_term_uncerts**2, axis=-1)
+        # uncertainty (squared) for the denomenator
+        denom_uncertsq = np.sum(flux_uncert**2)
+
+        sigma2_uncert = numerator/denom * np.sqrt(num_uncertsq * numerator**-2 +
+                                                  denom_uncertsq * denom**-2)
+
+        sigma.uncertainty = 0.5 * sigma2_uncert / sigma2 * sigma.unit
```
keflavich, line 254: it would be helpful to write out the math behind this in an associated latex document

keflavich: I've checked by hand and this all looks right to me though

specutils/analysis/width.py
```diff
@@ -212,11 +228,36 @@ def _compute_gaussian_sigma_width(spectrum, regions=None):
 
     if flux.ndim > 1:
         spectral_axis = np.broadcast_to(spectral_axis, flux.shape, subok=True)
+        centroid_result_uncert = centroid_result.uncertainty
         centroid_result = centroid_result[:, np.newaxis]
+        centroid_result.uncertainty = centroid_result_uncert[:, np.newaxis] if centroid_result_uncert is not None else None  # noqa
```
kecnry, line 233: This is where the centroid uncertainties were being dropped for multiple spectra.  Note that cases like this are just examples of why attaching uncertainties directly to the quantities is not an ideal solution as they are easily dropped (either internally or by users) and the workaround is a bit... ugly.

specutils/analysis/width.py
kecnry: This is a good idea (even though this should just follow normal uncertainty propagation rules, the code can definitely be difficult to parse, even with splitting into "steps" to try to keep it readable).  I think this probably should be done for all analysis functions that expose uncertainties in some consolidated place in the docs, and that might be better suited for a follow-up effort since its more broad?

If you/anyone thinks it should be included before merging this, let me know where it belongs and I can try to write it up and include it.

keflavich: Followup is fine by me.

rosteen (approved):
Dev job test failure is unrelated. Looks good to me, I'll open an issue for the documentation followup if there isn't one already.

PR #970, opened 2022-07-29
Fix moment with order > 1 and non-square spatial dimensions

### LINKED ISSUE #931: Revisit PR 912 that got reverted (moment transpose for order > 1)
Original problem:

* spacetelescope/jdaviz#947

Original fix:

* #912 (reverted by #929 and never released)
    * This patches code added in #781

Simple test case presented at https://github.com/astropy/specutils/pull/929#issuecomment-1043739843 did not convince @eteq , @rosteen , and myself that it is worth the risk being in a release without a better test case. @eteq proposed a following test case:

1. Create 12 Gaussian1D models with known sigma.
2. Pack them into a spectral cube that is 3 x 4 in spatial, uh, space.
3. Run moment calculations on the cube.
4. Do the results agree with theoretical values with and without #912?

Related issues:

* #930 
* #925

javerbukh:
Fixes https://github.com/spacetelescope/jdaviz/issues/947
Fixes #931 

I thought we would need to revisit the topic of dropping support for different axes, but then I found the `keepdims` paremeter in `np.sum` and now it all seems to work. I did have to update one test, hopefully someone can confirm if that fix is ok. Will definitely need to get a science review on this from @PatrickOgle if he is able.

### DIFF
--- a/specutils/analysis/moment.py
+++ b/specutils/analysis/moment.py
@@ -73,10 +73,12 @@ def _compute_moment(spectrum, regions=None, order=0, axis=-1):
 
     if order > 1:
         m0 = np.sum(flux, axis=axis)
-        m1 = np.sum(flux * dispersion, axis=axis) / np.sum(flux, axis=axis)
 
-        if len(flux.shape) > 1 and (axis == len(flux.shape)-1 or axis == -1):
-            _shape = flux.shape[-1:] + tuple(np.ones(flux.ndim - 1, dtype='i'))
-            m1 = np.tile(m1, _shape).T
+        # By setting keepdims to True, the axes which are reduced are
+        # left in the result as dimensions with size one. This means
+        # that we can broadcast m1 correctly against dispersion.
+        m1 = (np.sum(flux * spectral_axis, axis=axis, keepdims=True)
+              / np.sum(flux, axis=axis, keepdims=True))
 
-        return np.sum(flux * (dispersion - m1) ** order, axis=axis) / m0
+        mn = np.sum(flux * (dispersion - m1) ** order, axis=axis) / m0
+        return mn
--- a/specutils/analysis/moment.py
+++ b/specutils/analysis/moment.py
@@ -80,5 +80,4 @@ def _compute_moment(spectrum, regions=None, order=0, axis=-1):
         m1 = (np.sum(flux * spectral_axis, axis=axis, keepdims=True)
               / np.sum(flux, axis=axis, keepdims=True))
 
-        mn = np.sum(flux * (dispersion - m1) ** order, axis=axis) / m0
-        return mn
+        return np.sum(flux * (dispersion - m1) ** order, axis=axis) / m0
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,9 @@ New Features
 Bug Fixes
 ^^^^^^^^^
 
+- Fixed a bug with moment map orders greater than 1 not being able to handle
+  cubes with non-square spatial dimensions. [#970]
+
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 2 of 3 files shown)


PatrickOgle (approved):
Verified that moments 1, 2, 3, 4, 5, 6 give the correct numbers for a Gaussian with mean =50 Hz and sigma = 5 Hz:

Moment 1 = 50 Hz
Moment 2 = 25 Hz^2
Moment 3 = -1.067E-12  Hz^3 (should be zero, but it is close)
Moment 4 = 1.875E3 Hz^4 (should be 3*sigma^4 = 1875), so this is correct.
Moment 5  = -1.3E-10 Hz^5 (should be 0)
Moment 6  = 2.34375E5 Hz^6  (should be 15*sigma^6 =234,375), so this is correct

There may be a problem with Moment 0 (off by a factor equal to bin width), but this may be a pre-existing bug.

specutils/analysis/moment.py
```diff
@@ -73,10 +73,12 @@ def _compute_moment(spectrum, regions=None, order=0, axis=-1):
...
-            m1 = np.tile(m1, _shape).T
+        # By setting keepdims to True, the axes which are reduced are
+        # left in the result as dimensions with size one. This means
+        # that we can broadcast m1 correctly against dispersion.
+        m1 = (np.sum(flux * spectral_axis, axis=axis, keepdims=True)
+              / np.sum(flux, axis=axis, keepdims=True))
 
-        return np.sum(flux * (dispersion - m1) ** order, axis=axis) / m0
+        mn = np.sum(flux * (dispersion - m1) ** order, axis=axis) / m0
+        return mn
```
pllim, line 84: No need to change this part?
```suggestion
        return np.sum(flux * (dispersion - m1) ** order, axis=axis) / m0
```

specutils/analysis/moment.py
```diff
@@ -73,10 +73,12 @@ def _compute_moment(spectrum, regions=None, order=0, axis=-1):
...
         m0 = np.sum(flux, axis=axis)
-        m1 = np.sum(flux * dispersion, axis=axis) / np.sum(flux, axis=axis)
 
-        if len(flux.shape) > 1 and (axis == len(flux.shape)-1 or axis == -1):
-            _shape = flux.shape[-1:] + tuple(np.ones(flux.ndim - 1, dtype='i'))
-            m1 = np.tile(m1, _shape).T
+        # By setting keepdims to True, the axes which are reduced are
+        # left in the result as dimensions with size one. This means
+        # that we can broadcast m1 correctly against dispersion.
+        m1 = (np.sum(flux * spectral_axis, axis=axis, keepdims=True)
```
pllim, line 80: Nice! This is indeed much more elegant.

pllim (commented):
Probably need a change log.

Did you ever get around to constructing the gaussian test case as laid out in #931 ? That would be a good unit test to add to the test suite, as I am not convinced that the existing test is good enough.

Or at the very least implement the check that Patrick did as a proper unit test here?

pllim:
I looped in Adam G in case he is interested since this is related to spectral cube.

keflavich:
I don't think I understand the underlying issue (non-square dimensions).

Unfortunately this is not relevant to spectral-cube because spectral-cube defines the spectral axis to be the 0'th dimension, not the -1'th.  Our discussions of how to resolve this stalled.  We agreed  that it would be better to specify the dimensions directly (i.e., use a name, e.g., `specutils.SPECTRAL_AXIS=-1` and `spectral_cube.SPECTRAL_AXIS=0`) and always use a variable name, but we did not converge on how to do this or prioritize implementing it.

javerbukh:
@pllim Added the test @PatrickOgle used, please re-review when you have time!

javerbukh:
Code style check failures are related to an upstream change, not this PR.

pllim:
Test looks good, thanks! Still need a change log though.

https://github.com/astropy/specutils/blob/main/CHANGES.rst

javerbukh:
@pllim Done!

pllim (approved):
Thanks!

pllim:
p.s. Code style will be fixed with #971

rosteen (approved):
Looks great, and thanks for the additional test coverage. I need to remember that `keepdims` exists!

PR #972, opened 2022-08-08
Update common Data Central loaders

aragilar:
This adds some minor improvements to the common Data Central loaders,
which will be used in survey loader to come.


rosteen (approved):
Seems reasonable, I can see how this might be useful in the future.

PR #975, opened 2022-08-11
Change to deprecation warning for redshift/RV setters

### LINKED ISSUE #953: Properly deprecate redshift and radial_velocity setters
As of #946, the setters emit a `UserWarning` that usage is ambiguous. If indeed such usage should not be, uh, used, then the setters should be properly deprecated and then removed after a reasonable deprecation period.

https://github.com/astropy/specutils/blob/ec4fe50c6c032fc421c2cd0ee0dda11fd0f856cb/specutils/spectra/spectrum1d.py#L663-L678

rosteen:
Simple enough, updates these warnings from "don't use these" to an explicit deprecation warning.

Closes #953

### DIFF
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -4,6 +4,7 @@ from copy import deepcopy
 import numpy as np
 from astropy import units as u
 from astropy.utils.decorators import lazyproperty
+from astropy.utils.exceptions import AstropyDeprecationWarning
 from astropy.nddata import NDUncertainty, NDIOMixin, NDArithmeticMixin
 
 from .spectral_axis import SpectralAxis
@@ -662,19 +663,20 @@ class Spectrum1D(OneDSpectrumMixin, NDCube, NDIOMixin, NDArithmeticMixin):
 
     @redshift.setter
     def redshift(self, val):
-        warnings.warn(
-            "Setting the redshift of a spectrum is ambiguous, use either "
-            "set_redshift_to or shift_spectrum_to to be explicit."
-        )
+        warnings.warn(AstropyDeprecationWarning(
+            "Setting the redshift attribute of a spectrum directly is deprecated "
+            "as of version 1.8.0 and will be removed in a future release. Use "
+            "set_redshift_to or shift_spectrum_to instead."
+        ))
         self.shift_spectrum_to(redshift=val)
 
     @radial_velocity.setter
     def radial_velocity(self, val):
-        warnings.warn(
-            "Setting the radial velocity of a spectrum is ambiguous, use "
-            "either set_radial_velocity_to or shift_spectrum_to to be "
-            "explicit."
-        )
+        warnings.warn(AstropyDeprecationWarning(
+            "Setting the radial_velocity attribute of a spectrum directly is deprecated "
+            "as of version 1.8.0 and will be removed in a future release. Use "
+            "set_radial_velocity_to or shift_spectrum_to instead."
+        ))
         self.shift_spectrum_to(radial_velocity=val)
 
     def __add__(self, other):
--- a/specutils/spectra/spectrum1d.py
+++ b/specutils/spectra/spectrum1d.py
@@ -4,7 +4,7 @@ from copy import deepcopy
 import numpy as np
 from astropy import units as u
 from astropy.utils.decorators import lazyproperty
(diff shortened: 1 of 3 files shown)


specutils/spectra/spectrum1d.py
```diff
@@ -4,6 +4,7 @@
 import numpy as np
 from astropy import units as u
 from astropy.utils.decorators import lazyproperty
+from astropy.utils.exceptions import AstropyDeprecationWarning
```
pllim, line 7:
```suggestion
from astropy.utils.decorators import deprecated
```

specutils/spectra/spectrum1d.py
```diff
@@ -662,19 +663,20 @@ def shift_spectrum_to(self, *, redshift=None, radial_velocity=None):
...
     def redshift(self, val):
-        warnings.warn(
-            "Setting the redshift of a spectrum is ambiguous, use either "
-            "set_redshift_to or shift_spectrum_to to be explicit."
-        )
+        warnings.warn(AstropyDeprecationWarning(
+            "Setting the redshift attribute of a spectrum directly is deprecated "
+            "as of version 1.8.0 and will be removed in a future release. Use "
+            "set_redshift_to or shift_spectrum_to instead."
+        ))
```
pllim, line 670:
```suggestion
```

specutils/spectra/spectrum1d.py
```diff
@@ -662,19 +663,20 @@ def shift_spectrum_to(self, *, redshift=None, radial_velocity=None):
 
     @redshift.setter
     def redshift(self, val):
```
pllim, line 665: Does this not work for you?
```suggestion
    @deprecated('1.8.0', alternative='set_redshift_to or shift_spectrum_to')
    def redshift(self, val):
```

specutils/spectra/spectrum1d.py
```diff
@@ -662,19 +663,20 @@ def shift_spectrum_to(self, *, redshift=None, radial_velocity=None):
...
-        warnings.warn(
-            "Setting the radial velocity of a spectrum is ambiguous, use "
-            "either set_radial_velocity_to or shift_spectrum_to to be "
-            "explicit."
-        )
+        warnings.warn(AstropyDeprecationWarning(
+            "Setting the radial_velocity attribute of a spectrum directly is deprecated "
+            "as of version 1.8.0 and will be removed in a future release. Use "
+            "set_radial_velocity_to or shift_spectrum_to instead."
+        ))
```
pllim, line 679:
```suggestion
```

specutils/spectra/spectrum1d.py
```diff
@@ -662,19 +663,20 @@ def shift_spectrum_to(self, *, redshift=None, radial_velocity=None):
...
-        )
+        warnings.warn(AstropyDeprecationWarning(
+            "Setting the redshift attribute of a spectrum directly is deprecated "
+            "as of version 1.8.0 and will be removed in a future release. Use "
+            "set_redshift_to or shift_spectrum_to instead."
+        ))
         self.shift_spectrum_to(redshift=val)
 
     @radial_velocity.setter
     def radial_velocity(self, val):
```
pllim, line 674:
```suggestion
    @deprecated('1.8.0', alternative='set_radial_velocity_to or shift_spectrum_to')
    def radial_velocity(self, val):
```

specutils/spectra/spectrum1d.py
rosteen: I meant to say in the description, I tried the decorators and they didn't work (threw an error on missing `__name__` attribute), but maybe that was because I put the `deprecated` decorator before the `setter` decorator...I can try this locally and see if it works in this order.

pllim: Hmm maybe it just never works on a setter. It is okay to not use the decorator too. Pls lemme know if swapping works. 🤞

rosteen: Well, lo and behold, it does work with the decorators in the opposite order. Didn't think to try that until I saw your suggestion in that order.

pllim: My order is purely accident. Thanks for trying! 😸

pllim (commented):
LGTM but you also need change log under API change. Thanks!

pllim:
Also need to update tests.

pllim (approved):
All good now. 😉  Thanks!

PR #976, opened 2022-08-13
Fix uncertainty propagation in FluxConservingResampler

### LINKED ISSUE #792: FluxConservingResampler does not properly propagate the uncertainty
The FluxConservingResampler function does not properly propagate the uncertainties. This function, as stated in the documentation (https://specutils.readthedocs.io/en/stable/api/specutils.manipulation.FluxConservingResampler.html#specutils.manipulation.FluxConservingResampler.resample1d) should be based on formulas reported on Carnall+2017 paper
(https://arxiv.org/pdf/1705.05165.pdf). However, you can verify that the result obtained in a simple case, is different from the predictions of the native python package SpectRes (https://github.com/ACCarnall/SpectRes). The error in the source code is probably in computing the output uncertainties as

`        # Calculate output uncertainty
        if pixel_uncer is not None:
            pixel_uncer = pixel_uncer.reshape(new_flux_shape)
            out_variance = np.sum(pixel_uncer * resample_grid**2, axis=-1) / np.sum(
                resample_grid**2, axis=-1)
            out_uncertainty = InverseVariance(np.reciprocal(out_variance))`

While, following Equation (4) in Carnall+2017, the correct "out_variance" should be instead

`            out_variance = np.sum(pixel_uncer * resample_grid**2, axis=-1) / np.sum(
                resample_grid, axis=-1)**2`

Could you please check these lines?

hey-moon:
Fixed #792

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ Bug Fixes
 
 - Arithmetic with constants and Spectrum1D now works in either order. [#964]
 
+- Fixed uncertainty propagation in FluxConservingResampler. [#976]
+
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


rosteen (approved):
Thanks for the contribution! I went ahead and updated the expected test values, I'll add a changelog entry and get this merged.

PR #982, opened 2022-09-20
SpectrumList.read improvement for multiple FITS files

bmorris3:
Hi folks, 

I was working with some old code by @orifox which called `SpectrumList.read('directory/.')` in order to load all readable FITS files within `directory`. With the latest version of specutils this behavior produces an unexpected result: the `SpectrumList` that gets returned only has the _primary HDUs_ for each readable file within `directory`. This traces back to [this block of code](https://github.com/astropy/specutils/blob/a4038efa0a3cfdbaddf7a2e5bd5ef52c3c2969fd/specutils/io/default_loaders/jwst_reader.py#L281-L284) introduced in #838.

In this PR I've removed the line that extracts only primary HDU and put all available HDUs in the resulting `SpectrumList`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Bug Fixes
 ^^^^^^^^^
 
 - Fix bug in fitting with weights if weights argument is set to 'unc'. [#979]
+- Fix bug in reader which caused multi-extension files to load only the
+  primary HDU [#982]
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


rosteen (approved):
Thanks, I verified that it now loads data from multiple extensions in a JWST file. Would you please add a changelog entry?

CHANGES.rst
```diff
@@ -8,6 +8,8 @@ Bug Fixes
 ^^^^^^^^^
 
 - Fix bug in fitting with weights if weights argument is set to 'unc'. [#979]
+- Fix bug in reader which caused multi-extension files to load only the
```
rosteen, line 11:
```suggestion
- Fix bug in JWST reader which caused multi-extension files to load only the
```

PR #984, opened 2022-10-13
Replace all uncertainty conversion with represent_as()

### LINKED ISSUE #983: snr_threshold assumes that uncertainty is represented as a standard deviation
The snr_threshold function calculates SNR by dividing the flux by "spectrum.uncertainty.quantity". This assumes that the uncertainty is of type StdDevUncertainty. In some cases, if it is of type InverseVariance for example, this would yield an incorrect SNR. Changing "spectrum.uncertainty.quantity" to "spectrum.uncertainty.represent_as(StdDevUncertainty).quantity" and including "from astropy.nddata import StdDevUncertainty" in the imports should fix it.

rosteen:
As of astropy 5.1, `NDUncertainty` objects now have a built-in `represent_as` method to convert between types. This PR replaces the multiple ways we were doing conversion between uncertainty types with consistent use of this method, as well as converting between types in a few places where the uncertainty was assumed to be standard deviation, but not enforced. This definitely needs another set of eyes to check that my types are correct in those places where I added conversion.

Closes #983

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -56,9 +56,9 @@ jobs:
             toxenv: py38-test-alldeps
             toxposargs: --durations=50
 
-          - name: Python 3.8 doc build
+          - name: Python 3.9 doc build
             os: ubuntu-latest
-            python: 3.8
+            python: 3.9
             toxenv: build_docs
 
     steps:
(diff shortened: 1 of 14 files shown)


rosteen:
I'm also updating the CI tests here, since the version of astropy we now require dropped support for python 3.7.

rosteen:
Test failure is an unrelated remote data download error.

bmorris3 (approved):
This looks good!

PR #992, opened 2022-11-24
Fix SpectralAxis.with_observer_stationary_relative_to

StuartLittlefair:
`SpectralAxis.with_observer_stationary_relative_to` should call the function of same name defined on the parent class, and actually return the axis so created.


StuartLittlefair:
fixes #994

keflavich (approved):
should the indentation on the following lines be shifted?

nmearl:
Good catch!

nmearl (approved):

rosteen:
Thanks!

PR #998, opened 2022-12-01
Improve spectra arithmetic

### LINKED ISSUE #989: arithmetic operations don't check compatibility of spectral_axis
Arithmetic operations succeed so long as the flux arrays are of compatible shape, but do not check for equivalent spectral axis values.  For example,

```
import numpy as np

sp1 = Spectrum1D(np.random.random(10)*u.Jy, spectral_axis=np.linspace(0,1,10)*u.pix)
sp2 = Spectrum1D(np.random.random(10)*u.Jy, spectral_axis=np.linspace(2,8,10)*u.pix)
sp2 - sp1
```
succeeds but returns a `Spectrum1D` object with a spectral axis adopted directly from `sp2`.  Instead this should raise an error (or perhaps eventually support optional interpolation for cases in which there is sufficient overlap).

Related: #198

pllim:
Fix #989 and more.

[🐱](https://jira.stsci.edu/browse/JDAT-2928)

### TODO

- [ ] Rebase
- [ ] Consider implication of https://github.com/astropy/astropy/pull/13977

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ New Features
 Bug Fixes
 ^^^^^^^^^
 
+- Spectrum1D math operators no longer perform illogical math. [#998]
+
 1.9.1 (2022-11-22)
 ------------------
 
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -691,7 +691,8 @@ convenience functions to perform exactly this task.  An example is shown below.
     The normalized spectrum is simply the old spectrum devided by the
     fitted continuum, which returns a new object:
 
-    >>> spec_normalized = spectrum / y_continuum_fitted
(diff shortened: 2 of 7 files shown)


pllim:
I don't think the remote data failure is related. It has something to do with this file that I didn't touch.

https://data.sdss.org/sas/dr8/sdss/spectro/redux/26/0273/spPlate-0273-51957.fits

pllim:
If we ever come back to this, need to also consider the implication of https://github.com/astropy/astropy/pull/13977

pllim:
Is there still interest to pursue this or should I just close it without merge?

rosteen:
Yes, there's still interest, but I milestoned it to 2.0 so I haven't been in a rush to get it in.

PR #1001, opened 2022-12-22
Fix pixel region index retrieval

duytnguyendtn:
When a `u.pix` defined region attempted to `extract_region` on a Spectrum1D defined with a spectral axis defined with `u.pix`, the region index bound detection algorithm failed to resolve the correct indices, instead returning the actual value of the pixel quantity of the spectral region. This resulted in an empty list being returned due to the slicing indices being incorrect.

Quick test snippet:
```
import astropy.units as u
flux_unit = u.dimensionless_unscaled
spec_unit = u.pix

import numpy as np
flux = np.random.randn(200)*flux_unit
wavelength = np.arange(5100, 5300)*spec_unit
spec1d = Spectrum1D(spectral_axis=wavelength, flux=flux)

from specutils import SpectralRegion
from specutils.manipulation import extract_region
region = SpectralRegion.from_center(center=5200.5*spec_unit, width=100*spec_unit)
extract_region(spec1d, region)
```
should return:
```
Spectrum1D (length=101)
flux:             [ -0.14648, ..., 0.0035672 ],  mean=0.036819
spectral axis:    [ 5150.0 pix, ..., 5250.0 pix ],  mean=5200.0 pix
```
whereas on `main`:
```
Spectrum1D (length=0)
flux:             [ ],  mean= n/a
spectral axis:    [ ],  mean= n/a
```

`tox` test failures appear to be unrelated, possibly related to the same issues we've seen in https://github.com/spacetelescope/jdaviz/pull/1901. Related tests are all testing locally

### DIFF
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -75,7 +75,12 @@ def _subregion_to_edge_pixels(subregion, spectrum):
         if not spectral_axis.unit.is_equivalent(u.pix):
             left_index = floor(subregion[0].value)
         else:
-            left_index = int(np.ceil(left_func(subregion).value))
+            # Get index of closest value
+            # See https://stackoverflow.com/a/26026189 if performance becomes an issue
+            left_index = np.nanargmin((np.abs(spectral_axis - subregion[0])))
+            # Ensure index is inclusive of region bounds
+            if (spectral_axis[left_index] > subregion[0]) and (left_index <= 1):
+                left_index -= 1
     else:
         # Convert lower value to spectrum spectral_axis units
         left_reg_in_spec_unit = left_func(subregion).to(spectral_axis.unit,
@@ -87,7 +92,12 @@ def _subregion_to_edge_pixels(subregion, spectrum):
         if not spectral_axis.unit.is_equivalent(u.pix):
             right_index = ceil(subregion[1].value)
         else:
-            right_index = int(np.floor(right_func(subregion).value)) + 1
+            # Get index of closest value
+            # See https://stackoverflow.com/a/26026189 if performance becomes an issue
+            right_index = np.nanargmin((np.abs(spectral_axis - subregion[1])))
+            # Ensure index is inclusive of region bounds
+            if (spectral_axis[right_index] < subregion[1]) and (right_index < len(spectral_axis)):
+                right_index += 1
     else:
         # Convert upper value to spectrum spectral_axis units
         right_reg_in_spec_unit = right_func(subregion).to(spectral_axis.unit,
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -72,11 +72,10 @@ def _subregion_to_edge_pixels(subregion, spectrum):
 
     # Left/lower side of sub region
     if subregion[0].unit.is_equivalent(u.pix):
+        # Pixel handling assumes always ascending
         if not spectral_axis.unit.is_equivalent(u.pix):
             left_index = floor(subregion[0].value)
         else:
-            if order != "ascending":
-                raise ValueError("u.pix handling should always be ascending")
             # If the lower bound is larger than the largest value, immediately return nothing
             # Assuming ascending, both bounds are "out of bounds"
             if subregion[0] > spectral_axis[-1]:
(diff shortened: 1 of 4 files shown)


pllim:
As discussed, this could use more tests to cover edge cases.

What if region is completely out of bounds? Only 1 pix overlap? Within bounds but somehow falls between the gaps of sampled points?

pllim:
If you rebase after #1002 is merged, the CI should be green again unless this PR broke something.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -75,7 +75,19 @@ def _subregion_to_edge_pixels(subregion, spectrum):
         if not spectral_axis.unit.is_equivalent(u.pix):
             left_index = floor(subregion[0].value)
         else:
-            left_index = int(np.ceil(left_func(subregion).value))
+            if order != "ascending":
+                raise ValueError("u.pix handling should always be ascending")
```
rosteen, line 79:
```suggestion
                raise ValueError("u.pix spectral axes should always be ascending")
```

This check should probably be in `SpectralAxis.__new__()`, but either way this is clearer.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -75,7 +75,19 @@ def _subregion_to_edge_pixels(subregion, spectrum):
...
+            # If the lower bound is larger than the largest value, immediately return nothing
+            # Assuming ascending, both bounds are "out of bounds"
+            if subregion[0] > spectral_axis[-1]:
+                return None, None
+            else:
+                # Get index of closest value
+                # See https://stackoverflow.com/a/26026189 if performance becomes an issue
+                left_index = np.nanargmin((np.abs(spectral_axis - subregion[0])))
```
rosteen, line 87: Are there any cases where the spectral axis values would be NaN? That seems like something we should disallow in general.

specutils/manipulation/extract_spectral_region.py
```diff
@@ -87,7 +99,19 @@ def _subregion_to_edge_pixels(subregion, spectrum):
         if not spectral_axis.unit.is_equivalent(u.pix):
             right_index = ceil(subregion[1].value)
         else:
-            right_index = int(np.floor(right_func(subregion).value)) + 1
+            if order != "ascending":
+                raise ValueError("u.pix handling should always be ascending")
```
rosteen, line 103:
```suggestion
                raise ValueError("u.pix spectral axes should always be ascending")
```

Same comment as above.

rosteen (commented):
Mostly looks good, left a couple comments. Mainly I think the check on ascending pixel axis should go in `SpectralAxis.__new__` rather than appearing twice here.

specutils/manipulation/extract_spectral_region.py
duytnguyendtn: I dont' think there would be any "intended"/"expected" situations where there would be a NaN in the spectral axis, but unless there are preexisting checks that explicitly forbid this from happening, the default behavior of `np.argmin` to always select the `nan` is quite concerning.

Also, if we can futureproof against that case, then I say why not? `np.nanargmin` vs `np.argmin` is just one small substitution

specutils/manipulation/extract_spectral_region.py
duytnguyendtn: So the order is a local variable calculated in `_subregion_to_edge_pixels` in order to know how to calculate the edge-pixel. Are you suggesting we move the order calculation to `SpectralAxis.__new__` and store that as an attribute that `_subregion_to_edge_pixels` would access?

rosteen: I was thinking keep the ascending/descending check here, but add a similar check in `SpectralAxis` if units are u.pix to raise an error if spectral_axis[-1] < spectral_axis[0] since that is never allowed for u.pix.

rosteen: Then you don't need to check if order is `ascending` in these `u.pix` cases at all, since it's guaranteed.

duytnguyendtn: Haha I think I was massively overcomplicating this in my head. Is this what you were imagining? https://github.com/astropy/specutils/pull/1001/commits/efc106b233065996e551235ed201e9db5cd90193

rosteen (approved):
LGTM now, thanks!

PR #1002, opened 2022-12-23
TST: tox config revamp and modernization

pllim:
* Modernize tox settings
* Remove redundant jobs
* Remove options that do nothing
* Make cron job more useful (does anyone even check it?)
* Doc build is now completely done by RTD
* Grab nightly wheels from numpy, scipy, and astropy (it is faster to install but you do see a lag between dev changes and what you get from them)


.github/workflows/ci_cron_weekly.yml
```diff
@@ -21,28 +25,25 @@ jobs:
...
-          # releases.
-          - name: Python 3.10 with dev version of key dependencies
+          # We do not use remote data here, since
+          # that gives too many false positives due to URL timeouts.
+          - name: Python 3.11 with pre-release version of key dependencies and JWST
             os: ubuntu-latest
-            python: '3.10'
-            toxenv: py310-test-devdeps
+            python: '3.11'
+            toxenv: py311-test-predeps-external
```
pllim, line 33: TBH I am not sure how useful is including `jwst` here without remote data. 🤷

.github/workflows/ci_workflows.yml
```diff
@@ -18,56 +22,35 @@ jobs:
...
+          - name: Python 3.10 with astropy data and coverage
             os: ubuntu-latest
-            python: 3.8
-            toxenv: py38-test
+            python: '3.10'
+            toxenv: py310-test-cov
+            toxargs: -v
+            toxposargs: --open-files --remote-data=astropy
 
-          - name: Python 3.9 with all optional dependencies
```
pllim, line 26: I removed the redundant jobs. `alldeps` is a lie, so some of the jobs are actually equivalent. Quantity =/= quality

.github/workflows/ci_workflows.yml
```diff
@@ -18,56 +22,35 @@ jobs:
...
-          # NOTE: In the build below we also check that tests do not open and
-          # leave open any files. This has a performance impact on running the
-          # tests, hence why it is not enabled by default.
-          - name: Python 3.8 with oldest supported version of all dependencies
+          - name: Python 3.8 with oldest supported version of key dependencies
             os: ubuntu-18.04
             python: 3.8
             toxenv: py38-test-oldestdeps
 
-          - name: Python 3.8 with numpy 1.17 and full coverage
```
pllim, line 41: Numpy 1.17 pin here is also a lie. You require astropy>=5.1 which means you also require numpy>=1.18

.github/workflows/ci_workflows.yml
```diff
@@ -18,56 +22,35 @@ jobs:
...
-            python: 3.8
-            toxenv: py38-test-alldeps-numpy117-cov
-            toxposargs: --remote-data=astropy
-
-          - name: Python 3.9 with all optional dependencies (Windows)
+          - name: Python 3.9 (Windows)
             os: windows-latest
             python: 3.9
-            toxenv: py39-test-alldeps
-            toxposargs: --durations=50
```
pllim, line 51: Unless you have a campaign to weed out slow tests, there is no need to compute durations. They just slow things down.

.github/workflows/ci_workflows.yml
```diff
@@ -18,56 +22,35 @@ jobs:
...
+          - name: Python 3.8 (MacOS X)
             os: macos-latest
             python: 3.8
-            toxenv: py38-test-alldeps
-            toxposargs: --durations=50
-
-          - name: Python 3.9 doc build
-            os: ubuntu-latest
-            python: 3.9
-            toxenv: build_docs
```
pllim, line 62: Using tox to build docs is overkill. If you want to build docs locally, just do `cd docs; make html`. In the CI, RTD does the same thing and even has preview.

tox.ini
```diff
@@ -1,27 +1,18 @@
...
-indexserver =
-    NIGHTLY = https://pypi.anaconda.org/scipy-wheels-nightly/simple
 
 [testenv]
 # Suppress display of matplotlib plots generated during docs build
 setenv =
     MPLBACKEND=agg
-    # Disable the accelerate linear algebra library when running on macos as
-    # latest numpy versions do not work with it
-    NPY_BLAS_ORDER=
```
pllim, line 23: This was added 2 years ago and is no longer relevant, so I removed them.

tox.ini
```diff
@@ -40,51 +31,36 @@ changedir = .tmp/{envname}
...
     devdeps: git+https://github.com/spacetelescope/gwcs.git#egg=gwcs
 
-    external: asdf
-    external: git+https://github.com/spacetelescope/jwst@stable
+    external: jwst
 
 # The following indicates which extras_require from setup.cfg will be installed
 extras =
     test
-    alldeps: all
```
pllim, line 74: This does nothing because there is no `all` specification in your `setup.cfg`.

setup.cfg
```diff
@@ -30,9 +30,10 @@ test =
     matplotlib
     graphviz
     coverage
-    asdf
```
pllim, line 33: This is redundant because asdf is already in `install_requires`.

conftest.py
```diff
@@ -1,9 +1,34 @@
...
+    from pytest_astropy_header.display import PYTEST_HEADER_MODULES, TESTED_VERSIONS
+    ASTROPY_HEADER = True
+except ImportError:
+    ASTROPY_HEADER = False
+
 entry_points = []
 for entry_point in pkg_resources.iter_entry_points('pytest11'):
     entry_points.append(entry_point.name)
 
-if "asdf_schema_tester" not in entry_points and find_spec('asdf') is not None:
```
pllim, line 8: I don't think we need this since asdf is installed by `install_requires`? Unless this is doing some dark magic that I don't understand, @WilliamJamieson ?

WilliamJamieson: I believe this is redundant with any of the newer versions of asdf (2.8+).

pllim:
So looks like you never tested in Python 3.11 and also the coverage upload never worked. This PR would resolve both that also.

pllim:
Weekly cron somehow not reporting back but it is running at https://github.com/astropy/specutils/actions/runs/3767822486

rosteen (approved):
Looks good to me, thanks @pllim!

PR #1009, opened 2023-02-14
Fix and extend `wcs1d-fits` loader for multi-D WCS

### LINKED ISSUE #592: Writer needed for cube data
How do we save spectral cubes to file formats that are readable by other applications?
What type of api call do we implement for this?

[🐱](https://jira.stsci.edu/browse/JDAT-2199)

dhomeier:
To address https://github.com/astropy/specutils/issues/592#issuecomment-1428846666; quick fix for enabling writing of spectral cubes with celestial components in WCS.
Reading in with the full WCS is now supported as well, plus adding `mask` and `uncertainty` to the attributes read from file.
The latter at this point are still limited to the format storing them all in different ImageHDUs; an alternative common format storing the fields as sub-arrays of an N+1 image yet to be implemented.
Fixes #592

### DIFF
--- a/specutils/io/default_loaders/wcs_fits.py
+++ b/specutils/io/default_loaders/wcs_fits.py
@@ -3,7 +3,7 @@ import _io
 
 from astropy import units as u
 from astropy.io import fits
-from astropy.wcs import WCS, _wcs
+from astropy.wcs import WCS
 from astropy.modeling import models
 from astropy.utils.exceptions import AstropyUserWarning
 
@@ -36,7 +36,8 @@ def identify_wcs1d_fits(origin, *args, **kwargs):
     with read_fileobj_or_hdulist(*args, **kwargs) as hdulist:
         return (hdulist[hdu].header.get('WCSDIM', 1) == 1 and
                 (hdulist[hdu].header['NAXIS'] == 1 or
-                 hdulist[hdu].header.get('WCSAXES', 0) == 1 )and not
+                hdulist[hdu].header.get('WCSAXES', 0) == 1 or
+                hdulist[hdu].header.get('DISPAXIS', -1) >= 0 ) and not
                 hdulist[hdu].header.get('MSTITLE', 'n').startswith('2dF-SDSS LRG') and not
                 # Check in CTYPE1 key for linear solution (single spectral axis)
                 hdulist[hdu].header.get('CTYPE1', 'w').upper().startswith('MULTISPE'))
@@ -45,7 +46,7 @@ def identify_wcs1d_fits(origin, *args, **kwargs):
 @data_loader("wcs1d-fits", identifier=identify_wcs1d_fits,
              dtype=Spectrum1D, extensions=['fits', 'fit'], priority=5)
 def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
-                      hdu=0, verbose=False, **kwargs):
+                      hdu=None, verbose=False, **kwargs):
     """
     Loader for single spectrum-per-HDU spectra in FITS files, with the spectral
     axis stored in the header as FITS-WCS.  The flux unit of the spectrum is
@@ -67,9 +68,9 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
         Units of the flux for this spectrum. If not given (or None), the unit
         will be inferred from the BUNIT keyword in the header. Note that this
         unit will attempt to convert from BUNIT if BUNIT is present.
-    hdu : int
-        The index of the HDU to load into this spectrum.
-    verbose : bool
+    hdu : int, str or None. optional
+        The index or name of the HDU to load into this spectrum (default: search 1st).
+    verbose : bool. optional
         Print extra info.
     **kwargs
         Extra keywords for :func:`~specutils.io.parsing_utils.read_fileobj_or_hdulist`.
@@ -86,6 +87,17 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
         print("Spectrum file looks like wcs1d-fits")
 
     with read_fileobj_or_hdulist(file_obj, **kwargs) as hdulist:
+        if hdu is None:
+            for ext in ('FLUX', 'SCI', 'PRIMARY'):
+                # For now rely on extension containing spectral data.
+                if ext in hdulist and (
+                        isinstance(hdulist[ext], (fits.ImageHDU, fits.PrimaryHDU)) and
+                        hdulist[ext].data is not None):
+                    hdu = ext
+                    break
+            else:
+                raise ValueError('No HDU with spectral data found.')
+
         header = hdulist[hdu].header
         wcs = WCS(header)
 
@@ -95,6 +107,12 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
                 data = data.to(flux_unit)
         else:
             data = u.Quantity(hdulist[hdu].data, unit=flux_unit)
+
+        # TODO: add additional HDU data like 'IVAR'
+        if 'MASK' in hdulist:
+            mask = hdulist['MASK'].data
+        else:
+            mask = None
 
     if spectral_axis_unit is not None:
         wcs.wcs.cunit[0] = str(spectral_axis_unit)
@@ -119,11 +137,11 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
     if wcs.naxis > 4:
         raise ValueError('FITS file input to wcs1d_fits_loader is > 4D')
 
-    return Spectrum1D(flux=data, wcs=wcs, meta=meta)
+    return Spectrum1D(flux=data, wcs=wcs, mask=mask, meta=meta)
 
 
 @custom_writer("wcs1d-fits")
-def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False, **kwargs):
+def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False, flux_name='FLUX', **kwargs):
     """
     Write spectrum with spectral axis defined by its WCS to (primary)
     IMAGE_HDU of a FITS file.
@@ -133,14 +151,16 @@ def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False, **kwargs)
     spectrum : :class:`~specutils.Spectrum1D`
     file_name : str
         The path to the FITS file
-    hdu : int
+    hdu : int, optional
         Header Data Unit in FITS file to write to (base 0; default primary HDU)
-    update_header : bool
+    update_header : bool, optional
         Update FITS header with all compatible entries in `spectrum.meta`
-    unit : str or :class:`~astropy.units.Unit`
-        Unit for the flux (and associated uncertainty)
-    dtype : str or :class:`~numpy.dtype`
-        Floating point type for storing flux array
+    flux_name : str, optional
+        HDU name to store flux spectrum under (default 'FLUX')
+    unit : str or :class:`~astropy.units.Unit`, optional
+        Unit for the flux (and associated uncertainty; defaults to `spectrum.flux.unit`)
+    dtype : str or :class:`~numpy.dtype`, optional
+        Floating point type for storing flux array (defaults to `spectrum.flux.dtype`)
     """
     # Create HDU list from WCS
     try:
@@ -160,7 +180,7 @@ def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False, **kwargs)
     if np.abs(dwl).max() > 1.e-10:
         m = np.abs(dwl).argmax()
         raise ValueError('Relative difference between WCS spectral axis and'
-                         f'spectral_axis at {m:}: dwl[m]')
+                         f'spectral_axis at {m:}: {dwl[m]}')
 
     if update_header:
         hdr_types = (str, int, float, complex, bool,
@@ -180,6 +200,15 @@ def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False, **kwargs)
     funit = u.Unit(kwargs.pop('unit', spectrum.flux.unit))
     flux = spectrum.flux.to(funit, equivalencies=u.spectral_density(wl))
     hdulist[0].data = flux.value.astype(ftype)
+    if flux_name is not None:
+        hdulist[0].name = flux_name
+
+    # Append mask array (duplicate WCS for that extension)?
+    if spectrum.mask is not None:
+        hdulist.append(wcs.to_fits()[0])
+        hdulist[1].data = spectrum.mask
+        hdulist[1].name = 'MASK'
+        hdu += 1
 
     if hasattr(funit, 'long_names') and len(funit.long_names) > 0:
         comment = f'[{funit.long_names[0]}] {funit.physical_type}'
--- a/specutils/io/default_loaders/wcs_fits.py
+++ b/specutils/io/default_loaders/wcs_fits.py
@@ -118,12 +118,6 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
 
     if wcs.naxis > 4:
         raise ValueError('FITS file input to wcs1d_fits_loader is > 4D')
-    elif wcs.naxis > 1:
-        for i in range(wcs.naxis - 1, 0, -1):
-            try:
-                wcs = wcs.dropaxis(i)
-            except _wcs.NonseparableSubimageCoordinateSystemError as e:
-                raise ValueError(f'WCS cannot be reduced to 1D: {e!r} {wcs}')
 
     return Spectrum1D(flux=data, wcs=wcs, meta=meta)
 
--- a/specutils/io/default_loaders/wcs_fits.py
+++ b/specutils/io/default_loaders/wcs_fits.py
@@ -158,7 +158,11 @@ def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False, **kwargs)
 
     # Verify spectral axis constructed from WCS
     wl = spectrum.spectral_axis
-    dwl = (wcs.all_pix2world(np.arange(len(wl)), 0) - wl.value) / wl.value
+    # Not sure why the extra check is necessary for FITS WCS
+    if hasattr(wcs, 'celestial') and wcs.celestial.naxis > 0:
+        dwl = (wcs.spectral.all_pix2world(np.arange(len(wl)), 0) - wl.value) / wl.value
+    else:
+        dwl = (wcs.all_pix2world(np.arange(len(wl)), 0) - wl.value) / wl.value
     if np.abs(dwl).max() > 1.e-10:
         m = np.abs(dwl).argmax()
         raise ValueError('Relative difference between WCS spectral axis and'
--- a/specutils/io/default_loaders/wcs_fits.py
+++ b/specutils/io/default_loaders/wcs_fits.py
@@ -5,6 +5,7 @@ from astropy import units as u
 from astropy.io import fits
 from astropy.wcs import WCS
 from astropy.modeling import models
+from astropy.nddata import StdDevUncertainty, InverseVariance, VarianceUncertainty
 from astropy.utils.exceptions import AstropyUserWarning
 
 import numpy as np
@@ -15,6 +16,14 @@ from ..registers import data_loader, custom_writer
 from ..parsing_utils import read_fileobj_or_hdulist
 
 __all__ = ['wcs1d_fits_loader', 'non_linear_wcs1d_fits', 'non_linear_multispec_fits']
+
+UNCERT_REF = {'STD': StdDevUncertainty,
+              'ERR': StdDevUncertainty,
+              'UNCERT': StdDevUncertainty,
+              'VAR': VarianceUncertainty,
+              'IVAR': InverseVariance}
+
+UNCERT_EXP = {'std': 1, 'var': 2, 'ivar': -2}
 
 
 def identify_wcs1d_fits(origin, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


pllim:
So this fixes #592 ?

dhomeier:
This should work for your example.
Reading the full WCS in rather than only the spectral dimension may of course have consequences if `spectrum.wcs` is used at a later point. Another immediate change is that this will now only accept the official FITS spectral axis keywords recognised my WCSLIB, `FREQ`, `AFRQ`, `ENER`, `WAVN`,`VRAD`, `WAVE`, `VOPT`, `ZOPT`, `AWAV`, `VELO`, `BETA`.
The tests are adapted, but if downstream applications have used names like `wavelength` or `wavenumber`, this may need additional pre-processing of the header.

dhomeier:
I was somewhat surprised when updating the writer, that there are apparently WCS that have a `spectral` component, yet
`wcs.spectral.all_pix2world(pix)` returns a different (wrong) result than `wcs.all_pix2world(pix)` (the former fails on all other tests that only input a 1D, pure spectral WCS). But the check should probably be there anyway, if it ever has to deal with non-FITS WCS.

specutils/tests/test_loaders.py
```diff
@@ -707,11 +707,10 @@ def test_tabular_fits_compressed(compress, tmpdir):
 
 
 @pytest.mark.parametrize("spectral_axis",
-                         ['wavelength', 'frequency', 'energy', 'wavenumber'])
+                         ['WAVE', 'FREQ', 'ENER', 'WAVN'])
```
nmearl, line 710: Just some minor context: if I recall, the spectral axis names were meant to correspond with the class properties and not the FITS naming standard. Either way, it's not an issue, but figured the history of the usage might add some context.

dhomeier:
I have added read and write for `spectrum.mask` and given the HDUs names, handling it I think more similar to how the cubeviz loader treats its files. Things like detection strategy for the spectrum HDU (and mask? Should it be found under other names as well?) may need some more discussion, also perhaps more test coverage for the naming schemes.
In principle this could probably extended in a similar way to also handle uncertainties.

dhomeier:
Apparently something in oldestdeps is pulling numpy 1.24.2 in, overriding `numpy==1.19` and breaking the old scipy version. I guess that means 1.19 is not really supported anymore, but don't know what it is that's requiring a newer version.

dhomeier:
Marking this as ready for review now, as the multi-HDU format for masks and uncertainties is functional (unless there are more labels the loader needs to identify).
The format reading different attributes from `data` slices in a single HDU, which would work similarly to 6dfgs
https://github.com/astropy/specutils/blob/6eb7f96498072882c97763d4cd10e07cf81b6d33/specutils/io/default_loaders/sixdfgs_reader.py#L188-L189
could still be added here or in a follow-up PR.

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -27,25 +36,25 @@ def identify_wcs1d_fits(origin, *args, **kwargs):
...
+                (args[0].endswith(('.fits', '.fit')) and whdu == 0))
 
     hdu = kwargs.get('hdu', 0)
     # Check if number of axes is one and dimension of WCS is one
     with read_fileobj_or_hdulist(*args, **kwargs) as hdulist:
         return (hdulist[hdu].header.get('WCSDIM', 1) == 1 and
                 (hdulist[hdu].header['NAXIS'] == 1 or
-                 hdulist[hdu].header.get('WCSAXES', 0) == 1 )and not
+                hdulist[hdu].header.get('WCSAXES', 0) == 1 or
+                hdulist[hdu].header.get('DISPAXIS', -1) >= 0 ) and not
```
pllim, line 48: Indentation should match inner parenthesis.
```suggestion
                 hdulist[hdu].header.get('WCSAXES', 0) == 1 or
                 hdulist[hdu].header.get('DISPAXIS', -1) >= 0 ) and not
```

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -67,10 +76,17 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
         Units of the flux for this spectrum. If not given (or None), the unit
         will be inferred from the BUNIT keyword in the header. Note that this
         unit will attempt to convert from BUNIT if BUNIT is present.
-    hdu : int
-        The index of the HDU to load into this spectrum.
-    verbose : bool
+    hdu : int, str or None, optional
+        The index or name of the HDU to load into this spectrum
+        (default: find 1st applicable `ImageHDU`).
```
pllim, line 81: This confuses me. Is this loader only for *image*? A spectrum can also come as table HDU, right?

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -96,6 +130,49 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
...
 
+        # Read mask from specified HDU or try to auto-detect it.
+        mask = None
+        if mask_hdu is True:
+            for ext in ('MASK', 'DQ', 'QUALITY'):
+                if ext in hdulist and (isinstance(hdulist[ext], fits.ImageHDU) and
+                        hdulist[ext].data is not None):
+                    mask = hdulist[ext].data
+                    break
+        elif isinstance(mask_hdu, (int, str)):
```
pllim, line 141: Why would mask be represented as `str`?

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -139,14 +212,20 @@ def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False, **kwargs)
...
-    dtype : str or :class:`~numpy.dtype`
-        Floating point type for storing flux array
+    flux_name : str, optional
+        HDU name to store flux spectrum under (default 'FLUX')
+    mask_name : str or `None`, optional
+        HDU name to store mask under (default 'MASK'; `None`: do not save)
+    uncertainty_name : str or `None`, optional
+        HDU name to store uncertainty under (default set from type; `None`: do not save)
+    unit : str or :class:`~astropy.units.Unit`, optional
+        Unit for the flux (and associated uncertainty; defaults to `spectrum.flux.unit`)
```
pllim, line 226: Unless `spectrum.flux.unit` actually generates API doc, use double-backticks. I cannot check because RTD build failed on your PR. Also see #1028.

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -157,12 +236,16 @@ def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False, **kwargs)
         raise ValueError(f'Only Spectrum1D objects with valid WCS can be written as wcs1d: {err}')
 
     # Verify spectral axis constructed from WCS
-    wl = spectrum.spectral_axis
-    dwl = (wcs.all_pix2world(np.arange(len(wl)), 0) - wl.value) / wl.value
-    if np.abs(dwl).max() > 1.e-10:
-        m = np.abs(dwl).argmax()
+    disp = spectrum.spectral_axis
+    # Not sure why the extra check is necessary for FITS WCS
```
pllim, line 240: This comment does not inspire confidence. Can someone well-versed with spectrum WCS chime in here?

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -171,17 +254,46 @@ def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False, **kwargs)
...
     # Add flux array and unit
     ftype = kwargs.pop('dtype', spectrum.flux.dtype)
     funit = u.Unit(kwargs.pop('unit', spectrum.flux.unit))
-    flux = spectrum.flux.to(funit, equivalencies=u.spectral_density(wl))
+    flux = spectrum.flux.to(funit, equivalencies=u.spectral_density(disp))
     hdulist[0].data = flux.value.astype(ftype)
+    if flux_name is not None:
+        hdulist[0].name = flux_name
+
+    # Append mask array (duplicate WCS for that extension)?
```
pllim, line 265: Usually mask only makes sense in pixel space. Putting WCS on that extension would be misleading. In HST images, I think the WCS is only in the SCI extensions.

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -171,17 +254,46 @@ def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False, **kwargs)
...
+            hdulist[-1].name = 'MASK'
+        else:
+            hdulist[-1].name = mask_name
+        hdu += 1
+    # Warn if extension name for saving was explicitly chosen.
+    elif mask_name is not None and mask_name != '':
+        warnings.warn("No mask found in this Spectrum1D, none saved.", AstropyUserWarning)
+
+    # Append uncertainty array (duplicate WCS for that extension)?
+    if spectrum.uncertainty is not None and uncertainty_name is not None:
```
pllim, line 279: Should we even let user rename the uncertainty HDU? Would be misleading if `spectrum.uncertainty` is STD but then someone names it "IVAR".

specutils/tests/test_loaders.py
```diff
@@ -733,13 +747,98 @@ def test_wcs1d_fits_writer(tmpdir, spectral_axis):
...
+    flux = np.arange(1, 121).reshape((10, 4, 3))**2 * 1.e-14 * u.Jy
+    wlu = u.Unit(hdr['CUNIT1'])
+    wl0 = hdr['CRVAL1']
+    dwl = hdr['CDELT1']
+    disp = np.arange(wl0, wl0 + (flux.shape[2] - 0.5) * dwl, dwl) * wlu
+    tmpfile = str(tmpdir.join('_tst.fits'))
+
+    if with_mask:
+        die = np.random.Generator(np.random.MT19937(23))
+        mask = die.choice([0, 0, 0, 0, 0, 1], size=flux.shape).astype(np.uint16)
```
pllim, line 797: Isn't it simpler to start with all zeroes and then assign select slices to ones? Is a random number generator really necessary? Looks cool though.

pllim:
Thanks for working on this! I left some comments above.

Also, does this page need updating? It is not very clear to me exactly which data will go through this reader/writer.

https://specutils.readthedocs.io/en/latest/spectrum1d.html#list-of-loaders

specutils/io/default_loaders/wcs_fits.py
dhomeier: wcs1d handles only spectra with flux represented as image and spectral axis defined by the WCS; tables are covered by `tabular-fits`. That already supports masks and uncertainties, more or less, which are simply additional table columns in that representation.

specutils/io/default_loaders/wcs_fits.py
dhomeier: `mask_hdu` is the generalised index to its HDU, which for a HDUList can be the numerical index or the label from `EXTNAME`.

specutils/io/default_loaders/wcs_fits.py
dhomeier: [`spectrum.flux`](https://specutils.readthedocs.io/en/stable/api/specutils.Spectrum1D.html#specutils.Spectrum1D.flux) does, but not specifically `unit`, is that what you mean? I don't know if the loaders themselves generate API docs, or rather where they might be linked from ;-)
Thanks for pointing out the RTD failure; throws in all current PRs and does not seem related – from a quick glance some parsing error related to `reproject`, but needs further investigation.

specutils/io/default_loaders/wcs_fits.py
dhomeier: The comment refers to my expectation for a WCS with a single, spectral dimension, that has a `spectral` (and thus also a `celestial`) attribute, to return the same world coordinates via that `wcs.spectral` as from `wcs` directly.
But that case is also handled separately in
https://github.com/astropy/specutils/blob/main/specutils/spectra/spectrum1d.py#L277-L284
although perhaps the coverage there is incomplete, don't know.

specutils/io/default_loaders/wcs_fits.py
dhomeier: Yes, looking at the Mamba format I don't see any WCS info in MASK or IVAR; was just wondering what might happen when something is reading those HDUs separately, finding celestial axes without any coordinates.

specutils/io/default_loaders/wcs_fits.py
dhomeier: Good point. There are at least 3 commonly used designations for StdDev, and we probably should also expect things like 'VARIAN' or 'INVVAR' for the others, so if anyone insists they have always used naming convention X in their project and it needs to stay that way, we should not prohibit that. Ultimately should be the responsibility of the user to use sensible values there _if_ they override the defaults, but I should probably add a check that at least it does not conflict with `UNCERT_REF`.

specutils/tests/test_loaders.py
dhomeier: 😺 Maybe `5 * [0] + [1]` would be cooler still, but I did not really care as long as it fits on one line.
Do you mean is it necessary to use random numbers at all – probably not, just wanted to make up some more "realistic" example? Or to use `Generator` – that is how the `np.random` docs recommend it now.

dhomeier:
Thanks for the review!

> Also, does this page need updating? It is not very clear to me exactly which data will go through this reader/writer.

It's the same `wcs1d-fits`, and ideally should handle all WCS (ImageHDU) spectra that do not have a more specialised reader. Auto-identify will likely not work (yet) e.g. for WCS with non-spectral components, but I think that column generally is not meant to say that it works for _all_ cases.

specutils/io/default_loaders/wcs_fits.py
dhomeier: Actually it seems none of the functions in `default_loaders` have any docs built at all, though it would certainly be useful to have them and link them in https://specutils.readthedocs.io/en/latest/spectrum1d.html#list-of-loaders.

specutils/io/default_loaders/wcs_fits.py
dhomeier: Added to tests now, but nonstandard names do complicate everything quite a bit. Wondering if the uncertainty should actually written with the `BUNIT` – examples like MaNGA don't, but it would help with identifying the correct type (and also take care of the case that it might not match the flux unit in the file).

pllim:
Sorry, I think my fixes to the tests caused a conflict here, so you need to rebase and resolve the conflict.

specutils/io/default_loaders/wcs_fits.py
pllim: Yeah, trying to appease all the random file formats out there is going to be impossible.

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -356,7 +483,7 @@ def _read_non_linear_iraf_fits(file_obj, spectral_axis_unit=None, flux_unit=None
 
     Returns
     -------
-    Tuple of data to pass to SpectrumCollection() or Spectrum1D():
+    Tuple of data to pass to `SpectrumCollection`() or `Spectrum1D`():
```
pllim, line 486:
```suggestion
    Tuple of data to pass to `~specutils.SpectrumCollection` or `~specutils.Spectrum1D`:
```

pllim: p.s. Does not really matter because this is a hidden function, so not rendered by API doc. But might as well do it right?

specutils/tests/test_loaders.py
```diff
@@ -805,34 +955,35 @@ def test_wcs1d_fits_multid(tmp_path, spectral_axis):
...
-    tmpfile = str(tmp_path / f'_{2}d.fits')
+    tmpfile = tmp_path / f'wcs_{2}d.fits'
     spectrum.write(tmpfile, format='wcs1d-fits')
 
     # Reopen file and update header with off-diagonal element
-    with fits.open(tmpfile, mode='update') as hdulist:
-        hdulist[0].header.update([('PC1_2', 0.2)])
+    hdulist = fits.open(tmpfile, mode='update')
+    hdulist[0].header.update([('PC1_2', 0.2)])
+    hdulist.close()
```
pllim, line 983:
```suggestion
    with fits.open(tmpfile, mode='update') as hdulist:
        hdulist[0].header.update([('PC1_2', 0.2)])
```

specutils/tests/test_loaders.py
```diff
@@ -875,7 +1026,7 @@ def test_wcs1d_fits_compressed(compress, tmp_path):
     # Try again without compression suffix:
     with warnings.catch_warnings():
         warnings.simplefilter('ignore', FITSFixedWarning)
-        os.system(f'mv {tmpfile}{ext[compress]} {tmpfile}')
+        os.system(f'mv {tmpfile}.{ext[compress]} {tmpfile}')
```
pllim, line 1029: Side note: Why call `os.system` instead of `shutil`? Anyways, I guess fixing this is out of scope.

pllim (commented):
I left what general review comments I could. Now it is up to specutils maintainers to review. Thanks!

specutils/io/default_loaders/wcs_fits.py
dhomeier: Yes (none of the default_loaders are rendered anywhere afaics).

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -157,12 +246,16 @@ def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False, **kwargs)
         raise ValueError(f'Only Spectrum1D objects with valid WCS can be written as wcs1d: {err}')
```
rosteen, line 246: I'm not _that_ familiar with the workings of WCS (and especially `wcs.to_fits()`) - is there any realistic scenario for setting `hdu` to a value other than 0, given that the initial `hdulist` is always created from the spectrum's WCS anyway? I guess what I'm asking is, is there really utility in having empty HDUs in front of the specified `hdu` index? There are no options to specify the HDU numbers for uncertainty and mask, so it's not like the user could write out in order [uncertainty, mask, flux], for example. Maybe we just get rid of the `hdu` argument altogether?

specutils/io/default_loaders/wcs_fits.py
```diff
@@ -86,6 +106,24 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
...
+                if ext in hdulist and (
+                        isinstance(hdulist[ext], (fits.ImageHDU, fits.PrimaryHDU)) and
+                        hdulist[ext].data is not None):
+                    if hdu is None or hdulist[ext] == hdulist[hdu]:
+                        hdu = ext
+                        continue
+                    else:
+                        warnings.warn(f"Found multiple data HDUs '{hdu}' and '{ext}', "
+                                      f"will read '{hdu}'! Please use `hdu=<flux_hdu>` "
+                                      "to select a specific one.", AstropyUserWarning)
```
rosteen, line 121:
```suggestion
                                      f"will read '{hdu}'. Please use `hdu=<flux_hdu>` "
                                      "to select a specific extension.", AstropyUserWarning)
```
Minor quibble, would rather be specific ("extension" or "hdu") here than say "one".

specutils/io/default_loaders/wcs_fits.py
rosteen: Agreed, we can only accommodate so much.

specutils/tests/test_loaders.py
rosteen: +1 to this suggestion.

rosteen:
Thanks @dhomeier, this is a great improvement. I had one question and left one suggestion, and there's one suggestion from @pllim remaining, but other than that I think this looks good. I appreciate the test coverage. 

I was going to say, maybe we change the name from `wcs1d_fits` to just `wcs_fits` if it handles multi-D WCS now, but perhaps we go with the standard answer of "1D means one spectral dimension" like in `Spectrum1D`.

specutils/io/default_loaders/wcs_fits.py
pllim: As long as this is not considered "public" API and ok to break just like that, I agree.

specutils/tests/test_loaders.py
pllim: I don't see how "mv" cares about compression? https://docs.python.org/3/library/shutil.html#shutil.move

specutils/io/default_loaders/wcs_fits.py
dhomeier: > I'm not _that_ familiar with the workings of WCS (and especially `wcs.to_fits()`) - is there any realistic scenario for setting `hdu` to a value other than 0, given that the initial `hdulist` is always created from the spectrum's WCS anyway?

It's not as much associated to using a WCS, but rather to the flux being stored in an ImageHDU, which can be the PrimaryHDU or an extension, while tables can only be stored in extensions. However for symmetry reasons or whatever sometimes writing the image to an extension, too, and using just a basic PrimaryHDU, is desired (there was one comment somewhere sketching just this scenario, but I cannot find it either in the Slack discussion or any related issue now).
I think originally accepting any value for `hdu` was just introduced as a poor substitute for not being able to write multiple spectra to one file/hdulist, but there is probably no real use case for this. For `hdu=1` I'd say there is, and this should should still remain available.

specutils/tests/test_loaders.py
dhomeier: Not at all, I guess, but it's kind of the direct counterpart to `os.system(f'{compress} {tmpfile}')` just above.

specutils/io/default_loaders/wcs_fits.py
rosteen: Thanks for the explanation...I guess it makes sense to leave it in then for now.

rosteen:
@dhomeier Can I commit the last suggestions for you and merge this?

rosteen (approved):
Approved, need to commit last two suggestions before merging.

pllim:
Hmm does this mean we should ditch Jdaviz custom writer and use this?

rosteen:
> > @dhomeier Can I commit the last suggestions for you and merge this?
> 
> I've included them in some more local updates I had made; should be good to merge then.

Blast, I went ahead and committed and merged, thinking you might be done and logged off for the day. Are your other updates worth opening a follow-up PR for? Sorry about that.

dhomeier:
Ah, remote tests are taking too long here (spPlate really bogs everything down).
Will think about the follow-up, the extra test might be good to have, but now I am going to be offline for a while.

pllim:
Thanks, @dhomeier ! I am trying out this improved format over at https://github.com/spacetelescope/jdaviz/pull/2094 but I encountered error. Perhaps we need an extra step to convert bool mask to uint16 (I guess it could be uint8 but uint16 is common in bit-plane format)?

```
>       fitted_spectrum.write(out_fn, format="wcs1d-fits", overwrite=True)

jdaviz/configs/default/plugins/model_fitting/tests/test_fitting.py:229: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
astropy/nddata/mixins/ndio.py:103: in __call__
    self.registry.write(self._instance, *args, **kwargs)
astropy/io/registry/core.py:383: in write
    return writer(data, *args, **kwargs)
specutils/io/default_loaders/wcs_fits.py:278: in wcs1d_fits_writer
    hdulist[-1].data = spectrum.mask
astropy/utils/decorators.py:848: in __set__
    ret = self.fset(obj, val)
...
E           KeyError: 'bool'
```

https://github.com/spacetelescope/jdaviz/actions/runs/4451633462/jobs/7818514722?pr=2094

Issue #1011, opened 2023-02-15
ASDF legacy converter system is being deprecated


WilliamJamieson:
- [ ] Also revert https://github.com/astropy/specutils/pull/1021 if applicable
- [ ] Also revert https://github.com/astropy/specutils/pull/1024 if applicable

ASDF if planning on releasing one final version of ASDF 2, version 2.15 which will be an LTS release. Several things in ASDF 2 are slated to be removed in ASDF 3, and so deprecation warnings are being added to these things as part of the 2.15 release.

In particular, the `asdf.types.CustomType` class is being deprecated (see asdf-format/asdf#1359 for details), which `specutils` uses here https://github.com/astropy/specutils/blob/34b5e84fd107f31a3c2fe2d1aaf95517d68d95f1/specutils/io/asdf/types.py#L20 Subclassing `CustomType` is how the legacy converter system for extensions ASDF operates.

The new converter system, based on the `asdf.extension.Converter` class, is the replacement for this system. Effectively, this means much of the `specutils.io.asdf` ASDF extension will need to be refactored in a similar way to how https://github.com/astropy/asdf-astropy is a refactor of `astropy.io.misc.asdf` using this "new" converter system.

The ASDF maintainers are willing to assist `specutils` in performing this refactor.

[🐶](https://jira.stsci.edu/browse/AL-718)

Also see:

* #1006

pllim:
Is SCSB still willing to work on this part? If not, please give me an example PR with the changes I need to make. Thanks!

PR #1012, opened 2023-02-16
Catch warning from deprecation of `CustomType`

WilliamJamieson:
In issue #1011, it was mentioned the `CustomType` was going to be deprecated by ASDF. This has now occurred meaning testing with `specutils` installed under the development version of ASDF will now cause warnings to be emitted by `specutils` every time an ASDF file is loaded. This PR seeks to resolve this warning while a permanent fix  for #1011 can be created.

### DIFF
--- a/specutils/io/asdf/extension.py
+++ b/specutils/io/asdf/extension.py
@@ -3,14 +3,23 @@ Defines extension that is used by ASDF for recognizing specutils types
 """
 import os
 import urllib
+import warnings
 
 from asdf.util import filepath_to_url
 from asdf.extension import AsdfExtension
-
-from astropy.io.misc.asdf.extension import ASTROPY_SCHEMA_URI_BASE
+from asdf.exceptions import AsdfDeprecationWarning
 
 from .tags.spectra import *  # noqa
 from .types import _specutils_types
+
+
+with warnings.catch_warnings():
+    warnings.filterwarnings(
+        "ignore",
+        category=AsdfDeprecationWarning,
+        message=r".*from astropy.io.misc.asdf.* subclasses the deprecated CustomType .*",
+    )
+    from astropy.io.misc.asdf.extension import ASTROPY_SCHEMA_URI_BASE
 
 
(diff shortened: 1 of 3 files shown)


WilliamJamieson:
The failures in the CI seem irrelevant to the changes in this PR.

WilliamJamieson:
Also, since the CI does not test against dev dependencies, I manually tested this with the development version of ASDF and can confirm that this will catch the warning correctly.

specutils/io/asdf/extension.py
```diff
@@ -3,16 +3,25 @@
...
 from .types import _specutils_types
 
 
+with warnings.catch_warnings():
+    warnings.filterwarnings(
+        "ignore",
+        category=AsdfDeprecationWarning,
+        message=r".*from astropy.io.misc.asdf.* subclasses the deprecated CustomType .*",
+    )
+    from astropy.io.misc.asdf.extension import ASTROPY_SCHEMA_URI_BASE
```
pllim, line 22: Can this just import from `asdf-astropy` while you are at it?

WilliamJamieson: No, that does not exist in asdf-astropy.

pllim: Wait, what? But `astropy.io.misc.asdf` will go away. What should specutils use then?

WilliamJamieson: You don't need this for `asdf-astropy` to function. The entire extension system which supports `astropy.io.misc.asdf` and apparently `specutils` is deprecated and going away. If #1011 is resolved there is no need for this.

WilliamJamieson: To be clear, packages using `astropy.io.misc.asdf` cannot simply switch to using `asdf-astropy`. They will have to re-write their extensions to use the non-legacy ASDF extension system instead.

pllim (approved):
The diff looks uncontroversial. #1011 is a battle for another day. Failure is unrelated. Thanks!

Issue #849, opened 2021-08-09
TST: Do not skip valid doctest examples


pllim:
For example, I see you download real data in `docs/spectral_cube.rst` but then you skip testing actually using the object generated by the data.

```python
>>> sc.shape #doctest:+SKIP
(74, 74, 4563)
```

Why not just run the whole example with doctest?

dhomeier:
I agree checking the shape should be feasible at a minimum.
For the actual content might be tricky to set up the `+FLOAT_CMP` etc. the right way, but I think there are some examples of this in the docs.

PR #1015, opened 2023-02-21
Pin matplotlib in oldestdeps job

### LINKED ISSUE #1014: TST: oldest dep job fail because numpy upgraded when not supposed to
It is pulling in the latest numpy, not the oldest supported. Someone needs to fix the pins for that job. Maybe some other package not pinned needs to be pinned to an older version.

pllim:
matplotlib 3.7 bumped minversion of numpy.

Fix #1014

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -46,6 +46,7 @@ deps =
     oldestdeps: asdf-astropy==0.3.*
     oldestdeps: asdf==2.10.*
     oldestdeps: ndcube==2.0.*
+    oldestdeps: matplotlib==3.6.*
 
     devdeps: numpy>=0.0.dev0
     devdeps: scipy>=0.0.dev0
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -30,7 +30,7 @@ jobs:
             toxposargs: --open-files --remote-data=astropy
 
           - name: Python 3.8 with oldest supported version of key dependencies
-            os: ubuntu-18.04
+            os: ubuntu-20.04
             python: 3.8
             toxenv: py38-test-oldestdeps
 


tox.ini
```diff
@@ -46,6 +46,7 @@ deps =
     oldestdeps: asdf-astropy==0.3.*
     oldestdeps: asdf==2.10.*
     oldestdeps: ndcube==2.0.*
+    oldestdeps: matplotlib==3.6.*
```
dhomeier, line 49: If this is really the oldest supported version (and is it?), should that be reflected in the requirements in `setup.cfg`?
Asking since mamba wanted matplotlib 3.5.* for numpy 1.19, and tests are still passing with 3.5.3.

pllim: No, that is not the oldest supported version. It is actually the maxversion we can allow for matplotlib for the oldest supported version. According to setup.cfg , matplotlib is only a test dependency.

dhomeier: Yes, but if installed as test dependency (and not necessarily though tox), should not that also ensure a supported version is installed? And now that a version is specified in `oldestdeps`, I would expect, well, the actual oldest supported version to be tested (may be 3.1.0, which is the oldest I could reliably install and test with Python 3.8, or even older); just as it is done for numpy etc.

dhomeier: Well, actually the tests are running (with the same 435 passing w/o remote_data) without any matplotlib installed, as it is only ever imported in the docs. So maybe the question rather is whether it belongs into `test` deps at all (especially with no doctests configured).

pllim: Pretty sure specutils runs doctest because I just fixed those.

pllim: I don't want to get into the debate of minversion of matplotlib specutils is supposed to support. I don't think that is very relevant. Would a comment on this line in `tox.ini` enough?

dhomeier: If it is nowhere tested but just used for plotting out put in the docs, I suppose the version is not critical.
Still wondering if it is then needed as `test` dep at all, since `linkcheck` (which is the only doc-related test I could find) already pulls in `extras: docs` anyway.

dhomeier: (there existed a `build_docs` job at the time when #849 was open, but I cannot see any obvious replacement for that)

pllim: I think this setting pulls in `test` dependencies for all the jobs:

https://github.com/astropy/specutils/blob/6eb7f96498072882c97763d4cd10e07cf81b6d33/tox.ini#L62-L63

dhomeier: Yes, that is clear; was just wondering if anything running doctests would also pull in `extras = docs` as `linkcheck` does. Answer is 'no'; turns out I misread the
`!cov: pytest --pyargs specutils {toxinidir}/docs {posargs}`
command for a commented line! (Maybe one should go Spanish and use `¡` for `not` ;-)

pllim: So, do you still need a comment in `tox.ini`?

dhomeier: I have reconstructed now how it is a documentation dependency that is required for all tests; no strong opinion either way if no one else has stumbled over it.

PR #1017, opened 2023-02-21
basic code example and links to Spectrum1D.write

kecnry:
This PR adds a very brief example of how to use `Spectrum1D.write` and links to the API docs within the main spectrum1d landing page on the docs.

See [rendered docs from this PR](https://specutils--1017.org.readthedocs.build/en/1017/spectrum1d.html#writing-to-a-file)

### DIFF
--- a/docs/custom_loading.rst
+++ b/docs/custom_loading.rst
@@ -143,6 +143,8 @@ many `~specutils.Spectrum1D` objects. This method is available since
 `astropy.io.registry.read`).
 
 
+.. _custom_writer:
+
 Creating a Custom Writer
 ------------------------
 
--- a/docs/spectrum1d.rst
+++ b/docs/spectrum1d.rst
@@ -87,6 +87,18 @@ A full list of the supported formats is shown in the table below.
 
 | More information on creating custom loaders can be found in the :doc:`custom loading </custom_loading>` page.
 
+Writing to a File
+-----------------
+
+Similarly, a `~specutils.Spectrum1D` object can be saved to any file format supported by
+:class:`~astropy.nddata.CCDData` via the :meth:`specutils.Spectrum1D.write` method.
+
+.. code-block:: python
+
+    >>> from specutils import Spectrum1D
+    >>> spec1d = Spectrum1D.read("/path/to/input.fits")  # doctest: +SKIP
+    >>> spec1d.write("/path/to/output.fits")  # doctest: +SKIP
+
 
 Including Uncertainties
 -----------------------
--- a/docs/spectrum1d.rst
+++ b/docs/spectrum1d.rst
@@ -90,15 +90,14 @@ A full list of the supported formats is shown in the table below.
 Writing to a File
 -----------------
 
-Similarly, a `~specutils.Spectrum1D` object can be saved to any file format supported by
-:class:`~astropy.nddata.CCDData` via the :meth:`specutils.Spectrum1D.write` method.
+Similarly, a `~specutils.Spectrum1D` object can be saved to any of the supported formats using the
+:meth:`specutils.Spectrum1D.write` method.
 
 .. code-block:: python
 
-    >>> from specutils import Spectrum1D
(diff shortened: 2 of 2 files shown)


docs/spectrum1d.rst
```diff
@@ -87,6 +87,18 @@ A full list of the supported formats is shown in the table below.
 
 | More information on creating custom loaders can be found in the :doc:`custom loading </custom_loading>` page.
 
+Writing to a File
+-----------------
+
+Similarly, a `~specutils.Spectrum1D` object can be saved to any file format supported by
+:class:`~astropy.nddata.CCDData` via the :meth:`specutils.Spectrum1D.write` method.
```
pllim, line 94: Why is CCDData involved here?

https://github.com/astropy/specutils/blob/25cb121f8da53a7e79ebc10b94f665e28e0399d4/specutils/spectra/spectrum1d.py#L20

docs/spectrum1d.rst
```diff
@@ -87,6 +87,18 @@ A full list of the supported formats is shown in the table below.
...
+
+.. code-block:: python
+
+    >>> from specutils import Spectrum1D
+    >>> spec1d = Spectrum1D.read("/path/to/input.fits")  # doctest: +SKIP
```
pllim, line 99: Do we really need this? Reading is already covered in https://specutils.readthedocs.io/en/stable/spectrum1d.html#reading-from-a-file

```suggestion
```

pllim:
Maybe also want to link to https://specutils.readthedocs.io/en/stable/custom_loading.html#creating-a-custom-writer (using ref)?

docs/spectrum1d.rst
kecnry: It comes from [NDIOMixin](https://docs.astropy.org/en/stable/api/astropy.nddata.NDIOMixin.html) whose API docs state CCDData 🤷

docs/spectrum1d.rst
kecnry: Should the snippets be somewhat standalone or is it ok to assume at this point that the user already has a `spectrum1D` object loaded/created?

pllim: Depends on who you ask. 😆 

You can `ref` back to the "reading" section if you want, but I think it is overkill to tell people how to read again.

docs/spectrum1d.rst
pllim: I think NDIOMixin used CCDData as an example of implementation using the mixin. I don't think specutils has anything to do with CCDData.

kecnry: ok, would you rather that I just point directly to NDIOMixin then, or nothing at all and just leave the link to the API docs which inherits from NDIOMixin API?

pllim: Why not link to https://specutils.readthedocs.io/en/latest/spectrum1d.html#reading-from-a-file (with `ref`)?

kecnry: that is _immediately_ before this sub-section, so I'll just update the text to mention "supported formats".

docs/spectrum1d.rst
```diff
@@ -87,6 +87,18 @@ A full list of the supported formats is shown in the table below.
...
+Writing to a File
+-----------------
+
+Similarly, a `~specutils.Spectrum1D` object can be saved to any of the supported formats using the
+:meth:`specutils.Spectrum1D.write` method.
+
+.. code-block:: python
+
+    >>> spec1d.write("/path/to/output.fits")  # doctest: +SKIP
+
```
pllim, line 99:
```suggestion
```

docs/custom_loading.rst
```diff
@@ -143,6 +143,8 @@ many `~specutils.Spectrum1D` objects. This method is available since
 `astropy.io.registry.read`).
 
 
+.. _custom_writer
```
pllim, line 146:
```suggestion
.. _custom_writer:
```

pllim (approved):
LGTM now. Thanks!

No need to wait for CI. This just touch doc without doctest.

Issue #997, opened 2022-12-01
Flux with electron/s cannot be read back in from file (roundtrip problem)


pllim:
Consider this use case:

```python
import numpy as np
from astropy import units as u
from specutils import Spectrum1D

flux = np.random.random(10) * (u.electron / u.s)
wave = np.arange(10) * u.m
sp = Spectrum1D(flux=flux, spectral_axis=wave)  # Works fine
```

But then when writing it back out into a FITS file:

```python
>>> sp.write('flux_es.fits')
WARNING: The unit 'electron / s' could not be saved in native FITS format
and cannot be recovered in reading. It can roundtrip within astropy by using
QTable both to write and read back, though one has to enable the unit before
reading. [astropy.io.fits.convenience]
```

Now, that file cannot be read back in, as the warning indicated. But why can't `specutils` write it out in a way that can roundtrip?

```
>>> sp2 = Spectrum1D.read('flux_es.fits')
.../astropy/nddata/mixins/ndio.py:59, in NDDataRead.__call__(self, *args, **kwargs)
     58 def __call__(self, *args, **kwargs):
---> 59     return self.registry.read(self._cls, *args, **kwargs)

.../astropy/io/registry/core.py:219, in UnifiedInputRegistry.read(self, cls, format, cache, *args, **kwargs)
    214     format = self._get_valid_format(
    215         "read", cls, path, fileobj, args, kwargs
    216     )
    218 reader = self.get_reader(format, cls)
--> 219 data = reader(*args, **kwargs)
    221 if not isinstance(data, cls):
    222     # User has read with a subclass where only the parent class is
    223     # registered.  This returns the parent class, so try coercing
    224     # to desired subclass.
    225     try:

.../specutils/io/default_loaders/tabular_fits.py:86, in tabular_fits_loader(file_obj, column_mapping, hdu, **kwargs)
     83 # If no column mapping is given, attempt to parse the file using
     84 # unit information
     85 if column_mapping is None:
---> 86     return generic_spectrum_from_table(tab, wcs=wcs, **kwargs)
     88 return spectrum_from_column_mapping(tab, column_mapping, wcs=wcs)

.../specutils/io/parsing_utils.py:258, in generic_spectrum_from_table(table, wcs, **kwargs)
    256 flux_column = _find_spectral_column(table, colnames, spectral_axis)
    257 if flux_column is None:
--> 258     raise IOError("Could not identify column containing the flux")
    259 flux = table[flux_column].to(table[flux_column].unit)
[...]

PR #1018, opened 2023-02-21
support Spectrum1D reading from file with flux in counts

kecnry:
this PR implements an exception to support reading a Spectrum1D from a file with fluxes in counts.  This does not however address more general reading round-tripping such as the one reported in #997.  Note also that this is just hardcoding an additional exception, `Spectrum1D` currently doesn't seem to do any unit checking on the inputs for flux, and so other units might result in the same error (including non-sensical units, like fluxes with units of meters).

Example (that used to raise an `OSError: Could not identify column containing the flux`):

```
import specutils
import numpy as np
import astropy.units as u

sp = specutils.Spectrum1D(flux=np.linspace(0,1,11)*u.Unit('count'), spectral_axis=np.random.random(11)*u.Unit('J'))
sp.write("spec.fits", overwrite=True)

specutils.Spectrum1D.read("spec.fits")
```

### DIFF
--- a/specutils/tests/test_spectrum1d.py
+++ b/specutils/tests/test_spectrum1d.py
@@ -357,6 +357,18 @@ def test_create_with_uncertainty():
         Spectrum1D(spectral_axis=wavelengths*u.um, flux=flux, uncertainty=uncertainty)
 
 
+def test_flux_unit_io_roundtrip(tmpdir):
+    # regression test for https://github.com/astropy/specutils/pull/1018
+    fname = str(tmpdir.join('flux_unit_io_roundtrip.fits'))
+    for flux_unit in ('adu', 'ct/s', 'count'):
+        sp = Spectrum1D(flux=np.linspace(0,1,11)*u.Unit(flux_unit),
+                        spectral_axis=np.random.random(11)*u.Unit('Hz'))
+        sp.write(fname, overwrite=True)
+
+        sp_load = Spectrum1D.read(fname)
+        assert sp_load.flux.unit == sp.unit
+
+
 @pytest.mark.filterwarnings('ignore::astropy.io.fits.verify.VerifyWarning')
 @remote_access([{'id': '1481190', 'filename': 'L5g_0355+11_Cruz09.fits'}])
 def test_read_linear_solution(remote_data_path):
--- a/specutils/tests/test_spectrum1d.py
+++ b/specutils/tests/test_spectrum1d.py
@@ -357,16 +357,16 @@ def test_create_with_uncertainty():
         Spectrum1D(spectral_axis=wavelengths*u.um, flux=flux, uncertainty=uncertainty)
 
 
-def test_flux_unit_io_roundtrip(tmpdir):
+@pytest.mark.parametrize("flux_unit", ["adu", "ct/s", "count"])
+def test_flux_unit_io_roundtrip(tmp_path, flux_unit):
     # regression test for https://github.com/astropy/specutils/pull/1018
-    fname = str(tmpdir.join('flux_unit_io_roundtrip.fits'))
-    for flux_unit in ('adu', 'ct/s', 'count'):
-        sp = Spectrum1D(flux=np.linspace(0,1,11)*u.Unit(flux_unit),
-                        spectral_axis=np.random.random(11)*u.Unit('Hz'))
-        sp.write(fname, overwrite=True)
+    fname = str(tmp_path / 'flux_unit_io_roundtrip.fits')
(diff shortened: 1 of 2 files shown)


pllim:
> u.Unit('J')

You sure you didn't mean Jy?

pllim:
Are we able to turn that example you provided into a test?

kecnry:
> You sure you didn't mean Jy?

No, this is for spectral axis, not flux, so I just chose a random compatible unit for the spectral axis.  Swap out with Hz or um if you'd rather!

> Are we able to turn that example you provided into a test?

Sure, we could provide a regression test for this same hardcoded case, but it really would only test that someone doesn't revert this commit... so I'm wondering if it makes more sense to write general tests with the longer-term solution?  Happy to basically copy-paste this in to a test though just to make sure it doesn't raise errors.

pllim:
> No, this is for spectral axis, not flux

Doh! Ignore me... 😆 

> but it really would only test that someone doesn't revert this commit

Right, we wouldn't want regression, which could happen if we ever refactor the internals.

specutils/tests/test_spectrum1d.py
```diff
@@ -357,6 +357,18 @@ def test_create_with_uncertainty():
         Spectrum1D(spectral_axis=wavelengths*u.um, flux=flux, uncertainty=uncertainty)
 
 
+def test_flux_unit_io_roundtrip(tmpdir):
```
pllim, line 360:
```suggestion
def test_flux_unit_io_roundtrip(tmp_path):
```

specutils/tests/test_spectrum1d.py
```diff
@@ -357,6 +357,18 @@ def test_create_with_uncertainty():
...
+    # regression test for https://github.com/astropy/specutils/pull/1018
+    fname = str(tmpdir.join('flux_unit_io_roundtrip.fits'))
```
pllim, line 362:
```suggestion
    fname = str(tmp_path / 'flux_unit_io_roundtrip.fits')
```

specutils/tests/test_spectrum1d.py
pllim, line 360:
```suggestion
@pytest.mark.parametrize("flux_unit", ["adu", "ct/s", "count"])
def test_flux_unit_io_roundtrip(tmpdir, flux_unit):
```

specutils/tests/test_spectrum1d.py
```diff
@@ -357,6 +357,18 @@ def test_create_with_uncertainty():
...
+    for flux_unit in ('adu', 'ct/s', 'count'):
+        sp = Spectrum1D(flux=np.linspace(0,1,11)*u.Unit(flux_unit),
+                        spectral_axis=np.random.random(11)*u.Unit('Hz'))
+        sp.write(fname, overwrite=True)
+
+        sp_load = Spectrum1D.read(fname)
+        assert sp_load.flux.unit == sp.unit
```
pllim, line 369:
```suggestion
    sp = Spectrum1D(flux=np.ones(11) * u.Unit(flux_unit),
                    spectral_axis=np.arange(1, 12) * u.Unit('Hz'))
    sp.write(fname, overwrite=True)

    sp_load = Spectrum1D.read(fname)
    assert sp_load.flux.unit == sp.flux.unit
```

pllim (commented):
I think the test could be written using `parametrize` and there is no need for random number generator. Please see suggestions.

Not sure what's up with the change log, maybe @rosteen only updates it on release? https://github.com/astropy/specutils/blob/main/CHANGES.rst

rosteen:
> Not sure what's up with the change log, maybe @rosteen only updates it on release? https://github.com/astropy/specutils/blob/main/CHANGES.rst

Yeah, I've been updating it when I do a release. I'm certainly not opposed to people updating it when they make a PR though!

I'm going to approve this, but generally agree with your suggestions @pllim . I'm about to log off for vacation but feel free to merge this once those are addressed.

rosteen (approved):

dhomeier (changes requested):
I agree with the requested changes, otherwise this looks sensible. Yes, `Spectrum1D.__init__` does not make any checks on flux `PhysicalType`, but there you are directly passing your quantity. `parsing_utils` are searching for some compatible data fields in an input file, so it makes sense to be more conservative.

pllim:
@dhomeier , does the PR look okay to you now, or is there something else? Thanks!

dhomeier (approved):

dhomeier:
Everything you had requested as far as I can see; good to me too, thanks!

pllim:
Thanks, all!

Issue #1016, opened 2023-02-21
Reduce reliance on np.random in tests and docs, use new numpy Generator when you must


pllim:
Tests and doc examples are meant to be reproducible. It does not make sense to use random number generator. I suspect the reason behind such rampant usage was a rush to make fake data in a pinch, not because of true randomness was required. There are two things that need to be addressed:

1. Reduce the calls to `np.random`. For example, maybe you can generate it once and store as a test data file, and then just re-use the same data throughout your tests and doctests. Or you can generate it only once in `conftest.py` or use it as fixture, and then use that same stuff throughout in tests. Or better still, find real data (with reasonable file size) and use that as test data.
2. When you cannot help it, set a seed. Then refactor the existing calls to use the [new numpy random Generator](https://numpy.org/doc/stable/reference/random/generator.html). Stop using the legacy API.

PR #1019, opened 2023-02-21
Unignore and fix doctest

### LINKED ISSUE #849: TST: Do not skip valid doctest examples

pllim:
This PR is to address this issue:

* fix #849 

Also remove the doctest directive that used to ignore output. I actually uncovered some wrong code used in examples when I am doing this. Those are fixed here too.

I came across #1016 but that is a battle for another day.

### DIFF
--- a/specutils/manipulation/model_replace.py
+++ b/specutils/manipulation/model_replace.py
@@ -64,7 +64,8 @@ def model_replace(spectrum, replace_region, model=10, extrapolation_treatment='d
     >>> flux_val = np.array([2, 4, 6, 8, 10, 12, 14, 16, 18, 20])
     >>> input_spectrum = Spectrum1D(spectral_axis=wave_val * u.AA, flux=flux_val * u.mJy)
     >>> spline_knots = [3.5, 4.7, 6.8, 7.1] * u.AA
-    >>> result = model_replace(input_spectrum, None, spline_knots) # doctest: +IGNORE_OUTPUT
+    >>> model_replace(input_spectrum, None, spline_knots)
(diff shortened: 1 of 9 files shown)


specutils/manipulation/model_replace.py
```diff
@@ -64,7 +64,8 @@ def model_replace(spectrum, replace_region, model=10, extrapolation_treatment='d
     >>> flux_val = np.array([2, 4, 6, 8, 10, 12, 14, 16, 18, 20])
     >>> input_spectrum = Spectrum1D(spectral_axis=wave_val * u.AA, flux=flux_val * u.mJy)
     >>> spline_knots = [3.5, 4.7, 6.8, 7.1] * u.AA
-    >>> result = model_replace(input_spectrum, None, spline_knots) # doctest: +IGNORE_OUTPUT
+    >>> model_replace(input_spectrum, None, spline_knots)
+    <Spectrum1D(flux=<Quantity [ 2.,  4.,  6.,  8., 10., 12., 14., 16., 18., 20.] mJy>, spectral_axis=<SpectralAxis [ 1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.] Angstrom>)>
```
pllim, line 68: I dunno what this example is supposed to return. I don't have anything to refer to but the output is same as input, is that correct?

rosteen: The input spectrum is just a straight line, so I think a spline fit should just return a straight line as well. Makes sense to me, although it isn't a particularly enlightening example.

pllim: Thanks for the clarification!

I think making examples enlightening is out of scope here. 😆

rosteen (approved):
Looks good to me, docs pass and a quick look at the rendered docs looks good. Thanks!

pllim:
Thanks for the review!

PR #1022, opened 2023-02-22
Catch deprecation from AsdfExtension

WilliamJamieson:
asdf-format/asdf#1429 deprecated the `AsdfExtension` class. This PR catches that deprecation warning.


WilliamJamieson:
Since ASDF develop is not tested, the CI will not pickup this change. I have tested this locally with the head of ASDF development and this does fix the failing tests.

pllim (approved):

PR #1024, opened 2023-02-23
Catch deprecation warning for ASDF entry point

WilliamJamieson:
asdf-format/asdf#1361 officially deprecates the legacy entry-point which `specutils` uses for its ASDF extension. This PR ignores that deprecation.

Note, I have tested this locally and it successfully fixes failures due to the entrypoint deprecation warning.


pllim (approved):
Thanks!

PR #1025, opened 2023-02-23
change asdf.tests.helpers uses to asdf.testing.helpers

braingram:
asdf.tests.helpers will be deprecated:
https://github.com/asdf-format/asdf/pull/1440

The uses in specutils are limited to assert_roundtrip_tree

A similar roundtrip_object exists in asdf.testing.helpers which serializes then deserializes an object (without asserting equality). This commit replaces the uses of assert_roundtrip_tree with roundtrip_object and uses the equality comparison built into the custom asdf types implemented in this module to check for equality. These equality comparisons are the same checks that were used internally by assert_roundtrip_tree.

### DIFF
--- a/specutils/io/asdf/tags/tests/test_spectra.py
+++ b/specutils/io/asdf/tags/tests/test_spectra.py
@@ -8,10 +8,11 @@ import astropy.units as u  # noqa: E402
 from astropy.coordinates import FK5  # noqa: E402
 from astropy.nddata import StdDevUncertainty  # noqa: E402
 
-from asdf.tests.helpers import assert_roundtrip_tree  # noqa: E402
+from asdf.testing.helpers import roundtrip_object  # noqa: E402
 import asdf  # noqa: E402
 
 from specutils import Spectrum1D, SpectrumList, SpectralAxis  # noqa: E402
+from specutils.io.asdf.tags.spectra import Spectrum1DType, SpectrumListType
 
 
 def create_spectrum1d(xmin, xmax, uncertainty=None):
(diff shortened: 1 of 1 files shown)


specutils/io/asdf/tags/tests/test_spectra.py
```diff
@@ -26,27 +27,26 @@ def create_spectrum1d(xmin, xmax, uncertainty=None):
...
 @pytest.mark.xfail
 def test_asdf_spectralaxis(tmpdir):
 
     wavelengths  = np.arange(5100, 5300) * 0.1 * u.nm
     spectral_axis = SpectralAxis(wavelengths, bin_specification="edges")
-    tree = dict(spectral_axis=spectral_axis)
-    assert_roundtrip_tree(tree, tmpdir)
+    # there is no implemented asdf type for SpectralAxis and no defined
+    # equality comparison (assert_equal)
+    assert roundtrip_object(spectral_axis) == spectrum_axis
```
pllim, line 49: Will this actually pass now? This test is marked as xfail.

pllim:
Can u pls fix code style? Thx!

specutils/io/asdf/tags/tests/test_spectra.py
pllim: Nope, we have xfail strict on and CI passed, so this is still xfail

https://github.com/astropy/specutils/blob/ca3153f05b8d3613bba0b865e273187e8ff176c4/setup.cfg#L52

braingram:
> Can u pls fix code style? Thx!

I believe I fixed the code style issues. It pointed out a typo in the xfailed test (I used `spectrum_axis` instead of `spectral_axis`). I fixed that typo and the test still fails because the equality comparison fails.

Based on https://github.com/astropy/specutils/pull/645#issuecomment-614271632 SpectralAxis fails to round_trip and returns as SpectralCoord. I changed the test to fail due to a type comparison and added a link to the comment in the code.

pllim:
Thanks!

PR #1030, opened 2023-03-09
Bugfix: shuffled arg order in _fit_lines

bmorris3:
If you compare the ordered arguments to the `_fit_lines` function:
https://github.com/astropy/specutils/blob/6eb7f96498072882c97763d4cd10e07cf81b6d33/specutils/fitting/fitmodels.py#L369-L371
to the ordered arguments given to its call in `fit_lines`:
https://github.com/astropy/specutils/blob/6eb7f96498072882c97763d4cd10e07cf81b6d33/specutils/fitting/fitmodels.py#L355-L357
You'll see the order of `ignore_units` and `get_fit_info` are swapped in the latter.

This PR puts them in the correct order.

### DIFF
--- a/specutils/fitting/fitmodels.py
+++ b/specutils/fitting/fitmodels.py
@@ -354,7 +354,7 @@ def fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(calc_uncertainties
 
         fit_model = _fit_lines(spectrum, model_guess, fitter,
                                exclude_regions, weights, model_window,
-                               ignore_units, get_fit_info, **kwargs)
+                               get_fit_info, ignore_units, **kwargs)
         if model_guess.name is not None:
             fit_model.name = model_guess.name
 


pllim:
Those are supposedly keywords. Should the call use keywords to be explicit? I never like passing in keywords as posargs.

bmorris3:
@pllim Switched from relying on posargs to keywords in [cac2d77](https://github.com/astropy/specutils/pull/1030/commits/cac2d77078e962f48eb3a31f5ebdd6ac446af51d). Test failures look unrelated.

keflavich:
Those first two arguments are required, though, right?  What about:
```python
        fit_model = _fit_lines(
            spectrum,
            model,
            *,
            fitter=fitter,
...
```

keflavich (approved):

specutils/fitting/fitmodels.py
```diff
@@ -352,9 +352,18 @@ def fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(calc_uncertainties
 
         ignore_units = getattr(model_guess, model_guess.param_names[0]).unit is None
 
-        fit_model = _fit_lines(spectrum, model_guess, fitter,
-                               exclude_regions, weights, model_window,
-                               ignore_units, get_fit_info, **kwargs)
+        fit_model = _fit_lines(
+            spectrum=spectrum,
+            model=model_guess,
```
pllim, line 357:
```suggestion
            spectrum,
            model_guess,
```

pllim:
Failures are unrelated but...

* need a change log?
* also it is concerning that this bug was not caught in CI... do we need to add a test?

pllim (approved):
Thanks! Fix LGTM but I'll let specutils maintainers decide if they want a test or can merge.

rosteen (approved):
Good catch, thanks. I'll open an issue to improve the test coverage on this, for now we can get this in.

rosteen:
Test failures are unrelated, merging.

PR #1033, opened 2023-03-13
Allow spectral axis to be anywhere, instead of forcing it to be last

rosteen:
This replaces #999, merging into a new 2.0 development branch instead of `main` so that we can more thoroughly vet/test/discuss/etc big changes like this before releasing version 2.0. The context for the original PR is copied below, but  please see #999 for previous discussion.

------

The motivation for this is that the insistence on the spectral axis being last hinders integration with other packages that make different (or no) assumptions on the location of the spectral axis. The two biggest examples, currently, are that spectral-cube assumes that the spectral axis is first, and they would be interested in using specutils objects as the basis of their code but are hindered by our assumption. We don't need to switch to that assumption, I think being flexible here is the right call. jdaviz also had to do a lot of coding around Glue assumptions about where the spectral axis is as well, since they differ from ours.

I have also been concerned for a while that specutils putting the flux axis into a different order than a user would get out from astropy.io.fits is potentially confusing and/or a barrier to uptake of specutils. In my mind at least, this combined with the above outweighs the convenience of numpy broadcasting properly automatically if you index solely on the spectral axis when it is last for a multi-dimensional Spectrum1D, which if I'm recalling correctly was the motivation for forcing the spectral axis to be last.

This does provide a `move_spectral_axis` keyword when initializing a Spectrum1D so that users can either get the legacy behavior (force it to be last) or do the opposite (force first) if desired.


specutils/analysis/moment.py
```diff
@@ -43,10 +44,21 @@ def moment(spectrum, regions=None, order=0, axis=-1):
...
     """
     This is a helper function for the above `moment()` method.
     """
+    if axis is None:
+        if isinstance(spectrum, SpectrumCollection):
+            axes = [spec.spectral_axis_index for spec in spectrum]
+            if not np.all([x==axes[0] for x in axes]):
+                raise ValueError("All spectra in SpectrumCollection must have the same "
+                                 "spectral_axis_index for simultaneous moment calculation.")
+            axis = axes[0] + 1
```
keflavich, line 57: Why is this being done?  Isn't this setting the axis to be `axes[0].spectral_axis + 1`, which is not the spectral axis?  Or am I missing some logic?  (that happens easily during reviews...)

specutils/analysis/moment.py
```diff
@@ -64,9 +76,12 @@ def _compute_moment(spectrum, regions=None, order=0, axis=-1):
         return np.sum(flux, axis=axis)
 
     dispersion = spectral_axis
+    # We now have to account for the spectral axis being anywhere, not always last
     if len(flux.shape) > len(spectral_axis.shape):
-        _shape = flux.shape[:-1] + (1,)
-        dispersion = np.tile(spectral_axis, _shape)
+        for i in range(len(flux.shape)):
```
keflavich, line 81: I think this is the more idiomatic approach for numpy arrays
```suggestion
        for i in range(flux.ndim):
```

specutils/spectra/spectrum1d.py
```diff
@@ -163,27 +182,51 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
...
 
         # If flux and spectral axis are both specified, check that their lengths
-        # match or are off by one (implying the spectral axis stores bin edges)
+        # match or are off by one (implying the spectral axis stores bin edges).
+        # If we can't determine which flux axis corresponds to the spectral axis
+        # we raise an error.
         if flux is not None and spectral_axis is not None:
-            if spectral_axis.shape[0] == flux.shape[-1]:
+            if spectral_axis_index is None:
+                if len(flux.shape) == 1:
```
keflavich, line 190:
```suggestion
                if flux.ndim == 1:
```

specutils/spectra/spectrum1d.py
```diff
@@ -163,27 +182,51 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
...
+                    self._spectral_axis_index = 0
+                else:
+                    matching_axes = []
+                    if bin_specification == "centers":
+                        add_elements = [0,]
+                    elif bin_specification == "edges":
+                        add_elements = [1,]
+                    elif bin_specification is None:
+                        add_elements = [0,1]
+                    for i in range(len(flux.shape)):
```
keflavich, line 200:
```suggestion
                    for i in range(flux.ndim):
```

specutils/analysis/moment.py
keflavich: `axis` is zero-indexed, right?  not 1-indexed?

specutils/spectra/spectrum1d.py
```diff
@@ -196,33 +239,52 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
...
+                    else:
+                        raise ValueError("move_spectral_axis must be either 'first' or 'last'")
+
+                    if move_to_index != self._spectral_axis_index:
+                        wcs = wcs.swapaxes(self._spectral_axis_index, move_to_index)
+                        if flux is not None:
+                            flux = np.swapaxes(flux, self._spectral_axis_index, move_to_index)
+                        if "mask" in kwargs:
+                            if kwargs["mask"] is not None:
+                                kwargs["mask"] = np.swapaxes(kwargs["mask"],
```
keflavich, line 261: Just a note on this - I think it's fine as is, but we may need to add a convenience tool to reorder the celestial dimensions.  If you give a cube with dims [RA, Dec, Spectral] and move_to_first now, you'll get [Spectral, Dec, RA], which is not what anyone would normally expect (imo).  I don't think we should handle the logic here, but we might push a function in astropy.wcs that would `rearrange_celestial_to_have_longitude_first`

docs/spectrum1d.rst
```diff
@@ -321,72 +324,23 @@ value will apply to the lower bound input.
     ...          'SPECSYS': 'BARYCENT', 'RADESYS': 'ICRS', 'EQUINOX': 2000.0,
     ...          'LONPOLE': 180.0, 'LATPOLE': 27.004754})
     >>> spec = Spectrum1D(flux=np.random.default_rng(12345).random((20, 5, 10)) * u.Jy, wcs=w)  # doctest: +IGNORE_WARNINGS
-    >>> lower = [SpectralCoord(4.9, unit=u.um), SkyCoord(ra=205, dec=26, unit=u.deg)]
-    >>> upper = [SpectralCoord(4.9, unit=u.um), SkyCoord(ra=205.5, dec=27.5, unit=u.deg)]
+    >>> lower = [SkyCoord(ra=205, dec=26, unit=u.deg), SpectralCoord(4.9, unit=u.um)]
+    >>> upper = [SkyCoord(ra=205.5, dec=27.5, unit=u.deg), SpectralCoord(4.9, unit=u.um)]
```
keflavich, line 328: Question: Can we give these slices in any order now?  i.e., does specutils figure out which dimensions need to be sliced with which component?  That would be ideal but I'm not sure if it's the case (yet)

keflavich (changes requested):
I have a few minor suggestions on implementation and UI.

This looks good!

specutils/analysis/moment.py
keflavich: or is `spectral_axis_index` done in FITS convention and grabbed from the FITS header?  That seems like not what we want

specutils/spectra/spectrum1d.py
keflavich: spectral-cube does this:
https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/cube_utils.py#L204

docs/spectrum1d.rst
rosteen: Unfortunately no. I switched this here because the test was failing now that the spectral axis wasn't getting moved to last - the NDCube slicing machinery seems to expect the sky/spectral coords to be in the same order as in the WCS.

specutils/analysis/moment.py
rosteen: This is a great question, I'm going to have to think to remember what the heck I was thinking here - I coded this a while ago.

docs/spectral_cube.rst
```diff
@@ -144,9 +144,8 @@ Moments
 =======
 
 The `~specutils.analysis.moment` function can be used to compute moments of any order
-along one of the cube's axes. By default, ``axis=-1``, which computes moments
-along the spectral axis (remember that the spectral axis is always last in a
-:class:`~specutils.Spectrum1D`).
+along one of the cube's axes. By default, ``axis=None``, in which case the moment
```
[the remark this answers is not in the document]
rosteen: Agreed that that's clearer, I'll update it.

specutils/analysis/moment.py
rosteen: Ah right, I remember now - it's because my understanding is that a `SpectrumCollection` adds a leading axis when it stacks the individual spectra together.

keflavich: Oh, that makes sense!  Add a comment about it, then?

rosteen: Done!

specutils/spectra/spectrum1d.py
rosteen: Agreed - FYI this is currently what specutils does by default, so it's not any worse than it is currently 🙈 .

specutils/spectra/spectrum1d.py
```diff
@@ -163,27 +182,51 @@ def __init__(self, flux=None, spectral_axis=None, wcs=None,
...
+                        for add_element in add_elements:
+                            if spectral_axis.shape[0] == flux.shape[i] + add_element:
+                                matching_axes.append(i)
+
+                    if len(matching_axes) == 1:
+                        self._spectral_axis_index = matching_axes[0]
+                    else:
+                        raise ValueError("Unable to determine which flux axis corresponds to "
+                                         "the spectral axis. Please specify spectral_axis_index"
+                                         " or provide a spectral_axis matching a flux axis.")
```
eteq, line 210: While I like the general idea of "best guess" here, there's one failure mode that worries me: the case where there's a multi-d flux axis and one of those axes just by chance happens to match the spectral axis.  I can easily imagine someone writing code thinking the code just "knows" the correct axis, but then getting tripped up in a non-obvious way when there's an input that's ambiguous.

I'm not sure I have a clean solution though. A solution I don't love but *might* be good is to emit a warning whenever the guess is done, to hint to implementers they should really specify the axis if they can.  But that will probably lead to a lot of annoying warnings.

Maybe @keflavich or @Cadair have some idea how common this failure mode might be?  In optical-land I think it's rare enough it's not a big deal (since cubes are almost always a lot smaller than the spectral dimension) but I'm not as sure about other domains...

eteq (changes requested):
Broadly speaking... I think I'm convinced (good work on @keflavich and @Cadair wearing me down for years :wink: ). I do like the solution of still keeping `last` as an option but not forcing it.

There's two broad implementation concerns I want to raise though:

1. There's likely a lot of analysis/manipulation functions that still need to be made more flexible, right? E.g. the redshifting/template comparison stuff, and the resampling classes?  That doesn't necessarily need to all be fixed now, although *maybe* it will be possible to use my next suggestion to do a "quick fix"
2. For those who have already built code assuming last-is-spectral, I wonder if it makes sense to also make a convenience accessor?  Creating a new `Spectrum1D` with ``move_spectral_axis='last'``  is pretty syntactically annoying.  But it would be pretty straightforward to add a `get_last_spectral_axis_spectrum()` method or something that just does that for you. Even better, it could *cache* the result (at least, optionally), which then would basically allow transparently using the "last" convention at least in cases where it's not computationally too intensive to do.

specutils/spectra/spectrum1d.py
rosteen: Isn't it _required_ that at least one flux dimension to match the size of the spectral axis? I'm not sure it makes any sense to create a `Spectrum1D` with flux shape, say, `[10,10,200]` and provide a spectral axis of length 240. This error would get thrown in that case, and it would also be raised if one of the spatial dimensions coincidentally happened to also match the length of the spectral axis.

keflavich: Even in optical land, you'll work with data subsets, and these can have any shape.

If there's exactly _one_ match, you're OK.  If there are multiple matches or zero matches, error, right?

@rosteen What if you had x,y,spectral = [10,20,200] and a spectral axis was passed in with length 20?  That would not result in this error but would be wrong.  I'm .... not sure why a user would ever do this, though?

rosteen: Hm, I see the point now. I'm also not sure why a user would do that, but then again...there have been many times when I wasn't sure why a user did something 😆. I guess the question is "is this likely enough to justify the annoyance of either displaying a warning, or always requiring `spectral_axis_index`?" And I'm not sure of the answer.

keflavich: I'm loosely inclined toward no warning, but let's be prepared to post this conversation as a response to the first user error that results in this corner case.

specutils/spectra/spectrum1d.py
rosteen: I added a note to the docs for now saying that RA/Dec might not be in a sensible order if you move the spectral axis.

rosteen:
@eteq @keflavich I believe that I addressed all of your comments, mind taking a look and making sure that I did approximately what you were suggesting?

PR #1038, opened 2023-03-14
Use stdatamodels for all ASDF-in-FITS

pllim:
ASDF dev no longer can read ASDF-in-FITS. We need `stdatamodels` for that.


rosteen (approved):
Thanks! I added a note to the docs about the new dependency. I'll merge after tests pass.

pllim:
Thanks!

Issue #927, opened 2022-02-17
MNT: Un-ignore currently ignored test warnings


pllim:
I think this work would encompass multiple PRs. Right now, some warnings are in the blanket ignore like because they happen too often across the test suite or require time to track down. If possible, we should remove them one by one from the ignored list and handle them properly.

For example, when the code emits resource warning, either the test needs to gracefully close the file handler on exit (either pass or fail) or the code being tested is leaving a dangling file pointer (harder to fix).

https://github.com/astropy/specutils/blob/9a5bf9368586b420549c70e8279f31e93e567ad9/setup.cfg#L56

PR #1040, opened 2023-03-14
TST: Unignore FITSFixedWarning

pllim:
Address one warning from #927


pllim:
Could it be this easy? No changes to test code?!

pllim:
17 failed all from remote data, I see...

PR #1042, opened 2023-03-16
Compatibility with ASDF v3

### LINKED ISSUE #664: Make SpectralAxis serializable via ASDF.
See xfailed test written as part of #645.  Make it pass.

Basically, a new tag, schema and type need to be provided as part of the `specutils` `asdf` extension.

### LINKED ISSUE #1011: ASDF legacy converter system is being deprecated

pllim:
Fix #1011 and fix #664 

cc @braingram 

### TODO

- [x] Relax jsonschema pinning
- [x] Fix test failures

[🐶](https://jira.stsci.edu/browse/AL-718) [🐱](https://jira.stsci.edu/browse/JDAT-3199)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,7 @@ install_requires =
     astropy>=5.1
     gwcs>=0.18
     asdf-astropy>=0.3
-    asdf>=2.14
+    asdf>=2.14.4
     ndcube>=2.0
 
 [options.extras_require]
@@ -33,8 +33,6 @@ test =
     graphviz
     coverage
     spectral-cube
-    # https://github.com/asdf-format/asdf/issues/1485
-    jsonschema<4.18
 jwst =
     stdatamodels>=1.1.0
 docs =


specutils/io/asdf/extension.py
```diff
@@ -1,59 +1,40 @@
...
+    from pathlib import Path
+    from asdf.resource import DirectoryResourceMapping
+
+    resources_root = Path(__file__).resolve().parent
+    if not resources_root.is_dir():
+        raise RuntimeError(f"Missing resources directory: {resources_root}")
+
+    return [
+        DirectoryResourceMapping(
+            resources_root / "schemas" / "astropy.org" / "specutils" / "spectra",
```
WilliamJamieson, line 36:
```suggestion
            resources_root / "schemas" 
```

Then apply my suggestion below for moving the two schemas.

specutils/io/asdf/schemas/astropy.org/specutils/spectra/spectrum1d-1.0.0.yaml
```diff
@@ -2,7 +2,6 @@
 ---
 $schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
 id: "http://astropy.org/schemas/specutils/spectra/spectrum1d-1.0.0"
-tag: "tag:astropy.org:specutils/spectra/spectrum1d-1.0.0"
```
WilliamJamieson, line 5: Move this whole schema file into just `schemas` rather than the unnecessary set of folders.

specutils/io/asdf/schemas/astropy.org/specutils/spectra/spectrum_list-1.0.0.yaml
```diff
@@ -2,12 +2,11 @@
 ---
 $schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
 id: "http://astropy.org/schemas/specutils/spectra/spectrum_list-1.0.0"
-tag: "tag:astropy.org:specutils/spectra/spectrum_list-1.0.0"
```
WilliamJamieson, line 5: Same, move this to just the `schemas` directory.

specutils/io/asdf/extension.py
```diff
@@ -1,59 +1,40 @@
...
+    from asdf.extension import ManifestExtension
+    from specutils.io.asdf.tags.spectra import Spectrum1DType, SpectrumListType
+
+    SPECUTILS_TRANSFORM_CONVERTERS = [Spectrum1DType(), SpectrumListType()]
+
+    # The order here is important; asdf will prefer to use extensions
+    # that occur earlier in the list.
+    TRANSFORM_MANIFEST_URIS = [
+        "asdf://astropy.org/schemas/specutils/spectra/spectrum1d-1.0.0",
+        "asdf://astropy.org/schemas/specutils/spectra/spectrum_list-1.0.0"]
```
WilliamJamieson, line 16:
```suggestion
        "asdf://astropy.org/specutils/manifests/specutils-1.0.0"]
```

The URI's you listed are not for a manifest, but for the schemas referenced by the manifest. This means ASDF will not be able to properly locate schemas. Instead use the URI for the manifest you added instead

specutils/io/asdf/tags/tests/test_spectra.py
```diff
@@ -1,71 +1,69 @@
...
-def test_asdf_spectrum1d(tmpdir):
 
-    spectrum = create_spectrum1d(5100, 5300)
-    Spectrum1DType.assert_equal(spectrum, roundtrip_object(spectrum))
+@pytest.mark.parametrize('uncertainty', [False, True])
+def test_asdf_spectrum1d(tmp_path, uncertainty):
+    file_path = tmp_path / "test.asdf"
+    spectrum = create_spectrum1d(5100, 5300, uncertainty=uncertainty)
+    with asdf.AsdfFile() as af:
+        af["spectrum"] = spectrum
```
pllim, line 25: @WilliamJamieson , I applied your suggestions and schema lookup error is gone. However, now it fails to serialize here. I don't know why the existing serialization code does not work.

pllim:
I also don't know why schema tests still looking for the schemas at the old location.

```
FileNotFoundError: [Errno 2] No such file or directory: '.../astropy/io/misc/asdf/data/schemas/astropy.org/specutils/spectra/spectral_coord-1.0.0.yaml'
```

specutils/io/asdf/schemas/spectrum1d-1.0.0.yaml
```diff
@@ -19,7 +18,7 @@ properties:
     description: |
       SpectralCoord that represents the spectral axis of the spectrum
     anyOf:
-      - $ref: "spectral_coord-1.0.0"
+      - tag: tag:astropy.org:specutils/spectra/spectral_axis-1.0.0
```
pllim, line 21: @WilliamJamieson , can you please explain why there is a need to change `$ref` to `tag` here but not in `spectrum_list-1.0.0.yaml`? Thanks!

pllim:
* dev job is pulling in older scipy because of gwcs (see https://github.com/spacetelescope/gwcs/pull/447), nothing we can do here.
* jsonschema pin is due to https://github.com/asdf-format/asdf/issues/1485

specutils/io/asdf/tests/test_spectra.py
```diff
@@ -0,0 +1,64 @@
...
+    spectrum = create_spectrum1d(510, 530, uncertainty=uncertainty)
+    with asdf.AsdfFile() as af:
+        af["spectrum"] = spectrum
+        af.write_to(file_path)
+
+    with asdf.open(file_path) as af:
+        assert_spectrum1d_equal(af["spectrum"], spectrum)
+
+
+def test_asdf_spectralaxis(tmp_path):
```
pllim, line 32: Note to maintainer: This no longer xfail

pllim:
William has been an invaluable help to get things to really work here. Thanks, William!

specutils/io/asdf/schemas/spectrum1d-1.0.0.yaml
WilliamJamieson: There is no change from a $ref to a tag here.

The $ref that I removed was a local reference which pointed at the schema that was removed as part of this PR. Meaning this schema was broken by that removal. However, the removed schema was nearly identical to a schema in asdf-coordinates-schemas (I'll address the difference below). In fact, it duplicated the URI, which means one of the two versions would override the other. Which in general can cause confusion (no schemas should have the same URI). Indeed, looking at how ASDF loaded the schemas the asdf-coordinates-schema would always be discovered first meaning the procided here would never be used (except in the test generated by the pytest-asdf plug-in). The one difference was that it included a tag within the schema, which I will discuss why this is unnecessary below. 

Instead, what is going on here is I generated a new tag for an existing schema (multiple tags can point to the same schema), using the manifest. Moreover, the schema that tag points to is the one from asdf-coordinates schemas discussed above. 

In point of fact, this tag is "redundant" because the difference between a tag and a $ref, assuming the tag points to the same schema as the $ref, is that in addition addition to validating the data against the schema, the tag is on that sub tree is also checked. Note that the tag is also checked if the schema lists a tag; however, doing so is strongly discouraged because it prevents reuse of the schema by new tags (for new objects which have the same representation but different behaviors like the difference between SpectralAxis and SpectralCoord). That is it prevents the ability to have multiple tags pointing at the same schema, which is exactly what I did here to avoid unnecessary duplications.

To explicitly see the redundancy, observe the anyOf statement being applied to the list of statements This means any of the listed things is acceptable. Since the spectral_axis points to the schema which is referenced by the $ref on the next line. This means that spectral_axis tags would validate against the $ref anyways. As any tagged object (regardless of the tag itself) which satisfies the $ref will pass that line. Thus the tag added is purely of documentation value (this spectral axis tag is one possibility). Really, the spectralcood tag should be added for completeness. 

Note that in general, it is better to use a tag than a $ref in this situation because it makes explicit the object type(s) one expects, unless you are working with something which can accept a broad category of similar objects (in terms of how you actually write them down).

This means I would prefer to remove the $ref all together in this schema, but as this asdf extension was poorly maintained I fear existing files may fail to validate under these tighter restrictions. Meaning I did not do that (for now). Similarly, the spectral_list really ought to be of the tag for this schema not the $ref, but again the same caution is advised. 

I suggest that once this PR is merged, a second PR be opened. In this new PR new versions of the existing schemas be introduced (alongside a new manifest). These schemas would have the $ref to tag changes I discussed above. This would allow for reading/validating existing files via the old manifest, but strictly enforce things on any new files as ASDF will default to writing under the latest available tags/schema.

pllim: Thank you very much for the very detailed explanation!

setup.cfg
```diff
@@ -22,7 +22,7 @@ install_requires =
     astropy>=5.1
     gwcs>=0.18
     asdf-astropy>=0.3
-    asdf>=2.10,!=2.12.0
+    asdf>=2.14
```
WilliamJamieson, line 25:
```suggestion
    asdf>=2.14.4
```

setup.cfg
```diff
@@ -33,6 +33,8 @@ test =
     graphviz
     coverage
     spectral-cube
+    # https://github.com/asdf-format/asdf/issues/1485
+    jsonschema<4.18
```
WilliamJamieson, line 37: ASDF 2.14.4 resolves this issue.
```suggestion
```

pllim: Thank you!

rosteen (commented):
Thanks for this @pllim and @WilliamJamieson, sorry for taking so long to get around to reviewing it. I was able to install and run the tests locally with dev `asdf`, so that's a good sign. The only concern I have on my initial read-through of the code is with two things in `Spectrum1DConverter`:

1. We should round-trip the `mask` attribute as well as the `uncertainty`.
2. I think it would be better to round-trip the `wcs` rather than the `spectral_axis`, so that any spatial information in a multi-dimensional `Spectrum1D` also gets preserved (and the `spectral_axis` will be reconstructed from the `wcs`). I assume that `asdf` can encode both FITS WCS and GWCS, but I don't actually know that for sure off the top of my head/remember which direction dependencies go - is `asdf` only applicable in the GWCS case anyway? We currently aren't actually preserving spatial information from JWST cubes right now anyway, but it's at the top of my list `specutils` of to-dos in the next couple weeks.

The `SpectrumListConverter` looks like magic to me, but judging by the tests it works 😆.

pllim:
The roundtrip comments -- Was it happening with old version of asdf but not anymore, or is this more like a new feature request? I don't recall that this PR would remove any existing features.

rosteen:
> The roundtrip comments -- Was it happening with old version of asdf but not anymore, or is this more like a new feature request? I don't recall that this PR would remove any existing features.

I wasn't familiar with the old asdf code here, but it looks like these were also missing from the old code based on the diff. Still, this is a good chance to fix either of those if they are simple to add, but it could also be a follow-up PR if you don't want to implement here.

pllim:
It would not be simple for me, as I am unfamiliar with ASDF and cannot get to it in the next 2 weeks. I would prefer it to be a new ticket, especially since William has already rotated off the ASDF work and might not have time to jump in.

rosteen:
Alright, I opened a followup issue for that. I'll approve and merge this.

rosteen (approved):

PR #1044, opened 2023-03-27
Fix SNR computations with masks + regions

ivkram:
Calculating a S/N over a region of a **masked** spectrum throws an error:

`IndexError: boolean index did not match indexed array along dimension 0; dimension is 2601 but corresponding boolean dimension is 6800`

This pull request fixes this bug by replacing `spectrum.mask` with `calc_spectrum.mask` in the S/N helper function.


rosteen (approved):
Good catch! Looks good to me, thanks for adding a test to cover this case.

PR #1047, opened 2023-03-27
MNT: Remove non-existent package data

pllim:
There is no `data/`


pllim:
I am just gonna merge this when CI pass. I don't see how it is needed if CI does not fail.

PR #1048, opened 2023-03-28
Accelerate region extraction with `np.searchsorted()`

ivkram:
Current implementation of `extract_region()` relies on the `world_to_pixel()` method of the WCS. This method is quite slow when it comes to a relatively simple task of finding an index (pixel) where value should be inserted to maintain array (spectral axis) order. This pull request solves this issue by replacing `world_to_pixel()` with `np.searchsorted()`.

Reproducible example to demonstrate the performance improvement:

```
import astropy.units as u
import numpy as np
from specutils import Spectrum1D, SpectralRegion
from specutils.manipulation import extract_region

import timeit

flux = np.random.randn(500) * u.Jy
wavelength = np.arange(5000, 5500) * u.AA
spec1d = Spectrum1D(spectral_axis=wavelength, flux=flux)

reg = SpectralRegion(5100 * u.AA, 5200 * u.AA)


def extract():
    extract_region(spec1d, reg)


print(timeit.timeit("extract()", globals=locals(), number=100))
```

Using `astropy.wcs.WCS.world_to_pixel()`: 8.453231014998892s
Using `np.searchsorted()`: 0.11922155800129985s

### DIFF
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -12,34 +12,20 @@ __all__ = ['extract_region', 'extract_bounding_spectral_region', 'spectral_slab'
 
 def _edge_value_to_pixel(edge_value, spectrum, order, side):
     spectral_axis = spectrum.spectral_axis
-    if order == "ascending":
-        if edge_value > spectral_axis[-1]:
-            return len(spectral_axis)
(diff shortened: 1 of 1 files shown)


rosteen (approved):
Finally had a chance to give this a try, looks good to me! Definitely a performance improvement, thanks for the contribution.

PR #1051, opened 2023-03-29
Support write and read of boolean masks in `wcs1d-fits`

dhomeier:
This PR includes a test that was missed before merging #1009 and adds functionality to write `Spectrum1D` masks of `dtype = bool` in WCS1D format, resolving an issue that came up in https://github.com/astropy/specutils/pull/1009#issuecomment-1474424678.
As `bool` or `bit` cannot be written natively to FITS, enabling round-tripping requires encoding the original type, for which there also seems to be no standard. I propose to introduce the header keyword 'BFORM', which is neither defined nor reserved in the FITS standard, and use it analogously to 'TFORMn' in BINTABLE extensions, for which these values are allowed (excerpt):
> `TFORMn` value Description
    ’L’                    Logical
    ’X’                   Bit
    ’B’                   Unsigned byte
    ’I’                    16-bit integer

Here `BFORM` is only used for the 1-bit types, as other formats are already recognised, but it might be added for others for consistency.

### DIFF
--- a/specutils/io/default_loaders/wcs_fits.py
+++ b/specutils/io/default_loaders/wcs_fits.py
@@ -137,14 +137,19 @@ def wcs1d_fits_loader(file_obj, spectral_axis_unit=None, flux_unit=None,
         # Read mask from specified HDU or try to auto-detect it.
         mask = None
         if mask_hdu is True:
+            mask_hdu = None
             for ext in ('MASK', 'DQ', 'QUALITY'):
                 if ext in hdulist and (isinstance(hdulist[ext], fits.ImageHDU) and
                         hdulist[ext].data is not None):
-                    mask = hdulist[ext].data
+                    mask_hdu = ext
                     break
-        elif isinstance(mask_hdu, (int, str)):
+        if isinstance(mask_hdu, (int, str)):
             if mask_hdu in hdulist:
                 mask = hdulist[mask_hdu].data
+                # Do we have a mask originally converted from bool/bit?
+                if hdulist[mask_hdu].header.get('BFORM', 'B') in ('L', 'X'):
+                    # ToDo: check for overflow and warn on values != (0, 1)
+                    mask = mask.astype(bool)
             else:
                 warnings.warn(f"No HDU '{mask_hdu}' for mask found in file.",
                               AstropyUserWarning)
@@ -275,7 +280,13 @@ def wcs1d_fits_writer(spectrum, file_name, hdu=0, update_header=False,
     # Append mask array
     if spectrum.mask is not None and mask_name is not None:
         hdulist.append(fits.ImageHDU())
-        hdulist[-1].data = spectrum.mask
+        # No standard representation of bool in FITS;
+        # introducing 'BFORM' here in analogy to BINTABLE 'TFORMn'.
+        if spectrum.mask.dtype == bool:
+            hdulist[-1].data = spectrum.mask.astype(np.uint8)
+            hdulist[-1].header['BFORM'] = 'L'
+        else:
+            hdulist[-1].data = spectrum.mask
         if mask_name == '':
             hdulist[-1].name = 'MASK'
         else:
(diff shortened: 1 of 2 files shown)


dhomeier:
@rosteen do PRs need extra approval for the CI?

rosteen:
> @rosteen do PRs need extra approval for the CI?

They shouldn't, since you've contributed here before.

pllim (approved):
LGTM. Thanks!

pllim:
Actions was having problem. I am going to try to close/re-open.

pllim:
If CI does not kick off still, maybe just wait out the server problems. https://www.githubstatus.com/

pllim:
Not quite there. Looks like flux is written to EXT 0 even when there is mask, which is super confusing. Usually when there are multiple data extensions, EXT 0 is primary. Let's see this example:

```python
import numpy as np
from astropy.io import fits
from astropy.wcs import WCS

w = WCS(naxis=3)
w.wcs.ctype = 'FREQ', 'DEC--TAN', 'RA---TAN'
w.wcs.set()

sp = Spectrum1D(flux=np.ones((8, 9, 10)) * u.nJy, mask=np.zeros((8, 9, 10), dtype=np.int8), wcs=w)
sp.write("ztmp.fits", format="wcs1d-fits", overwrite=True)
fits.info("ztmp.fits")
```

With this patch, I get this:

```
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  FLUX          1 PrimaryHDU      29   (10, 9, 8)   float64
  1  MASK          1 ImageHDU        11   (10, 9, 8)   uint8 (rescales to int8)
```

But I want this:

```
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU     303   ()
  1  FLUX          1 ImageHDU        29   (10, 9, 8)   float64
  2  MASK          1 ImageHDU        11   (10, 9, 8)   uint8 (rescales to int8)
```

dhomeier:
> Not quite there. Looks like flux is written to EXT 0 even when there is mask, which is super confusing. Usually when there are multiple data extensions, EXT 0 is primary. Let's see this example:

Yes, if you want the flux written to a different extension, you need to specify (e.g.) `hdu=1`.
The default has always been to write `flux` to the first available HDU, which for ImageHDU format is the primary HDU.
Mask and uncertainty, if present, are written to the subsequent HDUs per the recent PRs, but flux will be in the same HDU it has always been. I don't quite see how this becomes more confusing when a mask HDU is appended – flux is in HDU 0, whether mask and/or uncertainty are there or not.

pllim:
> flux is in HDU 0, whether mask and/or uncertainty are there or not

But this is not how 90% of the data out there are laid out. I have never seen a real life example where the FLUX is still in primary extension if there are ERR or DQ tied to it.

dhomeier:
> But this is not how 90% of the data out there are laid out. I have never seen a real life example where the FLUX is still in primary extension if there are ERR or DQ tied to it.

The SDSS `spPlate` format e.g., with FLUX in the primary HDU and InvVar in the first extension.
SDSS-I/II `spSpec` also has FLUX in the primary HDU, though with ERR and MASK etc. folded into the same data "cube", and the generic IRAF reader does not even support reading from anything but HDU 0 (but also does not account for any uncertainty or mask data).
APOGEE uses a scheme more similar to what you are describing, but writing flux, StdErr and mask to extensions 1, 2 and 3, but it has its own dedicated loader – however perhaps `wcs1d` should adhere to that order as well, writing uncertainty before mask?
Bottom line, I don't have any robust percentage numbers on how all spectra out there are distributed, but for the basic image format using the primary HDU seems at least as common as the first extension. In any case the default for writing a flux-only spectrum has been `hdu=0` since specutils 1.1, and I don't think it's a good idea to change that now.
OTOH changing the default HDU to 1 depending on the presence of `uncertainty` and/or `mask` seems rather more confusing to me (e.g. what should it be if a mask exists, but is skipped in writing by `mask_name=None`?). But if that is what the majority of users find more logical, so be it...

pllim:
What do the other specutils maintainers think?

> should adhere to that order as well, writing uncertainty before mask?

Yes, I think it is common to have the `SCI`, `ERR`, `DQ` in that order.

pllim (approved):
I had to pass in extra options (`hdu=1, flux_name="SCI"`) to make an existing test pass over at https://github.com/spacetelescope/jdaviz/pull/2094 but it is not the end of the world.

LGTM. Thanks!

rosteen (approved):
Code looks fine to me, thanks!

pllim:
Please give me a ping when this ends up in a released version, so I can follow up over at Jdaviz to use this for Cubeviz. Thanks!

pllim:
When is the next release for specutils?

PR #1053, opened 2023-03-30
ASDF mask serialization

### LINKED ISSUE #1052: Round-trip mask through ASDF
Originally requested by @rosteen over at #1046 together with WCS, "In reviewing #1042 I noticed that the ASDF converters have never round-tripped the `mask` attribute. We should update the converters to do so."

[🐱](https://jira.stsci.edu/browse/JDAT-3236)

pllim:
Fix #1052

@braingram should also have a look to see if I did this wrong.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,7 +18,8 @@ Other Changes and Additions
 
 - ASDF tag for Spectrum1D is now compatible with ASDF v3.
   As a result, minversion of ``asdf`` has been bumped to 2.14.
-  Redundant ASDF schema for ``SpectralCoord`` is removed. [#1042]
+  Redundant ASDF schema for ``SpectralCoord`` is removed.
+  It also now supports ``mask`` serialization. [#1042, #1053]
 
 - JWST X1D reader will no longer raise a ``UnitWarning`` for surface brightness
   error. [#1050]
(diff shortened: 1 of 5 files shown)


braingram:
Overall these changes look good.

One thing to consider is how tag versioning will be handled. If a spectrum that contains a masked array is saved with an older version of specutils, the asdf file (with a 1.0.0 spectrum 1d tag) will not contain the mask (as noted in #1052). With the changes in this PR, the masked spectrum will be saved to the asdf file. However the file will contain the same 1.0.0 spectrum tag. If an older version of the software opens this newer file it will fail to load the mask and no errors or warnings will be raised (but the spectrum will not include the mask). If instead the tag was versioned (bumping it to 1.1.0) than specutils would generate a file with the 1.1.0 tag which would fail to convert for any older versions of specutils (and a AsdfConversionWarning would be issued). If this isn’t a concern (it looks like there has been no release since the converters were added) than a version increase doesn’t seem necessary.

pllim:
I am not wise in the ways of ASDF, so I will have to defer to specutils maintainers about version bump or not.

rosteen:
I don't think it's needed right now, since as said we haven't actually released the converters.

rosteen (approved):
LGTM, thanks!

PR #1057, opened 2023-05-15
Fix centroid and gaussian width analysis functions, add distribution-based implementation

rosteen:
We've had some reports that the uncertainties reported by `centroid`, `gaussian_sigma_width`, and `gaussian_fwhm` are suspect/wrong. This PR will update to using `astropy.uncertainty` Monte Carlo distributions to estimate the value and uncertainty for these quantities instead of the suspect analytic solutions, and give the option to use an updated analytic solution for `centroid` that @eteq re-derived. Opening this now to get eyes on my implementation, but I still need to update the gaussian measures.

### DIFF
--- a/docs/analysis.rst
+++ b/docs/analysis.rst
@@ -132,7 +132,13 @@ estimate the center of a spectral feature:
 
 While this example is "pre-subtracted", this function only performs well if the
 contiuum has already been subtracted, as for the other functions above and
-below.
+below. If the input spectrum has an ``uncertainty``, the result returned by
+`~specutils.analysis.centroid` will also have attached ``uncertainty`` and
+``uncertainty_type`` attributes. By default, the centroid and uncertainty results
+given are the analytical solution, but specifying ``analytic=False`` in the input
+to the function will instead return the mean and standard deviation of an
+`~astropy.uncertainty` Monte Carlo distribution generated using the ``uncertainty``
+values of the input spectrum's flux.
 
 
 Moment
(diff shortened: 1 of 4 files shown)


eteq:
For backstory, a picture of the derivation for the analytic solution is below.

centroid_unc_derivation

I cross-checked this solution and the implementation in this PR in a notebook here:
https://gist.github.com/eteq/37017d89ce3017e387de1e72edb40440

@rosteen, maybe the thing to do is to update this PR to include tests based on that notebook? That is, a test where you generate a fake spectrum with known uncertainty a bunch of times, compute the centroid, and then compare the stddev of those simulations to the uncertanty derived here? As long as you set a random number generator seed for that and a sufficient tolerance it shouldn't lead to mysterious failures due to bad luck of the draw.

eteq:
Oh, and given the growing complexity, maybe we want to change this PR to *just* the centroid, finish it up, and then do a separate PR for the others?

rosteen:
> Oh, and given the growing complexity, maybe we want to change this PR to _just_ the centroid, finish it up, and then do a separate PR for the others?

I'm going to try to get the gaussian width measures in here too, I'd rather not be inconsistent between those and the centroiding. 

A nice side effect of this is that I think I figured out the generalization of the centroiding math to multiple dimensions, so we aren't restricted to 1D or 2D like we were in the previous implementation of centroid.

rosteen:
@eteq  this is now ready for review. I ended up leaving the default as the analytic solution, since we're fairly certain it's correct now, and leaving the distribution-based solution as an option - I figured this is less disruptive/doesn't change a default that users might not notice. I noticed and fixed a couple other things along the way while adding/fixing tests, for example the sigma width uncertainty was previously just adding `sigma.unit` to the fractional uncertainty rather than correctly multiplying by `sigma` itself.

camipacifici:
From the specviz side: left is old, right is new. 
I'd say good job!! :)
Screen Shot 2023-06-13 at 2 34 02 PM

PR #1058, opened 2023-05-16
new implementation of FluxConservingResampler

### LINKED ISSUE #817: FluxConservingResampler shouldn't be O(n^2)
I ran some timing measurements down-sampling a spectrum with the various resamplers, and found `FluxConservingResampler` quickly becomes unusable for output sampling > a few thousand points.
FluxConservingResampler_timing

For comparison, `LinearInterpolatedResampler` and `SplineInterpolatedResampler` scale well in the (larger) domain checked here (note units are `s` instead of `min`):
resampler_timing

Since the output samples should be composable (at least with some overlap in the input range), I would expect the `FluxConservingResampler` algorithm to run in nearly linear time.  This algorithm is one of the main motivations for me to work with Specutils, so I hope we can improve its performance!

Computations run on a 2018 MacBook Pro 15" (2.9 GHz i9; 32GB RAM; OSX 10.15.7).

cshanahan1:
(Resolves issue #817, #1059)

This is a new version of FluxConservingResampler that addresses the issue described in #817. 

To summarize the changes made, the initial approach constructed a matrix of values that represented something about the fraction of overap between input wavelength bin i and output wavelength j. Most of the elements of this matrix will be zero, since only i_j values that have any overlap will be non-zero, so it is inefficient to use matrix operations here. 

In this new approach, it loops over each output bin, and then each input bin but excluding input bins determined in the previous iteration to have terminated in that output bin, and stoping when the front edge of the input bun exceeds the end of that output bin, therefore comparing the minimum number of bins required and calculating the output flux and errors in these bins.

As shown here, the speed increase for the new version is quite good:
Screen Shot 2023-05-15 at 11 07 36 PM

In the process of writing this, I noticed a subtle difference in flux and uncertainty outputs in certain cases when you have uneven sized output bins (it actually might be more general than this, and occur in cases of partial overlap). I opened an issue for this which has the justification and verification that the new results are correct (https://github.com/astropy/specutils/issues/1059). This PR solves that issue as well. Because of this, some of the truth values of the unit tests have changed.

EDIT: Fix #817, #1059

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,9 @@ Bug Fixes
 
 - Fix SNR calculations with both masks and regions. [#1044]
 
+- Reimplementation of FluxConservingResampler to fix incorrect flux and
+  uncertainties and increase speed. [#1058]
+
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -53,6 +56,7 @@ Other Changes and Additions
 
 - JWST X1D reader will no longer raise a ``UnitWarning`` for surface brightness
   error. [#1050]
+
 
 1.9.1 (2022-11-22)
 ------------------
--- a/specutils/manipulation/resample.py
+++ b/specutils/manipulation/resample.py
@@ -44,6 +44,7 @@ class ResamplerBase(ABC):
 
 
 class FluxConservingResampler(ResamplerBase):
+
     """
     This resampling algorithm conserves overall integrated flux (as opposed to
     flux density).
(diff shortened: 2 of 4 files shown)


pllim (commented):
Since you are going to speed here, I would change `var ** 2` to `var * var`.

Screenshot 2023-05-16 112121

And use `<<` to assign unit: https://docs.astropy.org/en/latest/units/index.html#performance-tips

pllim:
Can you also pls post the table that made the plot, so we have have concrete numbers to refer to? Thanks!

pllim:
Test failures appear related and should be fixed.

rosteen:
@sean-lockwood may also be interested in this, as the opener of the original issue.

sean-lockwood:
@cshanahan1 -
Based on reading your description, should there be a constraint that the wavelength arrays are sorted?

cshanahan1:
> @cshanahan1 - Based on reading your description, should there be a constraint that the wavelength arrays are sorted?

yes, and in ascending order. i'm considering how to check and warn for this, but maybe that can be addressed in a follow up PR since it's not currently checked anyway

cshanahan1:
> Test failures appear related and should be fixed.

Done, see https://github.com/astropy/specutils/issues/1059 for justification for changes.

cshanahan1:
> Can you also pls post the table that made the plot, so we have have concrete numbers to refer to? Thanks!

n old time new time
10 0.14875683400168782 1.982213707997289
100 2.3727758750028443 2.1637744170002406
250 5.837483707997308 2.1639707920039655
500 11.873500082998362 2.1583693330030655
1000 26.326482749995193 2.1945978329968057
2000 93.21111799999926 2.188627458002884
3000 260.69167562499933 2.286522624999634

I think I'm introducing some new overhead times somewhere, since its consistently ~2 seconds. For small # of bins this increases the time. I'll figure out where this is coming from. I think it might be because I'm converting back and forth between quantities/spectral axis types where the original function was able to work with the SpectralAxis objects directly to compute the resample matrix. I'll report back

cshanahan1:
my new function is much faster when you have 100+ output samples, but slower when less than that
Screen Shot 2023-05-17 at 8 44 25 PM

it asymptotes around 2s and the computation time grows slowly after that.

pllim:
@cshanahan1 , can you please try to profile it for the <100 data points and see where that bottleneck is? Maybe there is something we can do about it.

cshanahan1:
> @cshanahan1 , can you please try to profile it for the <100 data points and see where that bottleneck is? Maybe there is something we can do about it.

yes, working on that now. i was able to replace the inner loop with a while statement that increases the speed, but there is still a domain where the matrix approach is faster. i will post some exact times once i finish this

pllim:
I agree with others that this won't be a deal breaker, as I don't see anyone doing any real science with <100 bins... do they?

cshanahan1:
i'm going to close this PR and open a new one since most of this discussion is out of date and I don't want new reviewers to have to sift through it.

Issue #1059, opened 2023-05-16
FluxConservingResampler returning incorrect fluxes and errors


cshanahan1:
I am seeing a difference between the fluxes and errors returned by FluxConservingResampler and what is described in https://arxiv.org/abs/1705.05165, which this resampler should be implementing. Particularly, I see this in certain cases when there are uneven size bins (including the example that is used in the class docstring), but it may be more general than that and occur in any cases with partial overlap. 

For example lets say we want to resample wavelengths = [3, 5, 13]*u.nm, with fluxes [2.5, 4.5, 8.5]) * u.AB onto new output grid [ 2, 6, 10]  * u.AA. A visual representation of this is:

<img width="770" alt="Screen Shot 2023-05-16 at 5 04 55 PM" src="https://github.com/astropy/specutils/assets/20444899/5a3732d6-f92e-49f1-92a3-3eeb45c0e78e">

According the the flux conserving routine, the output fluxes should be [nan 4.5 7.5]. However, FluxConservingResampler is returning [nan, 4.5, 7.81].

I think the source of the error is the following: The resample matrix for this case as computed by FluxConservingResampler, which SHOULD represent, for each overlapping input, output bin the fractional overlap * the width of the input bin (i.e P_ij * w_i) is
[[ 0.  0.  0.]
 [ 0. 20.  0.]
 [ 0.  5. 24.]]
which appears to have an additional factor of w_i in it, and when you do the math you get a flux of 7.81

when this additional factor of w_i is removed, you get the correct flux of 
[[ 0.  0.  0.]
 [ 0. 4.  0.]
 [ 0.  1.  3.]]

((1 * 4.5) + (3 * 8.5)) / (3 + 1) = 7.5 for the last bin

A sanity check, stepping through each iteration in the method described in the paper, for each output bin, we have to find the fraction of each overlapping input bin that will contribute to the final flux.

**output bin 0 (centered at 2, edges 0 to 4)**
   - goes out of range of input bins. flux should be nan.
   **TOTAL FLUX IN OUTPUT BIN 0 = NaN**

**output bin 1 (centered at 6, edges 4 to 8)** 
   - does not overlap input bin 0. no contribution to flux
   - overlaps input bin 1 (centered at 5, edges 4 to 9)
       - what fraction of input bin 1 is inside output bin 1? 0.8.
       - (fraction of input bin in output (P_ij) * width of input bin (w_i) = 0.8 * 5 = 4.0
[...]

PR #1060, opened 2023-05-19
Fix incorrect fluxes and uncertainties returned by FluxConservingResampler, increase computation speed 

cshanahan1:
(Resolves issue https://github.com/astropy/specutils/issues/817, https://github.com/astropy/specutils/issues/1059)

As explained in #1059, the current implementation of FluxConservingResampler does not do the exact method described in https://arxiv.org/abs/1705.05165. The discrepancy becomes apparent in cases of partial overlapping bins with uneven bin sizes. This PR fixes that, and will return new (correct) fluxes and errors in those cases. For this reason, some of the 'truth' values in the tests are changed here.

Additionally, this fixes the issue described in #817 where the computation time is very long (especially in cases like the example there, resampling ~a million points down to a few hundred or thousand). With the new approach in this PR, the time to resample decreases from 4 minutes to about half a second. The following plot and table shows the speed comparison for the current version vs. this new one.
compare

| n_bins | old time | new time |
| --- | --- | --- |
| 2 |      0.0488 | 0.0046 |
| 3 |       0.0463 |  0.1264 |
| 20 |      0.291 |  0.2337 |
| 50  |     0.7302 |  0.2456 |
| 100  |    2.8603 |  0.2528 |
| 250  |    6.9634 |  0.253 |
| 500  |    14.6813 |  0.2578 |
| 1000  |   30.487 |  0.2655 |
| 2000  |   94.8554 |  0.3083 |
| 3000  |   255.93 |  0.4516 |

I do want to note that there is a small domain where the matrix approach is faster (a million down to 2-20 samples in that case, see table). We could consider eventually also supporting a matrix operation approach in addition to this method, and determining based on the input which should be faster, but for now I think this new version represents a significant improvement for most cases.

To summarize the changes, the current approach constructs a (n_input_samples) * (n_output_samples) array, where each matrix element (ij) represents the overlap between an input (i) and output (j) bin. Therefore, these matrices are huge with most elements 0.

The new approach calculates the new fluxes and errors iteratively (looping over output wavelengths), comparing the minimal number of input wavelengths possible. It starts by clipping out any bins from the output wavelength array that go out of range of the input wavelengths, and then another clip again on the input bin array after figuring out the minimal overlapping bin. From there, it will compare bins until the front edge of an output bin exceeds the final edge of that output bin. If an input bin terminates within an output bin, it will be excluded in the comparison next time, insuring there are no redundant comparisons. 

(I closed the first PR for this https://github.com/astropy/specutils/pull/1058)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ New Features
 
 Bug Fixes
 ^^^^^^^^^
+- Reimplementation of FluxConservingResampler to fix incorrect flux and
+  uncertainties and increase speed. [#1060]
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -38,6 +40,7 @@ Bug Fixes
 
 - Fix SNR calculations with both masks and regions. [#1044]
 
+
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -53,6 +56,7 @@ Other Changes and Additions
 
 - JWST X1D reader will no longer raise a ``UnitWarning`` for surface brightness
   error. [#1050]
+
 
 1.9.1 (2022-11-22)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,8 +6,7 @@ New Features
 
 Bug Fixes
 ^^^^^^^^^
-- Reimplementation of FluxConservingResampler to fix incorrect flux and
-  uncertainties and increase speed. [#1060]
+- Reimplementation of FluxConservingResampler. It is now faster and yields more accurate results. [#1060]
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/specutils/manipulation/resample.py
+++ b/specutils/manipulation/resample.py
@@ -77,19 +77,6 @@ class FluxConservingResampler(ResamplerBase):
     <Spectrum1D(flux=<Quantity [  nan,  3.  ,  6.  ,  7.  ,  6.25, 10.  , 20.  ,   nan,   nan] mJy>, spectral_axis=<SpectralAxis [ 1.,  5.,  9., 13., 14., 17., 21., 22., 23.] nm>)>
     """
 
-    @staticmethod
-    def _get_overlap_area(a1, a2, b1, b2):
-        # get intersection area of 2 bins bounded by
-        # ``bin1_bounds``=(a1, b1) and ``bin2_bounds``=(a2, b2)
-        # assuming its already known they overlap
-        if b1 >= a1:
-            if b2 > a2:
-                return a2 - b1
-            return b2 - b1
-        if b2 > a2:
-            return min(b2, a2) - max(b1, a1)
-        return b2 - a1
-
     def _fluxc_resample(self, input_bin_centers, output_bin_centers,
                         input_bin_fluxes, errs):
         """
--- a/specutils/manipulation/resample.py
+++ b/specutils/manipulation/resample.py
@@ -44,6 +44,7 @@ class ResamplerBase(ABC):
 
 
 class FluxConservingResampler(ResamplerBase):
+
     """
     This resampling algorithm conserves overall integrated flux (as opposed to
     flux density).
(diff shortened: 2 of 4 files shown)


pllim:
The timing stated in the table is very impressive. Good job! 👏 

Does @sean-lockwood want to take this PR branch for a spin with his original use case?

> I do want to note that there is a small domain where the matrix approach is faster (a million down to 2-20 samples in that case, see table)

I don't think `0.05 s -> 0.1 s` for that narrow range of samples is worth re-introducing matrix.

specutils/manipulation/resample.py
```diff
@@ -157,64 +248,45 @@ def resample1d(self, orig_spectrum, fin_spec_axis):
...
-        # Calculate output uncertainty
-        if pixel_uncer is not None:
-            pixel_uncer = pixel_uncer.reshape(new_flux_shape)
-
-            out_variance = np.sum(pixel_uncer * resample_grid**2, axis=-1) / np.sum(
-                resample_grid, axis=-1)**2
-            out_uncertainty = InverseVariance(np.reciprocal(out_variance))
+
+        # handle 2d flux inputs
+        if orig_spectrum.flux.ndim == 2:
```
rosteen, line 261: I think this is the only thing in this implementation that worries me - the previous one handled any higher dimensionality for flux, and I think that the most common cases are 1D and 3D. Could you generalize this at least to 3D?

cshanahan1: done!

CHANGES.rst
```diff
@@ -6,6 +6,8 @@ New Features
 
 Bug Fixes
 ^^^^^^^^^
+- Reimplementation of FluxConservingResampler to fix incorrect flux and
+  uncertainties and increase speed. [#1060]
```
eteq, line 10:
```suggestion
- Reimplementation of FluxConservingResampler. It is now faster and yields more accurate results. [#1060]
```
Rewording this to be more "user facing" - i.e., it tells what you get from this PR.

specutils/manipulation/resample.py
```diff
@@ -73,59 +74,151 @@ class FluxConservingResampler(ResamplerBase):
...
+            fill_val = 0
+
+        # get bin edges from centers
+        input_bin_edges = input_bin_centers.bin_edges.value
+        output_bin_edges = output_bin_centers.bin_edges.value
+
+        # convert to regular arrays
+        input_bin_centers = input_bin_centers.value
+        output_bin_centers = output_bin_centers.value
+        input_bin_fluxes = input_bin_fluxes.value
```
eteq, line 116: Is it actually necessary to use `.value` here? It doesn't matter too much given the conversion to match units has already happened, but if it doesn't have a negative consequence I think it's safer not to use `.value` since there's always the chance that this will catch some unit conversion issues that slip through.

eteq (approved):
Looking good!  I have a couple of take-it-or-leave-it-as-you-wish inline comments, @cshanahan1, and a few broader items, but all are optional some I'm approving this meaning "you can either make these changes or not and I'm ok with this either way".

General items:
 1) https://arxiv.org/abs/1705.05165 mentions the `spectres` backage.  I also mentioned this in https://github.com/astropy/specutils/issues/817#issuecomment-1497776023 but as @larrybradley pointed out in https://github.com/astropy/specutils/issues/817#issuecomment-1516745403 it is license-incompatible, so we cannot use the `spectres` code.  *however*, that doesn't stop us using `spectres` to build a test-case for this code (see https://www.gnu.org/licenses/gpl-faq.en.html#GPLOutput which highlights that the GPL cannot be applied to outputs of programs).  So I would encourage adding a test where you generate an *output* spectrum from spectres and test that this new version matches that. This will hopefully ensure that in the future we don't accidentally make breaking changes that violate the algorithm like what clearly happened originally here.

2) This is probably not a suggestion for *this* PR, rather a possible solution for the point you made that the matrix version is faster for small datasets: consider instead implementing this algorithm as a *sparse* matrix operation.  As I mentioned in https://github.com/astropy/specutils/issues/817#issuecomment-1497776023 I think that's the flaw in the original implementation - it's O(n^2) because almost all the time is spent multiplying zeros by zeros, and that's *exactly* the reason why sparse matrix libraries exist, and I suspect it will end up a lot faster than this for-loop based implementation if we can recast it that way.  But given this algorithm is still way way better than the previous one, this is definitely "future work" and shouldn't block this PR.

3) as an alternative to the above - a similar "how to speed it up for small n or other cases" approach might be to use numba or one of the other python jitting tools.  Those work especially well in cases like this where there's a for loop over an array, so it might be an easy way to get a good speedup.  Again, though, no need to block this PR unless you think you can do it really easily and quickly.

specutils/manipulation/resample.py
cshanahan1: shouldn't be necessary, i was getting a bunch of TypeErrors i didn't want to track down haha. i'll look into it

cshanahan1: Done!

cshanahan1:
Thanks for the review @eteq. I added a new test that takes care of item 1 on your list of suggestions - is that what you had in mind? I ran the code from the other package to generate output for the test rather than adding it as a test dependency and generating it each time.

As for items 2 and 3, I think those are good ideas and I will make follow up tickets for them but I want to get this PR merged first.

sean-lockwood:
I ran the calculations again, so the numbers won't exactly match the plots above.  Values are in seconds.

```
# out      1.10.0       0.1.dev2966+g7c71c41

 200     2.145341       0.179401
 250     2.625112       0.144806
 300     3.138709       0.150052
 350     3.728590       0.147948
 400     4.217818       0.147381
 450     4.709783       0.147746
 500     5.271168       0.152145
 550     5.813444       0.150349
 600     6.293068       0.156625
 650     6.798396       0.151181
 700     7.413938       0.151452
1200    12.702604       0.157407
1700    29.881622       0.164718
2200    50.650592       0.171427
2700    73.037056       0.181266
```

rosteen (approved):
Thank you for all the work on this, this is a great improvement. I'll hold off on hitting the merge button until tomorrow to see the answer to @pllim's last question (although I think that concern was addressed by checking against the `specres` output).

sean-lockwood:
Ah, I understand.  I just took a quick look at the output flux.  Most points do well, but a handful have up to 30% discrepancy in flux.  I would have to do further testing to understand the origin of the difference.

flux_comparison

cshanahan1:
@sean-lockwood if you can send me your input spectral axis, flux, and err values i can compare it against Spectres to make sure those differences are expected

sean-lockwood:
The uncertainty array was None in my test.

Input and outputs here:
[spec_files.zip](https://github.com/astropy/specutils/files/11617745/spec_files.zip)

Read output arrays via:
```python
import pickle

with open('v1.10.0_downsampled.pkl', 'rb') as f:
    v1_wave, v1_flux = pickle.load(f)

with open('new_downsampled.pkl', 'rb') as f:
    new_wave, new_flux = pickle.load(f)
```

Generated via:
```python
import astropy.units as u
import specutils
from specutils.manipulation import FluxConservingResampler

print(specutils.__version__)

orig = Spectrum1D.read('lte012.0-3.5-0.0a+0.0.BT-Settl.spec.fits')

resampler = FluxConservingResampler()
output_wavelengths = np.linspace(12.58e-4, 776247.11e-4, num=1500) * u.um
resampled_spec = resampler(orig, output_wavelengths)
```

cshanahan1:
the bad news is that I spent too much time chasing down a non-issue, but the good news is that i determined it was a non-issue: spectres and this new version of FluxConservingResampler match basically within floating point error. (my dumb mistake and what made me worried was that i saw 0.2-0.3 offsets and interpreted these as percent differences when they were in fact absolute differences). 

it makes sense that @sean-lockwood was seeing a general agreement but an occasional 20-30% difference between the old version and the new version - those were the cases of overlapping input+output bins that were previously being treated incorrectly that i described in the linked issue. those few instances can be seen here:
Screen Shot 2023-06-01 at 11 38 43 PM


what matters is that spectres and this new version agree, and they do:

Screen Shot 2023-06-01 at 11 36 53 PM

pllim:
How big is this "lte012.0-3.5-0.0a+0.0.BT-Settl.spec.fits" file? Maybe we should turn this Spectre use case into a unit test (as a follow-up PR) to prevent regression.

pllim (approved):
Based on Sean's response and the subsequent finding, I am satisfied with both performance improvement and bug fix.

Thank you very much, everyone! And excellent work, @cshanahan1 !

sean-lockwood:
Attaching a truncated version of the input file (now 847k uncompressed):
[truncated_lte012.0-3.5-0.0a+0.0.BT-Settl.spec.fits.zip](https://github.com/astropy/specutils/files/11636837/truncated_lte012.0-3.5-0.0a%2B0.0.BT-Settl.spec.fits.zip)

Still gives the discrepant point:
truncated_comparison

cshanahan1:
@sean-lockwood it makes sense there are some discrepant points when comparing the new version vs the old - this PR also includes a fix for some instances where fluxes and errors were being calculated incorrectly (when input and output bins overlap in a certain way).

sean-lockwood:
@cshanahan1 
Makes sense to me.  Was just attaching a smaller version of the file for the unit test.

pllim:
Thanks! I created https://github.com/astropy/specutils/issues/1064 to add a test, for whoever who wants to tackle it.

PR #1061, opened 2023-05-26
Assert spectral axis in Spectrum1D is sorted

cshanahan1:
This PR adds a constraint when initializing a Spectrum1D object that the spectral axis is either strictly ascending or descending, and raises an error if that condition is not met. A new attribute indicating which order the spectral axis is in (``spec1d.spectral_axis_order``) is added. This will be used, for example, by Specreduce wavelength calibration to make sure the spectral axis is ascending. I also changed some of the tests accordingly. 

I decided to do this on Spectrum1D rather than SpectralAxis - I thought that doing it there might be too restrictive? I'm not sure.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,7 @@ Bug Fixes
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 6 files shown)


rosteen (approved):
LGTM, thanks! Good to actually check for this rather than just saying it's a requirement in the docs.

PR #1066, opened 2023-06-21
Fix SDSS-I/II spSpec units

aragilar:
It looks like something changed (I've not yet worked out what) in how astropy/wcslib handled absent `CUNITa`, and now the loader is incorrectly using metres instead of using angstroms. As the WCS was already incorrectly encoded, we probably can't trust `CUNITa` anyway, so force angstroms.

This also adds a test to confirm that the unit does not change again.

### DIFF
--- a/specutils/io/default_loaders/sdss.py
+++ b/specutils/io/default_loaders/sdss.py
@@ -30,7 +30,10 @@ def _sdss_wcs_to_log_wcs(old_wcs):
     claims to be linear, but is logarithmic in base-10.
     The wavelength is given by:
     λ = 10^(w0 + w1 * i)
-    with i being the pixel index starting from 0.
+    with i being the pixel index starting from 0. This formula is documented at
+    https://classic.sdss.org/dr7/products/spectra/read_spSpec.php and appears to
+    be the same across all SDSS-I and SDSS-II data releases (replace dr7 with
+    dr<x> in the above URL).
 
     The FITS standard uses a natural log with a sightly different formulation,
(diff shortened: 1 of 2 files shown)


pllim (commented):
"The WCS in the SDSS files does not appear to follow the WCS standard" o_o

Since it is wonky to begin with, I guess this is fine, but please confirm with SDSS that this is universal for all their DRs. If not, you might also need to filter by DR number or something?

aragilar:
The WCS issue is only for the spSpec files, so only affects SDSS-I and SDSS-II, SDSS-III and later use a different format and hence loader, and are based on FITS tables rather than WCS, and so avoid all the WCS unit issues.

https://classic.sdss.org/dr7/products/spectra/read_spSpec.php and https://classic.sdss.org/dr7/dm/flatFiles/spSpec.php document the format, and appear to be similar for the earlier data releases as well, and are identical for the relevant parts documenting the wavelength scale. These emphasise the misuse (i.e. sticking the log parameters into a linear WCS) is intentional, it may be that the format (or at least the choice of it) predates the standardisation of logarithmic WCS?

I've added links to the above pages to the docstrings, so the next person who looks at that code knows what the format is documented as (rather than having to dig through issues).

pllim:
@aragilar , can you please rebase? The devdeps job should be all green unless this PR introduced new failures.

pllim (approved):
CI is green. Thanks!

pllim:
I dunno how this package do milestones, so I'll let @rosteen or @tepickering handle that.

PR #1068, opened 2023-06-23
Convert an automatic registration into an option

tbowers7:
Per the TODO suggestion in the code, this PR makes registering a SpectrumList reader for any `data_loader` that reads Spectrum1D objects optional (rather than automatic) via a boolean input flag.  The default behavior is identical to the existing code, but the new flag allows users to opt out in situations where such registration causes errors.

### DIFF
--- a/specutils/io/registers.py
+++ b/specutils/io/registers.py
@@ -26,7 +26,7 @@ def _astropy_has_priorities():
 
 
 def data_loader(label, identifier=None, dtype=Spectrum1D, extensions=None,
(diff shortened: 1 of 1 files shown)


aragilar:
This looks good to me, but I would call the option `autogenerate_spectrumlist`.

rosteen (approved):
Looks good to me, thanks. I just fixed the unrelated codestyle errors in another PR, they don't need to hold this up.

PR #1071, opened 2023-07-31
Fix some comparisons that should have been using `isinstance`

rosteen:
I'm not sure why these suddenly started getting caught by our codestyle checks, but here's the fix.


rosteen:
Test failure is unrelated, merging.

PR #1074, opened 2023-08-22
Improved GWCS handling in Spectrum1D

rosteen:
This PR allows Spectrum1D to keep the entire GWCS when reading a file (for example a JWST cube), rather than dropping any spatial WCS information as done previously. Probably most applicable beyond just JWST cubes but that's the use case I was targeting off the bat here. This has to go in v2.0 because GWCS doesn't have a `swapaxis` method like WCS does.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,7 @@ New Features
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-1.11.0 (unreleased)
+1.12.0 (unreleased)
 -------------------
 
 New Features
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
-
 2.0.0 (unreleased)
 ------------------
+
+New Features
+^^^^^^^^^^^^
 
 - Spectral axis can now be any axis, rather than being forced to be last. See docs
   for more details. [#1033]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,15 @@ New Features
 
 - Spectral axis can now be any axis, rather than being forced to be last. See docs
   for more details. [#1033]
-  
+
+- Spectrum1D now keeps the entire input GWCS instead of extracting only the spectral
+  information. [#1074]
+
+Other Changes and Additions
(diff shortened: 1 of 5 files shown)


specutils/io/default_loaders/jwst_reader.py
```diff
@@ -643,13 +644,10 @@ def _jwst_s3d_loader(filename, **kwargs):
 
             # get mask information
             mask_name = primary_header.get("MASKEXT", "DQ")
-            mask = hdulist[mask_name].data.T
+            mask = hdulist[mask_name].data
```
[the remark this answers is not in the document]
rosteen: Specutils currently demands that the spectral axis be last in a multidimensional spectrum. Specutils 2.0 is going to remove that requirements, so we'll be able to keep the data in the same shape as you would get if you read it in with say `astropy.io.fits`.

CHANGES.rst
```diff
@@ -7,10 +7,15 @@ New Features
...
   for more details. [#1033]
 
+- Spectrum1D now keeps the entire input GWCS instead of extracting only the spectral
+  information. [#1074]
+
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-1.11.0 (unreleased)
+- JWST reader no longer transposes the input data cube for 3D data. [#1074]
```
cshanahan1, line 16: correct me if im wrong, this was just my understanding from our convo earlier - but would a better description be that Spectrum1D now allows GWCS as input, and when using the JWST reader to load a Spectrum1D the full GWCS is now loaded?

cshanahan1 (approved):
looks good, just suggested a possible change to the change log entry to reflect the changes made here.

not essential as most lines are covered in this PR, but a test demonstrating creating a Spectrum1D with a GWCS might be appropriate too.

CHANGES.rst
rosteen: Good call, I'll update it.

rosteen: Does that sound better?

rosteen:
I'm going to defer making a test/documentation for creating a Spectrum1D with user-created GWCS, I'll open an issue.

PR #1078, opened 2023-09-11
added filter_non_finite flag to fit_lines

cshanahan1:
This PR adds the 'filter_non_finite' flag to specutils.fitting.fit_lines. By default, this flag is False, and it can only be True if the fitter is a `LevMarLSQFitter`. This fixes an issue when there are non-finite values in the weights array, and allows the fit to proceed with these filtered out.

### DIFF
--- a/specutils/fitting/fitmodels.py
+++ b/specutils/fitting/fitmodels.py
@@ -515,10 +515,6 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(calc_uncertaintie
             warnings.warn('`filter_non_finite` can only be True when fitter is LevMarLSQFitter. Setting to False.')
             filter_non_finite = False
 
-    print('FILTER NON FINITE', filter_non_finite)
-    print(flux)
-    print(weights)
-
     fit_model = fitter(model, dispersion,
                        flux, weights=weights, filter_non_finite=filter_non_finite, **kwargs)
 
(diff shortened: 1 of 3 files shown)


specutils/fitting/fitmodels.py
```diff
@@ -503,8 +508,19 @@ def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(calc_uncertaintie
...
+    if filter_non_finite is True:
+        if isinstance(fitter, fitting.LevMarLSQFitter):  # fine, just checking for clarity
+            pass 
+        else:
+            warnings.warn('`filter_non_finite` can only be True when fitter is LevMarLSQFitter. Setting to False.')
+            filter_non_finite = False
+
+    print('FILTER NON FINITE', filter_non_finite)
+    print(flux)
+    print(weights)
```
kecnry, line 520:
```suggestion
```

specutils/fitting/fitmodels.py
```diff
@@ -377,7 +382,7 @@ def fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(calc_uncertainties
 
 def _fit_lines(spectrum, model, fitter=fitting.LevMarLSQFitter(calc_uncertainties=True),
                exclude_regions=None, weights=None, window=None, get_fit_info=False,
-               ignore_units=False, **kwargs):
+               ignore_units=False, filter_non_finite=False, **kwargs):
```
rosteen, line 385: I think you'll need to add `filter_non_finite` to the `kwargs` in the `LevMarLSQFitter`, looks like tests are failing because other fitters don't expect that keyword argument.

cshanahan1: ah good point, I thought I caught this by setting it to False when the fitter isn't LevMarLSQFitter but those fitters won't expect the flag at all

PR #1079, opened 2023-09-12
Compatibility with unreleased ASDF stack for JWST s2d data

pllim:
This will hopefully address https://github.com/spacetelescope/jdaviz/issues/2446 . Minimally reproducible example that crashes without this patch using unreleased ASDF stack:

```python
from astropy.utils.data import download_file
fn = download_file('https://stsci.box.com/shared/static/exnkul627fcuhy5akf2gswytud5tazmw.fits', cache=True)
sp = Spectrum1D.read(fn)
```

[🐱](https://jira.stsci.edu/browse/JDAT-3786)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,17 +1,8 @@
-import pkg_resources
-
 try:
     from pytest_astropy_header.display import PYTEST_HEADER_MODULES, TESTED_VERSIONS
     ASTROPY_HEADER = True
 except ImportError:
     ASTROPY_HEADER = False
-
-entry_points = []
-for entry_point in pkg_resources.iter_entry_points('pytest11'):
-    entry_points.append(entry_point.name)
-
-if "asdf_schema_tester" not in entry_points:
-    pytest_plugins = ['asdf.tests.schema_tester']
 
 
 # Repeat this from specutils/conftest.py so tox picks it up.
--- a/specutils/io/default_loaders/jwst_reader.py
+++ b/specutils/io/default_loaders/jwst_reader.py
@@ -445,7 +445,7 @@ def _jwst_s2d_loader(filename, **kwargs):
     slits = None
 
     # Get a list of GWCS objects from the slits
-    with asdf_in_fits.open(filename) as af:
+    with fits.open(filename, memmap=False) as hdulist, asdf_in_fits.open(hdulist) as af:
         # Slits can be listed under "slits", "products" or "exposures"
         if "products" in af.tree:
             slits = "products"
@@ -460,8 +460,6 @@ def _jwst_s2d_loader(filename, **kwargs):
         # Create list of the GWCS objects, one for each slit
         if slits is not None:
             wcslist = [slit["meta"]["wcs"] for slit in af.tree[slits]]
-
-    with fits.open(filename, memmap=False) as hdulist:
 
         primary_header = hdulist["PRIMARY"].header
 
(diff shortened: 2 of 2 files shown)


braingram:
Looks good to me except I'm not sure the astropy dev testing change is working.
I see here it installed astropy 5.3.3:
https://github.com/astropy/specutils/actions/runs/6161782664/job/16721804170?pr=1079#step:6:35

pllim:
Good catch...

pllim:
But astropy 5.3.3 is also pulled into the devdeps job in unrelated PR (https://github.com/astropy/specutils/actions/runs/6147346407/job/16678644002). So something is pinning it but not caused by the changes here.

braingram:
It also doesn't appear to be grabbing numpy 2.0:
https://github.com/astropy/specutils/actions/runs/6161782664/job/16721804170?pr=1079#step:6:59
:-|

pllim:
I should investigate that separately...

pllim:
BTW numpy is fine, it is grabbing RC instead of dev but that is a known issue.

pllim:
For my own sanity, I moved the other fixes out to separate PRs. Now this PR only deals with the parser.

* https://github.com/astropy/specutils/pull/1080
* https://github.com/astropy/specutils/pull/1081

pllim:
Wut?

```
INTERNALERROR>   File "specutils/utils/wcs_utils.py", line 7, in <module>
INTERNALERROR>     from gwcs import WCS as GWCS
INTERNALERROR>   File "gwcs/__init__.py", line 72, in <module>
INTERNALERROR>     from .wcs import *   # noqa
INTERNALERROR>     ^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "gwcs/wcs.py", line 20, in <module>
INTERNALERROR>     from . import coordinate_frames as cf
INTERNALERROR>   File "gwcs/coordinate_frames.py", line 17, in <module>
INTERNALERROR>     from astropy.coordinates import StokesCoord
INTERNALERROR> ImportError: cannot import name 'StokesCoord' from 'astropy.coordinates' (astropy/coordinates/__init__.py)
```

pllim:
@braingram , what is this? Is this really necessary in specutils?

https://github.com/astropy/specutils/blob/85d07a98bf9ed6478d7ba43270e44e7ea990c43f/conftest.py#L9-L14

braingram:
> @braingram , what is this? Is this really necessary in specutils?
> 
> https://github.com/astropy/specutils/blob/85d07a98bf9ed6478d7ba43270e44e7ea990c43f/conftest.py#L9-L14

I don't think that's required for any of the asdf schema testing. The testing should be enabled by:
https://github.com/astropy/specutils/blob/85d07a98bf9ed6478d7ba43270e44e7ea990c43f/setup.cfg#L63-L64
See: https://asdf.readthedocs.io/en/latest/asdf/extending/schemas.html#testing-custom-schemas

Testing locally by removing the lines I still see the schemas tests. The CI does currently test the schemas, see:
https://github.com/astropy/specutils/actions/runs/6162339183/job/16723997336?pr=1079#step:6:155
I would expect that removing the lines shouldn't disable the `spectrum1d-1.0.0.yaml` etc tests.

pllim:
Good catch, @braingram . I pushed new changes. Does it look good now?

braingram (approved):
Looks great! Thanks for fixing this.

pllim:
I don't understand the GWCS error on devdeps job though. It seems unrelated but it is also blocking the test against asdf dev here.

braingram:
My suspicion is it has something to do with `astropy-iers-data`. The logs show attempts to install several astropy versions:
https://github.com/astropy/specutils/actions/runs/6163642724/job/16727823617?pr=1079#step:6:48
before 'succeeding' with `...dev94...` which is the newest version that doesn't list `astropy-iers-data` as a dependency. Why pip is doing this is a mystery to me.

pllim:
Oh, right... That would theoretically be fixed by https://github.com/astropy/specutils/pull/1081... So I guess I will merge this and worry about the devdeps in that PR. 🤪 

Thanks!

pllim:
asdf 3.0 is released, so we also need to release this... [🐱](https://jira.stsci.edu/browse/JDAT-3889)

pllim:
Thank YOU!

PR #1080, opened 2023-09-12
DOC: Fix RTD

pllim:
RTD is failing.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,7 @@
 version: 2
 
 build:
-  os: "ubuntu-20.04"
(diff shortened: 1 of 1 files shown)


pllim:
I don't even want to know why GWCS is failing devdeps here. This patch is only for ReadTheDocs, so the failure is unrelated.

pllim:
@rosteen , not sure why it takes so long. It is a simple fix. Should I just YOLO this?

rosteen (approved):
Could have sworn I already did this, must have been another package. Thanks.

pllim:
Thanks!

PR #1089, opened 2023-10-03
Update test envs now that python 3.12 is released

rosteen:
Updates our OS tests to all use python 3.12 and adds a test to keep checking against 3.11.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -25,20 +25,25 @@ jobs:
             python: 3.x
             toxenv: codestyle
 
-          - name: Python 3.11 with astropy data and coverage
+          - name: Python 3.12 with astropy data and coverage
             os: ubuntu-latest
-            python: '3.11'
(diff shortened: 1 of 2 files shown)


pllim (commented):
Hmm, don't you need quotes around python for python>=3.10 ?

```yaml
python: '3.11'
```

rosteen:
> Hmm, don't you need quotes around python for python>=3.10 ?

I thought the problem was specific to 3.10, since parsing that as a float would make it 3.1, whereas 3.11 would keep the full value.

pllim:
oldest deps failure is caused by #1091

devdeps failure is caused by something preventing scipy from upgrading to dev version: `numpy==2.0.0.dev0` vs `scipy==1.11.3` (while we're at it, looks like matplotlib also failed to upgrade, which is very annoying)

pllim:
I fixed oldest-deps, so you should rebase if you still want to pursue this.

PR #1095, opened 2023-11-14
Replace tmpdir with tmp_path

### LINKED ISSUE #1039: TST: Replace tmpdir with tmp_path in pytest
`tmpdir` is considered legacy now and should be replaced with `tmp_path`.

https://docs.pytest.org/en/stable/how-to/tmp_path.html

rosteen:
Closes #1039 .

### DIFF
--- a/specutils/io/default_loaders/tests/test_jwst_reader.py
+++ b/specutils/io/default_loaders/tests/test_jwst_reader.py
@@ -366,7 +366,7 @@ def generate_s3d_wcs():
 
 
 @pytest.fixture()
-def tmp_asdf(tmp_path):
+def tmp_asdf():
     # Create some data
     sequence = np.arange(100)
     squares  = sequence**2
--- a/specutils/io/default_loaders/tests/test_jwst_reader.py
+++ b/specutils/io/default_loaders/tests/test_jwst_reader.py
@@ -94,9 +94,9 @@ def spec_multi(request):
 @pytest.mark.parametrize('spec_multi, format',
                          [('EXTRACT1D', 'JWST x1d multi'),
                           ('COMBINE1D', 'JWST c1d multi')], indirect=['spec_multi'])
-def test_jwst_1d_multi_reader(tmpdir, spec_multi, format):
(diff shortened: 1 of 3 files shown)


specutils/io/default_loaders/tests/test_jwst_reader.py
```diff
@@ -366,7 +366,7 @@ def generate_s3d_wcs():
 
 
 @pytest.fixture()
-def tmp_asdf(tmpdir):
+def tmp_asdf(tmp_path):
```
pllim, line 369: But why? It looks unused.

```suggestion
def tmp_asdf():
```

pllim (commented):
Seems uncontroversial enough though maybe one of the tests does not need `tmp_path` in the first place.

specutils/io/default_loaders/tests/test_jwst_reader.py
rosteen: Hm, it does indeed 🤔. I'm surprised codestyle didn't catch that.

pllim: Some fixtures do magic behind the scenes but I think tmp_path is not needed unless it is explicitly used?

specutils/io/default_loaders/tests/test_jwst_reader.py
```diff
@@ -366,7 +366,7 @@ def generate_s3d_wcs():
 
 
 @pytest.fixture()
-def tmp_asdf(tmpdir):
+def tmp_asdf():
```
pllim, line 369: What hurts my head is `tmp_asdf` itself is a fixture but it took in `tmpdir` fixture. What was it trying to do?

pllim (approved):
LGTM now as long as CI agrees with me. Thanks!

pllim:
I will let you merge.

specutils/io/default_loaders/tests/test_jwst_reader.py
rosteen: It is a mystery 😕

PR #1097, opened 2023-11-15
Fix functions missing from __all__.

### LINKED ISSUE #1094: Why are SDSS spPlate readers "hidden" in the sdss.py loader?
The `__all__` object in `specutils.io.default_loaders.sdss` does not list `spPlate_identify` or `spPlate_loader`. Is this just an oversight?  If so, I'd be happy to submit a quick PR to just add those.

weaverba137:
This PR closes #1094, and adds a little bit more documentation to the docstring.


rosteen (approved):
Looks good to me, thanks. Would be nice to add an spPlate test as well that uses this, but that can be a separate follow-up (and you don't necessarily have to be the one to do it!)

PR #1104, opened 2023-11-15
add tabular-fits support for Spectrum1D mask

sbailey:
This PR fixes a bug in the Spectrum1D.read / write default tabular-fits format, which was dropping the .mask attribute.  It adds tests demonstrating the problem and failing before the fix, and with the fixes these tests now pass.

Detail: due to astropy/astropy#11963, by default int8 masks are written as bool thus losing information, so here I upcast int8 masks to int16 with the result that the dtype changes in the round trip but at least information isn't lost.


rosteen (approved):
I took the liberty of fixing the failing codestyle check, I think this is good to go in now. Thanks!

sbailey:
@rosteen thanks for proactively fixing the codestyle issues and merging this.  I've been busy enough with my day job that I hadn't even looked at what was failing, fearing it would be worse and take more time than I had available.

Issue #1096, opened 2023-11-14
Round tripping read/write does not preserve WCS


cshanahan1:
There are some issues with roundtripping of WCS when using Spectrum1D read and write.

If you create a spectrum 1D with a wcs:
`from astropy.wcs import WCS`
`from specutils import Spectrum1D`
`import astropy.units as u`

`wcs = WCS(header={ 'CDELT1': 1, 'CRVAL1': 6562.8, 'CUNIT1': 'Angstrom',
'CTYPE1': 'WAVE', 'RESTFRQ': 1400000000, 'CRPIX1': 25})`
`spec1d = Spectrum1D(spectral_axis = (1, 2, 3)*u.nm, flux=(1, 2, 3)*u.Jy, wcs=wcs)`

Use that wcs to transform the spectral axis:
`spec1d.wcs.pixel_to_world(spec1d.spectral_axis).value`
`output:
      array([6.5398e-07, 6.5408e-07, 6.5418e-07])`

Then write out that Spectrum1D to file:
`spec1d.write('test_output.fits')`

And it back in (either using Spectrum1D.read OR fits.open. Neither work because the header keywords weren't preserved),
the wcs on the resulting Spectrum1D isn't saved.
`spec1d.read('test_output.fits').wcs.pixel_to_world(spec1d.spectral_axis.value).value`
`output:
      array([ 2.,  3., nan])`

This issue was discovered when working on https://github.com/astropy/specutils/issues/905#issuecomment-1811320860 at a hack day with @kelle

pllim:
There are so many formats in the registry, when you invoke `write` without specifying the desired format, what format did it pick up? Maybe you want to give explicit format to write out to (as documented in https://specutils.readthedocs.io/en/latest/custom_loading.html#custom-writer)?

cshanahan1:
Spectrum1D.write allows 'tabular-fits' and 'wcs1d-fits', with the default being tabular-fits. Neither save the WCS at least in my test case.

pllim:
Looks like for tabular-fits, you need to force `update_header=True`, not sure why default is False:

https://github.com/astropy/specutils/blob/ba0ab6845095458036340645e5dd690c6d07cda2/specutils/io/default_loaders/tabular_fits.py#L93

Same with wcs1d-fits:

https://github.com/astropy/specutils/blob/ba0ab6845095458036340645e5dd690c6d07cda2/specutils/io/default_loaders/wcs_fits.py#L219

[...]

Issue #905, opened 2021-12-07
Revamp Spectrum1D tabular-fits writer


kelle:
I'm having a lot of trouble figuring out how to write a Spectrum1D object to a file. A few of my questions:

- What file types are supported? FITS, ASDF?
- How do I also make sure the header data is also saved?  

I'm happy to work on this once I understand it. :)

kelle:
As a result of digging into this, I discovered that header data does not currently roundtrip using `Spectrum1D.write`.  I propose that this be implemented using something like this:

```
# Make a dedicated HDU for the header and spectrum
hdu0 = fits.PrimaryHDU(header=header)
hdu1 = fits.BinTableHDU(data=NIR_table)
hdu1.header['EXTNAME'] = 'SPECTRUM'
hdu1.header.set('OBJECT', NIRSpec_dict['object_name'], 'Object Name')

# Combine the two HDUs into a multi-extension FITS (MEF)
spectrum_mef = fits.HDUList([hdu0, hdu1])  # hdu0 is header and hdu1 is data

# Write out the MEF to a file
spectrum_mef.writeto(fits_filename, output_verify="exception")

# Demonstrate how to read the file back in using specutils
spec1d_NIR = Spectrum1D.read(fits_filename, format='tabular-fits')
header_NIR = fits.getheader(fits_filename)
```

If folks agree, I would be happy to implement this given some guidance.

pllim:
@dhomeier might also be interested in this issue since he is working on:

* #1009

kelle:
Summarizing relevant discussions from Slack:
- _If someone modifies the spectrum, the keywords might become inaccurate._ My proposal is to add a warning that says something like, "Header keywords are the responsibility of the user and are not guaranteed to be accurate."  We can open issues and what not on specutils routines which modify header keywords and gradually make this less true. But I think that header round tripping  is important enough to tolerate this shortcoming. 

-  I am not sure how to implement this for Spectrum2D, 3D or SpectrumList. I am happy to just do it for Spectrum1D -- this would be immensely useful -- OR to figure out how to make it extendable to these other classes.

kelle:
We've tested the `spectrum.meta.update` method and it's pretty close! Problems include
- extra END keywords. This makes many header keywords inaccessible by other tools. For example,
```
header=fits.getheader(tmpfile)
```
results in a header which just has the minimal keywords and not any of the things we added.  (I assume this is because of the no-loneger-necessary END keyword)

[...]

Issue #1102, opened 2023-11-15
Inconsistant expectation for location of header data (vs. other metadata)


cshanahan1:
The location of header keywords is inconsistent - wcs1d_fits writer will update the output header with anything in the spectrum.meta dictionary. Tabular fits expects header-specific keys to be in meta.header, as a fits.Header object (and will ignore it if its a dictionary). I'm not sure what the other readers/writers do, but it is inconsistent at least in these. Because there can be other information in `meta` that might not belong in the fits file header, meta.header is preferable (either as dict or `Header`?)

 (Issue made after discussion with @kelle @havok2063, @tepickering)

dhomeier:
See previous discussion in #617 – roughly speaking I think half of the other specialised loaders handle it one way, the rest the other, and at the time the discussion never wrapped up to a conclusion.

kelle:
Also relevant to #1101 , which is about documentation. I think we should at least be honest about the situation in the documentation and give people a suggestion for what to do moving forward with new reader/writers.

Issue #617, opened 2020-03-23
Which (FITS) header should be stored in Spectrum1D.meta (and where)?


dhomeier:
Loaders for FITS files are saving the full information from the header cards in `Spectrum1D.meta['header']`. In https://github.com/astropy/specutils/pull/608#discussion_r395632586 I noted some ambiguity as to *which* header should be stored.
Most of the `default_loaders` for such formats, namely `apogee`, `hst_*`, `muscles`, `subaru_pfs` and `sdss` `spec_loader`, are pulling the info from the FITS header of the first HDU, while the spectral data are read from one (or several) of the extension HDUs.
This primary HDU header will usually contain general information on the target, observing run etc., while the HDU with the spectrum will include more specific info on the dataset. In principle both could be of interest. Should there be a general recommendation for new default (or custom) loaders (and a fix to existing ones) how to handle this?

I the following options:

1. Leave it at the primary HDU header as with the majority of current `default_loaders`.
Pros: no API changes
Cons: discarding potentially useful information; inconsistent with Astropy's `io.fits` behaviour for `Table` (see below)

2. Always read the header for the HDU containing the data.
This is already implicitly done by `generic_spectrum_from_table` and `spectrum_from_column_mapping` as they are loading `table.meta` under the hood.
Pros: Matches the `Table.meta` data; possibility to use `io.fits` mechanism to filter relevant keywords (i.e. excluding `REMOVE_KEYWORDS` and all, that have already been used in defining column formats and units).
Cons: Losing info from the primary HDU header (among the current formats with data in a BINTABLE extension, at least the Apogee, HST, MUSCLES and SDSS also have potentially interesting information in the primary HDU).
Some formats, e.g. `apogee`, are parsing spectrum data from several different extensions.

3. Combine all header info from primary HDU and any extension HDU from which data are read.
This is already implemented in the `jwst_reader` by using the (FITS) `header.extend` method to update the primary header with the data extension header.
Pros: No or minimal loss of information.
Cons: Possibly excessive accumulation of values in `spectrum.meta['header']`; writing the spectrum back will produce a FITS file with different headers (but currently there is no guarantee that any of the loaders will write back identically formatted files anyway).

I tend to the 3rd option, using a similar implementation to `jwst_reader`. It might still be preferable to use the Astropy `Table.meta` dict for updating to take advantage of the `io.fits` mechanism for stripping excess cards.

**Addendum**
[...]

PR #1105, opened 2023-11-15
Preserve fits primary header when writing out spectrum1d without wcs

cshanahan1:
(this is a suuuuper draft pr, just opening to track work done at Spectro 2023 hack day with @kelle)

The current issue (that we have been trying to wrap our heads around in #1096, #905, and has become clearer to us as we've explored this) is that tabular-fits writer, which converts Spectrum1D into a table and uses Table.write to save to file, does not support writing out to the PrimaryHDU. The alternative writer available for `Spectrum1D`, `wcs1d-fits', sort of does what we want (which is, a primary HDU with a header and a BinTableHDU in the 1st extension with spectrum info) but requires a valid wcs to work (since the initial HDUList is created from this WCS)

The goal of this PR is to bypass the tabular-fits writer to enable writing out meta.header info to the primary HDU when there is NO wcs.

### DIFF
--- a/specutils/io/default_loaders/mef_tabular_fits.py
+++ b/specutils/io/default_loaders/mef_tabular_fits.py
@@ -7,85 +7,90 @@ from astropy.io import fits
 import numpy as np
 from astropy.table import Table
 
-def mef_tabular_fits_writer(spectrum, file_name, hdu=0, update_header=False, **kwargs):
+def mef_tabular_fits_writer(spectrum, file_name='my_func_output.fits', hdu=0, update_header=False, **kwargs):
 
-	# no matter what the dimensionality of 'flux', there will always be a primary header and
-	# a single BinTableHDU extension. the `hdu` arg controls where `meta.header` is dumped.
-	hdulist = [fits.PrimaryHDU(), None]
+    # no matter what the dimensionality of 'flux', there will always be a primary header and
+    # a single BinTableHDU extension. the `hdu` arg controls where `meta.header` is dumped.
+    hdulist = [fits.PrimaryHDU(), None]
+    pri_header_initial = hdulist[0].header
 
-	if hdu > 1:
-	    raise ValueError('`hdu`, which controls which extension `meta.header` is written to, must either be 0 or 1')
-	    
-		# we expect anything that should be written out to the file header to be in `meta.header`
-		# This should be a `fits.header.Header` object, so convert to if meta.header is a dictionary
-		header = spectrum.meta.get('header', fits.header.Header())  # if no meta.header, create empty `fits.header.Header`
-		if not isinstance(header, fits.header.Header):
-		    if isinstance(header, dict):
-		        header = fits.header.Header(header)
-		    else:
-		        raise ValueError('`Spectrum1d.meta.header must be `fits.header.Header` or dictionary.')
+    if hdu > 1:
+        raise ValueError('`hdu`, which controls which extension `meta.header` is written to, must either be 0 or 1')
+        
+    # we expect anything that should be written out to the file header to be in `meta.header`
+    # This should be a `fits.header.Header` object, so convert to if meta.header is a dictionary
+    header = spectrum.meta.get('header', fits.header.Header())  # if no meta.header, create empty `fits.header.Header`
 
-		if update_header:
-		    hdr_types = (str, int, float, complex, bool,
-		                 np.floating, np.integer, np.complexfloating, np.bool_)
-		    header.update([keyword for keyword in spectrum.meta.items() if
-		                   isinstance(keyword[1], hdr_types)])
+    if hdu == 0:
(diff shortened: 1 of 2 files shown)


dhomeier:
> The current issue (that we have been trying to wrap our heads around in #1096, #905, and has become clearer to us as we've explored this) is that tabular-fits writer, which converts Spectrum1D into a table and uses Table.write to save to file, does not support writing out to the PrimaryHDU.

You cannot write tables to the PrimaryHDU, the standard forbids anything but IMAGE HDUs there. So yes, to store the data there you need a WCS, even if it's just the simplest case of `CRVAL1`, `CRPIX1`, `CDELT1` for constant sampling.

In principle you could probably get your info written to the Primary header by first creating an empty `fits.PrimaryHDU()` and then directly write to its header, adopting
https://github.com/astropy/specutils/blob/ba0ab6845095458036340645e5dd690c6d07cda2/specutils/io/default_loaders/wcs_fits.py#L265-L270

instead of passing it to `Table(..., meta=header)`. But I am not sure if `Table.write` will allow writing directly to the extension HDU (without overwriting your PrimaryHDU again) – see comments down there.
So reimplementing the writer without using the `astropy.table` interface might be the the way to go, but that will probably also require reimplementing much of `Table`'s IO registry.

cshanahan1:
Ultimately this PR is just a hack around the fact that BinTableHDU cannot be written to the 0th extension, as you mentioned @dhomeier. The only desired behavior (by both what this PR may become and by the existing `wcs1d-fits`) is to be able to control the placement of header data. 

`wcs1d-fits` is similar and hacks around this limitation by creating an HDUList from a wcs object. It then allows you to choose what extension this WCS header (and other metadata on your Spectrum1D) is written to, but the table that encodes flux/wave/uncert is stored in a non-0th (in fact, i think its ALWAYS the 1st since Spectrum1D.write means writing one spectrum, which belongs in one extension...). It goes through the Table machinery to convert Spectrum1d into a BinTableHDU, but doesn't use Table.write like the simply tabular-fits does.

This PR is basically the same thing, but it is for the case where there is no 1D WCS and the HDUList needs to be created from scratch. For this reason, I don't see the case for both of these existing - we need a 'tabular' reader/writer that contains the correct logic for headers (and therefore WCS). Theres also the related issue of where header data should be expected to be stored, which is not consistant amongst the reader/writers (https://github.com/astropy/specutils/issues/1102)

I can get this working to solve the case of 'I have a Spectrum1D without a WCS, and I want to stuff it in a fits file with a primary header', but I think there are some further concerns with this that need to be discussed.

dhomeier:
> `wcs1d-fits` is similar and hacks around this limitation by creating an HDUList from a wcs object. It then allows you to choose what extension this WCS header (and other metadata on your Spectrum1D) is written to, but the table that encodes

I probably would not call the way `wcs1d-fits` is handling this a hack – as it is writing header and data to an ImageHDU, it simply can (and by default does) write everything together to the PrimaryHDU, so there is normally no concern about having to store anything in a different HDU to begin with.

> flux/wave/uncert is stored in a non-0th (in fact, i think its ALWAYS the 1st since Spectrum1D.write means writing one spectrum, which belongs in one extension...). It goes through the Table machinery to convert Spectrum1d into a BinTableHDU, but doesn't use Table.write like the simply tabular-fits does.

I am not sure I can follow you here – are you still referring to the `wcs1d` writer? This does not make use of the `Table` interface at all (nor do any of the readers), and it also does not create any table-type HDUs.

> This PR is basically the same thing, but it is for the case where there is no 1D WCS and the HDUList needs to be created from scratch. For this reason, I don't see the case for both of these existing - we need a 'tabular' reader/writer that contains the correct logic for headers (and therefore WCS). Theres also the related issue of where header data should be expected to be stored, which is not consistant amongst the reader/writers (#1102)

If you are talking about the new loader from this PR and the old `tabular_fits` loader, I agree there should be no need to have both of them, except that it may be easier to provide backwards compatibility for a transition time by keeping the old writer available. Not sure of a tabular reader/writer actually needs to be able to handle WCS logic in the header – the only case I can imagine where this could occur in a tabular Spectrum1D would be if there existed a pure spatial/celestial WCS in addition. But that could probably be simply passed through without any processing.
The issue of a consistent scheme for storing header info in `Spectrum1D.meta` remains, and certainly both writers should ultimately follow what is concluded in https://github.com/astropy/specutils/issues/617#issuecomment-1814778481.
To provide some background: when I initially wrote `wcs1d_fits_writer` the design goal was simply to be able to write a format that the already existing `wcs1d_fits_loader` would expect and be able to read, and ideally be able to fully roundtrip on read/write. That loader was storing the header in `spectrum.meta['header']`, and there was no obvious majority for changing either this or the method used in `tabular_fits`.

kelle:
I think this can probably be closed.

PR #1111, opened 2023-11-22
TST: Fix devdeps and clean up matrix

### LINKED ISSUE #1036: Update docs/doctests for astropy 5.3 compatibility.
#1034 worked around this, but we should actually update the docs/doctests to account for https://github.com/astropy/astropy/pull/14439 at some point.

pllim:
Fix devdeps job and clean up the testing infrastructure in general (some outdated stuff and unnecessary stuff from copy-paste).

xref https://github.com/astropy/astropy/issues/15638

Close #1089 

Fix #1036


pllim:
I see numpy 1.x vs 2.x ABI warning and dateutil deprecation warning in the devdeps log here but somehow it did not fail pytest even though we are turning warnings into exceptions. I think maybe it is because in devdeps, we install quite a few things from source and some of those might be emitting these warnings, not us, so maybe it is okay to ignore as long as CI is green here.

setup.cfg
```diff
@@ -28,11 +27,7 @@ install_requires =
 [options.extras_require]
 test =
     pytest-astropy
-    pytest-cov
-    tox
```
rosteen, line 32: I just added `tox` here so that people can run `pip install -e .[test]` locally and get everything they need to check their code locally before opening a PR, e.g. running `tox -e codestyle` (or any of the other tests via `tox`). I would prefer to keep this in.

rosteen:
Thanks!

rosteen (approved):

setup.cfg
pllim: Okay but you don't really need tox to do that. You can run `flake8 specutils` locally and it is the same check.

PR #1112, opened 2023-11-22
One more fix to dev docs

kelle:
Needed single quotes in order for pip install command to work.


rosteen:
That's so strange, I've never needed quotes for the pip install to work. Maybe an OS thing. Either way, I tested and it doesn't break the install for me.

Good call on the 3.11 specification, 3.12 seems to still need dev versions of a couple dependencies to work.

rosteen (approved):

PR #1113, opened 2023-11-22
Fix for Tabular-fits reader/writer fix

kelle:
This PR modifies the `tabular-fits` writer to put the header information in BOTH HDU0 and HDU1 based on the conversation in #905 .

Other folks are welcome to modify this PR.

Broke out the testing of the `update_header` method into its own test.

`fits compressed` currently failing but I think that might be a file/memory handling thing which I may have accidentally broken in the test file.

Remaining Todo:
- [x] fix broken `fits_compressed` test
- [ ] add more tests
- [ ] add test that makes sure COMMENT and HISTORY keywords are perserved
- [x] make sure reader is doing what we expect
- [x]  fix `update_header` behavior and tests.


kelle:
At some point as part of the `read` method, the `Header` object gets converted into an `OrderedDict`.  Not sure where that's happening...I don't see that in the `tabular_fits_loader` function.

kelle:
I think the trouble is in `Table.read` which I guess changes everything to an ordered dictionary....FML.

kelle:
Added this to the reader and all is now well 
```python 
tab.meta = fits.getheader(file_obj, 0)

File: specutils/io/default_loaders/tabular_fits.py [hunk 1]
```diff
@@ -74,6 +74,9 @@ def tabular_fits_loader(file_obj, column_mapping=None, hdu=1, **kwargs):
 
     tab = Table.read(file_obj, format='fits', hdu=hdu)
 
+    # Get the header from HDU0
+    tab.meta = fits.getheader(file_obj, 0)
```
dhomeier, line 78: I think this is breaking compression support since `Table.read()` in those cases closes the `file_obj` on return, so you can no longer read from it. Probably need to get away without using the `table` io at all here as well, i.e. use `read_fileobj_or_hdulist` as `wcs1d_fits_loader` does and read both `tab` and `header` from `hdulist` (not sure if `Table` directly supports this).

specutils/tests/test_loaders.py
```diff
@@ -599,27 +599,47 @@ def test_tabular_fits_header(tmp_path):
...
+                              'OBSERVER': 'Parsons'})
 
     spectrum = Spectrum1D(flux=flux, spectral_axis=disp, meta={'header': hdr})
     tmpfile = str(tmp_path / '_tst.fits')
     spectrum.write(tmpfile, format='tabular-fits')
 
+    print(f"spetrum meta: \n {repr(spectrum.meta['header'])}")
     # Read it in and check against the original
     with fits.open(tmpfile) as hdulist:
-        assert hdulist[0].header['NAXIS'] == 0
```
dhomeier, line 610: Is this no longer working? Would be worrying, because an "empty" PrimaryHDU should always have `NAXIS = 0`!

kelle: I'll put it back in!

kelle:
This was just discussed in the Astropy Coordination Meeting. It would be really great if someone could take this up. I think it's most of the way there!

simontorres:
So, the failing test has to do with `xz` compression. It is my understanding that `xz` support was not implemented. Is it safe to just remove `xz` from the test parameters?

simontorres:
Also I was wondering whether specifying the format was the best way to fix the compressed fits file test. I thought of improving the format identifier function but I think the current one is good enough and adding more would increase specificity beyond an appropriate level. I'm curious on other's opinion regarding this.

dhomeier:
If you look at the docstring of the test, automatic format detection is one of the features it is testing, so explicitly passing the format is not an option. Also does not fix anything, and the other compression types did work even with auto-detection – the format is a matter of the internal file structure, so should only come into play _after_ the file compression is correctly recognised and handled.
And `xz` does work with `nddata`, if you check other recent test runs.
My guess is that it broke here due to the additional accesses to the file handle, as [commented](https://github.com/astropy/specutils/pull/1113#discussion_r1402778197) before, although I did not expect this to break for `xz` only, and don't quite understand why it is already failing on the first `read_fileobj_or_hdulist`...

simontorres:
Thanks @dhomeier for your reply.
I find testing multiple subjects at once a bit confusing, so I would have a specific test for format identification and a different one for the round-tripping part. But yes, it makes more sense that the test is for identification.

My understanding of compressed files is very limited and I don't think I'll find the time to get deep into it, so I think someone else should give it a try.

rosteen:
I also don't have any real expertise with compression, but I'll see what I can figure out today.

dhomeier:
> I find testing multiple subjects at once a bit confusing, so I would have a specific test for format identification and a different one for the round-tripping part. But yes, it makes more sense that the test is for identification.

`test_tabular_fits_compressed` is all about compression support (and automatic detection of the various formats based on their magic bytes) and not about round-tripping (which is not well supported so far, but hopefully will be better with the new functionality tested here in `test_tabular_fits_update_header`).

My suggestion (cannot put it directly in as it's outside of the covered code region here) would be something like
```python
    with read_fileobj_or_hdulist(file_obj, **kwargs) as hdulist:
        wcs = WCS(hdulist[hdu].header)
        tab = Table.read(hdulist[hdu])
        tab.meta['header'] = hdulist[0].header
```

Note that this is following closely the solution in the existing MUSCLES reader
https://github.com/astropy/specutils/blob/d16d91a077417a7aacdee06d2b259fa887908b14/specutils/io/default_loaders/muscles_sed.py#L47-L50
which has working xz support, so if accessing `hdulist[0].header` above should not work for some reason, you may need to follow that example further by reading the header into a temporary object _before_ calling `Table.read()`.

rosteen:
I put the table read in the block @dhomeier suggested to read the already-opened hdulist, and the `xz` compression test now passes. Interestingly though, it _fails_ if the format (`tabular-fits`) is explicitly specified in the `Spectrum1D.read` call. I suspect that whatever is handling the compression happens before the extracted file is then directed to the appropriate reader, and setting the format explicitly bypasses this and sends the compressed file straight to the reader, but I'm not sure since I don't understand what is handling the decompression.

rosteen:
I was right, in `astropy/io/registry/core.py` there's this block of code that handles decompressing the file, but _only_ if `format=None`. The key here is that `get_readable_fileobj` handles compressed files, but if `format` is specified then the file gets sent straight to the specified reader.

```
             if format is None:
                 path = None
                 fileobj = None

                 if len(args):
                     if isinstance(args[0], PATH_TYPES) and not os.path.isdir(args[0]):
                         from astropy.utils.data import get_readable_fileobj

                         # path might be a os.PathLike object
                         if isinstance(args[0], os.PathLike):
                             args = (os.fspath(args[0]),) + args[1:]
                         path = args[0]
                         try:
                             ctx = get_readable_fileobj(
                                 args[0], encoding="binary", cache=cache
                             )
                             fileobj = ctx.__enter__()
                         except OSError:
                             raise
                         except Exception:
                             fileobj = None
                         else:
                             args = [fileobj] + list(args[1:])
                     elif hasattr(args[0], "read"):
                         path = None
                         fileobj = args[0]

                 format = self._get_valid_format(
                     "read", cls, path, fileobj, args, kwargs
                 )
```

dhomeier:
Yes, that seems to confirm my speculations, in the Slack discussion I think. Still works for gzip and bzip2 since those are also directly supported by `io.fits`.

rosteen:
@dhomeier any further objections to this? I think all your previous comments are addressed and this is (at least close to being) ready to merge.

specutils/tests/test_loaders.py
```diff
@@ -662,27 +662,44 @@ def test_tabular_fits_maskheader(tmp_path):
...
+
+        # Test HDU0 header
         assert hdulist[0].header['NAXIS'] == 0
+        assert hdulist[0].header['OBSERVER'] == 'Parsons'
+
+        # keys relevant to datashape are in HDU1 header
         assert hdulist[1].header['NAXIS'] == 2
         assert hdulist[1].header['NAXIS2'] == disp.shape[0]
+
+        # Test legacy behaviour of metadata is HDU1 header
```
dhomeier, line 682: Not sure I understand "metadata is HDU1 header" here – "metadata written to HDU1 header" perhaps?

specutils/io/default_loaders/tabular_fits.py
```diff
@@ -181,8 +183,12 @@ def tabular_fits_writer(spectrum, file_name, hdu=1, update_header=False, **kwarg
...
-    tab = Table(columns, names=colnames, meta=header)
+    tab = Table(columns, names=colnames)
+    hdu1 = fits.BinTableHDU(data=tab, header=header)
+    hdu1.header['EXTNAME'] = 'DATA'
+    hdu0 = fits.PrimaryHDU(header=header)
 
-    # Todo: support writing to other HDUs than the default (1st)
-    #       and an 'update' mode so different HDUs can be written to separately
-    tab.write(file_name, format="fits", **kwargs)
+    hdulist = fits.HDUList([hdu0, hdu1])
```
dhomeier, line 191: Is writing to 2nd or subsequent extensions then no longer of interest?

dhomeier (commented):
Everything appears to be working again, that's good. I could not find the updates to the _writer_ though @simontorres was about to make.
It is also not clear if this is now allows true round-tripping of tabular spectra to FITS, which was (at least one of) the original motivations for this change – I don't think header and data are tested in full in the tests yet.
Finally making these changes to the existing reader is a breaking change, as there would be no default loader anymore that would be able to return header info from the data HDUs; which was seen as a serious issue in the [discussion before](https://github.com/astropy/specutils/issues/905#issuecomment-1823414872).
It might be better after all to relegate this to a new loader or at least add a keyword to allow compliance with the previous behaviour.

specutils/io/default_loaders/tabular_fits.py
kelle: Writing more HDUs can be addressed in a future pull request. I'm happy to open an issue once this gets merged.

kelle:
What would the keyword be in the reader?  `header_ext = 0` would be the default but `header_ext = 1` would preserve the legacy behavior?

kelle:
Are we concerned about the Mac 3.12 test failing?

rosteen:
> Are we concerned about the Mac 3.12 test failing?

It's unrelated, being discussed in https://github.com/astropy/specutils/issues/1146.

CHANGES.rst
```diff
@@ -10,6 +10,8 @@ Bug Fixes
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+- Changed the ``tabular-fits`` reader to round-trip the primary header. [#1113]
```
rosteen, line 13:
```suggestion
- Changed the ``tabular-fits`` reader/writer to round-trip the primary header. [#1113]
```

rosteen:
> Everything appears to be working again, that's good. I could not find the updates to the _writer_ though @simontorres was about to make. It is also not clear if this is now allows true round-tripping of tabular spectra to FITS, which was (at least one of) the original motivations for this change – I don't think header and data are tested in full in the tests yet. Finally making these changes to the existing reader is a breaking change, as there would be no default loader anymore that would be able to return header info from the data HDUs; which was seen as a serious issue in the [discussion before](https://github.com/astropy/specutils/issues/905#issuecomment-1823414872). It might be better after all to relegate this to a new loader or at least add a keyword to allow compliance with the previous behaviour.

@kelle @dhomeier I added a keyword `store_data_header` to store as table.meta the header from whichever data HDU is specified instead of the primary header. Does this sufficiently address the concern raised in the quoted comment?

specutils/io/default_loaders/tabular_fits.py
```diff
@@ -71,8 +74,13 @@ def tabular_fits_loader(file_obj, column_mapping=None, hdu=1, **kwargs):
     # routines to search for spectral axis information in the file.
     with read_fileobj_or_hdulist(file_obj, **kwargs) as hdulist:
         wcs = WCS(hdulist[hdu].header)
+        primary_header = hdulist[0].header
+        tab = Table.read(hdulist[hdu])
 
-    tab = Table.read(file_obj, format='fits', hdu=hdu)
+    if store_data_header:
+        tab.meta = hdulist[hdu].header
```
dhomeier, line 81: I am quite puzzled how this is working at all, accessing `hdulist` outside of its context (but apparently it does at least in my local installation). Wondering then why reading `hdulist[0].header` would not work without the temporary variable...
In principle there should not even be another read required, since this is what `Table` sets by default, but its reader is converting `meta` to `dict`, which is arguably less convenient than the `fits.Header`.
Highlights that the new option is not tested yet, as in fact there are no tests on the `meta['header']` returned on the _reader_ side.

CHANGES.rst
```diff
@@ -10,6 +10,8 @@ Bug Fixes
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+- Changed the ``tabular-fits`` reader/writer to round-trip the primary header. [#1113]
```
dhomeier, line 13:
```suggestion
- Changed the ``tabular-fits`` reader/writer to round-trip the primary header.
   The reader now by default reads the primary header into ``meta``; the old behaviour
   can be restored by setting the option ``store_data_header=True``. The writer is writing
   copies of ``meta['header']`` to both primary and data extension headers. [#1113]
```

dhomeier (changes requested):
Adding the option as a switch seems indeed better than specifying the extension directly, as there should be no reason to use the header from anything other than primary or data HDU.
But as I don't understand why reading from the closed `hdulist` is working here, or if it will in all contexts, this should certainly get test coverage.

dhomeier:
> @kelle @dhomeier I added a keyword `store_data_header` to store as table.meta the header from whichever data HDU is specified instead of the primary header. Does this sufficiently address the concern raised in the quoted comment?

Can't speak for the concerns from @sbailey or others at the November meeting, but since there is a documented way to reprocude the current functionality, this looks OK to me.
I am wondering now if the duplicate header _writing_ is still required with this – by default header metadata info should now stick with the primary HDU? Unless there is concern about older specutils version round-tripping with files written by the new version; but even that would not truly round-trip, since the old reader is storing a dictionary, not the header object.

specutils/io/default_loaders/tabular_fits.py
rosteen: You know, I'm confused how it's working too 😅 . I'm going to move it into the context to at least make me feel better.

specutils/io/default_loaders/tabular_fits.py
```diff
@@ -181,8 +187,12 @@ def tabular_fits_writer(spectrum, file_name, hdu=1, update_header=False, **kwarg
         if columns[c].ndim > 1:
             columns[c] = columns[c].T
 
-    tab = Table(columns, names=colnames, meta=header)
+    tab = Table(columns, names=colnames)
+    hdu1 = fits.BinTableHDU(data=tab, header=header)
```
dhomeier, line 191:
```suggestion
    hdu1 = fits.BinTableHDU(data=tab)
```
Now that the reader is by default always just reading from the primary header, perhaps just give this a try and see if it still passes the round-trip tests? Although to be fully compatible, the writer should probably then get a `store_data_header` option as well.

dhomeier (approved):

kelle:
So happy to see this approved -- Woot!!  What else needs to be done before it can be merged? (I've got several projects waiting to use this new and improved functionality.)

specutils/tests/test_loaders.py
```diff
@@ -662,27 +662,44 @@ def test_tabular_fits_maskheader(tmp_path):
...
+    disp = np.linspace(1, 1.2, 21) * u.AA
+    flux = np.random.normal(0., 1.0e-14, disp.shape[0]) * u.erg / (u.s * u.cm**2 * u.AA)
+    hdr = fits.header.Header({'TELESCOP': 'Crystal', 'OBSERVER': 'Cruz'})
+    spec = Spectrum1D(flux=flux, spectral_axis=disp, meta={'header': hdr})
+    tmpfile = str(tmp_path / '_tst2.fits')
+    spec.write(tmpfile, format='tabular-fits')
+    spectrum = Spectrum1D.read(tmpfile, format='tabular-fits')
+
     # Now write with updated header information from spectrum.meta
     spectrum.meta.update({'OBSERVER': 'Rosse', 'EXPTIME': 32.1, 'NAXIS2': 12})
```
dhomeier, line 696: I've been looking into what needed to be updated in the tests to account for the `store_data_header` option, and began to wonder if this is really testing what we expect it to do. Basically we now have constellations like
```python
spectrum.meta = {'OBSERVER': 'Rosse', 'header': {'TELESCOP': 'Crystal', 'OBSERVER': 'Cruz'}}
```
which is bound to become very confusing, at a minimum. So if it is now decided that we don't want to encourage the "flat" `meta` layout, perhaps we should consider deprecating the `update_header` option altogether.

dhomeier:
I've checked what it takes for supporting a corresponding `store_data_header` option (defaulting to `False` as well) in the writer, and it's basically just two of the existing tests that would have to be updated to the new default.
The changes are just a bit too complex to put in as suggestions here, so I'd need to make another PR against this one.

kbwestfall:
Hi all.  Just a heads up that, related to #1149 and https://github.com/astropy/astropy/issues/16648, I was just looking at what a PR into specutils would look like, building from https://github.com/astropy/astropy/pull/16690.  I would want to start with the changes made here, and so I'm planning to wait until this clears.

rosteen:
> Hi all. Just a heads up that, related to #1149 and [astropy/astropy#16648](https://github.com/astropy/astropy/issues/16648), I was just looking at what a PR into specutils would look like, building from [astropy/astropy#16690](https://github.com/astropy/astropy/pull/16690). I would want to start with the changes made here, and so I'm planning to wait until this clears.

@dhomeier Any ETA on your PR? Given @kbwestfall's comment, would it be reasonable to merge this now and have your follow up be a PR to `main` rather than @kelle's branch?

kelle:
I would love for this to be merged. I've got folks eager to use it.

dhomeier:
Will push my changes later today; this may not test full round-tripping yet, and not sure if it addresses all the previous comments on header location, but hopefully would be mergeable, and completion of the tests could certainly be done in a follow-up.

That should certainly not interfere with a new `covariance` property added through astropy/astropy#16690 as it just concerns handling of header and `meta` data.
Also that PR presently is using a combination of `ImageHDU` and `BinTableHDU` for storing the covariance in `output_hdus`
https://github.com/astropy/astropy/blob/dc3c6eab63136fd04b2f958791b20f1d21b2c38e/astropy/nddata/covariance.py#L860-L863
so that may not be best connected to the tabular loader here after all, but perhaps instead be implemented in `wcs1d-fits`, or a new loader altogether.

rosteen:
> Will push my changes later today; this may not test full round-tripping yet, and not sure if it addresses all the previous comments on header location, but hopefully would be mergeable, and completion of the tests could certainly be done in a follow-up.

Ok, I'll wait for your changes then!

kbwestfall:
Just to clarify, I had planned change `tabular_fits_writer` to:
 - Allow `spectrum.uncertainty` to be a `Covariance` instance and check for it.
 - Use a new method in `Covariance` that outputs a 1D array with the variances (diagonal of the covariance) and a `Table` with the off-diagonal covariance data.
 - Replace the `uncertainty` column in the spectrum output table with the variances, and add the covariance table as another HDU in the output file.

So the edits here shouldn't be constrained by the current functionality in the draft https://github.com/astropy/astropy/pull/16690 PR.

dhomeier:
> Just to clarify, I had planned change `tabular_fits_writer` to:
> 
> * Allow `spectrum.uncertainty` to be a `Covariance` instance and check for it.

That would of course require changes to `Spectrum1D` as well, which currently checks for, and expects one of the 3 `NDUncertainty` subclasses. But it might be possible to adapt its operations on `uncertainty`, as your `Covariance.variance` provides a compatible quantity (should perhaps be explicitly cast as `VarianceUncertainty`).
I thought it might be safer to add this as a new property along with `uncertainty` in its accustomed definition, but that can be discussed in your forthcoming PR.

The only thing to take care of should be to make sure, if you are writing your `_hdr` to the primary HDU, it is not interfering with or overwriting the `meta` data written per this PR.

dhomeier:
> Ok, I'll wait for your changes then!

Added in kelle/specutils#1

rosteen:
Test failures are known/unrelated/exist on main, I'll merge this.

kelle:
Words cannot describe my joy! Thanks to all who helped out on this!  Super excited to start using it!

dhomeier:
Alas, there was still lurking in my head the issue of updating the docs (i.e. changelog at least) ot make the changes in (default) API clearer; actually the bit about writing `meta` header data to both HDUs has now become factually wrong.
Shall I fix this with a follow-up PR?

kelle:
YES - I was thinking the same thing. We need to update the docs...If you open the PR, I can help.

rosteen:
If you would open a follow-up PR that would be great. I won't do a release with this until that's in.

PR #1114, opened 2023-11-26
feat: add correlation mode to template_correlation

zhukgleb:
Added the ability to select a correlation method. The direct method is often slower; on weaker computers, calculating the spectrum offset can take hours. The "direct" flag in the scipy correlation source code refers to direct correlation using numpy. Forward compatibility is enabled by default.

### DIFF
--- a/specutils/analysis/correlation.py
+++ b/specutils/analysis/correlation.py
@@ -52,7 +52,7 @@ def template_correlate(observed_spectrum, template_spectrum, lag_units=_KMS,
         a sensible resampling).
     method: str
         If you choose "FFT", the correlation will be done through the use
-        of convolution and will be calculated faster (for small spectral 
+        of convolution and will be calculated faster (for small spectral
         resolutions it is often correct), otherwise the correlation is determined
         directly from sums (the "direct" method in `~scipy.signal.correlate`).
 
@@ -91,11 +91,11 @@ def template_correlate(observed_spectrum, template_spectrum, lag_units=_KMS,
 
     # Correlate
     if method.lower() == "fft":
-        corr = correlate(observed_log_spectrum.flux.value, 
+        corr = correlate(observed_log_spectrum.flux.value,
                          (template_log_spectrum.flux.value * normalization),
                          method="fft")
     else:
-        corr = correlate(observed_log_spectrum.flux.value, 
+        corr = correlate(observed_log_spectrum.flux.value,
                          (template_log_spectrum.flux.value * normalization),
                          method="direct")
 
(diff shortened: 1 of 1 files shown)


specutils/analysis/correlation.py
```diff
@@ -84,9 +90,14 @@ def template_correlate(observed_spectrum, template_spectrum, lag_units=_KMS,
         normalization = 1.
 
     # Correlate
-    corr = np.correlate(observed_log_spectrum.flux.value,
-                        (template_log_spectrum.flux.value * normalization),
-                        mode='full')
+    if method == "fft":
```
rosteen, line 93:
```suggestion
    if method.lower() == "fft":
```

specutils/analysis/correlation.py
```diff
@@ -49,6 +50,11 @@ def template_correlate(observed_spectrum, template_spectrum, lag_units=_KMS,
         ``template_logwl_resample(spectrum, template, delta_log_wavelength=.1)``.
         If False, *no* resampling is performed (and the user is responsible for
         a sensible resampling).
+    mode: str
+        If you choose FFT method, the correlation will be done through the use
+        of convolution and will be calculated faster (for small spectral 
+        resolutions it is often correct). In other cases, the direct method
+        from the numpy will be used.
```
rosteen, line 57:
```suggestion
    method: str
        If you choose "FFT", the correlation will be done through the use
        of convolution and will be calculated faster (for small spectral 
        resolutions it is often correct), otherwise the correlation is determined
        directly from sums (the "direct" method in `~scipy.signal.correlate`).
```

rosteen (commented):
Thanks for the contribution, it will be nice to provide this option! I made a couple small suggestions, if you don't have any objections I'll commit those and approve.

specutils/analysis/correlation.py
zhukgleb: Thanks for corrections! Now it's look much better :)

specutils/analysis/correlation.py
```diff
@@ -49,6 +50,11 @@ def template_correlate(observed_spectrum, template_spectrum, lag_units=_KMS,
         ``template_logwl_resample(spectrum, template, delta_log_wavelength=.1)``.
         If False, *no* resampling is performed (and the user is responsible for
         a sensible resampling).
+    method: str
+        If you choose "FFT", the correlation will be done through the use
+        of convolution and will be calculated faster (for small spectral
```
rosteen, line 55:
```suggestion
        of convolution and will be calculated faster (for small spectral
```

PR #1116, opened 2023-12-01
Add a DESI file reader

### LINKED ISSUE #1103: Create a reader for DESI data files
This issue is simply a reminder to add support for reading DESI spectrum files.

There are two types:

1. `spectra` files may contain multiple observations of the same astronomical object.
2. `coadd` files co-add all observations of the same astronomical object into one spectrum.

Both files have very similar structure, the only real difference would be metadata.

Feel free to assign this to me.

weaverba137:
This PR closes #1103.

I'd like to add more tests before merging, but people can start taking a look at this.


weaverba137:
@eteq, @rosteen, @nmearl, @keflavich, I'm looking for advice on adding additional tests to this PR.  Specifically:

1. Are there examples of how to test file identifiers, in addition to file readers?
2. Is the existing test configuration for the DESI *reader* sufficient, even though it will only ever be activated with *e.g.* `--remote-data`, or would you prefer smaller versions of the DESI data files to be added to the package itself?

Also, not a question but a comment: The code coverage configuration in setup.cfg,  `specutils/*/tests/*`, does *not* exclude `specutils/io/asdf/tests` or `specutils/io/default_loaders/tests`, so I added `specutils/*/*/tests/*`.

weaverba137:
@eteq, @rosteen, @nmearl, @keflavich, post-holiday ping.

rosteen:
> @eteq, @rosteen, @nmearl, @keflavich, post-holiday ping.

@weaverba137 thanks for the ping, I'll try to respond to your questions later today.

rosteen:
>     1. Are there examples of how to test file identifiers, in addition to file readers?

I think the best way to do this is by simply calling `Spectrum1D.read()` or `SpectrumList.read()` (whichever is appropriate) rather than calling `coadd_loader` and `spectra_loader` directly in your tests. This way the tests will first go through the identifier and then get routed to the relevant loader.

>     2. Is the existing test configuration for the DESI _reader_ sufficient, even though it will only ever be activated with _e.g._ `--remote-data`, or would you prefer smaller versions of the DESI data files to be added to the package itself?

Most of the existing tests, if I'm remembering correctly, create a fake FITS file on the fly programmatically that mimic the structure of the relevant files - see for example the [JWST reader tests](https://github.com/astropy/specutils/blob/main/specutils/io/default_loaders/tests/test_jwst_reader.py). It seems like the DESI files are pretty large, so I would say that either that strategy or creating a minimal DESI file (by, e.g., opening a real file with `astropy.io.fits` and deleting/shrinking the data arrays) to include in the repo (or host somewhere and have it still be remote data) would be preferable.

weaverba137:
@rosteen, thank you, those are good suggestions.

weaverba137:
I think this is now ready for final review.

specutils/io/default_loaders/tests/test_desi.py
```diff
@@ -0,0 +1,141 @@
...
+                                                  pytest.param('spectra-5-169-thru20210419.fits', 'DESI spectra', marks=pytest.mark.remote_data)], indirect=['remote_filename'])
+def test_SpectrumList_reader_remote(tmp_path, remote_filename, fmt):
+    """Remote data test of class method.
+    """
+    with set_temp_cache(path=str(tmp_path)):
+        filename = download_file(remote_filename, cache=True)
+        spectrum = SpectrumList.read(filename, format=fmt)  # noqa
+        assert spectrum[0].meta['band'] == 'b'
+        assert spectrum[1].meta['band'] == 'r'
+        assert spectrum[2].meta['band'] == 'z'
```
rosteen, line 117:
```suggestion
```

I thought the local file tests were supposed to replace these remote tests - they take a very long time, I don't think we want to keep them.

rosteen (commented):
I think this is almost good to go but needs a couple more small changes. As commented above it would be nice to remove the remote data tests in favor of the local file tests rather than having both. I also recommend renaming `generate_desi.py` to `generate_desi_test_data.py` and the `t` directory to `desi_test_data`, to be a little more descriptive.

weaverba137:
@rosteen, thank you, those are all reasonable suggestions, I'll get those in soon.

specutils/io/default_loaders/tests/test_desi.py
weaverba137: Remote tests have been removed. I added some documentation to the test file on the origin of the local files.

weaverba137:
All requested changes are in. Tests are failing due to a temporary failure in name resolution:
```
astropy.utils.iers.iers.IERSWarning: failed to download https://datacenter.iers.org/data/9/finals2000A.all: <urlopen error [Errno -3] Temporary failure in name resolution>
```
which is unrelated to this PR.

rosteen (approved):
I confirmed the test pass for me locally where I have those files cached. Looks good to me now, thanks for the contribution!

PR #1119, opened 2024-01-22
Fix `with_spectral_unit`

### LINKED ISSUE #995: Spectrum1D.with_spectral_unit broken
I believe this is due to a bug here: https://github.com/astropy/specutils/blob/fc04cd6c948698c70ec9bd15b3354b58e4dbbcc2/specutils/spectra/spectrum_mixin.py#L274

For example:

```python
import numpy as np

spec = Spectrum1D(spectral_axis=np.linspace(7000,7200,100)<<u.AA, flux=np.ones(100)*u.mJy)
spec.with_spectral_unit(u.km/u.s, velocity_convention='optical', rest_value=7100*u.AA)
TypeError                                 Traceback (most recent call last)
Input In [30], in <cell line: 1>()
----> 1 spec.with_spectral_unit(kms, velocity_convention='optical', rest_value=7200*u.AA)

File ~/mambaforge/envs/py310/lib/python3.10/site-packages/specutils/spectra/spectrum_mixin.py:211, in OneDSpectrumMixin.with_spectral_unit(self, unit, velocity_convention, rest_value)
    187 def with_spectral_unit(self, unit, velocity_convention=None,
    188                        rest_value=None):
    189     """
    190     Returns a new spectrum with a different spectral axis unit.
    191 
   (...)
    209 
    210     """
--> 211     new_wcs, new_meta = self._new_spectral_wcs(
    212         unit=unit,
    213         velocity_convention=velocity_convention or self.velocity_convention,
    214         rest_value=rest_value or self.rest_value)
    216     spectrum = self.__class__(flux=self.flux, wcs=new_wcs, meta=new_meta)
    218     return spectrum

File ~/mambaforge/envs/py310/lib/python3.10/site-packages/specutils/spectra/spectrum_mixin.py:274, in OneDSpectrumMixin._new_spectral_wcs(self, unit, velocity_convention, rest_value)
    272 if velocity_convention is not None:
    273     equiv = getattr(u, 'doppler_{0}'.format(velocity_convention))
--> 274     rest_value.to(unit, equivalencies=equiv)
    276 # Store the original unit information for posterity
    277 meta = self._meta.copy()

File ~/mambaforge/envs/py310/lib/python3.10/site-packages/astropy/units/quantity.py:846, in Quantity.to(self, unit, equivalencies, copy)
    842 unit = Unit(unit)
    843 if copy:
    844     # Avoid using to_value to ensure that we make a copy. We also
    845     # don't want to slow down this method (esp. the scalar case).
--> 846     value = self._to_value(unit, equivalencies)
    847 else:
    848     # to_value only copies if necessary
    849     value = self.to_value(unit, equivalencies)

File ~/mambaforge/envs/py310/lib/python3.10/site-packages/astropy/units/quantity.py:800, in Quantity._to_value(self, unit, equivalencies)
    797     equivalencies = self._equivalencies
    798 if not self.dtype.names or isinstance(self.unit, StructuredUnit):
    799     # Standard path, let unit to do work.
--> 800     return self.unit.to(unit, self.view(np.ndarray),
    801                         equivalencies=equivalencies)
    803 else:
    804     # The .to() method of a simple unit cannot convert a structured
    805     # dtype, so we work around it, by recursing.
    806     # TODO: deprecate this?
    807     # Convert simple to Structured on initialization?
    808     result = np.empty_like(self.view(np.ndarray))

File ~/mambaforge/envs/py310/lib/python3.10/site-packages/astropy/units/core.py:1130, in UnitBase.to(self, other, value, equivalencies)
   1128     return UNITY
   1129 else:
-> 1130     return self._get_converter(Unit(other),
   1131                                equivalencies=equivalencies)(value)

File ~/mambaforge/envs/py310/lib/python3.10/site-packages/astropy/units/core.py:1047, in UnitBase._get_converter(self, other, equivalencies)
   1044 # if that doesn't work, maybe we can do it with equivalencies?
   1045 try:
   1046     return self._apply_equivalencies(
-> 1047         self, other, self._normalize_equivalencies(equivalencies))
   1048 except UnitsError as exc:
   1049     # Last hope: maybe other knows how to do it?
   1050     # We assume the equivalencies have the unit itself as first item.
   1051     # TODO: maybe better for other to have a `_back_converter` method?
   1052     if hasattr(other, 'equivalencies'):

File ~/mambaforge/envs/py310/lib/python3.10/site-packages/astropy/units/core.py:769, in UnitBase._normalize_equivalencies(equivalencies)
    748 @staticmethod
    749 def _normalize_equivalencies(equivalencies):
    750     """
    751     Normalizes equivalencies, ensuring each is a 4-tuple of the form::
    752 
   (...)
    767     ValueError if an equivalency cannot be interpreted
    768     """
--> 769     normalized = _normalize_equivalencies(equivalencies)
    770     if equivalencies is not None:
    771         normalized += get_current_unit_registry().equivalencies

File ~/mambaforge/envs/py310/lib/python3.10/site-packages/astropy/units/core.py:83, in _normalize_equivalencies(equivalencies)
     79     return []
     81 normalized = []
---> 83 for i, equiv in enumerate(equivalencies):
     84     if len(equiv) == 2:
     85         funit, tunit = equiv

TypeError: 'function' object is not iterable
```

rosteen:
This has been needed for a long time - intended to fix #995 , #889 , #676. My current changes fix the behavior for a `Spectrum1D` initialized with a `spectral_axis`, but I think it needs a little more work to also fix it for one initialized with a `wcs`. Note that this also moves the `SpectralGWCS` out of `gwcs_from_array` so that it's actually accessible to check instances, since there are times we may want to do something different if that's the WCS type.

### DIFF
--- a/specutils/analysis/moment.py
+++ b/specutils/analysis/moment.py
@@ -21,7 +21,7 @@ def moment(spectrum, regions=None, order=0, axis=-1):
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
         The spectrum object over which the width will be calculated.
 
-    regions: `~specutils.utils.SpectralRegion` or list of `~specutils.utils.SpectralRegion`
+    regions: `~specutils.SpectralRegion` or list of `~specutils.SpectralRegion`
         Region within the spectrum to calculate the gaussian sigma width. If
         regions is `None`, computation is performed over entire spectrum.
 
--- a/specutils/analysis/uncertainty.py
+++ b/specutils/analysis/uncertainty.py
@@ -22,7 +22,7 @@ def snr(spectrum, region=None):
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
         The spectrum object overwhich the equivalent width will be calculated.
 
-    region: `~specutils.utils.SpectralRegion` or list of `~specutils.utils.SpectralRegion`
+    region: `~specutils.SpectralRegion` or list of `~specutils.SpectralRegion`
         Region within the spectrum to calculate the SNR.
 
     Returns
@@ -67,7 +67,7 @@ def _snr_single_region(spectrum, region=None):
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
         The spectrum object overwhich the equivalent width will be calculated.
 
-    region: `~specutils.utils.SpectralRegion`
+    region: `~specutils.SpectralRegion`
         Region within the spectrum to calculate the SNR.
 
     Returns
@@ -109,7 +109,7 @@ def snr_derived(spectrum, region=None):
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
         The spectrum object overwhich the equivalent width will be calculated.
 
-    region: `~specutils.utils.SpectralRegion`
+    region: `~specutils.SpectralRegion`
         Region within the spectrum to calculate the SNR.
 
     Returns
@@ -155,7 +155,7 @@ def _snr_derived(spectrum, region=None):
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
         The spectrum object overwhich the equivalent width will be calculated.
 
-    region: `~specutils.utils.SpectralRegion`
+    region: `~specutils.SpectralRegion`
         Region within the spectrum to calculate the SNR.
 
     Returns
--- a/specutils/spectra/spectrum_mixin.py
+++ b/specutils/spectra/spectrum_mixin.py
@@ -229,7 +229,7 @@ class OneDSpectrumMixin():
             meta['original_unit'] = orig_unit
 
         if 'original_wcs' not in self.meta:
-            meta['original_wcs'] = self.wcs.copy()
+            meta['original_wcs'] = self.wcs.deepcopy()
 
         new_spectral_axis = self.spectral_axis.to(unit, doppler_convention=velocity_convention,
                                                   doppler_rest=rest_value)
--- a/specutils/spectra/spectrum_mixin.py
+++ b/specutils/spectra/spectrum_mixin.py
@@ -193,8 +193,9 @@ class OneDSpectrumMixin():
         """
         Returns a new spectrum with a different spectral axis unit. Note that this creates a new
         object using the converted spectral axis and thus drops the original WCS, if it existed,
-        replacing it with a :class:`~specutils.utils.wcs_utils.SpectralGWCS`. The original WCS
-        will be stored in the ``original_wcs`` entry of the new object's ``meta`` dictionary.
+        replacing it with a lookup-table :class:`~gwcs.wcs.WCS` based on the new spectral axis. The
+        original WCS will be stored in the ``original_wcs`` entry of the new object's ``meta``
+        dictionary.
 
         Parameters
         ----------
--- a/specutils/spectra/spectrum_mixin.py
+++ b/specutils/spectra/spectrum_mixin.py
@@ -192,9 +192,9 @@ class OneDSpectrumMixin():
                            rest_value=None):
         """
         Returns a new spectrum with a different spectral axis unit. Note that this creates a new
-        object using the converted spectral axis and thus drops the original WCS, if it existed.
-        The original WCS will be stored in the ``original_wcs`` entry of the new object's ``meta``
-        dictionary.
+        object using the converted spectral axis and thus drops the original WCS, if it existed,
+        replacing it with a :class:`~specutils.utils.wcs_utils.SpectralGWCS`. The original WCS
+        will be stored in the ``original_wcs`` entry of the new object's ``meta`` dictionary.
 
         Parameters
         ----------
--- a/specutils/spectra/spectrum_mixin.py
+++ b/specutils/spectra/spectrum_mixin.py
@@ -5,7 +5,7 @@ import astropy.units.equivalencies as eq
 from astropy import units as u
 from astropy.utils.decorators import deprecated
 
-from specutils.utils.wcs_utils import gwcs_from_array
+from specutils.utils.wcs_utils import gwcs_from_array, SpectralGWCS
 
 DOPPLER_CONVENTIONS = {}
 DOPPLER_CONVENTIONS['radio'] = u.doppler_radio
@@ -208,12 +208,31 @@ class OneDSpectrumMixin():
                      even if your spectrum has air wavelength units
 
         """
-        new_wcs, new_meta = self._new_spectral_wcs(
-            unit=unit,
-            velocity_convention=velocity_convention or self.velocity_convention,
-            rest_value=rest_value or self.rest_value)
+        velocity_convention = velocity_convention if velocity_convention is not None else self.velocity_convention # noqa
+        rest_value = rest_value if rest_value is not None else self.rest_value
 
-        spectrum = self.__class__(flux=self.flux, wcs=new_wcs, meta=new_meta)
+        # Store the original unit information for posterity
+        meta = self._meta.copy()
+        orig_unit = self.wcs.unit[0] if hasattr(self.wcs, 'unit') else self.spectral_axis.unit
+        if 'original_unit' not in self._meta:
+            meta['original_unit'] = orig_unit
+
+        # In this case the original Spectrum1D was instantiated with a spectral_axis and no wcs
+        if isinstance(self.wcs, SpectralGWCS):
+            new_spectral_axis = self.spectral_axis.to(unit,
+                                                      doppler_convention=velocity_convention,
+                                                      doppler_rest=rest_value)
+
+            return self.__class__(flux=self.flux, spectral_axis=new_spectral_axis, meta=meta,
+                                  uncertainty=self.uncertainty, mask=self.mask)
+
+        # Otherwise we handle the original wcs properly
+        new_wcs = self._new_spectral_wcs(unit=unit, orig_unit=orig_unit,
+                                         velocity_convention=velocity_convention,
+                                         rest_value=rest_value)
+
+        spectrum = self.__class__(flux=self.flux, wcs=new_wcs, meta=meta,
+                                  uncertainty=self.uncertainty, mask=self.mask)
 
         return spectrum
 
@@ -242,7 +261,7 @@ class OneDSpectrumMixin():
 
         return unit
 
-    def _new_spectral_wcs(self, unit, velocity_convention=None,
+    def _new_spectral_wcs(self, unit, orig_unit, velocity_convention=None,
                           rest_value=None):
         """
         Returns a new WCS with a different Spectral Axis unit.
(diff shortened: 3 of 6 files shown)


rosteen:
Ok, I realized the existing WCS-initialized spectrum case was just acting on the `spectral_axis` as well, so I didn't need the a special case for the `spectral_axis`-initialized case and handle both the same way. It would probably be more ideal to preserve as much of the original WCS as possible while changing the unit, but that can be work for another day. This at least restores the previous (intended) functionality.

specutils/utils/wcs_utils.py
```diff
@@ -8,6 +8,18 @@
...
+        self.original_unit = kwargs.pop("original_unit", "")
+        super().__init__(*args, **kwargs)
+
+    def pixel_to_world(self, *args, **kwargs):
+        if self.original_unit == '':
+            return u.Quantity(super().pixel_to_world_values(*args, **kwargs))
+        return super().pixel_to_world(*args, **kwargs).to(
+            self.original_unit, equivalencies=u.spectral())
+
+
```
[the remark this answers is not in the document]
rosteen: Weird, I thought I ran the tests locally after adding that, guess not. Thanks for the link.

rosteen:
Looks like the remaining failures are a (hopefully transient) download failure:

```
astropy.utils.iers.iers.IERSWarning: failed to download https://datacenter.iers.org/data/9/finals2000A.all: <urlopen error [Errno -3] Temporary failure in name resolution>
```

dhomeier:
Yes, seen those frequently, although they did not show up in the previous run; unlikely this was introduced by adding the copy methods. Can maybe check if astropy has done something to avoid these connection failures.

specutils/utils/wcs_utils.py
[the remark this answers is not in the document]
rosteen: Actually the tests do pass locally, I must have the IERS files cached.

dhomeier:
Seems the last solution was just to wait till DNS is working again... astropy/astropy#15456

specutils/spectra/spectrum_mixin.py
```diff
@@ -208,14 +215,27 @@ def with_spectral_unit(self, unit, velocity_convention=None,
...
+
+        # Store the original unit information and WCS for posterity
+        meta = self._meta.copy()
+
+        if 'original_unit' not in self._meta:
+            orig_unit = self.wcs.unit[0] if hasattr(self.wcs, 'unit') else self.spectral_axis.unit
+            meta['original_unit'] = orig_unit
+
+        if 'original_wcs' not in self.meta:
+            meta['original_wcs'] = self.wcs.copy()
```
dhomeier, line 232:
```suggestion
            meta['original_wcs'] = self.wcs.deepcopy()
```
Since this brought up the issue of the convenience copy method, you might actually want this, if you really treat the  created spectrum as a new object; otherwise any changes to `new_spec.wcs` will modify the original as well.
Should perhaps be added to the test below for an instance with wcs.

rosteen: Good call, I can never keep straight when copy vs deepcopy is the right answer.

rosteen:
@dhomeier I switched to `deepcopy` and added a test to check that the stored copy of the original WCS has the original unit. If this looks good to you now would you approve so I can merge? Thanks!

dhomeier:
The test I did locally to verify it really needs `deepcopy` was (kinda brute force) setting `new_spec.wcs.cdelt` (with an Astropy WCS) to different values and check if the original was changed as well. But `SpectralGWCS` might not be modified that easily, so maybe that's overkill.

rosteen:
> check if the original `spec.wcs.cdelt` was changed by this

Doesn't checking that the `world_axis_units` is different between the original WCS and new WCS confirm that they are different objects? I think this test is accomplishing the same thing as yours.

> Just noted that the entire `meta` is copied as well; being a dictionary, this would probably need a `deepcopy` as well to ensure you have separate objects.

The meta is an `OrderedDict`, which doesn't seem to have `deepcopy`. I just confirmed that `meta is self._meta` returns False after the `copy`, so I think this is ok.

specutils/spectra/spectrum_mixin.py
```diff
@@ -208,14 +215,27 @@ def with_spectral_unit(self, unit, velocity_convention=None,
...
-            rest_value=rest_value or self.rest_value)
 
-        spectrum = self.__class__(flux=self.flux, wcs=new_wcs, meta=new_meta)
+        velocity_convention = velocity_convention if velocity_convention is not None else self.velocity_convention # noqa
+        rest_value = rest_value if rest_value is not None else self.rest_value
+        unit = self._new_wcs_argument_validation(unit, velocity_convention,
+                                                 rest_value)
+
+        # Store the original unit information and WCS for posterity
+        meta = self._meta.copy()
```
dhomeier, line 225:
```suggestion
        meta = self._meta.deepcopy()
```
This would also be needed if `meta` itself includes nested objects (see test example)

specutils/tests/test_spectrum1d.py
```diff
@@ -150,10 +150,17 @@ def test_spectral_axis_conversions():
...
 
-    new_spec = spec.with_spectral_unit(u.GHz)  # noqa
+    new_spec = spec.with_spectral_axis_unit(u.km/u.s, rest_value=125*u.um,
+                                       velocity_convention="relativistic")
+
+    assert new_spec.spectral_axis.unit == u.km/u.s
+    assert new_spec.wcs.world_axis_units[0] == "km.s**-1"
+    # Make sure meta stored the old WCS correctly
+    assert new_spec.meta["original_wcs"].world_axis_units[0] == "nm"
+    assert new_spec.meta["original_spectral_axis_unit"] == "nm"
```
dhomeier, line 163:
```suggestion
    assert new_spec.meta["original_spectral_axis_unit"] == "nm"

    wcs_dict = {"CTYPE1": "WAVE", "CRVAL1": 3.622e3, "CDELT1, "CDELT3": 8e-2, "CRPIX1": 0, "CUNIT1": "Angstrom"}
    wcs_spec = Spectrum1D(flux=np.random.randn(49) * u.Jy, wcs=WCS(wcs_dict),
                          meta={'header': wcs_dict.copy()})
    new_spec = wcs_spec.with_spectral_axis_unit(u.km/u.s, rest_value=125*u.um,
                                       velocity_convention="relativistic")
    new_spec.wcs.wcs.crval = [3777]
    new_spec.meta['header']['CRVAL1'] = 3777.0

    assert wcs_spec.wcs.wcs.crval[0] == 3.622e-7
    assert wcs_spec.meta['header']['CRVAL1'] == 3622.
```
This should provide a basic test that the `new_spec` components are independent and not propagating changes back to `wcs_spec` (the second one should require the `self._meta.deepcopy()` change above).

dhomeier (commented):
Tried to set up an example of the deepcopy tests I commented on

rosteen:
@dhomeier I did update the docstring already to explain what happens to the original WCS, is that enough to address the concern you brought up about documentation earlier?

specutils/spectra/spectrum_mixin.py
```diff
@@ -184,10 +182,19 @@ def velocity(self):
...
+                                     rest_value=rest_value)
+
+    def with_spectral_axis_unit(self, unit, velocity_convention=None,
+                           rest_value=None):
         """
-        Returns a new spectrum with a different spectral axis unit.
+        Returns a new spectrum with a different spectral axis unit. Note that this creates a new
+        object using the converted spectral axis and thus drops the original WCS, if it existed.
+        The original WCS will be stored in the ``original_wcs`` entry of the new object's ``meta``
+        dictionary.
```
dhomeier, line 197:
```suggestion
        object using the converted spectral axis and thus drops the original WCS, if it existed,
        replacing it with a :class:`~specutils.utils.wcs_utils.SpectralGWCS`. The original WCS
        will be stored in the ``original_wcs`` entry of the new object's ``meta`` dictionary.
```
That's about my only point left about the documentation, to make it clear that `new_spec.wcs` will be a different class of object. Maybe `SpectralGWCS` could also use a basic docstring, if only to link to `~gwcs.WCS`...

dhomeier (approved):
Thanks; leaving some final remarks on the docs, but this looks good now!

dhomeier:
Not sure – do we get
> WARNING: py:class reference target not found: specutils.utils.wcs_utils.SpectralGWCS

because that version is not in the online docs yet, or actually because `SpectralGWCS` does not have a docstring yet?

But I am seeing other [broken links](https://specutils.readthedocs.io/en/latest/api/specutils.analysis.moment.html#specutils.analysis.moment) into `utils` that apparently have not broken the build.

specutils/analysis/moment.py
```diff
@@ -21,7 +21,7 @@ def moment(spectrum, regions=None, order=0, axis=-1):
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
         The spectrum object over which the width will be calculated.
 
-    regions: `~specutils.utils.SpectralRegion` or list of `~specutils.utils.SpectralRegion`
+    regions: `~specutils.SpectralRegion` or list of `~specutils.SpectralRegion`
```
dhomeier, line 24:
```suggestion
    regions: `~specutils.spectra.spectral_region.SpectralRegion` or list of `~specutils.spectra.spectral_region.SpectralRegion`
```
looks like these need the full path, even if they are exposed at the top level (cf. Spectrum1D reference above).

specutils/analysis/uncertainty.py
```diff
@@ -22,7 +22,7 @@ def snr(spectrum, region=None):
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
         The spectrum object overwhich the equivalent width will be calculated.
 
-    region: `~specutils.utils.SpectralRegion` or list of `~specutils.utils.SpectralRegion`
+    region: `~specutils.SpectralRegion` or list of `~specutils.SpectralRegion`
```
dhomeier, line 25:
```suggestion
    region: `~specutils.spectra.spectral_region.SpectralRegion` or list of `~specutils.spectra.spectral_region.SpectralRegion`
```

specutils/analysis/uncertainty.py
```diff
@@ -67,7 +67,7 @@ def _snr_single_region(spectrum, region=None):
     spectrum : `~specutils.spectra.spectrum1d.Spectrum1D`
         The spectrum object overwhich the equivalent width will be calculated.
 
-    region: `~specutils.utils.SpectralRegion`
+    region: `~specutils.SpectralRegion`
```
dhomeier, line 70:
```suggestion
    region: `~specutils.spectra.spectral_region.SpectralRegion`
```

specutils/analysis/uncertainty.py
```diff
@@ -109,7 +109,7 @@ def snr_derived(spectrum, region=None):
...
```
dhomeier, line 112:
```suggestion
    region: `~specutils.spectra.spectral_region.SpectralRegion`
```

specutils/analysis/uncertainty.py
```diff
@@ -155,7 +155,7 @@ def _snr_derived(spectrum, region=None):
...
```
dhomeier, line 158:
```suggestion
    region: `~specutils.spectra.spectral_region.SpectralRegion`
```

dhomeier (commented):
Original problem solved, new barrel opened!
At least the docs are getting a good overhaul now. :D

rosteen:
@dhomeier After some tinkering I couldn't get that doc link to work (I was breaking more things...) so I decided to just make the docstring a little more descriptive here for now as a compromise. I pointed those other `utils` doc links to the proper place, but they still seem to be broken rather than linking 🤔. I'll come back to that at some point but I have more pressing concerns at the moment.

rosteen:
@dhomeier As I commented above on one of your suggestions, you can see [here](https://specutils--1119.org.readthedocs.build/en/1119/api/specutils.analysis.moment.html#specutils.analysis.moment) that using the full path to `SpectralRegion` doesn't fix the broken link. If you want to debug that and open a follow-up PR that would be awesome, but I don't think it should hold up this one.

dhomeier:
Sorry, was a bit too fast in committing then – maybe you should force-push again to revert.
Sphinx keeps puzzling me; I just noted that in the docs themselves the recommended path is
```python
from specutils.spectra import SpectralRegion
```
but this can wait for another PR as well.

PR #1121, opened 2024-02-02
Truncate in template_comparison by default rather than nan_fill

rosteen:
Implements the suggestion given in #653. One thing I changed here, which I'm not sure of, is that the returned template is now resampled onto the observed spectrum's spectral axis, in addition to being normalized as previous. Opening as draft since I still need to update the docs, and I'll probably need to add a bit more test coverage.

### DIFF
--- a/specutils/analysis/template_comparison.py
+++ b/specutils/analysis/template_comparison.py
@@ -108,10 +108,6 @@ def _chi_square_for_templates(observed_spectrum, template_spectrum, resample_met
 
     observed_truncated = observed_spectrum[matching_indices.min():matching_indices.max()+1]
 
-    # If there was no overlap, we raise an error
-    #if len(template_obswavelength.flux) == 0:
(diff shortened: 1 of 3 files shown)


camipacifici (commented):
Tested now. I see it uses only the region of overlap between the template and the observed spectrum to calculate the chi^2, but it also returns the template only in the overlapping region, while it would be useful to get out the full original template properly redshifted and normalized. Here are two screenshots: one shows the observed spectrum and one of the template in its full range; the other shows the observed spectrum and the selected template after template_comparison.
Screenshot 2024-02-06 at 12 03 46 PM
Screenshot 2024-02-06 at 12 03 56 PM

rosteen:
@camipacifici Thanks for the comment, I reverted back to returning the full template. I also fixed an existing bug in the redshifting so that it will now work on any unit of spectral axis (e.g., frequency), not just wavelengths.

camipacifici (approved):
Functionality looks good to me!

PR #1126, opened 2024-03-26
Update spectrum class name from Spectrum1D to Spectrum

rosteen:
Does what it says on the tin - based on the result of https://github.com/astropy/specutils/discussions/1087. `Spectrum1D` will raise a deprecation warning when initialized for some deprecation period.


rosteen:
Looks like the test failure is unrelated, I wonder if that MANGA file was removed or changed url or something...

rosteen:
This is probably the most important part if anyone reviews:

```
@deprecated(since="2.0", alternative="Spectrum")
class Spectrum1D(Spectrum):

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
```

Issue #1129, opened 2024-04-05
Copy semantics change in numpy 2.0 caused incompatibility with astropy v6.1rc 


pllim:
Please see:

* https://github.com/astropy/astropy/issues/16167
* https://github.com/astropy/astropy-integration-testing/pull/16
* https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword

```
    def test_quantity_model():
        c = modeling.models.Chebyshev1D(3)
        uc = QuantityModel(c, u.AA, u.km)
    
>       assert uc(10*u.nm).to(u.m) == 0*u.m

specutils/tests/test_utils.py:26: 
...
E       ValueError: Unable to avoid copy while creating an array as requested.
E       If using `np.array(obj, copy=False)` replace it with `np.asarray(obj)` to allow a copy when needed (no behavior change in NumPy 1.x).
E       For more details, see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword.
```

Example logs:

* https://github.com/astropy/astropy-integration-testing/actions/runs/8568043943/job/23485760695?pr=16 (`specutils==1.13.0`)
* https://github.com/astropy/astropy-integration-testing/actions/runs/8568043943/job/23485761014?pr=16 (`specutils @ git+https://github.com/astropy/specutils.git@dac04cafe03079e487c9b5351a3713f1da09f6a7`)

cc @astrofrog @neutrinoceros

pllim:
I might be able to patch this...

pllim:
I can reproduce this error in this repo now that astropy nightly wheel upload is fixed. A previously green job now fails: https://github.com/astropy/specutils/actions/runs/8531304724/job/23500820517

PR #1130, opened 2024-04-05
MNT: Compat with numpy 2 and astropy 6.1

pllim:
Similar patch as https://github.com/astropy/astropy-healpix/pull/215

Fix https://github.com/astropy/specutils/issues/1129


rosteen (approved):
I see green checks, thanks!

PR #1138, opened 2024-05-14
Separate out the codecov upload to a separate job in CI

rosteen:
This will let us see a test failure if the coverage upload fails. Thanks to @pllim for the example.


rosteen:
It works, merging.

pllim:
Over at https://github.com/OpenAstronomy/github-actions-workflows/pull/199 , I found a bug in this setup where if we have two jobs that generate coverage, one will clobber the other. Since we only have one coverage job here, it doesn't matter. But if the situation change in the future, we can either switch to OpenAstronomy or I apply the same patch here. FYI.

PR #1141, opened 2024-05-24
Fix moment calculations

rosteen:
This solves a bug that was reported offline by @camipacifici - moment 0 should be in units of flux * spectral axis, being the flux integrated over the spectral range, matching the units of line flux. This PR changes from doing a simple sum of flux for moment 0, to summing flux*dx as in `line_flux`.

The case for `SpectrumCollection` is because the `spectral_axis` of a `SpectrumCollection` comes out as `SpectralCoord`, and we need `bin_edges` from `SpectralAxis` in each individual member of the collection.

### DIFF
--- a/docs/analysis.rst
+++ b/docs/analysis.rst
@@ -150,7 +150,7 @@ The `~specutils.analysis.moment` function computes moments of any order:
 
     >>> from specutils.analysis import moment
     >>> moment(noisy_gaussian, SpectralRegion(7*u.GHz, 3*u.GHz)) # doctest:+FLOAT_CMP
-    <Quantity 98.26318995 Jy>
+    <Quantity 4.93784874 GHz Jy>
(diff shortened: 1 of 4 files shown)


rosteen:
There's an unrelated deprecation warning in devdeps that I'll address separately.

camipacifici (approved):
It is now correct in the 1D case for flux density and it seems to propagate well in Cubeviz. Thank you!

rosteen:
> There's an unrelated deprecation warning in devdeps that I'll address separately.

Ah, FWIW this is upstream in GWCS: https://github.com/spacetelescope/gwcs/issues/499.

pllim:
Looks like this broke tests over at Jdaviz, see https://github.com/spacetelescope/jdaviz/pull/2888#issuecomment-2140975061

PR #1145, opened 2024-06-19
BUG: adjust copy semantics for compatibility with numpy 2

neutrinoceros:
See https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword

This fixes a test failure that I saw locally, though oddly enough it doesn't seem to have shown in CI yet
Reprod:
```
pytest specutils/tests/test_utils.py::test_quantity_model
```

### DIFF
--- a/specutils/utils/quantity_model.py
+++ b/specutils/utils/quantity_model.py
@@ -1,6 +1,9 @@
+import numpy as np
 from astropy import units as u
 
 __all__ = ['QuantityModel']
+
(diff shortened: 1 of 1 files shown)


rosteen:
It looks like all test failures are unrelated, I think we can get this in for future-proofing. it is odd that this didn't show up in the CI and only locally, thanks for noticing and fixing it!

rosteen (approved):

PR #1152, opened 2024-07-12
Update changelog for #1113 and loader documentation

dhomeier:
Following up on #1113 with some clarifications of its changelog that I missed to bring up to date with the last commits.
With the new options I am wondering if they should not rather go into _New Features_.

This is also attempting to document the reader/writer usage a bit better in the `Spectrum1D` loaders docs.
No fix to the general problem that there is no narrative or API documentation for the loaders themselves – the docs so far are focussing a lot on creating custom loaders, which from experience, almost no one is using (since most people needing a specific loader are contributing it to `default_loaders` right away), and which is not in a good shape itself IMO.
So the changes here (essentially just the first paragraphs, the rest is just formatting fixes) are trying to point the users more to the builtin help functions.

Another quirk I noticed is the [read/write docs](https://specutils.readthedocs.io/en/latest/api/specutils.Spectrum1D.html#specutils.Spectrum1D.read) everywhere referring to `CCDData`. I suppose because they are directly inherited from `NDIOMixin`, and `CCDData` is probably the oldest subclass actually implementing those I/O methods, but I don't see an easy way to fix this.

### DIFF
--- a/docs/spectrum1d.rst
+++ b/docs/spectrum1d.rst
@@ -21,8 +21,8 @@ create it explicitly from arrays or `~astropy.units.Quantity` objects:
     >>> import astropy.units as u
     >>> import matplotlib.pyplot as plt
     >>> from specutils import Spectrum1D
-    >>> flux = np.random.randn(200)*u.Jy
-    >>> wavelength = np.arange(5100, 5300)*u.AA
+    >>> flux = np.random.randn(200) * u.Jy
+    >>> wavelength = np.arange(5100, 5300) * u.AA
     >>> spec1d = Spectrum1D(spectral_axis=wavelength, flux=flux)
     >>> ax = plt.subplots()[1]  # doctest: +SKIP
     >>> ax.plot(spec1d.spectral_axis, spec1d.flux)  # doctest: +SKIP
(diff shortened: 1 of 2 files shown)


docs/spectrum1d.rst
```diff
@@ -81,24 +85,37 @@ List of Loaders
...
 
 .. code-block:: python
 
     >>> spec1d.write("/path/to/output.fits")  # doctest: +SKIP
 
+Note that the above example, calling ``write()`` without specifying
+any format, will default to the ``wcs1d-fits`` loader if the `~specutils.Spectrum1D`
+has a compatible WCS, and to ``tabular-fits`` otherwise, or if writing
+to another than the primary HDU (``hdu=0``) has been selected.
+For better control of the file type, the ``format`` parameter should be explicitly passed.
```
kelle, line 117: does header need to be specified? or does it default to using what's in meta?

kelle (changes requested):
Document how to write out the header.

docs/spectrum1d.rst
dhomeier: Per default it is using `meta['header']` both in `tabular-fits` and `wcs1d-fits` (in principle it depends on the loader, but those 2 are the only ones that provide a writer at the moment).
With `update_header=True` it treats the entire `meta` as a dictionary to update the default header from (i.e. pulls in all items that are serialisable and do not conflict with FITS keyword rules).

kelle:
Just gonna throw some notes in here as we work through some examples.
- `overwrite` argument doesn't seem to be documented for `write`.

rosteen (approved):
Thanks for updating this, let's get it in so I can do a release soon. We can always make further clarifications/expand the docs more later if desired.

dhomeier:
> Just gonna throw some notes in here as we work through some examples.
> 
> - `overwrite` argument doesn't seem to be documented for `write`.

This was merged before I pushed my responses to this comment, will submit them in another PR.

rosteen:
> This was merged before I pushed my responses to this comment, will submit them in another PR.

I may have misunderstood your comments last time we spoke offline, I thought this was your plan. Apologies if it wasn't!

dhomeier:
I may not have stated a clear plan; I think I mentioned some further updates to docstrings, but not being sure how much of the narrative docs I'd get to work on. Does not matter if they go into this PR or another follow-up though, no problems!

PR #1156, opened 2024-07-22
Bugfix: ensure flux dtype passed to medfilt is supported

bmorris3:
specutils' `median_smooth` method passes the `Spectrum1D.flux` attribute to scipy's `medfilt` method. `medfilt` [only supports a few dtypes](https://github.com/scipy/scipy/blob/44e4ebaac992fde33f04638b99629d23973cb9b2/scipy/signal/_signaltools.py#L1573-L1580), but the `Spectrum1D` parser doesn't normalize the dtype of the fluxes.

https://github.com/astropy/astropy-workshop/pull/241 is failing on calling `median_smooth` because the SDSS spectrum [at this URL](https://data.sdss.org/sas/dr16/sdss/spectro/redux/26/spectra/1323/spec-1323-52797-0012.fits) gets read into a numpy array with dtype `>f4`, which is not one of the supported dtypes in `medfilt`.

This PR tries to catch fluxes that are not supported and cast them to floats before passing them to `medfilt`. I've added a test that makes sure a few dtypes are supported.

### DIFF
--- a/specutils/manipulation/smoothing.py
+++ b/specutils/manipulation/smoothing.py
@@ -267,6 +267,12 @@ def median_smooth(spectrum, width):
     # Get the flux of the input spectrum
     flux = spectrum.flux
 
+    if (
+            not isinstance(flux.dtype, (float, int)) or
+            not np.issubdtype(flux.dtype, (np.floating, np.integer))
+    ):
+        flux = flux.astype(float)
+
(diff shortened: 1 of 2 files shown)


rosteen (approved):
Failures are unrelated, thanks for the fix!

PR #1157, opened 2024-07-31
Add documentation for kwargs passed to reader+writer backends

dhomeier:
Additional follow up to #1152 in response to https://github.com/astropy/specutils/pull/1152#issuecomment-2231742477 – @kelle, those are generally backend options and thus depend on what interface the various loaders are using; I have completed the docstrings for the two FITS loaders to specifically point to that. Again, since these are not linked to the processed docs, I put a few examples into the narrative docs (might be preferable to link to the utility functions docs, but `parsing_utils` again are not included even in the API docs – would be useful to have them also for the _custom loader_ section).

I noticed that we are linking to the astropy-dev docs rather than stable, is this by design?

### DIFF
--- a/docs/spectrum1d.rst
+++ b/docs/spectrum1d.rst
@@ -3,15 +3,15 @@ Working with Spectrum1Ds
 ========================
 
 As described in more detail in :doc:`types_of_spectra`, the core data class in
-specutils for a single spectrum is `~specutils.Spectrum1D`.  This object
+specutils for a single spectrum is :class:`~specutils.Spectrum1D`.  This object
 can represent either one or many spectra, all with the same ``spectral_axis``.
 This section describes some of the basic features of this class.
 
 Basic Spectrum Creation
 -----------------------
 
-The simplest way to create a `~specutils.Spectrum1D` is to
-create it explicitly from arrays or `~astropy.units.Quantity` objects:
+The simplest way to create a :class:`~specutils.Spectrum1D` is to create
+it explicitly from arrays or :class:`~astropy.units.Quantity` objects:
 
 .. plot::
     :include-source:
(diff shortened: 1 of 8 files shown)


kelle:
I think we should add an **entire section** dedicated to header with the best practices and caveats. The information you provided in a response to me is a great start.  I think the section should also include an example of header round-tripping with `tabular-fits`. E.g, an example of adding a header to a Spectrum1D object, writing it out, and reading it back in.

> Per default it is using meta['header'] both in tabular-fits and wcs1d-fits (in principle it depends on the loader, but those 2 are the only ones that provide a writer at the moment).
> With update_header=True it treats the entire meta as a dictionary to update the default header from (i.e. pulls in all items that are serialisable and do not conflict with FITS keyword rules).
>

kelle (changes requested):
I think we should add a section dedicated to headers.

rosteen:
> I noticed that we are linking to the astropy-dev docs rather than stable, is this by design?

That surprises me, certainly not by design on my part. I actually am not even sure where that's specified.

dhomeier:
The first intersphinx links, for astropy and gwcs, were using `latest`, those added later `stable`
https://github.com/astropy/specutils/blob/f77c8dfc386d6b5a3febf673a4d116468227eead/docs/conf.py#L60-L63
so might be accidental.

dhomeier:
> I think we should add an **entire section** dedicated to headers with the best practices and caveats.

Yes, I think that would be a good follow up PR (i.e. for a future release) for someone with a good grasp of what are now considered best practices, maybe @simontorres?
For this docs fixup I have added a small paragraph with the basics so it is at least documented that a `meta` exists. ;-)

docs/spectrum1d.rst
```diff
@@ -115,9 +118,25 @@ any format, will default to the ``wcs1d-fits`` loader if the `~specutils.Spectru
...
+The :attr:`specutils.Spectrum1D.meta` attribute provides a dictionary to store
+additional information on the data, like origin, date and other circumstances.
+For spectra read from files containing header-like attributes like a FITS
+:class:`~astropy.io.fits.Header` or :attr:`astropy.table.Table.meta`,
+loaders are conventionally storing this in ``Spectrum1D.meta['header']``.
+
+Conversely the two provided FITS writers are saving the contents of any
+``Spectrum1D.meta['header']`` (which should be an :class:`astropy.io.fits.Header`
+or any object like a `dict` that can instantiate one) as the header of the
+:class:`~astropy.io.fits.hdu.PrimaryHDU`.
```
kelle, line 138:
```suggestion
The two provided FITS writers (`fits-tabular` and `wcs-fits`) save the contents of
``Spectrum1D.meta['header']`` (which should be an :class:`astropy.io.fits.Header`
or any object like a `dict` that can instantiate one) as the header of the
:class:`~astropy.io.fits.hdu.PrimaryHDU`.
```

kelle: Please also add a link to the docstrings of the `tabular_fits` and the `wcs_fits` writers. I cant seem to find those currently included in the specutils docs. Maybe that's also a separate PR?

kelle (approved):
Minor suggestions.

docs/spectrum1d.rst
dhomeier: I've added the suggestions; there are no API docs to any of the loaders one could link to, and I am afraid this will be non-trivial to amend since they are dynamically loaded through the unified IO framework.
I suppose one could build upon `_list_of_loaders` to then recursively call `Spectrum1D.read.help(format)` from its list entries to generate a doc page of all the readers (and writers)...
As mentioned above the `parsing_utils` functions are missing as well, but that should be more straightforward to fix; just don't know where to add the right hook for Sphinx. But those are more of interest for the _Custom Loader_ documentation. Definitely more than enough stuff for another PR, or two.

rosteen (approved):
Thanks again, LGTM.

PR #1159, opened 2024-08-08
Compromise and print out the estimated parameters individually

### LINKED ISSUE #1146: TST: a failing doctest on MacOS + numpy 2
There were a number of intentional changes for MacOS arm64 in numpy 2.0, which was released 3 days ago.
It seems that as a side effect of the performance optimizations conducted a table printed in `docs/fitting.rst` changed; `# doctest: +FLOAT_CMP` helpfully allows to ignore discrepancies in lower digits, but it doesn't protect against what's happening here: the *number* of digits printed changed.
See [example logs](https://github.com/astropy/specutils/actions/runs/9574096855/job/26396745973).

I tried to make this doctest more portable using the `+ELLIPSIS` directive and ended up with the following patch
```patch
diff --git a/docs/fitting.rst b/docs/fitting.rst
index b4a3786..4db9e35 100644
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -147,16 +147,13 @@ For example, based on the spectrum defined above we can first select a region:
 
 Then estimate the line  parameters it it for a Gaussian line profile::
 
-   >>> print(estimate_line_parameters(sub_spectrum, models.Gaussian1D()))  # doctest: +FLOAT_CMP
+   >>> print(estimate_line_parameters(sub_spectrum, models.Gaussian1D()))  # doctest: +FLOAT_CMP, +ELLIPSIS
       Model: Gaussian1D
       Inputs: ('x',)
       Outputs: ('y',)
       Model set size: 1
       Parameters:
-              amplitude            mean             stddev
-                  Jy                um                um
-          ------------------ ---------------- -------------------
-          1.1845669151078486 4.57517271067525 0.19373372929165977
+      ...
 
 
 If an `~astropy.modeling.Model` is used that does not have the predefined
```
I could confirm it fixes the test on macos arm64, but it's inconvenient that I just ended up hiding the _whole_ table; unfortunately my attempts to only hide parts of it did not succeed.
Would this patch be accepted as is ?

rosteen:
Trying to fix the annoying Mac doctest failure (which I can't replicate on my own Mac...)

Fixes #1146

### DIFF
--- a/docs/fitting.rst
+++ b/docs/fitting.rst
@@ -145,19 +145,17 @@ For example, based on the spectrum defined above we can first select a region:
    >>> sub_region = SpectralRegion(4*u.um, 5*u.um)
    >>> sub_spectrum = extract_region(spectrum, sub_region)
 
-Then estimate the line  parameters it it for a Gaussian line profile::
+Then estimate the line parameters for a Gaussian line profile::
 
-   >>> print(estimate_line_parameters(sub_spectrum, models.Gaussian1D()))  # doctest: +FLOAT_CMP
-      Model: Gaussian1D
-      Inputs: ('x',)
-      Outputs: ('y',)
-      Model set size: 1
-      Parameters:
-              amplitude            mean             stddev
-                  Jy                um                um
-          ------------------ ---------------- -------------------
-          1.1845669151078486 4.57517271067525 0.19373372929165977
+   >>> result = estimate_line_parameters(sub_spectrum, models.Gaussian1D())
+   >>> print(result.amplitude)  # doctest: +FLOAT_CMP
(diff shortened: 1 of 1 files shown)


pllim (approved):
Works for me. Thanks!

pllim:
Since there is nothing you can do about the deprecation warning in devdeps, maybe you just want to ignore it?

rosteen:
> Since there is nothing you can do about the deprecation warning in devdeps, maybe you just want to ignore it?

Eh, it's an allowed failure, I'm inclined to leave it as-is for now. I don't want to forget about it in case it actually starts causing problems later on...

PR #1164, opened 2024-09-04
Fix fit_lines with single quantity as window value

### LINKED ISSUE #1160: fit_lines center window does not work
Using a single value for `window` in `fit_lines` causes the program to use the entire spectrum as opposed to centering around the mean of the gaussian. Defining the window bounds manually works as expected. I've tried putting in some extreme constants into `window` such as np.inf and negative numbers, all of which have no effect on the line fit.

Defining a window of 10 by setting bounds manually:
```python
g_init = models.Gaussian1D(amplitude=2000, mean=6563 * u.AA, stddev=3 * u.AA)
g_init += models.Const1D(0)
g_fit = fit_lines(spectrum, g_init, window=(6553 * u.AA, 6573 * u.AA))
```
bounds_window

Defining a window of 10 automatically:
```python
g_init = models.Gaussian1D(amplitude=2000, mean=6563 * u.AA, stddev=3 * u.AA)
g_init += models.Const1D(0)
g_fit = fit_lines(spectrum, g_init, window=10 * u.AA)
```
center_window

rosteen:
Fixes #1160. Currently on main, giving a single Quantity as input for `window` gets ignored, since a Quantity is neither an `int` nor a `float`. You can see below that one of the fits in the fitting tests is actually very bad, since it's fitting the whole spectrum and doesn't limit to a window around the left gaussian:

Screenshot 2024-09-04 at 1 23 31 PM

With this fix, the fit for the left gaussian in the test case properly gets limited to the defined window and is fit well:

Screenshot 2024-09-04 at 1 24 04 PM


rosteen:
Looks like the one test failure is a remote data timeout.

@TurboSlayer would you check to see if this fixes your use case?

TurboSlayer:
Thanks - I managed to work around it by setting manual bounds anyways. I trust from the screenshots that you've fixed the issue haha.

pllim (approved):
Adding test to fit in frequency space might be good, or a test where spectral_axis is wavelength but the window in frequency (is that even allowed) but you can also argue expanding test suite is out of scope here. 

Needs a change log?

rosteen:
> Needs a change log?

You're not my boss!

(yes, yes, I'll add one momentarily)

pllim:
> You're not my boss!

Indeed I am not. But I can unleash the change log bot here if you want. 😉

CHANGES.rst
```diff
@@ -7,6 +7,8 @@ New Features
 Bug Fixes
 ^^^^^^^^^
 
+- Fixed specifying a single value for ``window`` in ``analysis.fit_lines``. [#1164]
```
pllim, line 10: Actually, do you want to add a new test for this particular bug?

rosteen: The existing tests would theoretically have caught it, if they were set to check for the incorrect values. Now they're looking for the values resulting from the fit actually doing the right thing.

rosteen: ` yl_double_fit_expected` is the big difference.

pllim: Gotcha, thanks!

Issue #1167, opened 2024-09-17
excise_regions() is not inclusive on the first boundary


etsmit:
I am one of the developers for [dysh](https://github.com/GreenBankObservatory/dysh), a project using specutils to perfom single-dish radio telescope calibration and data reduction. Upon testing baseline subtraction, we came across a discrepancy caused by the fact that the specutils excise_regions() function (or, either linear_exciser or true_exciser) are not inclusive on one boundary of each exclusion region, leading to a single channel unintentionally getting through. There are more details/discussions in the [dysh issue](https://github.com/GreenBankObservatory/dysh/issues/378), but is summarized by the pre- and post- excision spectra below:

![367334702-30c028fa-506b-4e75-8a42-7b8b3b93235d](https://github.com/user-attachments/assets/b626d1d5-b51e-46fc-85f7-9157613bb783)

![367335288-6ed54659-6fcf-446e-aee9-3c6401175553](https://github.com/user-attachments/assets/eee39fd1-51be-446f-9ae5-a8bf6c47b43e)

where channel 0* gets missed by the true_exciser() function and dramatically shows up as the large spike to the right. Although only a small difference, this has implications for baseline fitting unit tests where we are trying to compare accuracy to a previous software suite.

Note: There are technically 3 excision regions in this example, so an extra 3 channels are unintentionally included by the excision. For the first two regions, one can simply compensate by increasing the boundary length by 1 channel. However, how does one change the boundary to be lower than 0?

I am adding a small PR with the change to the boundary conditions.

* This is lower sideband data, which means that the frequency axis is flipped - the lower channel is the second boundary in frequency space.

rosteen:
I'm going to have to think about this one for a bit - I'm assuming that this was coded this way to be consistent with general Python indexing, where the lower bound is inclusive and the upper bound is exclusive. But it should at least be well documented, and we should make sure  similar operations in the package are all consistent. 

My initial thought is that, rather than changing it to always have an inclusive upper bound like your PR, it would be reasonable to keep the current Python-like indexing, but add a `inclusive_upper` argument to the excise functions to enable cases like yours. Otherwise you would need to hack around this in your `exclude_to_region` function (perhaps by adding a slight offset to the upper bound [here](https://github.com/GreenBankObservatory/dysh/blob/c444eca2221e3f88cedbd50e229ddf19f5678fac/src/dysh/spectra/core.py#L184)), which I understand isn't ideal.

rosteen:
[...]

PR #1168, opened 2024-09-17
fit_lines exclusion regions now have inclusive boundaries

etsmit:
Details in [Issue 1167](https://github.com/astropy/specutils/issues/1167)


rosteen:
Thanks for the contribution. As noted elsewhere, I don't think I want to change the default behavior, so I opened #1171 to add this as an optional toggle instead.

PR #1169, opened 2024-09-19
Change default refraction_index method 

### LINKED ISSUE #1162: Is the default implementation in`specutils.utils.wcs_utils.air_to_vac` incorrect?
Hi folks,

I make a lot of model/data comparisons where model spectra are in vacuum and data are in wavelengths.

Today I noticed that the default method in specutils (`scheme='inversion', method='Griesen2006'`) for converting air wavelengths to vacuum seems to be either incorrect, or inaccurate. 

I attach a figure showing a ESO/HARPS spectrum of alpha Cen A, where the ESO/HARPS spectrum is recorded in air wavelengths. In the top panel I compare the data (in air wavelengths) with some approximate spectral model (in vacuum wavelengths) in pink. The subsequent panels also show the model in pink, but the data are shifted to vacuum wavelengths using different arguments in `specutils.utils.wcs_utils.air_to_vac`. I have added a small y-offset so we can see the conversion using different methods.

specutils-air-to-vac-issue

You can see that in either scheme (inversion/iteration), the Greisen 2006 conversion is vastly different from all other methods. This seems to be inaccurate or incorrect, and the unfortunately the Greisen 2006 method is the default in `specutils.utils.wcs_utils.air_to_vac`. The docstring for `air_to_vac` says that 'Griesen et al. (2006) report that the error in naively inverting is less than 10^-9.'. The difference in the figure between Greisen 2006 and the other methods is about 0.07 Angstroms.. far higher than 10^-9!

I haven't cross-checked the equations in the Greisen 2006 paper to see whether the implementation in `specutils` is correct or not. I wanted to open this issue for discussion first, particularly since users will get the wrong result by using the default parameters, and debugging air/vacuum issues with noisy RV/BERV shifts can be fiddly!

rosteen:
Closes #1162. It seems to me that we should use one of the consistent methods when possible, rather than the outlier (having no reason to believe that it's `Griesen2006` that is correct while all the other methods are wrong). This makes us consistent with the IAU, SDSS, VALD, etc.

### DIFF
--- a/specutils/utils/wcs_utils.py
+++ b/specutils/utils/wcs_utils.py
@@ -90,7 +90,7 @@ def refraction_index(wavelength, method='Morton2000', co2=None):
     VALID_METHODS = ['Griesen2006', 'Edlen1953', 'Edlen1966', 'Morton2000',
                      'PeckReeder1972', 'Ciddor1996']
     assert isinstance(method, str), 'method must be a string'
-    if method != 'Griesen2006' and wavelength.min() < 250 * u.nm:
+    if method != 'Griesen2006' and wavelength.min() < 200 * u.nm:
         raise ValueError("The chosen method is invalid for wavelengths below 250 nm."
                          " 'Griesen2006' is the only option for this wavelength range -"
                          " see the specutils.utils.wcs_utils.refraction_index docstring"
--- a/specutils/utils/wcs_utils.py
+++ b/specutils/utils/wcs_utils.py
@@ -46,7 +46,7 @@ class SpectralGWCS(GWCS):
             self.original_unit, equivalencies=u.spectral())
 
 
-def refraction_index(wavelength, method='Griesen2006', co2=None):
+def refraction_index(wavelength, method='Morton2000', co2=None):
     """
     Calculates the index of refraction of dry air at standard temperature
     and pressure, at different wavelengths, using different methods.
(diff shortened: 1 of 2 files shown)


keflavich:
I don't feel qualified to assess this without doing a pretty deep dive.  The research in https://github.com/astropy/specutils/issues/1162 is helpful, but it doesn't really establish why there is a difference or the range of validity of the approximations.   So I'm not going to approve, even though this seems fine to me - I just don't know enough to offer meaningful judgement right now.

@andycasey noted in that PR that the quoted 0.25 ppm accuracy point for Greisen was 650-850 nm; it would be helpful to include information like that in the docstrings and warnings in addition to the note already added here about singularities.

rosteen:
> I don't feel qualified to assess this without doing a pretty deep dive

Same really (and now I'm in the middle of doing that deep dive 😅 ).

specutils/utils/wcs_utils.py
```diff
@@ -85,6 +90,11 @@ def refraction_index(wavelength, method='Griesen2006', co2=None):
     VALID_METHODS = ['Griesen2006', 'Edlen1953', 'Edlen1966', 'Morton2000',
                      'PeckReeder1972', 'Ciddor1996']
     assert isinstance(method, str), 'method must be a string'
+    if method != 'Griesen2006' and wavelength.min() < 250 * u.nm:
```
rosteen, line 93:
```suggestion
    if method != 'Griesen2006' and wavelength.min() < 200 * u.nm:
```
The literature puts the limit for this method at 200 nm, might as well match that.

rosteen:
After some more reading, I don't see any reason to not do this, so I'm going to go ahead and merge. I updated the wavelength limit before throwing an error for the default method to match the literature.

PR #1170, opened 2024-09-19
Fix bug handling subregions in true_excise

rosteen:
Looking into #1167 led me to realize that the subregion handling in `true_excise` was actually using the upper and lower bounds from the full region rather than each subregion. This fixes that. I'll add a test tomorrow that would have caught this, since right now it looks like all tests still pass with the change.


rosteen:
I'm actually going to close this without merging, I didn't realize `true_excise` is not available to call independently from `excise_regions`, and the current behavior is as documented. Might be worth a deeper revisit sometime but for now I'll leave it as-is.

PR #1171, opened 2024-09-24
Add option to have inclusive upper boundaries in fitting/excision.

### LINKED ISSUE #1167: excise_regions() is not inclusive on the first boundary

rosteen:
Supersedes #1168 by making this behavior an optional toggle in `fit_models`, `fit_continuum`, and `excise_regions` rather than changing the default. 

Closes #1167.

### DIFF
--- a/docs/spectral_regions.rst
+++ b/docs/spectral_regions.rst
@@ -291,8 +291,13 @@ finding functions:
       (7.690954773869347 um, 8.690954773869347 um)
 
 This can be fed into the ``exclude_regions`` argument of the `~specutils.fitting.fit_generic_continuum` or
-`~specutils.fitting.fit_continuum` functions to avoid fitting regions that contain line features. Conversely, users can
-also invert the spectral region
+`~specutils.fitting.fit_continuum` functions to avoid fitting regions that contain line features. Note that,
+by default, this uses pythonic slicing, i.e., spectral values greater than or equal to the lower bound and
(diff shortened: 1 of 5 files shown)


rosteen:
Test failures are unrelated, I'm going to go ahead and merge.

PR #1175, opened 2024-10-03
Add docs about air to vac conversion

rosteen:
Adding some context about the refractive index default method change to the narrative docs. The `air_to_vac` stuff was never documented there as far as I can tell...I'm also wondering if these functions would fit better under `manipulation` than where they are now 🤔 .


rosteen:
The test failure is unrelated (and perhaps a sign that we should drop 3.9 support, we're behind other packages on that front).

Issue #1172, opened 2024-09-26
Replace LevMarLSQFitter with something else?


pllim:
As per this upstream PR:

* https://github.com/astropy/astropy/pull/16983

Looks like it is used here to fit lines and continuum. Also see https://github.com/astropy/photutils/issues/1895

[🐱](https://jira.stsci.edu/browse/JDAT-4812)

PR #1180, opened 2024-10-07
Replace LevMarLSQFitter with TRFLSQFitter

pllim:
Related PRs:

* https://github.com/astropy/astropy/pull/16983
* https://github.com/astropy/photutils/pull/1899
* https://github.com/spacetelescope/jdaviz/pull/3202

`TRFLSQFitter` also has `calc_uncertainties` and should respect bounds.

Fix https://github.com/astropy/specutils/issues/1172

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ Bug Fixes
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+- Replaced ``LevMarLSQFitter`` with ``TRFLSQFitter`` as the former is no longer
+  recommended by ``astropy``. [#1180]
+
 1.17.0 (2024-10-04)
 -------------------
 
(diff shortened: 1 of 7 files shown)


keflavich (approved):
lgtm.  Could you confirm that those numerical changes were caused by switching default fitter?

pllim:
> Could you confirm that those numerical changes were caused by switching default fitter?

How should I do that? That is the only change and then the numbers change. Not sure what else I can do. Hope you can advise. Thanks.

keflavich:
That's pretty much all I was asking.  I don't see any other differences.  I would like to know what these changes imply about the absolute accuracy of the two methods, or perhaps about their default tolerances, but that desire shouldn't hold back merging this.

astrofrog:
The differences here seem pretty small considering it is a completely different fitting algorithm. Note that TRFLSQFitter does have ``calc_uncertainties`` which you can use to find the uncertainties so it would be interesting to try this to see if the differences are indeed within the uncertainties.

pllim:
I'll leave that exercise for the future but I am *uncertain* who will do it. 😏 

Thanks, all! Merging.

PR #1181, opened 2024-10-08
Upstream fix for "Investigate replacing spectrum1d unit conversion in cubeviz parser"

pllim:
If accepted, after merge, will need a PR over at Jdaviz to use this and bump specutils pin.

If rejected, hack-fix Jdaviz to overwrite uncertainty after calling `with_flux_unit` from specutils. No need to bump specutils pin.

[🐱](https://jira.stsci.edu/browse/JDAT-4788)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,9 @@ New Features
 Bug Fixes
 ^^^^^^^^^
 
+- Fixed ``Spectrum1D.with_flux_unit()`` not converting uncertainty along
+  with flux unit. [#1181]
+
 Other Changes and Additions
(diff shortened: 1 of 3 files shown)


rosteen (approved):
Looks reasonable to me, I appreciate the added tests. Thanks!

pllim:
Thanks for the review and approval!

On the Jdaviz side, what specutils version should we bump to to get this patch?

pllim:
Actually maybe do a 1.18.0 instead after https://github.com/astropy/specutils/pull/1184 is resolved? Thanks!

PR #1184, opened 2024-10-11
FEAT: with_spectral_axis_and_flux_units

pllim:
Follow-up of https://github.com/astropy/specutils/pull/1181 . I think it would really benefit Cubeviz parser if we can do both spectral axis and unit conversions at the same time, to avoid copying an object twice unnecessarily.

I also couldn't help myself by cleaning up some existing test code after noticing unnecessary usage of Numpy random number generator.

[🐱](https://jira.stsci.edu/browse/JDAT-4788)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@
 New Features
 ^^^^^^^^^^^^
 
+- New ``Spectrum1D.with_spectral_axis_and_flux_units`` method to convert both
+  spectral axis and flux units at the same time. [#1184]
+
 Bug Fixes
 ^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


pllim:
devdeps failure is unrelated: `http.client.RemoteDisconnected: Remote end closed connection without response`

keflavich:
This makes some sense to me - with single 1D spectra, copying the spectral axis (which is needed for changing the spectral unit) is just as expensive as copying the data (changing the flux unit).  In spectral-cube, we have considered copies of the spectral axis 'cheap' under the assumption that the 1D axis is always << the 3D data.  So I'm onboard with this - but I had to think through it to see why there's value.

rosteen (approved):
LGTM, thanks for the test cleanups.

Issue #1182, opened 2024-10-10
sdss-v mwmVisit data loaders broken


havok2063:
The SDSS-V data loader for `mwmVisit` fails to load `mwmVisit` files that only have BOSS-extension data in them.  It loads fine for `mwmVisit` files with only APOGEE-extension data.  I think the issue is during metadata collection, it looks for info specific to certain extensions but not all.  e.g. see https://github.com/astropy/specutils/blob/67bea40efb4e4a01672241123db6c225b3954d14/specutils/io/default_loaders/sdss_v.py#L589

We could either remove the metadata entirely as it's not strictly necessary to load the spectrum, or adjust this code.  This is a blocker for https://github.com/sdss/sdss_solara/issues/9

Either I or relevant folks can fix, just creating an issue for tracking.

havok2063:
@rileythai do you have time to look into this?  If not, I can.  A relevant id is "54459273".   Separately, we also need to add a new loader for the astra aspcap model spectra.

rileythai:
I have time this weekend -- this should be a simple fix.

rosteen:
> I have time this weekend -- this should be a simple fix.

Great - we just merged another bug fix, I can wait to do a bug fix release until early next week if you can get this done by then. Thanks!

havok2063:
Thanks @rosteen I'd also like to get https://github.com/astropy/specutils/issues/1183 in that fix if possible.  If not, I'd like to get it in another release soon after.

PR #1185, opened 2024-10-12
fix: SDSS-V SpectrumList format ambiguity, mwmVisit BOSS load fail

rileythai:
This PR addresses #1182 , and an issue mentioned in #1107 about `SpectrumList` loader ambiguity.

Changes are:
- `mwmVisit` files now correctly load BOSS-only files.
   - the `mwmVisit` default loader no longer loads metadata about the "date observed" -- only the MJD of the observation (which is enough)
- Format is now successfully automatically detected and no longer require a manual specification when loading as a `SpectrumList`.
    - Forces override of default SDSS-V `SpectrumList` loaders generated from the `Spectrum1D` default loaders.
       - **BREAKING:** if anything previously manually specified any SDSS-V "multi" loaders for `SpectrumList`, they no longer exist and it will fail. 
       - This is only really relevant for [sdss_solara](https://github.com/sdss/sdss_solara)'s Jdaviz component @havok2063 . 
- Now warns users when using `Spectrum1D` loader for SDSS-V files with multiple spectra and specific HDU is not specified.
- Loaders now fetch `sdss_id` for `meta` field.
- Relevant tests in `tests/test_sdss_v.py` have been adjusted
   - tests now correctly assess whether the `mwm` filetype is `mwmVisit` or `mwmStar`
   - tests have been adjusted to remove temporary file when complete (via `os.remove`)
   - test cases have been added for BOSS-only `mwmVisit` and `mwmStar` files.

### DIFF
--- a/specutils/io/default_loaders/sdss_v.py
+++ b/specutils/io/default_loaders/sdss_v.py
@@ -517,9 +517,6 @@ def load_sdss_mwm_list(file_obj, **kwargs):
         for hdu in range(1, len(hdulist)):
             if hdulist[hdu].header.get("DATASUM") == "0":
                 # Skip zero data HDU's
-                warnings.warn(
-                    "WARNING: HDU{} ({}) is empty.".format(
-                        hdu, hdulist[hdu].name), AstropyUserWarning)
                 continue
             spectra.append(_load_mwmVisit_or_mwmStar_hdu(hdulist, hdu))
     return spectra
--- a/specutils/io/default_loaders/sdss_v.py
+++ b/specutils/io/default_loaders/sdss_v.py
@@ -312,6 +312,7 @@ def load_sdss_apVisit_list(file_obj, **kwargs):
 
 # BOSS REDUX products (specLite, specFull, custom coadd files, etc)
 
+
 @data_loader(
     "SDSS-V spec",
     identifier=spec_sdss5_identify,
@@ -578,16 +579,16 @@ def _load_mwmVisit_or_mwmStar_hdu(hdulist: HDUList, hdu: int, **kwargs):
     meta["snr"] = np.array(hdulist[hdu].data["snr"])
 
     # Add identifiers (obj, telescope, mjd, datatype)
-    # TODO: need to see what metadata we're interested in for the MWM files.
     meta["telescope"] = hdulist[hdu].data["telescope"]
     meta["instrument"] = hdulist[hdu].header.get("INSTRMNT")
-    try:
+    try:  # get obj if exists
         meta["obj"] = hdulist[hdu].data["obj"]
     except KeyError:
         pass
+
+    # choose between mwmVisit/Star via KeyError except
     try:
-        meta["date"] = hdulist[hdu].data["date_obs"]
-        meta["mjd"] = hdulist[hdu].data["mjd"]
+        meta['mjd'] = hdulist[hdu].data['mjd']
         meta["datatype"] = "mwmVisit"
     except KeyError:
         meta["mjd"] = (str(hdulist[hdu].data["min_mjd"][0]) + "->" +
@@ -595,6 +596,7 @@ def _load_mwmVisit_or_mwmStar_hdu(hdulist: HDUList, hdu: int, **kwargs):
         meta["datatype"] = "mwmStar"
     finally:
         meta["name"] = hdulist[hdu].name
+        meta["sdss_id"] = hdulist[hdu].data['sdss_id']
 
     return Spectrum1D(
         spectral_axis=spectral_axis,
--- a/specutils/io/default_loaders/sdss_v.py
+++ b/specutils/io/default_loaders/sdss_v.py
@@ -183,9 +183,10 @@ def load_sdss_apStar_1D(file_obj, idx: int = 0, **kwargs):
 
 
 @data_loader(
-    "SDSS-V apStar multi",
+    "SDSS-V apStar",
     identifier=apStar_identify,
     dtype=SpectrumList,
+    force=True,
     priority=10,
     extensions=["fits"],
 )
@@ -259,9 +260,10 @@ def load_sdss_apVisit_1D(file_obj, **kwargs):
 
 
 @data_loader(
-    "SDSS-V apVisit multi",
+    "SDSS-V apVisit",
     identifier=apVisit_identify,
     dtype=SpectrumList,
+    force=True,
     priority=10,
     extensions=["fits"],
 )
@@ -338,6 +340,7 @@ def load_sdss_spec_1D(file_obj, *args, hdu: Optional[int] = None, **kwargs):
     """
     if hdu is None:
         # TODO: how should we handle this -- multiple things in file, but the user cannot choose.
+        print('HDU not specified. Loading coadd spectrum (HDU1)')
         hdu = 1  # defaulting to coadd
         # raise ValueError("HDU not specified! Please specify a HDU to load.")
     elif hdu in [2, 3, 4]:
@@ -348,9 +351,10 @@ def load_sdss_spec_1D(file_obj, *args, hdu: Optional[int] = None, **kwargs):
 
 
 @data_loader(
-    "SDSS-V spec multi",
+    "SDSS-V spec",
     identifier=spec_sdss5_identify,
     dtype=SpectrumList,
+    force=True,
     priority=5,
     extensions=["fits"],
 )
@@ -463,14 +467,17 @@ def load_sdss_mwm_1d(file_obj, hdu: Optional[int] = None, **kwargs):
             for i in range(len(hdulist)):
                 if hdulist[i].header.get("DATASUM") != "0":
                     hdu = i
+                    print('HDU not specified. Loading spectrum at (HDU{})'.
+                          format(i))
                     break
 
         return _load_mwmVisit_or_mwmStar_hdu(hdulist, hdu, **kwargs)
 
 
 @data_loader(
-    "SDSS-V mwm multi",
+    "SDSS-V mwm",
     identifier=mwm_identify,
+    force=True,
     dtype=SpectrumList,
     priority=20,
     extensions=["fits"],
--- a/specutils/io/default_loaders/sdss_v.py
+++ b/specutils/io/default_loaders/sdss_v.py
@@ -3,19 +3,23 @@ import warnings
 from typing import Optional
 
 import numpy as np
-from astropy.io.fits import BinTableHDU, HDUList, ImageHDU
-from astropy.nddata import InverseVariance, StdDevUncertainty
-from astropy.units import Angstrom, Quantity, Unit
+from astropy.units import Unit, Quantity, Angstrom
+from astropy.nddata import StdDevUncertainty, InverseVariance
+from astropy.io.fits import HDUList, BinTableHDU, ImageHDU
 from astropy.utils.exceptions import AstropyUserWarning
 
 from ...spectra import Spectrum1D, SpectrumList
-from ..parsing_utils import read_fileobj_or_hdulist
 from ..registers import data_loader
+from ..parsing_utils import read_fileobj_or_hdulist
 
 __all__ = [
+    "load_sdss_apVisit_1D",
     "load_sdss_apVisit_list",
+    "load_sdss_apStar_1D",
     "load_sdss_apStar_list",
+    "load_sdss_spec_1D",
     "load_sdss_spec_list",
+    "load_sdss_mwm_1d",
     "load_sdss_mwm_list",
 ]
 
@@ -117,9 +121,16 @@ def _fetch_flux_unit(hdu):
 
 
 # APOGEE files
-def _load_sdss_apStar_1D(file_obj, idx: int = 0, **kwargs):
+@data_loader(
+    "SDSS-V apStar",
+    identifier=apStar_identify,
+    dtype=Spectrum1D,
+    priority=10,
+    extensions=["fits"],
+)
+def load_sdss_apStar_1D(file_obj, idx: int = 0, **kwargs):
     """
-    Subloader to load an apStar file as a Spectrum1D.
+    Load an apStar file as a Spectrum1D.
 
     Parameters
     ----------
@@ -197,13 +208,57 @@ def load_sdss_apStar_list(file_obj, **kwargs):
     """
     with read_fileobj_or_hdulist(file_obj, **kwargs) as hdulist:
         nvisits = hdulist[0].header.get("NVISITS")
-        if nvisits < 1:
-            raise RuntimeError(
-                "No visits in this file. Use Spectrum1D.read() instead.")
+        if nvisits <= 1:
+            raise ValueError(
+                "Only 1 visit in this file. Use Spectrum1D.read() instead.")
         return SpectrumList([
-            _load_sdss_apStar_1D(file_obj, idx=i, **kwargs)
+            load_sdss_apStar_1D(file_obj, idx=i, **kwargs)
             for i in range(nvisits)
         ])
+
+
+@data_loader(
+    "SDSS-V apVisit",
+    identifier=apVisit_identify,
+    dtype=Spectrum1D,
+    priority=10,
+    extensions=["fits"],
+)
+def load_sdss_apVisit_1D(file_obj, **kwargs):
+    """
+    Load an apVisit file as a Spectrum1D.
+
+    Parameters
+    ----------
+    file_obj : str, file-like, or HDUList
+        FITS file name, file object, or HDUList.
+
+    Returns
+    -------
+    Spectrum1D
+        The spectrum contained in the file
+    """
+    flux_unit = Unit("1e-17 erg / (Angstrom cm2 s)")
+
+    with read_fileobj_or_hdulist(file_obj, memmap=False, **kwargs) as hdulist:
+        # Fetch and flatten spectrum parameters
+        spectral_axis = Quantity(hdulist[4].data.flatten(), unit=Angstrom)
+        flux_unit = _fetch_flux_unit(hdulist[1])
+        flux = Quantity(hdulist[1].data.flatten(), unit=flux_unit)
+        e_flux = StdDevUncertainty(hdulist[2].data.flatten())
+
+        # Get metadata and attach bitmask and MJD
+        meta = dict()
+        meta["header"] = hdulist[0].header
+        mask = hdulist[3].data.flatten()
+        # NOTE: specutils considers 0/False as valid values, simlar to numpy convention
+        mask = mask != 0
+
+    return Spectrum1D(spectral_axis=spectral_axis,
+                      flux=flux,
+                      mask=mask,
+                      uncertainty=e_flux,
+                      meta=meta)
 
 
 @data_loader(
@@ -263,9 +318,45 @@ def load_sdss_apVisit_list(file_obj, **kwargs):
 @data_loader(
     "SDSS-V spec",
     identifier=spec_sdss5_identify,
+    dtype=Spectrum1D,
+    priority=5,
+    extensions=["fits"],
+)
+def load_sdss_spec_1D(file_obj, *args, hdu: Optional[int] = None, **kwargs):
+    """
+    Load a given BOSS spec file as a Spectrum1D object.
+
+    Parameters
+    ----------
+    file_obj : str, file-like, or HDUList
+        FITS file name, file object, or HDUList..
+    hdu : int
+        The specified HDU to load a given spectra from.
+
+    Returns
+    -------
+    Spectrum1D
+        The spectrum contained in the file at the specified HDU.
+    """
+    if hdu is None:
+        # TODO: how should we handle this -- multiple things in file, but the user cannot choose.
+        warnings.warn('HDU not specified. Loading coadd spectrum (HDU1)',
+                      AstropyUserWarning)
+        hdu = 1  # defaulting to coadd
+        # raise ValueError("HDU not specified! Please specify a HDU to load.")
+    elif hdu in [2, 3, 4]:
+        raise ValueError("Invalid HDU! HDU{} is not spectra.".format(hdu))
+    with read_fileobj_or_hdulist(file_obj, memmap=False, **kwargs) as hdulist:
+        # directly load the coadd at HDU1
+        return _load_BOSS_HDU(hdulist, hdu, **kwargs)
+
+
+@data_loader(
+    "SDSS-V spec",
+    identifier=spec_sdss5_identify,
     dtype=SpectrumList,
     force=True,
-    priority=10,
+    priority=5,
     extensions=["fits"],
 )
 def load_sdss_spec_list(file_obj, **kwargs):
@@ -344,9 +435,68 @@ def _load_BOSS_HDU(hdulist: HDUList, hdu: int, **kwargs):
 @data_loader(
     "SDSS-V mwm",
     identifier=mwm_identify,
+    dtype=Spectrum1D,
+    priority=20,
+    extensions=["fits"],
+)
+def load_sdss_mwm_1d(file_obj,
+                     hdu: Optional[int] = None,
+                     visit: Optional[int] = None,
+                     **kwargs):
+    """
+    Load an unspecified spec file as a Spectrum1D.
+
+    Parameters
+    ----------
+    file_obj : str, file-like, or HDUList
+        FITS file name, file object, or HDUList..
+    hdu : Optional[int]
+        Specified HDU to load.
+    visit : Optional[int]
+        Specified visit index to load.
+
+    Returns
+    -------
+    Spectrum1D
+        The spectrum contained in the file
+    """
+    with read_fileobj_or_hdulist(file_obj, memmap=False, **kwargs) as hdulist:
+        # Check if file is empty first
+        datasums = []
+        for i in range(1, len(hdulist)):
+            datasums.append(int(hdulist[i].header.get("DATASUM")))
+        if (np.array(datasums) == 0).all():
+            raise ValueError("Specified file is empty.")
+
+        # TODO: how should we handle this -- multiple things in file, but the user cannot choose.
+        if hdu is None:
+            for i in range(len(hdulist)):
+                if hdulist[i].header.get("DATASUM") != "0":
+                    hdu = i
+                    warnings.warn(
+                        'HDU not specified. Loading spectrum at (HDU{})'.
+                        format(i), AstropyUserWarning)
+                    break
+
+        # load spectra
+        spectra_list = _load_mwmVisit_or_mwmStar_hdu(hdulist, hdu)
+
+        # if mwmVisit AND visit not specified, load first visit in the HDU
+        if visit is None:
+            if spectra_list[0].meta['datatype'].lower() == 'mwmvisit':
+                warnings.warn(
+                    'Visit to load not specified. Loading spectrum at index 0. Specify with (visit=...)',
+                    AstropyUserWarning)
+            visit = 0
+        return spectra_list[visit]
+
+
+@data_loader(
+    "SDSS-V mwm",
+    identifier=mwm_identify,
     force=True,
     dtype=SpectrumList,
-    priority=10,
+    priority=20,
     extensions=["fits"],
 )
 def load_sdss_mwm_list(file_obj, **kwargs):
@@ -360,8 +510,8 @@ def load_sdss_mwm_list(file_obj, **kwargs):
 
     Returns
     -------
-    SpectrumList
-        The spectra contained in the file, each of 1D flux.
+    SpectrumList[Spectrum1D]
+        A list spectra from each visit with each instrument at each observatory (mwmVisit), or the coadd from each instrument/observatory (mwmStar).
     """
     spectra = SpectrumList()
     with read_fileobj_or_hdulist(file_obj, memmap=False, **kwargs) as hdulist:
@@ -370,7 +520,7 @@ def load_sdss_mwm_list(file_obj, **kwargs):
         for hdu in range(1, len(hdulist)):
             datasums.append(int(hdulist[hdu].header.get("DATASUM")))
         if (np.array(datasums) == 0).all():
-            raise RuntimeError("Specified file is empty.")
+            raise ValueError("Specified file is empty.")
 
         # Now load file
         for hdu in range(1, len(hdulist)):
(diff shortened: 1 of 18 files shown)


pllim:
Please add a change log.

@weaverba137 , would you be interested to review this or ping someone who would be?

Thanks!

specutils/io/default_loaders/sdss_v.py
```diff
@@ -347,9 +351,10 @@ def load_sdss_spec_1D(file_obj, *args, hdu: Optional[int] = None, **kwargs):
 
 
 @data_loader(
-    "SDSS-V spec multi",
+    "SDSS-V spec",
     identifier=spec_sdss5_identify,
     dtype=SpectrumList,
+    force=True,
```
havok2063, line 357: With the `multi` removed, what is the difference between this data loader and the above data loader at line 318?   Same with the `mwm` data loaders?  

With this removal, what would be the correct `format` to specify to load all data extensions?  Is it now the default?  Jdaviz only supports using `format` to during data load to provide hints to the type of filepath.

specutils/io/default_loaders/sdss_v.py
```diff
@@ -462,14 +467,17 @@ def load_sdss_mwm_1d(file_obj, hdu: Optional[int] = None, **kwargs):
             for i in range(len(hdulist)):
                 if hdulist[i].header.get("DATASUM") != "0":
                     hdu = i
+                    print('HDU not specified. Loading spectrum at (HDU{})'.
+                          format(i))
```
havok2063, line 471:
```suggestion
                    print(f'HDU not specified. Loading spectrum at (HDU{i}).')
```

specutils/io/default_loaders/sdss_v.py
```diff
@@ -462,14 +467,17 @@ def load_sdss_mwm_1d(file_obj, hdu: Optional[int] = None, **kwargs):
             for i in range(len(hdulist)):
                 if hdulist[i].header.get("DATASUM") != "0":
                     hdu = i
+                    print('HDU not specified. Loading spectrum at (HDU{})'.
```
havok2063, line 470: Is this just a warning or is specifying an HDU required?   For production, we use Jdaviz to load the file, and want by default all data extensions loaded, without having to specify an HDU, which I think will the most common use case by users as well.  Does this result in many constant warning messages printed to the user?

havok2063:
I think we might need the `multi` loader.  With this change, Specviz now fails to load all the extensions for the `mwmStar` file.  It now only loads the first data extension found.  
```
from jdaviz import Specviz
s = Specviz()
s.show()
spec.load_data("mwmStar-0.6.0-61731453.fits", format='SDSS-V mwm')
```
Before I could load all extensions with spec.load_data("mwmStar-0.6.0-61731453.fits", format='SDSS-V mwm multi')

I can load a `mwmVisit` file in specutils with `ss = Spectrum1D.read('mwmVisit-0.6.0-109635459.fits', format='SDSS-V mwm')`.  However, when I load into Specviz, it fails
```
s = Specviz()
s.show()
s.load_data(mwmVisit-0.6.0-109635459.fits, format='SDSS-V mwm')
```
with 
```
WARNING: UnitsWarning: The unit 'Angstrom' has been deprecated in the VOUnit standard. Suggested: 0.1nm. [astropy.units.format.utils]
WARNING:astropy:UnitsWarning: The unit 'Angstrom' has been deprecated in the VOUnit standard. Suggested: 0.1nm.
[ERROR]: Traceback (most recent call last):
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3577, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "/var/folders/sg/r2s67pm50tv6ydrrhs1c3rmw0000gn/T/ipykernel_88658/1622468986.py", line 1, in <module>
    spec.load_data(files[-7], format='SDSS-V mwm')
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/configs/specviz/helper.py", line 64, in load_data
    super().load_data(data,
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/helpers.py", line 110, in load_data
    self.app.load_data(data, parser_reference=parser_reference, **kwargs)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/app.py", line 828, in load_data
    parser(self, file_obj, **kwargs)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/configs/specviz/plugins/parsers.py", line 149, in specviz_spectrum1d_parser
    app.add_data_to_viewer(spectrum_viewer_reference_name, data_label[i])
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/app.py", line 1547, in add_data_to_viewer
    self.set_data_visibility(viewer_item, data_label, visible=visible, replace=clear_other_data)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/app.py", line 2092, in set_data_visibility
    self.hub.broadcast(add_data_message)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/glue/core/hub.py", line 215, in broadcast
    handler(message)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/app.py", line 427, in _on_add_data_message
    self._on_layers_changed(msg)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/app.py", line 573, in _on_layers_changed
    self.state.layer_icons = {
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/echo/core.py", line 261, in __setattr__
    super(HasCallbackProperties, self).__setattr__(attribute, value)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/echo/core.py", line 76, in __set__
    self.notify(instance, old, new)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/echo/core.py", line 125, in notify
    cback(new)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 3275, in <lambda>
    self.app.state.add_callback('layer_icons', lambda _: self._on_data_changed())
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 3433, in _on_data_changed
    self._apply_default_selection()
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 917, in _apply_default_selection
    self.selected = self.labels[0] if len(self.labels) else default_empty
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 624, in __setattr__
    return setattr(self._plugin, self._plugin_traitlets.get(attr), value)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 716, in __set__
    self.set(obj, value)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 706, in set
    obj._notify_trait(self.name, old_value, new_value)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 1513, in _notify_trait
    self.notify_change(
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/ipywidgets/widgets/widget.py", line 701, in notify_change
    super().notify_change(change)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 1525, in notify_change
    return self._notify_observers(change)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 1568, in _notify_observers
    c(event)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 2751, in _check_non_finite_uncertainty_mismatch
    spec = self._get_1d_spectrum()
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/configs/default/plugins/model_fitting/model_fitting.py", line 318, in _get_1d_spectrum
    return self.dataset.selected_spectrum_for_spatial_subset(self.spatial_subset_selected,
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 3347, in selected_spectrum_for_spatial_subset
    return self.plugin._specviz_helper.get_data(data_label=self.selected,
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/configs/specviz/helper.py", line 333, in get_data
    return self._get_data(data_label=data_label, spatial_subset=spatial_subset,
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/helpers.py", line 558, in _get_data
    data = data.get_object(cls=cls, **object_kwargs)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/glue/core/data.py", line 298, in get_object
    return handler.to_object(self, **kwargs)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/glue_astronomy/translators/nddata.py", line 148, in to_object
    attribute = _get_attribute(attribute, data)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/glue_astronomy/translators/nddata.py", line 23, in _get_attribute
    raise ValueError("Data object has more than one attribute, so "
ValueError: Data object has more than one attribute, so you will need to specify which one to use as the flux for the spectrum using the attribute= keyword argument.

ERROR:sdss_access:Traceback (most recent call last):
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3577, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "/var/folders/sg/r2s67pm50tv6ydrrhs1c3rmw0000gn/T/ipykernel_88658/1622468986.py", line 1, in <module>
    spec.load_data(files[-7], format='SDSS-V mwm')
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/configs/specviz/helper.py", line 64, in load_data
    super().load_data(data,
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/helpers.py", line 110, in load_data
    self.app.load_data(data, parser_reference=parser_reference, **kwargs)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/app.py", line 828, in load_data
    parser(self, file_obj, **kwargs)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/configs/specviz/plugins/parsers.py", line 149, in specviz_spectrum1d_parser
    app.add_data_to_viewer(spectrum_viewer_reference_name, data_label[i])
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/app.py", line 1547, in add_data_to_viewer
    self.set_data_visibility(viewer_item, data_label, visible=visible, replace=clear_other_data)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/app.py", line 2092, in set_data_visibility
    self.hub.broadcast(add_data_message)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/glue/core/hub.py", line 215, in broadcast
    handler(message)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/app.py", line 427, in _on_add_data_message
    self._on_layers_changed(msg)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/app.py", line 573, in _on_layers_changed
    self.state.layer_icons = {
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/echo/core.py", line 261, in __setattr__
    super(HasCallbackProperties, self).__setattr__(attribute, value)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/echo/core.py", line 76, in __set__
    self.notify(instance, old, new)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/echo/core.py", line 125, in notify
    cback(new)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 3275, in <lambda>
    self.app.state.add_callback('layer_icons', lambda _: self._on_data_changed())
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 3433, in _on_data_changed
    self._apply_default_selection()
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 917, in _apply_default_selection
    self.selected = self.labels[0] if len(self.labels) else default_empty
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 624, in __setattr__
    return setattr(self._plugin, self._plugin_traitlets.get(attr), value)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 716, in __set__
    self.set(obj, value)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 706, in set
    obj._notify_trait(self.name, old_value, new_value)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 1513, in _notify_trait
    self.notify_change(
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/ipywidgets/widgets/widget.py", line 701, in notify_change
    super().notify_change(change)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 1525, in notify_change
    return self._notify_observers(change)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 1568, in _notify_observers
    c(event)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 2751, in _check_non_finite_uncertainty_mismatch
    spec = self._get_1d_spectrum()
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/configs/default/plugins/model_fitting/model_fitting.py", line 318, in _get_1d_spectrum
    return self.dataset.selected_spectrum_for_spatial_subset(self.spatial_subset_selected,
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/template_mixin.py", line 3347, in selected_spectrum_for_spatial_subset
    return self.plugin._specviz_helper.get_data(data_label=self.selected,
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/configs/specviz/helper.py", line 333, in get_data
    return self._get_data(data_label=data_label, spatial_subset=spatial_subset,
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/jdaviz/core/helpers.py", line 558, in _get_data
    data = data.get_object(cls=cls, **object_kwargs)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/glue/core/data.py", line 298, in get_object
    return handler.to_object(self, **kwargs)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/glue_astronomy/translators/nddata.py", line 148, in to_object
    attribute = _get_attribute(attribute, data)
  File "/Users/brian/anaconda3/envs/valis_solara/lib/python3.10/site-packages/glue_astronomy/translators/nddata.py", line 23, in _get_attribute
    raise ValueError("Data object has more than one attribute, so "
ValueError: Data object has more than one attribute, so you will need to specify which one to use as the flux for the spectrum using the attribute= keyword argument.

HDU not specified. Loading spectrum at (HDU1)
```

specutils/io/default_loaders/sdss_v.py
rileythai: My goal with this was to simplify the interface. The difference previously was that `multi` loaded every valid HDU extension, whereas the single `SpectrumList` ones would load just a specified HDU. Since it's a `SpectrumList`, I felt the default loader for that type should just load every extension.

Across both the `Spectrum1D` and `SpectrumList` loaders there should be a single format (`SDSS-V [DATATYPE]`), which should be detected automagically and no longer needs to be manually specified on `SpectrumList`.

specutils/io/default_loaders/sdss_v.py
rileythai: I'm not sure how we want to handle this case generally. `mwm` files contain a max of 4 extensions which can each have N spectra. These shouldn't really be loaded into a single `Spectrum1D` object because APOGEE/BOSS are at different wavelength ranges.

havok2063: Yeah I agree that the APOGEE/BOSS extensions should not be loaded into a single Spectrum1D.  They should be loaded into a SpectrumList.  I think the warning is fine.  It's useful for anyone using `specutils` directly, and will be buried within Specviz and Zora.

rileythai:
> ValueError: Data object has more than one attribute, so you will need to specify which one to use as the flux for the spectrum using the attribute= keyword argument.

This is again partly that same issue we run into previously where jdaviz can't load `Spectrum1D` objects with 2D flux arrays. I've opened a `jdaviz` issue at https://github.com/spacetelescope/jdaviz/issues/3223 (I think I forgot to last time).

> With this change, Specviz now fails to load all the extensions for the mwmStar file. It now only loads the first data extension found.

The helper for `Specviz`'s default I/O will try and open it first as a `Spectrum1D`, and then it will try do it as a `SpectrumList` _only_ if it throws an `IORegistryError`. 

For both of these, I've been debating how to keep I/O simple without waiting for any jdaviz fixes, so I've done this:
1. There is a single SDSS-V loader for each datatype. It is always a `SpectrumList` because every SDSS-V file contains multiple spectra.
2. It always loads each spectra in that list as a `Spectrum1D` with 1D flux (bypass jdaviz issues and can be changed later if necessary/desired)
3. Users can no longer load `Spectrum1D` instances with the `hdu` specification (avoids the complexity of users having to know our datatypes to load as `Spectrum1D`; just send them to `SpectrumList` I/O)

This is how the DESI default loaders handle I/O, and this avoids any confusion on a user (student's) part who is unfamiliar with the SDSS-V datatype structure.

I'm not sure if this is exactly what we want for `specutils` though. Should it be that all files load only as `SpectrumList` objects if they have multiple spectra in them?

havok2063:
> The helper for `Specviz`'s default I/O will try and open it first as a `Spectrum1D`, and then it will try do it as a `SpectrumList` _only_ if it throws an `IORegistryError`. This means it'll default to a non-specified `hdu` keyword case and just load that first extension.

This is exactly the problem that was occurring, and why we need an explicitly different format for a list of spectra, until we can fix the Specviz data loader.  

> 1. There is a single SDSS-V loader for each datatype. It is always a `SpectrumList` because every SDSS-V file contains multiple spectra.

This is NOT true.  The majority of BHM spectra are single spectra.  Only the spec-full files contain multiple spectra for individual exposures, and most users do not use these files.    

> 3\. Users can no longer load `Spectrum1D` instances with the `hdu` specification (avoids the complexity of users having to know our datatypes to load as `Spectrum1D`; just send them to `SpectrumList` I/O)

IMO, we should avoid doing this, and I'm hesitant to remove it.  `Spectrum1D` is the primary user entry point for handling spectra.  `SpectrumList` is rarely mentioned in the `specutils` documentation, beyond that it exists, and for precisely dealing with [loading multiple spectra](https://specutils.readthedocs.io/en/stable/custom_loading.html#loading-multiple-spectra).    `Spectrum1D` will work for the majority of SDSS spectra, and `SpectrumList` will work for the Astra outputs.  

I agree that it'd be nice to have a single format for each SDSS data product.  I think it's ok to have users specify `hdu=x` to get a specific extension.  The default `hdu` value should be 1.  The following examples should be supported
```
ss = Spectrum1D.read("spec-lite.fits", format='SDSS-V spec') # loads the coadd spectrum in hdu 1
ss = Spectrum1D.read("spec-full.fits", format='SDSS-V spec', hdu=6) # loads the exposure spectrum in hdu 6
sl = SpectrumList.read("spec-full.fits", format='SDSS-V spec') # loads all the extensions 
ss = Spectrum1D.read("mwmStar.fits", format='SDSS-V mwm', hdu=3) # loads the apogee/apo spectrum in hdu 3
sl = SpectrumList.read("mwmStar.fits", format='SDSS-V mwm') # loads all extensions as a list of Spectrum1D
ss = Spectrum1D.read("mwmVisit.fits", format='SDSS-V mwm') # loads only the boss/apo spectrum in hdu 1
sl = SpectrumList.read("mwmVisit.fits", format='SDSS-V mwm') # loads all the extensions
```
For extensions with many spectra on the same wavelength grid, a single `Spectrum1D` object with flux `[n_spectra, n_wave]` is the correct format.   Until the Specviz loader can be fixed, you're current approach of unpacking any n-dimensional flux arrays into individual `Spectrum1D` objects within the `SpectrumList` is the right one here.  

It's not ideal, but if we also add back in the explicit `multi` on the format for SpectrumList, then, with your current fix, we could load into the current Specviz.  

> **edit:** with this change the `mwm` files can load in Jdaviz as well. @havok2063 , and the print warnings now go through the `AstropyUserWarning` class.

While indeed this does "fix" Specviz data loading, it does so at the expense of usability within the `specutils` package, and potentially other downstream applications.

specutils/io/default_loaders/sdss_v.py
```diff
@@ -578,28 +441,71 @@ def _load_mwmVisit_or_mwmStar_hdu(hdulist: HDUList, hdu: int, **kwargs):
...
+    return [
+        Spectrum1D(
+            spectral_axis=spectral_axis,
+            flux=flux[i],
+            uncertainty=e_flux[i],
+            mask=mask[i],
+            meta=metadicts[i],
+        ) for i in range(flux.shape[0])
+    ]
+
```
havok2063, line 474: we will need to do a similar thing for the `SDSS-V spec` loader, but only for any of the spec-full files.

specutils/io/default_loaders/sdss_v.py
rosteen: I'm planning to catch up on this this afternoon, I might have thoughts then.

specutils/io/default_loaders/sdss_v.py
rileythai: Aren't `spec-full` visits dispersed across the HDUs past the `spall` and `zall` HDU's? Or am I mistaken about the format?

rosteen:
> ```
> 
> ```
> 
> 
>     
>   
> 
> For extensions with many spectra on the same wavelength grid, a single `Spectrum1D` object with flux `[n_spectra, n_wave]` is the correct format.   Until the Specviz loader can be fixed, your current approach of unpacking any n-dimensional flux arrays into individual `Spectrum1D` objects within the `SpectrumList` is the right one here.
> 
> It's not ideal, but if we also add back in the explicit `multi` on the format for SpectrumList, then, with your current fix, we could load into the current Specviz.
> 
> > **edit:** with this change the `mwm` files can load in Jdaviz as well. @havok2063 , and the print warnings now go through the `AstropyUserWarning` class.
> 
> 
> While indeed this does "fix" Specviz data loading, it does so at the expense of usability within the `specutils` package, and potentially other downstream applications.

I agree with @havok2063's points here - we don't want to sacrifice `specutils` usability.

If my understanding is correct, at this point the main sticking point seems to be Specviz not handling 2D flux arrays. I [opened a PR](https://github.com/spacetelescope/jdaviz/pull/3229) in Jdaviz with a quick patch to do that, it simply unpacks the 2D array into separate `Spectrum1D`s before continuing to load the data. Is this sufficient for this use case for now? I see potential downsides, mainly that with a 2D array unpacked like this, there probably isn't a great way to attach the proper metadata about each of the 1D spectra (we could just copy the full `meta` to all of the split out spectra). That PR needs a change to also handle a 2D array coming from a file, I can add that if this seems like a direction that will enable your use case here.

specutils/io/default_loaders/sdss_v.py
havok2063: Yeah that's right.  They're appended at the end as 
```
5  MJD_EXP_60052-00    1 BinTableHDU    273   4648R x 8C   [E, E, E, J, J, E, E, E]   
6  MJD_EXP_60052-01    1 BinTableHDU    271   4648R x 8C   [E, E, E, J, J, E, E, E]   
7  MJD_EXP_60052-02    1 BinTableHDU    273   4648R x 8C   [E, E, E, J, J, E, E, E]   
8  MJD_EXP_60052-03    1 BinTableHDU    273   4648R x 8C   [E, E, E, J, J, E, E, E] 
...  
```
So, yeah, a change isn't needed here since they are 1d spectra in separate extensions.

havok2063:
> I'd like to keep the interface consistent -- if we output 1D flux here, we output it everywhere. So users should also have to specify the specific `visit` as well from the `mwm` interface for `mwmVisit` files. For example:
Spectrum1D.load('mwmVisit.fits', visit=2) # loads 3rd (indexing) visit in 1st HDU w/ data 
Spectrum1D.load('mwmVisit.fits', hdu=2, visit=2) # loads 2nd visit in HDU2

That's fine.  Although `Spectrum1D` is also meant to hold many spectra with the same spectral axis, so I don't think it's necessary to force the visit spectrum to be 1d.  I would say if someone did `Spectrum1D.load('mwmVisit.fits', hdu=2)` it loads all visits, and they can optionally specify `visit` to extract just a single visit.  

Does adding the `visit` keyword here make it available generally on all `Spectrum.load` calls, or just the visit specific data loader?   It may be confusing for users if `visit` is now an allowed keyword when loading hst or jwst spectra, for example. 

> I would really like to keep this change since it fixes a `specutils` side auto-detection bug: i.e. any other 3rd party code loads `SpectrumList` won't automatically work. I'm fine to remove it if it's an absolute blocker.

Then we'll also need to change the specviz loading logic at https://github.com/spacetelescope/jdaviz/blob/main/jdaviz/configs/specviz/plugins/parsers.py#L78-L87

> Is it possible in the case of `sdss_solara` to load them as `Spectrum` objects in the solara app itself and pass that object in?

Yeah it is, but I'd like to avoid that.  I'd prefer to pass the file and have `jdaviz` and `specutils` handle the data loading.  If we move that logic into the solara app, then users would need to do the same thing when loading data locally into jdaviz.  Which to me defeats the advantage of jdaviz/specutils.

havok2063:
@rosteen 
> If my understanding is correct, at this point the main sticking point seems to be Specviz not handling 2D flux arrays. I [opened a PR](https://github.com/spacetelescope/jdaviz/pull/3229) in Jdaviz with a quick patch to do that, it simply unpacks the 2D array into separate `Spectrum1D`s before continuing to load the data. Is this sufficient for this use case for now? 

Thanks! I will test this out.  If we use your PR, then would we likely need to rollback the equivalent changes here?

The other issue is the logic of the Specviz data loader when the input is a filepath, at https://github.com/spacetelescope/jdaviz/blob/main/jdaviz/configs/specviz/plugins/parsers.py#L78-L87.   It always tries Spectrum1D first and only on fail, does it try SpectrumList.  Previously we had loader formats `SDSS-V mwm` and `SDSS-V mwm multi` for explicitly loading data as `Spectrum1D` and `SpectrumList`, respectively.  This worked with Specviz because I could specify the `multi` format and trigger line 87.  This PR would consolidate the specutils formats into a single `SDSS-V mwm` format for both, which would break the loading since line 82 always gets called, since it's always using the Spectrum1D loader first.     

I'm not sure the right solution here.  Is there a preference in `specutils` for creating specifically named loader formats for Spectrum1D vs SpectrumList?    One option is to go back to different format names here.  Another option could be to allow `Specviz.load_data` to specify a `dtype` keyword to prefer loading with either Spectrum1D or SpectrumList?   

> I see potential downsides, mainly that with a 2D array unpacked like this, there probably isn't a great way to attach the proper metadata about each of the 1D spectra (we could just copy the full `meta` to all of the split out spectra). 

I don't have a strong opinion on the metadata.  I'd be fine copying the full dict.  I'm not aware of any visit-specific metadata but @rileythai might know.   @rileythai what are you currently copying over for metadata?

> That PR needs a change to also handle a 2D array coming from a file, I can add that if this seems like a direction that will enable your use case here.

Yeah our use case has the nd array coming from an extension in a file.

rosteen:
Hm, how about adding a `force_load_as_list` keyword to the Specviz loader that forces it to use the `SpectrumList` instead of `Spectrum1D` read function? That would be a minimal change on our end and I think would be sufficient to replace the `multi` specification.

havok2063:
> Hm, how about adding a `force_load_as_list` keyword to the Specviz loader that forces it to use the `SpectrumList` instead of `Spectrum1D` read function? That would be a minimal change on our end and I think would be sufficient to replace the `multi` specification.

That works for me!

rosteen:
Great, I implemented it in that PR as `load_as_list`. Once @rileythai confirms that my PR is working with this I'll move it out of draft to get reviewed.

rileythai:
> @rileythai what are you currently copying over for metadata?
I'm copying over some visit-specific stuff (i.e. SNR, MJD, telescope) that will let users.

@rosteen check [my PR for your PR](https://github.com/rosteen/jdaviz/pull/6), which has some fixes and simplifies the cases a bit. It works for our files.

In this case, when we keep the `Spectrum1D` functionality as originally implemented, this means that SDSS-V data loaded through the `jdaviz` solara instance (import data button) won't load all the spectra in the file, just the first one it comes across. See my comment about that at https://github.com/spacetelescope/jdaviz/pull/3229#issuecomment-2423483719

rileythai:
This PR should be good to go now @rosteen .

rosteen (approved):
I think this looks good now, but I'm going to wait until I merge https://github.com/spacetelescope/jdaviz/pull/3229 to merge it.

rosteen:
Just merged the downstream PR, I have one more bugfix I'd like to get in the next `specutils` release but I'll try to get this out soon.

PR #1190, opened 2024-11-04
Fix spline interpolation between different unit physical types

### LINKED ISSUE #1186: Spline interpolation no longer works for binning on velocity grid
Due to changes added in https://github.com/astropy/specutils/pull/1121, the SplineInterpolatedResampler is no longer suitable for binning a spectrum onto a spectral axis which is uniform in velocity space. Prior to this, the following code would work (e.g works in 1.12.0 but not in 1.17.0):

```python
rest_wavelength = 656.2 * u.nm

# make a fake spectrum
wavelengths = np.linspace(640, 672, 1000) * u.nm
flux = np.random.normal(1.0, 0.1, 1000) * u.mJy
spec1d = Spectrum1D(spectral_axis=wavelengths, velocity_convention="optical", flux=flux)
spec1d.spectral_axis.doppler_rest = rest_wavelength

velocities = np.linspace(-1000, 1000, 500) * u.km/u.s
velocity_grid = SpectralAxis(
        velocities, doppler_rest=rest_wavelength, doppler_convention="optical"
)
resample_fn = SplineInterpolatedResampler()
velocity_binned = resample_fn(spec1d, velocity_grid)
```

This now fails because the line below

https://github.com/astropy/specutils/blob/a859dd6c9b540b6f651a3b0d882be6b418aecde7/specutils/manipulation/resample.py#L465

assumes that the bin edges and the `fin_spec_axis` are both in wavelength units. The only resampler that works for this use case is the LinearInterpolatedResampler, but this is not accurate enough for velocity binning.

rosteen:
Fixes #1186. Part of this accidentally snuck into a commit on `main` while I was doing the last release 😬 (changing from index 0 to using min and max) but this completes the fix. @StuartLittlefair would you confirm that you get what you expected in #1186 with this fix?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,9 @@ Bug Fixes
 
 - Fixed automatic format detection for SDSS-V ``SpectrumList`` default loaders. [#1185]
 
+- Fixed ``SplineInterpolatedResampler`` when input and output spectral axes are different
+  physical types, e.g. wavelength and velocity. [#1190]
+
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


StuartLittlefair:
Thanks @rosteen - this is perfect

PR #1195, opened 2024-11-19
Use the APE 14 specified API in some of the tests

WilliamJamieson:
spacetelescope/gwcs#457 corrects an oversight in the APE 14 related API, where "low-level" inputs were accepted by the "high-level" api. This PR simply swaps the usage in specutils from using the "high-level" with "low-level" inputs to using the proper "low-level" API.

Also, I updated the tests so that files generated by the tests get cleaned up by pytest.

### DIFF
--- a/specutils/tests/test_spectrum1d.py
+++ b/specutils/tests/test_spectrum1d.py
@@ -318,7 +318,7 @@ def test_wcs_transformations():
     spec = Spectrum1D(spectral_axis=np.arange(1, 50) * u.nm,
                       flux=np.ones(49) * u.Jy)
 
-    pix_axis = spec.wcs.world_to_pixel(np.arange(20, 30) * u.nm)
+    pix_axis = spec.wcs.world_to_pixel_values(np.arange(20, 30) * u.nm)
     disp_axis = spec.wcs.pixel_to_world(np.arange(20, 30))
 
     assert isinstance(pix_axis, np.ndarray)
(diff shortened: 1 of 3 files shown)


pllim (approved):
Thanks for the clean up!

specutils/tests/test_spectrum1d.py
```diff
@@ -318,15 +318,15 @@ def test_wcs_transformations():
     spec = Spectrum1D(spectral_axis=np.arange(1, 50) * u.nm,
                       flux=np.ones(49) * u.Jy)
 
-    pix_axis = spec.wcs.world_to_pixel(np.arange(20, 30) * u.nm)
+    pix_axis = spec.wcs.world_to_pixel_values(np.arange(20, 30) * u.nm)
```
pllim, line 321: `FAILED ../../specutils/tests/test_spectrum1d.py::test_wcs_transformations - AttributeError: 'numpy.ndarray' object has no attribute 'to_value'`

I don't think this patch is correct?

WilliamJamieson: Yeah I'll have to adjust things. It's because units have been passed in.

WilliamJamieson: spacetelescope/gwcs#457 also corrects things so that the `*_values` can accept quantity inputs. GWCS was inconsistent about this. For now its better to just wrap it in a `SpectralCoord` object like it should be for the function as it is used in the test.

pllim:

specutils/tests/test_spectrum1d.py
```diff
@@ -318,15 +318,19 @@ def test_wcs_transformations():
     spec = Spectrum1D(spectral_axis=np.arange(1, 50) * u.nm,
                       flux=np.ones(49) * u.Jy)
 
-    pix_axis = spec.wcs.world_to_pixel(np.arange(20, 30) * u.nm)
+    # After spacetelescope/gwcs#457 is merged and released, this can be changed to
+    # pix_axis = spec.wcs.world_to_pixel_values(np.arange(20, 30) * u.nm)
+    pix_axis = spec.wcs.world_to_pixel(SpectralCoord(np.arange(20, 30) * u.nm))
```
pllim, line 323: This is fine. I think it makes sense for specutils to use high level API.

Issue #1201, opened 2024-12-04
Subaur-pfsObject data loader fails with the latest datamodel


monodera:
The `Subaru-pfsObject` data loader fails with the latest datamodel as follows.

```
from specutils import Spectrum1D

filename = "pfsObject-00009-00000-0,0-0000000000000067-002-0x60241444a9af118c.fits"

spec = Spectrum1D.read(filename, format="Subaru-pfsObject")
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 spec = Spectrum1D.read(filename, format="Subaru-pfsObject")

File ~/tmp/test_specutils/.venv/lib/python3.11/site-packages/astropy/nddata/mixins/ndio.py:59, in NDDataRead.__call__(self, *args, **kwargs)
     58 def __call__(self, *args, **kwargs):
---> 59     return self.registry.read(self._cls, *args, **kwargs)

File ~/tmp/test_specutils/.venv/lib/python3.11/site-packages/astropy/io/registry/core.py:221, in UnifiedInputRegistry.read(self, cls, format, cache, *args, **kwargs)
    218         kwargs.update({"filename": path})
    220 reader = self.get_reader(format, cls)
--> 221 data = reader(*args, **kwargs)
    223 if not isinstance(data, cls):
    224     # User has read with a subclass where only the parent class is
    225     # registered.  This returns the parent class, so try coercing
    226     # to desired subclass.
    227     try:

File ~/tmp/test_specutils/.venv/lib/python3.11/site-packages/specutils/io/default_loaders/subaru_pfs_spec.py:68, in pfs_spec_loader(file_obj, **kwargs)
     65 with read_fileobj_or_hdulist(file_obj, **kwargs) as hdulist:
     66     header = hdulist[0].header
     67     meta = {'header': header,
---> 68             'tract': m['tract'],
     69             'patch': m['patch'],
     70             'catId': m['catId'],
     71             'objId': m['objId'],
     72             'nVisit': m['nVisit'],
     73             'pfsVisitHash': m['pfsVisitHash']}
     75     # spectrum is in HDU 2
     76     data = hdulist[2].data['flux']

TypeError: 'NoneType' object is not subscriptable
```

This is due to significant updates on the datamodel (https://github.com/Subaru-PFS/datamodel/blob/244bdeacf0e062e13b75d8d541e962b52c22bffb/datamodel.txt#L866)

```
Combined spectra.

[...]

PR #1202, opened 2024-12-04
Update data loader to the latest PFS datamodel (as of Nov 2024)

monodera:
In the last 2yrs since the data loader for Subaru-pfsObject is created, significant changes has been made for the datamodel. Although it's still actively developed, I'd like to initiate updating the data loader.

Note that I didn't keep the old loader, so no backward compatibility is maintained. 

Unfortunately, at this moment, the actual pfsObject file is still proprietary, but I would like to share an example file once it's publicly available (openuse operation will start in early 2025).

The changes also include code formatting by black.

Reference:
- PFS datamodel: https://github.com/Subaru-PFS/datamodel/blob/master/datamodel.txt

### DIFF
--- a/specutils/io/default_loaders/subaru_pfs_spec.py
+++ b/specutils/io/default_loaders/subaru_pfs_spec.py
@@ -3,27 +3,26 @@ Loader for PFS spectrum files.
 
 https://github.com/Subaru-PFS/datamodel/blob/master/datamodel.txt
 """
+
 import os
 import re
 
-from astropy.units import Unit
-from astropy.nddata import StdDevUncertainty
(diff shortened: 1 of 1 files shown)


monodera:
Related issue: #1201

pllim:
Thank you! 

While this does need a change log in `CHANGES.rst`, that can probably wait until you are ready to take this PR out of draft for final review.

pllim:
CI failure about "Fits" deprecation warning is unrelated and can be ignored for now, though that also means you would have to rebase this PR in the future when it is fixed upstream. FYI.

PR #1204, opened 2024-12-12
Fix for gwcs inverse transform with a bounding_box

nden:
This PR fixes a test where the input passed to the inverse transform is out of bounds. It should be merged after gwcs#498.
More specifically the footprint defined by the test WCS is [1, 49]*u.nm. 
The input passed is in GHz and well outside this footprint. As a result the Tabular model in the WCS errors with "out of bounds" . This PR adjust the input values to be within the footprint.
As a side note, I found it is best to define Tabular model with `bounds_error=False, fill_value=np.nan`. In this case an error is not raised and the value returned when out of bounds is NaN which is more in line with the WCS philosophy.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,7 +19,7 @@ install_requires =
     numpy>=1.24
     scipy>=1.3
     astropy>=5.1
-    gwcs>=0.18
+    gwcs @ git+https://github.com/nden/gwcs.git@inverse-bbox
     asdf-astropy>=0.3
     asdf>=2.14.4
     ndcube>=2.0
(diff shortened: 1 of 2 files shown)


rosteen:
Thanks - you're planning to update the pin from your branch to 0.22 once that's released, right? Probably should keep this as draft until then. @nden do you have any idea when 0.22 might release? We could really use a release that fixes the FITS deprecation error being raised in CI by GWCS due to astropy 7.0.

setup.cfg
```diff
@@ -19,7 +19,7 @@ install_requires =
     numpy>=1.24
     scipy>=1.3
     astropy>=5.1
-    gwcs>=0.18
+    gwcs @ git+https://github.com/nden/gwcs.git@inverse-bbox
```
pllim, line 22: Is this a PR on gwcs?

rosteen: This is what I was asking about above as well.

pllim: Oh I think it is https://github.com/spacetelescope/gwcs/pull/498 that is already merged

pllim:
Cleaned up version in https://github.com/astropy/specutils/pull/1207

PR #1211, opened 2025-02-04
Refactor gwcs_from_array to provide ND GWCS in ND flux case

rosteen:
Currently, `gwcs_from_array` always gives a 1D purely spectral GWCS, even in the case where the flux array is multi-dimensional. This PR updates that function to return an ND GWCS, so that the dimensionality of a `Spectrum`'s GWCS always matches that of the flux array, even if the spatial dimensions are simply returning the pixel values. This will simplify and solve a lot of problems down stream in e.g. `glue-astronomy` and `jdaviz`.

There were a lot of changes here that were needed to accommodate this change, but I think it's all improvements. I have one final failing test that I need to debug, so I'm opening this as a draft while I resolve that.

### DIFF
--- a/specutils/tests/test_arithmetic.py
+++ b/specutils/tests/test_arithmetic.py
@@ -1,6 +1,7 @@
 import astropy.units as u
 from astropy.tests.helper import assert_quantity_allclose
 import numpy as np
+import pytest
 
 from ..spectra.spectrum import Spectrum
 
@@ -89,8 +90,9 @@ def test_multiplication_basic_spectra(simulated_spectra):
 
 def test_add_diff_spectral_axis(simulated_spectra):
 
-    # Calculate using the spectrum/nddata code
-    spec3 = simulated_spectra.s1_um_mJy_e1 + simulated_spectra.s1_AA_mJy_e3  # noqa
+    # We now raise an error if the spectra aren't on the same spectral axis
+    with pytest.raises(ValueError, match="Spectral axis of both operands must match"):
+        spec3 = simulated_spectra.s1_um_mJy_e1 + simulated_spectra.s1_AA_mJy_e3  # noqa
 
 
 def test_masks(simulated_spectra):
--- a/specutils/spectra/spectrum.py
+++ b/specutils/spectra/spectrum.py
@@ -832,14 +832,16 @@ class Spectrum(OneDSpectrumMixin, NDCube, NDIOMixin, NDArithmeticMixin):
     def _other_as_correct_class(self, other):
         # NDArithmetic mixin will try to turn other into a Spectrum, which will fail
         # sometimes because of not specifiying the spectral axis index
-        flux = u.Quantity(other, unit=self.unit)
-        if flux.ndim > 1 and flux.shape == self.shape:
-            return Spectrum(flux=flux, spectral_axis=self.spectral_axis,
+        if not isinstance(other, u.Quantity):
+            other = u.Quantity(other, unit=self.unit)
+        if other.ndim > 1 and other.shape == self.shape:
+            print("Making other operator a Spectrum")
+            return Spectrum(flux=other, spectral_axis=self.spectral_axis,
                             spectral_axis_index=self.spectral_axis_index)
-        return flux
+        return other
 
     def __add__(self, other):
-        if not isinstance(other, (NDCube, u.Quantity)):
+        if not isinstance(other, (Spectrum)):
             try:
                 other = self._other_as_correct_class(other)
             except TypeError:
@@ -848,7 +850,7 @@ class Spectrum(OneDSpectrumMixin, NDCube, NDIOMixin, NDArithmeticMixin):
         return self._return_with_redshift(self.add(other))
 
     def __sub__(self, other):
-        if not isinstance(other, NDCube):
+        if not isinstance(other, Spectrum):
             try:
                 other = self._other_as_correct_class(other)
             except TypeError:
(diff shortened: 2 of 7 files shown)


docs/spectrum.rst
```diff
@@ -406,9 +406,6 @@ spectral axis, or 'spatial', which will collapse along all non-spectral axes.
     >>> spec.mean(axis='spatial')  # doctest: +FLOAT_CMP
     <Spectrum(flux=<Quantity [0.37273938, 0.53843905, 0.61351648, 0.57311623, 0.44339915,
                0.66084728, 0.45881921, 0.38715911, 0.39967185, 0.53257671] Jy> (shape=(10,), mean=0.49803 Jy); spectral_axis=<SpectralAxis
-       (observer to target:
-          radial_velocity=0.0 km / s
-          redshift=0.0)
```
eteq, line 411: Why was this removed?  Is it an incidental thing you noticed at the time (no prob if so) or a result of this change somehow?

specutils/tests/test_arithmetic.py
```diff
@@ -89,8 +90,9 @@ def test_multiplication_basic_spectra(simulated_spectra):
 
 def test_add_diff_spectral_axis(simulated_spectra):
 
-    # Calculate using the spectrum/nddata code
-    spec3 = simulated_spectra.s1_um_mJy_e1 + simulated_spectra.s1_AA_mJy_e3  # noqa
+    # We now raise an error if the spectra aren't on the same spectral axis
+    with pytest.raises(ValueError, match="Spectral axis of both operands must match"):
+        spec3 = simulated_spectra.s1_um_mJy_e1 + simulated_spectra.s1_AA_mJy_e3  # noqa
```
eteq, line 95: I'm confused by this - I would have thought this failed previously - i.e. the shapes did *not* change.  So is this also an incidental improvement that comes from these changes that's not strictly about the gwcs?

If so, that's good, but should be in the changelog as an API change I think (since users might encounter this as an error where they didn't before)

eteq (commented):
I have a few minor comments but after taking a first look I realized one major concern/confusion so I'm going to start there and give another more thorough review depending on the outcome.

Specifically, it looks to me like the outcome of this change is essentially breaking the conceptual "link" between the `spectral_axis` and the `wcs`.  What I mean is that in my head the way Spectrum1D is *supposed* to work is exactly as shown in this diagram:

https://specutils.readthedocs.io/en/latest/_images/specutils_classes_diagrams.png

in that diagram, the spectral_axis is basically just exactly what you'd get if you do `spectrum.wcs.pixel_to_world(all_of_the_pixels_in_the_spectrum)`.  (I know that's not literally how it's implemented but in my mind the API "contract" is that this is so) This would seem to break that link because now the dimensionality of `wcs` and `spectral_axis` are different.  Is that the intent?  If so, I have to ask: why not instead use `SpectrumCollection` for use cases where that's what's needed?

rosteen:
> in that diagram, the spectral_axis is basically just exactly what you'd get if you do `spectrum.wcs.pixel_to_world(all_of_the_pixels_in_the_spectrum)`. (I know that's not literally how it's implemented but in my mind the API "contract" is that this is so) This would seem to break that link because now the dimensionality of `wcs` and `spectral_axis` are different. Is that the intent? If so, I have to ask: why not instead use `SpectrumCollection` for use cases where that's what's needed?

This is already not the case for things like JWST cubes, which provide a full 3D GWCS solution that encodes both the spectral axis and spatial coordinates. The issue is that `specutils` is inconsistent in whether a multidimensional spectral cube can be expected to have a WCS/GWCS of matching dimensionality (when you're reading a FITS file, generally) or a 1D GWCS (the case where you're manually creating a `Spectrum1D` with just a `spectral_axis` specified). The latter case causes all sorts of problems with downstream packages that expect the dimensionality of a WCS/GWCS to match the dimensionality of the data - for example there is an ugly work-around in `glue-astronomy` to pad out the WCS in this case with dummy dimensions (similar to what I'm implementing here) so that Glue doesn't choke on the dimensionality mismatch when we load something further downstream in Jdaviz.

My goal here is to bring the case where we generate a dummy GWCS for a spectrum created with a `spectral_axis` rather than `wcs` specified in line with the (reasonable, IMO) expectation that a multidimensional flux array would have a multidimensional WCS/GWCS. This implementation does that by just passing pixel values through as the spatial coordinates, which should allow a lot of code simplification downstream in a couple packages. This does mean that the diagram you referenced isn't accurate and should be updated, but as I said, it already wasn't really accurate for any multidimensional `Spectrum1D` created by reading a FITS file, where the dimensionality of the WCS _already_ matches the flux array, not the spectral axis.

specutils/tests/test_arithmetic.py
rosteen: Agreed that this should be in the changelog too. Previously spectrum arithmetic didn't care if the spectral axis values were different, only that the shape of the arrays were the same. It now checks to see that the values actually match as well.

docs/spectrum.rst
rosteen: I don't recall, I'll double check.

rosteen: I think this was due to creating the new collapsed `Spectrum` with `self.spectral_axis` rather than `self.wcs` now, and is actually probably more correct than the previously printed out `repr`, which was replacing `None` values for these quantities with `0` in the process of collapsing (implying information we actually don't have).

rosteen:
@eteq I'm going to go ahead and merge this into the v2.0-dev branch so I can release an RC for people to start testing 2.0. I'll start a discussion issue where people can post concerns about 2.0 to address before an actual release, we can continue the discussion of this here or take it to that issue if my replies didn't alleviate your concerns.

PR #1222, opened 2025-03-27
Address problems found in specreduce compatibility check

rosteen:
https://github.com/astropy/specreduce/pull/260 exposed a couple of errors that needed addressing here. I reworked the arithmetic updates slightly to take advantage of `NDDataArray`'s mask and uncertainty handling by operating on the operand fluxes and then reinstantiating a new `Spectrum` afterwards, since we can't pass `spectral_axis_index` through the astropy arithmetic machinery. This also fixes handling for `pixel` unit spectral axes in the updated `gwcs_from_array` function to fix an error that was being raised when trying to instantiate `SpectralCoord` with `pixel` units.

Opening as draft, need to do something similar to what I did for subtraction with the other arithmetic operations.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,7 +19,7 @@ Other Changes and Additions
 
 - Initializing a ``Spectrum`` with only a spectral axis (not full WCS) will now
   result in a GWCS matching the dimensionality of the flux array, rather than a
-  1D spectral GWCS in all cases. [#1211]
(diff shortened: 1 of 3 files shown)


rosteen:
Going to go ahead and merge this so we can make the associated updates in https://github.com/astropy/specreduce/pull/260 and see tests pass there.

PR #1228, opened 2025-04-22
Fix empty returned list of template chi2 values

### LINKED ISSUE #1227: List of chi2 values from template_match is empty
I'm using the template_match function and it appears to work very well. However the chi2_list output parameter, which according to the docs is "A list with all chi2 values found for each template spectrum", is returned as an empty list. Is this a known issue? Would be useful to be able to recover the full set of chi2 values for all templates.

rosteen:
Closes #1227

Thanks @skendrew for filing a bug ticket for this, looks like it was just an oversight in the code not appending to the list in one case. Despite the comment in the test saying it was checking for this, it actually wasn't, so I added a check in the test as well.


rosteen:
I think this is uncontroversial enough that I'll just merge it.

PR #1233, opened 2025-05-02
Improvements to the HST/STIS data loaders

sean-lockwood:
### Echelle datasets
HST/STIS can have multivalued wavelengths where echelle orders overlap.  Simply sorting by
wavelength will interleave high- and low-SNR data near the edges of these orders.  Rather
than solving the complex task of combining orders into a single 1D spectrum, we created
a second loader, "HST/STIS multi", to handle loading individual orders into rows of a
SpectrumCollection object.

Orders (each with independent wavelength solutions) overlap, so simply sorting by wavelength and combining all orders leads to non-physical Δλ.  Here's an example showing the overlaps:
image

The edges of orders have especially variable SNR (mainly due to the blaze function shape), so interleaving in the overlap regions would be non-physical:
image

### Multi-read datasets
Additionally, some STIS datasets contain multiple reads (i.e. when REPEATOBS > 1 or
OCRREJECT="OMIT").  These reads are stored as additional FITS extensions.  Any multi-read
dataset also produces a SpectrumCollection object via the "HST/STIS multi" reader.

### Metadata
We supplemented the meta information with the relevant SCI extension header, as well as
the scalar table values describing the 1D extraction parameters for each order.

### Data masking
We used the STIS data quality (DQ) array, masked via bitwise-and with an SDQFLAGS
(serious DQ flags) scalar value to produce a boolean mask.  STIS DQ values are
integers, allowing the user to choose which DQ flags are relevant to their analysis via
the selection of the SDQFLAGS value.  By default, this value is read from the relevant
SCI extension header.  We also provided the capability for users to override this value
by specifying "sdqflags=<int>" to the STIS data readers.

### Caveat about echelle wavelength ordering
Note that while the wavelengths of any individual STIS echelle order are in ascending
ordered, the order of the wavelengths across echelle orders is descending (corresponding
to increasing SPORDER number).  This was chosen to remain consistent with the source
data.

### Flux units
While "Unit('erg/cm**2 Angstrom s')" appears to evaluate as intended, we
used a more explicit formulation.

### Identifiers
We modified the identifiers to not raise an exception upon the non-detection
of a FITS header keyword.

### Testing
Updated HST/STIS test_loaders tests.
New test data covering different file dimensionality are available at:
https://zenodo.org/records/15320389

### Updated `CHANGES.rst`

### DIFF
--- a/specutils/io/default_loaders/hst_stis.py
+++ b/specutils/io/default_loaders/hst_stis.py
@@ -1,28 +1,54 @@
-from astropy.units import Unit
+from copy import deepcopy
+import numpy as np
+import astropy.units as u
 from astropy.nddata import StdDevUncertainty
 
-from ...spectra import Spectrum1D
+from ...spectra import Spectrum1D, SpectrumCollection
 from ..registers import data_loader
 from ..parsing_utils import read_fileobj_or_hdulist
 
-__all__ = ['stis_identify', 'stis_spectrum_loader']
+__all__ = ['identify_stis_1storder_and_singleread',
+           'identify_stis_multiorder_or_multiread',
+           'stis_single_spectrum_loader',
+           'stis_multi_spectrum_loader',]
+
+SDQFLAGS_DEFAULT = 31743  # = 0b11110111111111
+SCALAR_COLUMNS = ['SPORDER', 'NELEM', 'A2CENTER', 'EXTRSIZE', 'MAXSRCH', 'BK1SIZE',
+                  'BK2SIZE', 'BK1OFFST', 'BK2OFFST', 'OFFSET',]
+FLUX_UNIT = u.erg / (u.s * u.cm**2 * u.Angstrom)
+DISP_UNIT = u.Angstrom
 
 
-def stis_identify(origin, *args, **kwargs):
-    """Check whether given file contains HST/STIS spectral data."""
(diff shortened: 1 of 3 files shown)


specutils/io/default_loaders/hst_stis.py
```diff
@@ -1,57 +1,175 @@
...
+
+        multi_spec = _construct_Spectrum1D_list(hdulist, sdqflags=sdqflags)
+
+    return SpectrumCollection.from_spectra(multi_spec)
+
+
+def _construct_Spectrum1D_list(hdulist, sdqflags=None):
+    """Construct a list of Spectrum1D objects from an HST/STIS FITS HDUList.
+    """
+    meta = {'header_primary': hdulist[0].header}
```
sean-lockwood, line 134: I just noticed that the `IRAF` FITS write assumes the primary header is included as "header".  Should I change this?

sean-lockwood: Will do.

There can be some discrepancy between the primary and SCI headers for multi-read STIS observation.  E.g., the start time listed in the primary header won't match the second+ SCI read, primary header's exposure time is the sum of the SCI exposures.  I think we'd need complex logic to combine them, so I'm happy keeping them separate.

rosteen (approved):

PR #1234, opened 2025-05-02
identify-tabular-fits more strict JWST test

kelle:
Currently, `tabular-fits` won't identify on any spectrum with TELESCOP=JWST.  This is far too broad since folks are re-reducing JWST data and outputting tabular-fits format spectra.  I've modified the identifier to only ignore JWST spectra with specific HDU names that have their own JWST loaders.

### DIFF
--- a/specutils/tests/test_loaders.py
+++ b/specutils/tests/test_loaders.py
@@ -870,6 +870,7 @@ def test_tabular_fits_compressed(compress, tmp_path):
     assert quantity_allclose(spec.flux, spectrum.flux)
 
 
+@pytest.mark.remote_data
 def test_tabular_fits_jwst():
     """Test reading of tabular FITS file of a JWST spectrum."""
     spec = Spectrum.read("https://bdnyc.s3.us-east-1.amazonaws.com/Beiler24/SDSSpJ1346-00NIR.fits")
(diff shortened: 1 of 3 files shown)


kelle:
I'm not quite sure where to write the test. In the meantime, here's a test spectrum which should be auto-identified by `tabular-fits`:
[CWISEPJ1446-23NIR.fits.zip](https://github.com/user-attachments/files/20094587/CWISEPJ1446-23NIR.fits.zip)

specutils/tests/test_loaders.py
```diff
@@ -870,6 +870,17 @@ def test_tabular_fits_compressed(compress, tmp_path):
     assert quantity_allclose(spec.flux, spectrum.flux)
 
 
+def test_tabular_fits_jwst():
```
keflavich, line 873: add the `@pytest.mark.remote_data` tag here

specutils/tests/test_loaders.py
```diff
@@ -870,6 +870,17 @@ def test_tabular_fits_compressed(compress, tmp_path):
...
+    """Test reading of tabular FITS file of a JWST spectrum."""
+    spec = Spectrum.read("https://bdnyc.s3.us-east-1.amazonaws.com/Beiler24/SDSSpJ1346-00NIR.fits")
```
keflavich, line 875: this spectrum has an uncertainty column; want to check that that gets loaded too?

kelle:
Edits made based on review! please take another look.

PR #1236, opened 2025-05-05
Use simple alias for future classname instead of subclass

### LINKED ISSUE #1235: bug in .spectral_axis in v1.20?
I am getting a weird error with v1.20 when I try to access the spectral_axis:
```Python
>>> spectrum = Spectrum.read(spectrum_path, format='tabular-fits')
WARNING: Standard Deviation has values of 0 or less [specutils.io.parsing_utils]
>>> spectrum.flux.unit
Unit("erg / (Angstrom s cm2)")
>>> spectrum.spectral_axis
Traceback (most recent call last):
  File "<python-input-10>", line 1, in <module>
    spectrum.spectral_axis
  File "/Users/kelle/miniforge3/envs/simple-db/lib/python3.13/site-packages/specutils/spectra/spectrum_mixin.py", line 62, in spectral_axis
    return self._spectral_axis
           ^^^^^^^^^^^^^^^^^^^
AttributeError: 'Spectrum' object has no attribute '_spectral_axis'. Did you mean: 'spectral_axis'?
```
I downgraded to v1.19 and this error disappears.  Not sure what's going on.

rosteen:
Fixes #1235. I hadn't fully considered the ramifications of subclassing `Spectrum1D` for `Spectrum`, specifically that private variables don't get inherited in the subclass. It seems like simply aliasing `Spectrum1D` to `Spectrum` is the way to go for enabling the new 2.x name here. I also added a test to make sure that accessing the spectral axis actually works, which I should have done to begin with.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,11 +1,11 @@
-1.21.0 (unreleased)
+1.20.1 (unreleased)
 -------------------
-
-New Features
-^^^^^^^^^^^^
 
 Bug Fixes
 ^^^^^^^^^
+
+- Switch to using an alias rather than a subclass for the temporary 1.x access
+  to the new 2.x class name (``Spectrum``). [#1236]
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


rosteen:

PR #1244, opened 2025-06-27
Allow specifying flux column in JWST x1d reader

### LINKED ISSUE #1144: Add flux column keyword to Spectrum1D.read
Follow-up from #1143 : it would be nice in `Spectrum1D.read()` to have a keyword specifying that you want the `flux` column rather than `surface brightness` even if the `SRCTYPE` is "EXTENDED" for JWST (and probably other) data, and vice-versa.

rosteen:
Closes #1144 . Unfortunately it's not possible to do this across all readers simultaneously, so I implemented it in the JWST reader since that's the one that is causing problems. If this comes up for other missions we can add the keyword to those as well.

### DIFF
--- a/specutils/io/default_loaders/jwst_reader.py
+++ b/specutils/io/default_loaders/jwst_reader.py
@@ -284,7 +284,7 @@ def jwst_x1d_miri_mrs_loader(input, missing="raise", **kwargs):
     return SpectrumList(chain.from_iterable(spectra))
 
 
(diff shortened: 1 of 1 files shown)


kelle:
I'm not sure if this is asking too much but can we get a release with this fix please?

rosteen:
> I'm not sure if this is asking too much but can we get a release with this fix please?

I'll try to get a release out later this week, since there have been a couple PRs merged.

PR #1253, opened 2025-07-18
Updates the sdss mwmStar loader to fix bug with file datasums

havok2063:
This PR updates the SDSS-V mwm data loaders to account for a bug in a subset of the DR19 SDSS-V mwmStar files.  A subset of files have extensions with empty data arrays and non-zero DATASUM keywords.  This was due to some data needing to be cleaned out but the header wasn't fully scrubbed/corrected.  

The previous loader would attempt to load first extension found with non-zero DATASUM and would crash on the empty data.  This fix allows the loader to correctly skip over it and load the first real data extension it finds. 

It would be great if this can be backported to `specutils 1.x`, so it can be used currently in Jdaviz.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,12 @@ Other Changes and Additions
 - Updated the JWST data readers to accomodate new WFSS file formats
   with ``SpectrumList.read()``. [#1252]
 
+Bug Fixes
+^^^^^^^^^
+
+- Fixed/updated SDSS-V mwm data loader to account for a subset of targets with bad datasums. [#1253]
+
+
 2.0.0 (2025-06-12)
 ------------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,17 +11,16 @@ New Features
   solutions; at the same time the IRAF equispec format gains the ability to read log-linear
   wavelength solutions. [#1254]
 
-Other Changes and Additions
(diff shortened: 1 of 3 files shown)


rosteen:
Thanks for the contribution @havok2063, unfortunately it looks like there's a real test failure. Could you address that, please? Once we get this in I can backport and do a bugfix release, although with the next Jdaviz release we'll be using specutils 2 there so it won't be needed for too long.

rosteen:
> @rosteen Are you referring to the Python 3.12 Windows run? Unfortunately, I don't have a windows machine to test on. I'll see if I can run the action locally. I just pushed fixes for the Code style checks. ~Do I need to worry about the rtd or release action failures?~ (looks like they are passing now)

Yup, I was talking about the Windows run, looks like the others were transient problems with remote data. I have a Windows machine at home I can try debugging on tomorrow or Friday, unfortunately I just went into the office for the rest of the day today.

rosteen:
I was able to reproduce the test failure locally, it's due to `arr.itemsize` being 0 at https://github.com/astropy/astropy/blob/412db1e64cca4e4d05cd373dc3cbe12ff2d6a257/astropy/io/fits/util.py#L589. That code is working around a bug but is 11 years old, so I wonder if its still relevant. You also may be able to change the test to work around it but I'm don't 100% understand the datasum issue to change it myself. I'm trying to figure out on my Windows machine if that `fits` code is even needed anymore...

Issue #229, opened 2018-06-12
Read in IRAF FITS format (including multispec)


andycasey:
👋 

I have [some code](https://github.com/andycasey/smhr/blob/master/smh/specutils/spectrum.py#L838-L995) (that @alexji and I maintain) that reads in IRAF FITS format of various types (chebyshev, legendre, polynomial, etc) that I'd be happy to update and create a pull request to bring it into the astropy ecosystem. It handles various dispersion mappings and multispec formats.

Would this be useful? If so, am I right in assuming that I should introduce it as an adapter in the specutils/wcs?

nmearl:
Hey @andycasey! We do have an implementation for reading IRAF FITS format [in the previous specutils iteration](http://specutils.readthedocs.io/en/doc-testing/_modules/specutils/io/read_fits.html#read_fits_spectrum1d), although it has not been brought over to the new specutils yet.

If you're interested, I would encourage you to see if merging the two implementations to perhaps improve upon what we both have is possible.

You would not need to create a new wcs adapter, as it's still a FITS file, and should follow the FITS wcs paradigm.

crawfordsm:
I think @simontorres was also working on this in #192

simontorres:
Yes, what I have done is mostly for reading non-linear wavelength solutions.

travegre:
Hi all,

I was wondering if there is already a way (an implementation) to read multispec IRAF spectra with the new astropy/specutils versions?

simontorres:
I implemented some of the features some time ago but I'm not sure how it's working now or whether it survived all the recent changes in the package or maybe someone improved it.

Either way I must say it's tricky, at least that was my experience.

kelle:
Hey Andy,  do you by chance have a suite of test data which we could use to kick the tires of specutils current functionality for reading in these various types of IRAF FITS files?

andycasey:
Hi @kelle, this issue thread is a blast from the past! I can dig up some data for Magellan/MIKE, and other sources with various levels of effort. I'll try to do this by the end of next week.

PR #1254, opened 2025-07-20
Add linear and loglinear wavelength solutions to (IRAF) fits-wcs

hamogu:
This adds linear and log-linear solutions to IRAF MULTISPEC format. Since the implementation is shares with the simpler IRAF formats (which had only linear, but not log-linear before), they gain the log-linear solutions as well.

This is one step towards #229. The MUTLISPEC format is often used for echelle spectra and linear solutions for that are common in the wild (e.g. KECK HIRES pipeline).

### DIFF
--- a/specutils/io/default_loaders/tests/test_wcs_fits.py
+++ b/specutils/io/default_loaders/tests/test_wcs_fits.py
@@ -1,0 +1,67 @@
+from astropy.io.fits import Header
+import pytest
+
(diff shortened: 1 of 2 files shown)


rosteen:
Thanks for the contribution, I haven't gotten a chance to look at the code yet but I'll review and get this in before the next release (hopefully later this week or early next week).

rosteen (approved):
Thanks for the contribution, I rebased to main and removed one more reference to log-linear not being implemented.

Issue #1262, opened 2025-08-02
extract_region should return a sliced WCS when possible


keflavich:
In extract region, there is a line:
```
        # NOTE: WCS is intentionally dropped, which will then fallback on lookup
```

This is unexpected behavior: the type of the underlying WCS has changed and all metadata lost.  In order to make a slice preserving the rest frequency, for example, I had to do the following:

```python
    spectrum = Spectrum(flux=u.Quantity(data), wcs=wcs, meta=meta, velocity_convention='radio')
    def v_to_f(v):
        return spectrum.frequency[np.argmin(np.abs(spectrum.velocity - v))]

    left_pix, right_pix = _subregion_to_edge_pixels(SpectralRegion(v_to_f(-500*u.km/u.s), v_to_f(500*u.km/u.s)).subregions[0], spectrum)

    spectrum = Spectrum(flux=spectrum.flux[left_pix:right_pix], wcs=spectrum.wcs[left_pix:right_pix], meta=spectrum.meta, velocity_convention='radio')
```

I expected to simply do:
```python
spectrum = specutils.manipulation.extract_region(spectrum, SpectralRegion(-500*u.km/u.s, 500*u.km/u.s))
```

rosteen:
Closed by #1263

Issue #1261, opened 2025-08-02
Region slicing should work with non-native units


keflavich:
Attempting to slice a Spectrum defined with frequency units, for example, with a velocity region results in an error:

```
UnitConversionError: 'km / s' (speed/velocity) and 'Hz' (frequency) are not convertible
````

Given that `spectrum.velocity` is defined in this case, spectralregions should be able to slice in velocity space.

PR #1263, opened 2025-08-07
extract_region: preserve WCS for single regions, add velocity-region support

harry353:
This update addresses issues #1261 and #1262, changing how the `extract_region` function behaves in two ways:

* When `preserve_wcs=True`, the WCS is now kept for single-region extractions.
* Velocity regions can now be extracted from spectra with a frequency axis (and the other way around).

It also includes a new test script to cover these cases.

### DIFF
--- a/specutils/manipulation/extract_spectral_region.py
+++ b/specutils/manipulation/extract_spectral_region.py
@@ -8,7 +8,7 @@ from ..spectra import Spectrum, SpectralRegion
 __all__ = ['extract_region', 'extract_bounding_spectral_region', 'spectral_slab']
 
 
-def _edge_value_to_pixel(edge_value, spectrum, order, side, axis=None):    
+def _edge_value_to_pixel(edge_value, spectrum, order, side, axis=None):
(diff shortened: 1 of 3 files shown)


keflavich (approved):

specutils/manipulation/extract_spectral_region.py
```diff
@@ -128,6 +139,10 @@ def extract_region(spectrum, region, return_single_spectrum=False):
         instead of multiple `~specutils.Spectrum` objects.  The returned spectrum
         will be a unique, concatenated, spectrum of all sub-regions.
 
+    preserve_wcs: `bool`
+        If True, the WCS will be adjusted and retained in the output spectrum(s).
+        If False (default), WCS will be dropped.
```
rosteen, line 144:
```suggestion
        If False (default), input WCS will be dropped and replaced by a lookuptable WCS.
```

rosteen (approved):
Approved, thanks!

rosteen:
It looks like all the test failures are unrelated remote data download failures. merging.

PR #1267, opened 2025-08-26
Bump min python version to 3.11, test 3.13

rosteen:
#1266 uses some astropy behavior that needs astropy 7.x, which also means we need to drop python 3.10 support. We're behind other packages on that, so it's time anyway.


minouHub:
I noticed today that Astropy stopped the support of Python 3.10. I understand that some news features require to move on and drop old python versions but this is 1 year earlier than the official end of support, see https://devguide.python.org/versions/. Are you planning on being ahead of schedule or was it exceptional?

PR #1273, opened 2025-09-24
Keep original units from the WCS for spectral axis

rosteen:
Using `wcs.spectral` converts the spectral axis to meters. This PR instead keeps the spectral axis in whatever units were specified in the input WCS in this case.


rosteen:
This isn't needed due to a fix in https://github.com/astropy/astropy/pull/18655, closing.

PR #1275, opened 2025-10-03
Avoid a private `EarthLocation` method in a test

eerovaher:
A test is relying on a private `EarthLocation` method that just got changed in `astropy` `main` (in https://github.com/astropy/astropy/commit/32f7e02966e2a5960387133c29197fa5e397f2a1) so it will not work in the future. Luckily the private method call is very simple to avoid and the replacement is compatible with older `astropy` versions.


rosteen (approved):
Seems reasonable enough, thanks!

PR #1282, opened 2025-11-17
Add CompoundSpectralRegion class to build regions with operators

### LINKED ISSUE #1205: Compound spectral region
Desired outcome: Have something like [compound regions in astropy regions package](https://astropy-regions.readthedocs.io/en/stable/compound.html) but for `SpectralRegion` in specutils.

Motivation: Currently [jdaviz has to maintain complicated and recursive logic to turn composite subsets in glue into SpectralRegion[(https://github.com/spacetelescope/jdaviz/blob/05e9afb0fdc11d062e52b483b11ee5ac1e8577ec/jdaviz/app.py#L1126). In addition to maintenance burden, such functionality would be indefinitely more useful even to non-Jdaviz users if implemented natively in specutils.

When this exists here, Jdaviz can then retire the complicated code and add [proper translators to glue-astronomy](https://github.com/glue-viz/glue-astronomy/blob/main/glue_astronomy/translators/spectrum1d.py) because then `glue` operators will [combine them natively via operator overloading](https://github.com/glue-viz/glue/blob/0c8f5e88c66c513b2132a123b11fc6bbc72b7f1d/glue/core/subset_group.py#L207).

rosteen:
This PR allows the user to create a `CompoundSpectralRegion` by combining `SpectralRegions` with operators (and, or, xor) and convert these back to a simple `SpectralRegion` defined by the equivalent subregions. This will be useful for downstream applications that do things like creating a `SpectralRegion` though UI actions. I also added a `contains` method to `SpectralRegion` for convenience, to check if a single value is within any of the subregions of the `Spectralregion`.

Closes #1205


pllim (commented):
Not all operators are supported, right? Should there be ValueError or something when unsupported operator is used?

Need a change log and some user facing doc.

Looks exciting. Thanks!

PR #1297, opened 2026-01-02
Classes diagram: Spectrum1D -> Spectrum

keflavich:
all in the title.

Process was: remove two characters in pptx, run slideshow, take screenshot, crop, save.


rosteen (approved):
Thanks!

Issue #1301, opened 2026-01-13
Convolution smoothing propagates uncertainty incorrectly


fcotizelati:
Uncertainty propagation in convolution smoothing uses the kernel rather than the kernel squared, which is inconsistent with variance propagation for a linear filter with independent errors. The current code convolves inverse variance (or variance/standard deviation) with `kernel`, not `kernel**2`, leading to underestimated or incorrect uncertainties.

rosteen:
Whoops, this should have been closed by #1305, the fix has been released.

PR #1305, opened 2026-01-16
Fix error propagation for smoothing and resampling

### LINKED ISSUE #1300: FluxConservingResampler uncertainty propagation mixes variance/stddev and inverts incorrectly
In `FluxConservingResampler`, uncertainty propagation in the overlap case appears to mix up variance and standard deviation. The input `errs` array comes from `VarianceUncertainty`, so it already contains variances.

For multi-bin overlap, the code effectively treats `errs` as if they were standard deviations: it squares them, then takes a square root, and finally passes the result to `InverseVariance` as though it were still a variance. This makes the multi-bin case inconsistent with the single-bin case and leads to incorrect uncertainties.

Conceptually, variance for a weighted average should be propagated directly and only then inverted. A simple fix would be to keep everything in variance space in the overlap case and ensure that the array passed to `InverseVariance` is always a variance.

rosteen:
Closes #1300 and #1301, thanks @fcotizelati for pointing out these bugs. I'm milestoning to 2.3 since I don't plan to do a 2.2.1 release before 2.3 at this point.

### DIFF
--- a/specutils/manipulation/resample.py
+++ b/specutils/manipulation/resample.py
@@ -210,7 +210,7 @@ class FluxConservingResampler(ResamplerBase):
                 output_fluxes[i] = final_flux
 
                 if errs is not None:
-                    output_errs[i] = np.sum((errs[first_bin:final_bin+1] * p_ij**2)) / (sum_pij * sum_pij)  # noqa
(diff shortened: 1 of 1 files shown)


rosteen:
Test failures are unrelated, due to the GWCS release yesterday. I'm working on fixing those elsewhere.

PR #1315, opened 2026-02-03
Deprecate old spelling of air-to-vac method

rosteen:
@pllim I decided you were right that the old spelling should be deprecated for a release before being removed.


pllim (approved):
LGTM. devdeps failures are unrelated.

I'll let you merge when you are ready. Thanks!

PR #1324, opened 2026-04-08
properly handle units in SpectralRegion.invert

cshanahan1:
The invert method for SpectralRegion had hard coded units of u.um applied to upper_bound and lower_bound inputs. These inputs are quantity objects themselves so their own units should be used. 

This PR removes the hard coded units and uses the actual units of upper_bound and lower_bound - if these units differ, upper_bound is converted to the unit of lower_bound (and this is explained in the docstring). If the units are not compatible, then an error is raised. Test coverage is added as well.

### DIFF
--- a/specutils/spectra/spectral_region.py
+++ b/specutils/spectra/spectral_region.py
@@ -369,16 +369,17 @@ class SpectralRegion:
         ----------
         lower_bound : `~astropy.units.Quantity`
            The lower bound of the region. Can be scalar with pixel or any
-           valid ``spectral_axis`` unit
+           valid ``spectral_axis`` unit that is equal to or compatible with the
+           unit of ``lower_bound``.
         upper_bound : `~astropy.units.Quantity`
            The upper bound of the region. Can be scalar with pixel or any
-              valid ``spectral_axis`` unit. If compatible with ``lower_bound``,
-              this value is converted to ``lower_bound`` units.
+           valid ``spectral_axis`` unit that is equal to or compatible with the
+           unit of ``lower_bound``.
 
         Returns
         -------
         spectral_region : `~specutils.SpectralRegion`
-           Spectral region of the non-selected regions
+           Spectral region of the non-selected regions.
 
         Raises
         ------
@@ -411,7 +412,7 @@ class SpectralRegion:
         #
         min_num = -sys.maxsize-1
         max_num = sys.maxsize
-        
+
         rs = self._subregions + [(min_num* lower_bound.unit, lower_bound),
                                  (upper_bound, max_num * upper_bound.unit)]
 
--- a/specutils/spectra/spectral_region.py
+++ b/specutils/spectra/spectral_region.py
@@ -413,7 +413,7 @@ class SpectralRegion:
         min_num = -sys.maxsize-1
         max_num = sys.maxsize
 
-        rs = self._subregions + [(min_num* lower_bound.unit, lower_bound),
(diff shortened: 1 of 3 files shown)


specutils/spectra/spectral_region.py
```diff
@@ -394,13 +402,19 @@ def invert(self, lower_bound, upper_bound):
...
+
         #
         # Create 'rs' region list with left and right extra ranges.
         #
         min_num = -sys.maxsize-1
         max_num = sys.maxsize
-        rs = self._subregions + [(min_num*u.um, lower_bound),
-                                 (upper_bound, max_num*u.um)]
+
+        rs = self._subregions + [(min_num* lower_bound.unit, lower_bound),
```
keflavich:
```suggestion
        rs = self._subregions + [(min_num * lower_bound.unit, lower_bound),
```

specutils/tests/test_regions.py
```diff
@@ -172,6 +172,42 @@ def test_invert():
...
+          SpectralRegion(0.3*u.um, 0.4*u.um) +
+          SpectralRegion(0.45*u.um, 0.6*u.um) +
+          SpectralRegion(0.8*u.um, 0.9*u.um) +
+          SpectralRegion(1.0*u.um, 1.2*u.um) +
+          SpectralRegion(1.3*u.um, 1.5*u.um))
+
+    sr_inverted_expected = [(0.05*u.um, 0.15*u.um), (0.2*u.um, 0.3*u.um),
+                            (0.4*u.um, 0.45*u.um), (0.6*u.um, 0.8*u.um),
+                            (0.9*u.um, 1.0*u.um), (1.2*u.um, 1.3*u.um),
+                            (1.5*u.um, 3.0*u.um)]
```
keflavich: some of these tests should use different units, otherwise this isn't serving as a regression test

keflavich (commented):
I sort of approve, but I'm not super clear on there being a problem to begin with.

The hard-coded micron unit is only being used for sys.maxsize values - i.e., it's serving the roles of "zero" and "infinity".  The choice of unit - micron here - is somewhat arbitrary, but we kind of expect all spectra to start and end many orders of magnitude above the minimum and below the maximum expressible value, such that this change should have no effect for any theoretically possible observations.

What am I missing?

(suggestion is a minor style thing)

cshanahan1:
> I sort of approve, but I'm not super clear on there being a problem to begin with.
> 
> The hard-coded micron unit is only being used for sys.maxsize values - i.e., it's serving the roles of "zero" and "infinity". The choice of unit - micron here - is somewhat arbitrary, but we kind of expect all spectra to start and end many orders of magnitude above the minimum and below the maximum expressible value, such that this change should have no effect for any theoretically possible observations.
> 
> What am I missing?
> 
> (suggestion is a minor style thing)

The issue arrises when the units of upper_bound and lower_bound are not able to be converted to u.um (e.g are in units of 'pixel' or require an equivalency). I think using the input units is the easiest solution for this rather than detecting what type of unit is present to choose which arbitrary unit to use.

keflavich:
OK, got it, that makes sense.  Maybe add a note to `def test_invert_incompatible_bounds_units_raises():` that it is a regression test for this specific case (which you have encountered in practice).

rosteen:
@keflavich Look good to you now?

keflavich (approved):

rosteen:
Docs failure was just a remote data timeout, merging.

PR #1332, opened 2026-06-10
Adding redshift and radial velocity properties to SpectrumCollection

### LINKED ISSUE #1318: How do I apply `shift_spectrum_to(radial_velocity=...)` to a `SpectrumCollection`?
`Spectrum.shift_spectrum_to(...)` updates the wavelength axis in place, but I cannot find the equivalent for a `SpectrumCollection`:

```Python
from astropy.io import fits
filename = 'https://data.sdss.org/sas/dr16/sdss/spectro/redux/26/spectra/1323/spec-1323-52797-0012.fits'
# The spectrum is in the second HDU of this file.
with fits.open(filename) as f:
    specdata = f[1].data
from specutils import Spectrum
lamb = 10**specdata['loglam'] * u.AA
flux = specdata['flux'] * 10**-17 * u.Unit('erg cm-2 s-1 AA-1')
spec = Spectrum(spectral_axis=lamb, flux=flux)
spec_collection = SpectrumCollection.from_spectra([spec])

for s in spec_collection:
    print(s.wavelength)
    s.shift_spectrum_to(radial_velocity=-2700.5 * u.km / u.s)
    print(s.wavelength)
print(spec_collection[0].wavelength)
```
This gives me the following output:
```
[3815.0483 3815.926  3816.806  ... 9202.379  9204.495  9206.613 ] Angstrom
[3780.8362 3781.706  3782.578  ... 9119.8545 9121.952  9124.051 ] Angstrom
[3815.0483 3815.926  3816.806  ... 9202.379  9204.495  9206.613 ] Angstrom
```

Note how the wavelength for `s` is updated, but once I leave the loop, it's gone and `spec_collection[0]` leaves me with the old data.

It seems that iterating over a `SpectrumCollection` doesn't give me the spectra that are actually in it, but copies of them?
I'm not sure if that's a bug or a feature. If that's intended, I'm happy to do a PR to highlight that in the documentation. Looking at the existing documentation, there are sentences like "SpectrumCollection objects can be treated just like Spectrum objects;" which doesn't sound like "You can update one of them in place, but in the other you always need to make a new object".

In fact, I think it would be useful to give `SpectrumCollection` its own `shift_spectrum_to` method. At least the way I use it, the collections are often spectra from the same observation (e.g. different echelle orders) so they naturally need the same heliocentric correction and the same redshift. Even when the collection contains spectra from different observations, they are typically related in some way, e.g. they are spectra from the same object taken at different times (same radial velocity correction) or for different stars in e.g. a cluster, observed with multiple fibers in the same field (same observation time and almost the same pointing, thus the same heliocentric correction).

rosteen:
Currently only allow scalar values (rather than an array of values matching the number of spectra), and allows shifting the spectral axes with `shift_spectrum_to` like `Spectrum` does. I also added some clarification to the narrative docs about what exactly is happening when you access a single spectrum from a SpectrumCollection. Allowing array redshifts (one per spectrum in the collection) requires some upstream work in Astropy first.

Closes #1318.

### DIFF
--- a/docs/spectrum_collection.rst
+++ b/docs/spectrum_collection.rst
@@ -96,7 +96,7 @@ and ``M`` is the length of the output dispersion grid.
     >>> print(spec_coll.flux.shape)
     (2, 50)
 
-It is important to understand that `~specutils.SpectrumCollection` does *not* store
+~specutils.SpectrumCollection` does *not* store
 the data as a collection of individual `~specutils.Spectrum`
 objects, but rather contains single multi-dimensional flux and spectral axis arrays, and
 constructs a `~specutils.Spectrum` on the fly from the corresponding slices of these
--- a/docs/spectrum_collection.rst
+++ b/docs/spectrum_collection.rst
@@ -96,6 +96,14 @@ and ``M`` is the length of the output dispersion grid.
     >>> print(spec_coll.flux.shape)
     (2, 50)
 
+It is important to understand that `~specutils.SpectrumCollection` does *not* store
+the data as a collection of individual `~specutils.Spectrum`
+objects, but rather contains single multi-dimensional flux and spectral axis arrays, and
+constructs a `~specutils.Spectrum` on the fly from the corresponding slices of these
+arrays when the collection is indexed. Thus, doing operations on the returned
+`~specutils.Spectrum` returned by indexing a `~specutils.SpectrumCollection`
+in this case does not affect the `~specutils.SpectrumCollection` object itself.
+
 
 Reference/API
 -------------
--- a/specutils/spectra/spectrum_collection.py
+++ b/specutils/spectra/spectrum_collection.py
@@ -78,7 +78,8 @@ class SpectrumCollection(NDIOMixin, RedshiftMixin):
             raise ValueError("Cannot set both radial_velocity and redshift at "
                              "the same time.")
 
-        if redshift is not None and not isinstance(redshift, (int, float)):
+        if redshift is not None and not (isinstance(redshift, (int, float)) or
+                (isinstance(redshift, u.Quantity) and redshift.ndim == 0)):
             raise ValueError("Only single-value redshifts are supported at this time.")
 
         if radial_velocity is not None and (not isinstance(radial_velocity, u.Quantity)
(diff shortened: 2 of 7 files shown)


docs/spectrum_collection.rst
```diff
@@ -96,6 +96,14 @@ and ``M`` is the length of the output dispersion grid.
     >>> print(spec_coll.flux.shape)
     (2, 50)
 
+It is important to understand that `~specutils.SpectrumCollection` does *not* store
```
keflavich:
```suggestion
~specutils.SpectrumCollection` does *not* store
```

specutils/spectra/spectrum_collection.py
```diff
@@ -48,20 +49,58 @@ class SpectrumCollection(NDIOMixin):
     mask : ndarray or None
         The n-dimensional mask information associated with each spectrum. If
         present, must match the dimensionality of ``flux``.
+    redshift : float or None
+        A single value to apply to every spectrum in the collection.
+        Cannot be specified if radial_velocity is specified.
```
keflavich:
```suggestion
        Cannot be specified if ``radial_velocity`` is specified.
```

specutils/spectra/spectrum_collection.py
```diff
@@ -48,20 +49,58 @@ class SpectrumCollection(NDIOMixin):
...
+    radial_velocity : `~astropy.units.Quantity`
+        A single value to apply to all spectra in the collection.
+        Cannot be specified if redshift is specified.
```
keflavich:
```suggestion
        Cannot be specified if ``redshift`` is specified.
```

specutils/spectra/spectrum_collection.py
```diff
@@ -48,20 +49,58 @@ class SpectrumCollection(NDIOMixin):
...
             raise u.UnitsError("Flux must be a `Quantity`.")
 
+        # Ensure that only one or neither of these parameters is set
+        if redshift is not None and radial_velocity is not None:
+            raise ValueError("Cannot set both radial_velocity and redshift at "
+                             "the same time.")
+
+        if redshift is not None and not (isinstance(redshift, (int, float)) or
+                (isinstance(redshift, u.Quantity) and redshift.ndim == 0)):
+            raise ValueError("Only single-value redshifts are supported at this time.")
```
keflavich: maybe `np.isscalar` can replace this logic?

specutils/spectra/spectrum_collection.py
```diff
@@ -181,13 +222,18 @@ def from_spectra(cls, spectra):
             warnings.warn("Not all spectra have associated masks, "
                           "skipping masks.")
 
+        # Check that all redshifts are equal
+        redshift = spectra[0].redshift
+        if (not all((x.redshift == redshift for x in spectra))):
+            warnings.warn("Not all redshifts are equal, using redshift of first spectrum.")
```
keflavich: maybe this should raise an exception?  I'm not sure when using the first would be acceptable

keflavich (commented):
I have some minor wording suggestions.

I'm not sure it is ever right to adopt the first redshift out of a collection if they're not all equal

specutils/spectra/spectrum_collection.py
rosteen: I was iffy on this, I'll change it to an exception.

specutils/spectra/spectrum_collection.py
rosteen: Unfortunately it doesn't work on Quantities, so it can't replace everything here.

specutils/spectra/spectrum_collection.py
rosteen: Oh, turns out this already would have been caught by an earlier check of `radial_velocity` (which raises an exception), I removed this.

rosteen:
@keflavich I addressed your comments and fixed the failing docs build, I think this is ready for another look.

keflavich (approved):

PR #1339, opened 2026-07-29
Adds a flag for returning the model spectra for SDSS BHM spec files

havok2063:
This PR adds a `model` flag to the loader for SDSS BHM spec files to optionally return the model spectrum.  The model is stored as a separate column in the same table with the flux spectrum.  This flag sets the model data as the Spectrum "flux".  There is an added metadata boolean keyword `is_model` whether the returned Spectrum object is a model or not.  This is a placeholder implementation until specutils adds support for model spectra.  

```
# returns original flux spectrum
flux = Spectrum.read(file, format='SDSS-V spec')
flux.meta["is_model"]
False

# returns model spectrum
model = Spectrum.read(file, format='SDSS-V spec',  model=True)
model.meta["is_model"]
True
```
Pinging @albireox for review as well.


havok2063:
@rosteen Any chance on getting this reviewed soon?

rosteen:
> @rosteen Any chance on getting this reviewed soon?

I'll try to get to this (and your other PR) tomorrow or Monday.

rosteen (approved):
This looks reasonable to me, but I'm curious what your vision is for future model support. Loading both automatically?

havok2063:
@rosteen  Yeah, I was imagining that `model` would live as an optional attribute on `Spectrum`, a la `mask`, that one could set within a data_loader when constructing the Spectrum object.  Then it would live alongside flux, and other arrays.  This would work fine for model spectra within the same file as the flux.  

One complication would be for cases where the model is defined in a separate file.  In that case, maybe you would have a separate data loader for the model, which could be loaded separately to the flux Spectrum.  Or maybe there could be a `Spectrum.load_model` method that could either load a model manually or call an exiting data_loader to link and associate the flux+model Spectrum objects together.  

SDSS has both cases that would need to be supported.

rosteen:
> @rosteen Yeah, I was imagining that `model` would live as an optional attribute on `Spectrum`, a la `mask`, that one could set within a data_loader when constructing the Spectrum object. Then it would live alongside flux, and other arrays. This would work fine for model spectra within the same file as the flux.
> 
> One complication would be for cases where the model is defined in a separate file. In that case, maybe you would have a separate data loader for the model, which could be loaded separately to the flux Spectrum. Or maybe there could be a `Spectrum.load_model` method that could either load a model manually or call an exiting data_loader to link and associate the flux+model Spectrum objects together.
> 
> SDSS has both cases that would need to be supported.

Got it, thanks for the details - that seems like a good plan to me and this makes sense as a stepping stone in that direction. I'll go ahead and merge after changelog and codestyle checks are resolved - want me to go ahead and fix those or do you have time?

havok2063:
If they are quick and easy and you have the time, go for it.  Otherwise I could get to it maybe on Thursday.

rosteen:
Docs failure is unrelated.

PR #1341, opened 2026-08-20
Simplify cubemodel test

braingram:
This simplifies the cubemodel test using the stdatadmodels datamodels API instead of asdf_in_fits. This produces a test file that is closer to real files (the pipeline uses the datamodels API to produce files) and simplifies the code.


braingram:
The docs failure looks unrelated. I don't think this needs a changelog but can't add labels.

PR #1348, opened 2026-09-17
Allow SpectralAxis to represent medium (e.g., air/vac), reference frame, source position, observer position and time

andycasey:
Hi,

Every time I have some new project that involves spectra I go back to see whether `specutils` natively has everything that I would need. My desiderata are mostly about metadata on how spectra are represented, and less about analysis utilities.

This PR addresses that desiderata, and is **a draft PR** intended to spark discussion about the best way to implement these features.

In summary the features introduced are to represent:
- the medium that a spectrum was recorded in (e.g., vacuum/air), as a frozen `SpectralMedium` record;
- the reference frame of the observed spectrum (e.g., observed, at rest, etc);
- the source location (this was already possible, but not really used because of inconsistency in how coordinates are stored in spectrum files);
- the time of observation (variance in header keys, again); and 
- the observer location.

An example use case is one where I am fitting spectra from different instruments (eg air/vac varies), my forward model is in vacuum, and I want to simultaneously fit telluric features and stellar features. To fit the tellurics correctly I need to know the reference frame and the medium, and for example to account for barycentric or radial motion I need to know the source location, the observatory location, and the time of observations.

That's a lot of metadata that `specutils` currently glosses over, sometimes for good reason. Header keys are a mess for representing observer location, time, and even coordinates. It's a little more straightforward for recording the medium (there is a standard FITS keyword), but currently `specutils` ignores that and has no standard way to store it.

In the implementation for this PR, the information is stored on the `SpectralAxis` so it survives slicing, arithmetic, resampling, region extraction, and copying. The introduced attributes include:
- `medium` (a frozen `SpectralMedium` record, with the refraction formula and optional temperature, pressure, humidity, and CO2
- `frame` using the FITS SPECSYS vocabulary, with `SOURCE` as the rest frame
- `target`, `obstime`, `location`: the observer is build from frame, location, and time, and stays consistent with the frame through any transformations

I would appreciate any input on both:
(1) whether you think this functionality is worthwhile for consideration in specutils (it is for me, and others I know, but that is a biased representation);
(2) whether this implementation is suitable, or if some other way is better.


rosteen:
Thanks for the contribution! I don't have time to review this right away, but I think @keflavich is going to take a look next week.

keflavich:
I'm starting a review.  At the least, I agree with the overall idea and principles and would find the metadata described very useful.

docs/spectrum.rst
```diff
@@ -224,6 +224,168 @@ An example of the different treatments of the ``spectral_axis`` is shown below.
...
+Medium, Reference Frame and Observation Metadata
+------------------------------------------------
+
+The values of a ``spectral_axis`` only have a definite meaning once you know
+the medium the wavelengths refer to (vacuum or air), the reference frame in
+which they were measured (the telescope, the solar system barycentre, the
+rest frame of the source, ...), and, to convert between frames, where the
+source is and when and where it was observed. :class:`~specutils.Spectrum`
+records all of these on its :class:`~specutils.SpectralAxis`, and they are
+carried through slicing, arithmetic, resampling and region extraction:
```
keflavich: we should clarify here what happens if spectra are combined in some way (averaged over two exposures, say, taken at different times/temperatures).  I don't know immediately how that happens.

docs/spectrum.rst
```diff
@@ -224,6 +224,168 @@ An example of the different treatments of the ``spectral_axis`` is shown below.
...
+    >>> spec.medium
+    SpectralMedium(kind='air', refraction_method='Morton2000')
+    >>> spec.frame
+    'TOPOCENT'
+    >>> spec.in_rest_frame
+    False
+    >>> spec[2:5].frame
+    'TOPOCENT'
+
+The metadata is:
```
keflavich: are?

keflavich: there are multiple entries below, so no matter ones' opinion on the plurality of data, I think this should be 'are'

docs/spectrum.rst
```diff
@@ -224,6 +224,168 @@ An example of the different treatments of the ``spectral_axis`` is shown below.
...
+
+* ``medium``: a :class:`~specutils.SpectralMedium`, or simply ``'vacuum'`` or
+  ``'air'``. An air medium can also record which formula for the refractive
+  index of air the wavelengths correspond to, and the air ``temperature``,
+  ``pressure``, ``humidity`` and ``co2`` concentration (see
+  `~specutils.utils.wcs_utils.refraction_index`). Air wavelengths cannot be
+  converted to frequency, energy or velocity, since those relations only hold
+  in vacuum; convert the spectrum to vacuum first.
```
keflavich: maybe a note-to-selves: we should make sure this is enforced

docs/spectrum.rst
```diff
@@ -224,6 +224,168 @@ An example of the different treatments of the ``spectral_axis`` is shown below.
...
+
+    >>> spec.radial_velocity  # doctest: +FLOAT_CMP
+    <Quantity 9.4102236 km / s>
+    >>> spec.barycentric_correction  # doctest: +FLOAT_CMP
+    <Quantity -9.4102236 km / s>
+
+Converting between media and frames
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:meth:`~specutils.Spectrum.with_medium` returns a copy of the spectrum with
```
keflavich: I like this approach; it's similar to what we do in spectral-cube.

We should clarify whether the spectrum _data_ are copied or if there's just a reference to the same in-memory or on-disk array (I would naively assume the latter)

docs/spectrum.rst
```diff
@@ -224,6 +224,168 @@ An example of the different treatments of the ``spectral_axis`` is shown below.
...
+.. code-block:: python
+
+    >>> obs = Spectrum(spectral_axis=np.linspace(5000, 5010, 11) * u.AA,
+    ...                flux=np.ones(11) * u.Jy, frame='TOPOCENT',
+    ...                radial_velocity=42 * u.km / u.s)
+    >>> bary = obs.with_frame('BARYCENT', velocity=-9.41 * u.km / u.s)
+    >>> bary.radial_velocity  # doctest: +FLOAT_CMP
+    <Quantity 32.59 km / s>
+    >>> bary.to_rest().radial_velocity
+    <Quantity 0. km / s>
```
keflavich: I find this a bit confusing.

We have a source with a topocentric radial velocity of 42 km/s, then we change it to a barycentric frame by subtracting 9.41 km/s.  Normally the topo->bary conversion would be computed using the observer location and time - is this `with_frame` call overriding that by specifying the topo->bary conversion?  i.e., is this what you might use if you do not have the observer location but you _do_ know the topo<->bary conversion?

keflavich (commented):
I started reviewing this but haven't been able to finish - I'm posting my partial review with the note-to-self that I was apparently on line 145 of spectral_frame.py when I stopped.  Remind me after the JWST deadline if you don't hear from me!

